PR #2, opened 2018-09-10
StormEurope class and utility

mmyrte:
Hi Gabriela,

after a lot of Haareraufen because of git, I finally get around to submitting a PR. Maybe you can just give me commit rights to the original repo? It would avoid the messy commit structure.

By the way: I implemented your advice right away regarding the adm0 centroids attribute. I also switched the centroids logic around a bit, since I didn't want to commit the default WISC centroids pickle. Now it defaults to constructing a barebones Centroids object from the first netcdf file in the list.

best
Jan


gabrielaznar:
Thanks Jan! Now you can push already.
We'll keep in touch,
Gabriela

PR #17, opened 2020-04-23
read_raster: correctly handle windows with float bounds

tovogt:
The utility function `read_raster` in `climada.util.coordinates` can't handle windows that have non-integer bounds. The tests only cover the case of integer bounds so far: https://github.com/CLIMADA-project/climada_python/blob/7ec2cde13d762d09f88cced8d6a7973e0dad9143/climada/util/test/test_coordinates.py#L301

A common usecase would be if you get the window from coordinate bounds, like in the following:
```
import rasterio
from climada.util.coordinates import read_raster
from climada.util.constants import HAZ_DEMO_FL

with rasterio.open('/vsigzip/' + HAZ_DEMO_FL, "r") as src:
    win = rasterio.windows.from_bounds(-69.3, 10.0, -69.0, 10.4, transform=src.transform)
    
meta, inten_ras = read_raster(HAZ_DEMO_FL, window=win)
```

This will fail with a `KeyError` because the second argument to `reshape` is non-integer valued in the following line: https://github.com/CLIMADA-project/climada_python/blob/7ec2cde13d762d09f88cced8d6a7973e0dad9143/climada/util/coordinates.py#L485

Before you accept my PR, we should double check the following:

* Is there any reason why the second argument to reshape isn't set to `-1`?
* Is there any reason why the `meta['width']` and `meta['height']` aren't set depending on the array shape returned by the read operation in https://github.com/CLIMADA-project/climada_python/blob/7ec2cde13d762d09f88cced8d6a7973e0dad9143/climada/util/coordinates.py#L479?


tovogt:
@ThomasRoosli If you think this is okay, maybe you can merge? Or maybe I misunderstood your comment in today's call?

PR #29, opened 2020-05-28
RCLIPER model to compute rainfield from track

aleeciu:
Added an hazard class: rainfield from tropical cyclone tracks based on the RCLIPER model. Code based on the Matlab implementation.


aleeciu:
done, also added testing with results from MATLAB version and applied PEP related changes

PR #47, opened 2020-07-31
Feature: tc extent

mmyrte:
How about this impementation? Paging @tovogt 

I changed the `tc_tracks_synth.calc_random_walk` to only pass back the data, while still being able to use the instance for its `get_extent` method. Vielen Dank für deine Hilfe :)

### DIFF
--- a/climada/hazard/tc_tracks_synth.py
+++ b/climada/hazard/tc_tracks_synth.py
@@ -29,6 +29,7 @@ from numba import jit
 import numpy as np
 
 from climada.util.config import CONFIG
+import climada.util.coordinates
 import climada.hazard.tc_tracks
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


climada/hazard/tc_tracks_synth.py
```diff
@@ -76,6 +77,10 @@ def calc_random_walk(tracks,
             Default: configuration file
         decay (bool, optional): compute land decay in probabilistic tracks.
             Default: True
+
+
+    Returns:
+        List of xarray.Dataset instances representing cyclones
```
tovogt, line 83: Return None (remove the `Returns: ...` part of the docstring and make note in the body of the docstring that the input object is modified in place).

climada/hazard/tc_tracks_synth.py
```diff
@@ -100,22 +105,27 @@ def calc_random_walk(tracks,
...
-                                            pool=tracks.pool)
+                tracks.data = _apply_land_decay(tracks.data, v_rel, p_rel,
+                                                land_geom, pool=tracks.pool)
             except ValueError:
                 LOGGER.info('No land decay coefficients could be applied.')
         else:
             LOGGER.error('No historical tracks contained. '
                          'Historical tracks are needed for land decay.')
-    return new_ens
+    return tracks.data
```
tovogt, line 128: Remove the `return ...` line (input object is modified in place).

mmyrte:
You can WHAT, this is magic, my god, this OO stuff is whacky coming from R's functional style 🙈 

Thanks so much, the new commit should resolve it.

PR #49, opened 2020-08-06
Centroids: select by extent (aka clipping)

mmyrte:
@tovogt I also added a little test for `Centroids().from_base_grid()`. Could you give this a look?

### DIFF
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -560,18 +560,31 @@ class Centroids():
         sel_cen = geom_wkb.drop_duplicates().index
         return self.select(sel_cen=sel_cen)
 
-    def select(self, reg_id=None, sel_cen=None):
+    def select(self, reg_id=None, extent=None, sel_cen=None):
         """Return Centroids with points in the given reg_id or within mask
 
         Parameters:
             reg_id (int): region to filter according to region_id values
-            sel_cen (np.array): 1-dim mask
+            extent (tuple): Format (min_lon, max_lon, min_lat, max_lat) tuple.
+            sel_cen (np.array): 1-dim mask, overrides reg_id and extent
 
         Returns:
             Centroids
         """
         if sel_cen is None:
-            sel_cen = np.isin(self.region_id, reg_id)
+            if reg_id:
+                sel_reg = np.isin(self.region_id, reg_id)
+            if extent:
+                sel_ext = \
+                    (extent[0] < self.lon) & (extent[1] > self.lon) & \
+                    (extent[2] < self.lat) & (extent[3] > self.lat)
+
+            if extent and reg_id:
+                sel_cen = np.logical_and(sel_reg, sel_ext)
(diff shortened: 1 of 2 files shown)


sameberenz:
Thanks, that's a usefull feature.
In many modules, what you call "extent" here is the parameter "bbox" (for "bounding box").
Maybe this is already not consistent throughout CLIMADA, but I would suggest sticking with one name for this parameter.

climada/hazard/centroids/centr.py
```diff
@@ -560,18 +560,31 @@ def remove_duplicate_points(self, scheduler=None):
...
+                sel_ext = \
+                    (extent[0] < self.lon) & (extent[1] > self.lon) & \
+                    (extent[2] < self.lat) & (extent[3] > self.lat)
+
+            if extent and reg_id:
+                sel_cen = np.logical_and(sel_reg, sel_ext)
+            elif extent and not reg_id:
+                sel_cen = sel_ext
+            else:
+                sel_cen = sel_reg
```
tovogt, line 587:
```suggestion
            sel_cen = np.ones_like(self.region_id, dtype=bool)
            if reg_id:
                sel_cen &= np.isin(self.region_id, reg_id)
            if extent:
                sel_cen &= ((extent[0] < self.lon) & (extent[1] > self.lon)
                            & (extent[2] < self.lat) & (extent[3] > self.lat))
```

climada/hazard/centroids/centr.py
```diff
@@ -560,18 +560,31 @@ def remove_duplicate_points(self, scheduler=None):
...
         return self.select(sel_cen=sel_cen)
 
-    def select(self, reg_id=None, sel_cen=None):
+    def select(self, reg_id=None, extent=None, sel_cen=None):
         """Return Centroids with points in the given reg_id or within mask
 
         Parameters:
             reg_id (int): region to filter according to region_id values
-            sel_cen (np.array): 1-dim mask
+            extent (tuple): Format (min_lon, max_lon, min_lat, max_lat) tuple.
```
tovogt, line 568: You should consider renaming it `bbox` (like in many other CLIMADA modules) or `bounds` (like in rasterio and many CLIMADA modules). But make sure to use the order `(lon_min, lat_min, lon_max, lat_max)` in that case. The name `extent` with the order `(lon_min, lon_max, lat_min, lat_max)` is mainly used in matplotlib contexts, as far as I know.

mmyrte:
Yeah the terminology is a problem. All three (bounding box, bounds, extent) are around in climada, AFAIK. It seems that [we're not the only ones](https://gis.stackexchange.com/questions/240979/difference-between-bounding-box-envelope-extent-bounds) struggling.

I think I'll stick with extent for now (ag gives me 92 matches on `extent`, 339 on `bbox`, 184 on `bounds`); we should coordinate a cleanup at a later time. That would then include a translation function from the bbox tuple order to matplotlib extent. Sounds good?

PR #50, opened 2020-08-07
Feature/isimip crop

sameberenz:
@chahank , could you please review our pull request for the crop features.
this tutorial gives an overview over the core functionalities:
https://github.com/CLIMADA-project/climada_python/blob/feature/isimip_crop/doc/tutorial/climada_hazard_entity_Crop.ipynb

ps: parts of the crop modules were developed still on _develop_ before the new git flow was implemented. this feature branch was therefore started after start of the project.


climada/entity/exposures/crop_production.py
```diff
@@ -37,27 +37,32 @@
 logging.root.setLevel(logging.DEBUG)
 LOGGER = logging.getLogger(__name__)
 
-DEF_HAZ_TYPE = 'CP'
+DEF_HAZ_TYPE = 'RC'
```
chahank, line 40: Why 'RC'?

carmensteinmann: RC stands for "Relative Cropyield" (we changed the hazard from crop_potential to relative_cropyield).

climada/entity/exposures/crop_production.py
```diff
@@ -37,27 +37,32 @@
...
 """Default hazard type used in impact functions id."""
 
 BBOX = np.array([-180, -85, 180, 85])  # [Lon min, lat min, lon max, lat max]
+"""Default geographical bounding box"""
```
chahank, line 44: Why these values? Comment should explain.

climada/entity/exposures/crop_production.py
```diff
@@ -371,76 +389,85 @@ def set_to_usd(self, input_dir=INPUT_DIR, yearrange=YEARS_FAO):
...
-        fao_year = fao.Year.values
-        fao_price = fao.Value.values
+        fao = dict()
+        fao['file'] = pd.read_csv(os.path.join(input_dir, FAO_FILE))
+        fao['crops'] = fao['file'].Item.values
+        fao['year'] = fao['file'].Year.values
+        fao['price'] = fao['file'].Value.values
 
-        fao_country = co.country_faocode2iso(fao_area)
+        fao_country = co.country_faocode2iso(getattr(fao['file'], 'Area Code').values)
```
chahank, line 418: `coord. `would be easire to read than `co.`

climada/entity/exposures/crop_production.py
```diff
@@ -371,76 +389,85 @@ def set_to_usd(self, input_dir=INPUT_DIR, yearrange=YEARS_FAO):
...
+
         # iterate over all countries that are covered in the exposure, extract the according price
-        # and calculate the produced yield in USD/y
+        # and calculate the crop production in USD/y
         area_price = np.zeros(self.value.size)
-        for item, _ in enumerate(list_countries):
-            country = list_countries[item]
+        for country in list_countries:
             if country != 'No country':
                 if country == 'Other country':
```
chahank, line 440: Can be made with one `if` with an `and`

climada/entity/exposures/crop_production.py
```diff
@@ -481,57 +511,65 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
-        filename_saveto = ('cropyield_isimip_' + ((item[2]).split('-'))[0] + '-'
-                           + (((item[2]).split('-'))[1]).split('.')[0] + '_'
-                           + str(yearrange[0]) + '-' + str(yearrange[1]) + '.hdf5')
-        filename_list.append(filename_saveto)
-        output_list.append(cropyield)
-        cropyield.write_hdf5(os.path.join(output_dir, 'Exposure', filename_saveto))
-
-    if returns == 'filename_list':
+    for file in filenames:
+        item = file.split('_')
```
chahank, line 515: `item` is not very clear as name here.

climada/entity/exposures/crop_production.py
```diff
@@ -541,62 +579,78 @@ def normalize_with_fao_cropyield(exp_firr, exp_noirr, input_dir=INPUT_DIR,
...
     fao_country = co.country_iso2faocode(country_list)
 
-    fao_yield = np.zeros(len(country_list))
+    fao_crop_production = np.zeros(len(country_list))
     ratio = np.ones(len(country_list))
-    exp_firr_norm = CropyieldIsimip()
-    exp_firr_norm = exp_firr
-    exp_noirr_norm = CropyieldIsimip()
-    exp_noirr_norm = exp_noirr
+    exp_firr_norm = CropProduction()
```
chahank, line 584: Use

chahank: You can use the copy module: 
`import copy`
`exp_firr_norm = copy.deepcopy(exp_firr)`

climada/entity/exposures/crop_production.py
```diff
@@ -541,62 +579,78 @@ def normalize_with_fao_cropyield(exp_firr, exp_noirr, input_dir=INPUT_DIR,
...
+        return_data (str): returned output
+            True: lists containing data for each exposure file. Lists: crops, country list,
+            ratio = FAO/ISIMIP, normalized exposures, crop production per country as documented
+            by the FAO and calculated by the ISIMIP dataset
+            False: lists containing data for each exposure file. Lists: crops, country list,
             ratio = FAO/ISIMIP, normalized exposures
-        Returns:
 
+        Returns:
+            c.f. normalize_with_fao_cp
```
chahank, line 653: What does c.f. normalize_with_fao_cp mean?

climada/entity/exposures/test/test_crop_production.py
```diff
@@ -78,7 +79,29 @@ def test_set_to_usd_unnecessary(self):
...
+        self.assertAlmostEqual(fao_crop_production[2], 673416.4)
+        self.assertAlmostEqual(fao_crop_production[11], 160328.7)
+        self.assertAlmostEqual(np.nanmax(exp_firr_norm.value.values), 220735.69212710857)
+        self.assertAlmostEqual(np.nanmax(exp_firr_norm.value.values), np.nanmax(exp_noirr_norm.value.values))
+        self.assertAlmostEqual(np.nanmax(exp.value.values), 284244.81023404596)
+        self.assertAlmostEqual(np.nansum(exp_noirr_norm.value.values) + np.nansum(exp_firr_norm.value.values), np.nansum(fao_crop_production), places=1)
+
+        self.assertListEqual(list(country_list), [0, 40, 56, 70, 191, 203, 208, 250,
+                                                  276, 380, 442, 528, 616, 705, 724, 756, 826])
+
```
chahank, line 103: Add tests for all methods of the class CropProduction.

climada/entity/impact_funcs/relative_cropyield.py
```diff
@@ -14,30 +14,30 @@
...
+        """Impact functions defining the impact as intensity"""
+        self.haz_type = 'RC'
         self.id = 1
-        self.name = 'Crop Potential ISIMIP'
+        self.name = 'Relative Cropyield ISIMIP'
         self.intensity_unit = ''
-        self.intensity = np.arange(0, 11)
-        self.mdr = (self.intensity - 1)
-        self.mdd = (self.intensity - 1)
+        self.intensity = np.arange(-1, 10)
```
chahank, line 40: Maybe comment on why it starts at -1.

climada/entity/exposures/test/test_crop_production.py
chahank, line 103: Add tests for all methods of the class CropProduction.

chahank:
Suggestions for the tutorial:
- A short description of what the module does in laymen terms at the top would be nice (in particular what Relative Crop Yield and Crop Production are, from what models they are computed and how)
- In[2] remove the commented out code, fix print outpus
- In[14] use ' ' and " " consistently
- In[14] why use the DATA_DIR from Climada to save computation outputs?
- In[14] input_haz -> input_haz_dir    (do this for all the lines where it is)
- CropProduction Exposure: Define the variables USD, t, y, ha (and use latex to format the formulae maybe?)
- In[42] Explain what the normalization of the data does
- Explain what IFRelative Cropyield is

climada/hazard/relative_cropyield.py
```diff
@@ -166,32 +167,33 @@ def set_from_single_run(self, input_dir=None, bbox=BBOX,
...
         # define indexes of the netcdf-bands to be extracted, and the
         # corresponding event names and dates
         # corrected indexes due to the bands in input starting with the index=1
         id_bands = np.arange(yearrange[0] - yearchunk['startyear'] + 1,
                              yearrange[1] - yearchunk['startyear'] + 2).tolist()
-        event_list = [str(n) for n in range(int(yearrange[0]), int(yearrange[1] + 1))]
 
         # hazard setup: set attributes
         self.set_raster([filename], band=id_bands,
                         geometry=list([shapely.geometry.box(bbox[0], bbox[1], bbox[2], bbox[3])]))
```
chahank, line 185: An easier to read way to do:
`[Lon min, lat min, lon max, lat max] = bbox
  self.set_raster([filename], band=id_bands,
         geometry=list([shapely.geometry.box(Lon min, lat min, lon max, lat max)]))`

climada/hazard/relative_cropyield.py
```diff
@@ -201,13 +203,16 @@ def set_from_single_run(self, input_dir=None, bbox=BBOX,
...
+            save (boolean): save mean to file? default: False
+            output_dir (str): path of output directory
 
             Returns:
                 hist_mean(array): contains mean value over the given reference
-                time period for each centroid
+                    time period for each centroid
         """
         time_array = np.array(self.event_name, dtype=int)
         idx_time = np.zeros(2, dtype=int)
```
chahank, line 218: The idx_time is not really needed. This part can be written in a simpler/more straightforward way.

Also, it is not ideal to have hardcoded indices on the yearrange variable. If in the future another yearrange is added, the code would be broken. Can this be written in a more flexible way?

climada/hazard/relative_cropyield.py
```diff
@@ -247,18 +252,15 @@ def set_rel_yield_to_int(self, hist_mean):
         idx = np.where(hist_mean != 0)[0]
 
         # initialize new hazard_matrix
-        hazard_matrix = np.empty(self.intensity.shape)
-        hazard_matrix[:, :] = np.nan
+        hazard_matrix = np.zeros(self.intensity.shape, dtype=np.float32)
 
         # compute relative yield for each event:
         for event in range(len(self.event_id)):
```
chahank, line 258: Why first make a non-sparse matric and then convert it to a sparse matrix? This seems not very efficient...

climada/hazard/relative_cropyield.py
```diff
@@ -247,18 +252,15 @@ def set_rel_yield_to_int(self, hist_mean):
...
-            hazard_matrix[event, idx] = self.intensity[event, idx] / hist_mean[idx]
+            hazard_matrix[event, idx] = (self.intensity[event, idx] / hist_mean[idx])-1
 
         self.intensity = sparse.csr_matrix(hazard_matrix)
         self.intensity_def = 'Relative Yield'
         self.units = ''
-        self.intensity.max = np.nanmax(self.intensity.toarray())
-        self.intensity.min = np.nanmin(self.intensity.toarray())
 
         return self
```
chahank, line 266: The double loop in `def set_percentile_to_int(self, reference_intensity=None):` can be simplified.

climada/hazard/relative_cropyield.py
```diff
@@ -323,21 +328,24 @@ def plot_intensity_cp(self, event=None, dif=False, axis=None, **kwargs):
     def plot_time_series(self, years=None):
         """Plots a time series of intensities (a series of sub plots)
 
+        Optional Parameters:
+            event (int or str): event_id or event_name
```
chahank, line 332: There is no parameter event, but only years

chahank, line 332: There is no parameter event, but only years

climada/hazard/relative_cropyield.py
```diff
@@ -423,18 +431,19 @@ def plot_comparing_maps(self, his, fut, axes, nr_cli_models=1, model=1):
...
             [lon min, lat min, lon max, lat max]
         yearrange (int tuple): year range for hazard set, f.i. (2001, 2005)
-        returns (str): returned output
-        'filename_list': returns list of filenames only, else returns also list of data
+        return_data (str): returned output
+            False: returns list of filenames only
+            True: returns also list of data
 
     """
     filenames = [f for f in listdir(input_dir) if (isfile(join(input_dir, f))) if not
```
chahank, line 449: Simplify the loops for the filenames.

climada/hazard/relative_cropyield.py
```diff
@@ -495,7 +504,7 @@ def init_full_hazard_set(input_dir=INPUT_DIR, output_dir=OUTPUT_DIR, bbox=BBOX,
         # historic file
         crop_irr = (((items[i])[6]).split('-'))[1] + '-' + (((items[i])[6]).split('-'))[2]
 
-        cp_his = CropPotential()
+        cp_his = RelativeCropyield()
         cp_his.set_from_single_run(input_dir=input_dir, bbox=bbox, yearrange=yearrange,
                                    ag_model=(items[i])[0], cl_model=(items[i])[1],
                                    scenario=(items[i])[3], soc=(items[i])[4],
```
chahank, line 510: Is it possible to reduce the number of hard-coded list indexing (in all the lines below)? The hardcoded indices are completely obscure and are not easy to handle in future changes of the code.

climada/hazard/test/test_relative_cropyield.py
```diff
@@ -48,19 +56,23 @@ def test_load_EU_all(self):
...
+        self.assertEqual(haz.intensity_def, 'Relative Yield')
+
         self.assertAlmostEqual(np.max(hist_mean), 8.098413, places=5)
         self.assertEqual(haz.intensity.shape, (5, 1092))
-        self.assertAlmostEqual(np.nanmax(haz.intensity.toarray()), 4.0, places=5)
-        self.assertAlmostEqual(haz.intensity.max, 4.0, places=5)
-        self.assertAlmostEqual(haz.intensity.min, 0.0, places=5)
+        self.assertAlmostEqual(np.nanmax(haz.intensity.toarray()), 3.0, places=5)
+        self.assertAlmostEqual(haz.intensity.max(), 3.0, places=5)
+        self.assertAlmostEqual(haz.intensity.min(), -1.0, places=5)
```
chahank, line 74: Add tests for all the methods in the class RelativeCropyiel (except maybe the plot methods)

climada/hazard/relative_cropyield.py
chahank: An article on hardcoded index.

https://treyhunner.com/2018/03/tuple-unpacking-improves-python-code-readability/

climada/entity/exposures/crop_production.py
carmensteinmann: Do you have an idea for a better word? I just came up with element or component, but they seem so long..

chahank: Ideally, you should use a variable name that makes it clear what it is... In this case, what are the elements in file.split('_')? It looks like maybe they are crop names? Or irrigation types? So, something like crop_prop (for crop properties) maybe? Also, try to avoid the hard-coded indexes... it is completely unclear for someone reading the code what is contained in the variable item[0], item[1] and item[2].

carmensteinmann: Yes, it's the crop and irrigation type and the time period of the historical mean, that is used - so maybe histmean_prop? 

I'll try to think of an alternative to match the ISIMIP exposure files to the corresponding historical mean without using hard-coded indexes.. we used them a lot, also in the relative_cropyield hazard.

climada/entity/exposures/crop_production.py
carmensteinmann: I think it refers to the returns of the method "normalise_with_fao_cp" (as they are similar). However, they are not exactly the same, so I adapted this docstring.

climada/hazard/relative_cropyield.py
carmensteinmann: Would it make more sense to create a copy of self.intensity?

climada/entity/exposures/crop_production.py
sameberenz: maybe if we leave it hard coded but change it to be a dictionary, e.g. item['crop_model'], item['irrigation_type'], etc? that should be less confusing

climada/hazard/relative_cropyield.py
sameberenz: > Why first make a non-sparse matric and then convert it to a sparse matrix? This seems not very efficient...

which lines are you referring to exactly?

climada/entity/exposures/crop_production.py
chahank: This is probably suboptimal, as dictionaries are mutable.

sameberenz:
Regarding the [suggestions for the tutorial](https://github.com/CLIMADA-project/climada_python/pull/50#issuecomment-674123429), I have implemented most suggestions in [bb172bd](https://github.com/CLIMADA-project/climada_python/pull/50/commits/bb172bd41434cf3a212947f58163a3d76bc03f37).

Open question:
>     * In[14] why use the DATA_DIR from Climada to save computation outputs?
I haven't changed the data dir yet. Reason: DATA_DIR is a directory available for all users without customizing the code in the tutorial to a local path.
Which directory would you propose to set as an output directory, @chahank ?

chahank:
The use of DATA_DIR seems then reasonable for the tutorial.

sameberenz:
Hi @chahank,
thank you very much for all your suggestions!

We have implemented all changes requested by you with one exception:
_You asked why we do not initiate the hazard.intensity matrix as sparse matrix directly. We did not find a viable soluton for this and left the matrix initiation as it was. Reason: The comparison of each event's intensity matrix to hist_mean matrix is difficult to match to a sparse matrix not representing all centroids. With the first initiation of the intensity matrix as dense we don't have this problem._

Once you approve to the pull request, we are happy to do the merge. Please let us know!

best,
Carmen and Sam

climada/entity/exposures/crop_production.py
chahank: Was this point addressed? What do the default numbers refer to? Is this the whole world? The global South? The global North?

climada/entity/exposures/crop_production.py
```diff
@@ -105,42 +92,45 @@
...
 # by default the hist_mean files created by climada_python/hazard/crop_potential are saved in
 # climada_python/data/ISIMIP_crop/Output/hist_mean/
 HIST_MEAN_PATH = os.path.join(DATA_DIR, 'ISIMIP_crop', 'Output', 'Hist_mean')
 OUTPUT_DIR = os.path.join(DATA_DIR, 'ISIMIP_crop', 'Output')
 
 
-class CropyieldIsimip(Exposures):
+class CropProduction(Exposures):
     """Defines agriculture exposures from ISIMIP input data and
     FAO crop price data"""
```
chahank, line 107: Please add the class attributes in the docstring

climada/entity/exposures/crop_production.py
```diff
@@ -105,42 +92,45 @@
...
             yearrange (int tuple): year range for exposure set
                 f.i. (1990, 2010)
             scenario (string): climate change and socio economic scenario
-                f.i. 'histsoc' or 'rcp60soc'
-            cl_model (string): abbrev. climate model (only when landuse data
-            is future projection)
-                f.i. 'gfdl-esm2m' etc.
+                f.i. '1860soc', 'histsoc', '2005soc', 'rcp26soc','rcp60soc','2100rcp26soc'
+            cl_model (string): abbrev. climate model (only for future projections of lu data)
+                f.i. 'gfdl-esm2m', 'hadgem2-es', 'ipsl-cm5a-lr','miroc5'
```
chahank, line 133: The change from global variables to this docstring is very good. Could one be more precise here and not use 'f.i.' only? The examples given here refer to one particular dataset right?

climada/entity/exposures/crop_production.py
```diff
@@ -167,31 +160,30 @@ def set_from_single_run(self, input_dir=INPUT_DIR, filename=None, hist_mean=HIST
...
         else:
-            items = filename.split('_')
             if 'histsoc' or '1860soc' in filename:
-                cl_model = None
-                scenario = items[0]
+                scenario, _, _, _ = filename.split('_')
             else:
-                cl_model = items[0]
-                scenario = items[1]
+                scenario, _, _, _, _, _ = filename.split('_')
```
chahank, line 173: For this type of assigment, use `scenario, *rest = filename.split('_')`

climada/entity/exposures/crop_production.py
```diff
@@ -371,76 +367,86 @@ def set_to_usd(self, input_dir=INPUT_DIR, yearrange=YEARS_FAO):
...
                 can also be set to a single year
             crop (str): crop type
                 f.i. 'mai', 'ric', 'whe', 'soy'
+
+        Returns:
+            Exposure
         """
 
+        # the exposure in t/y is saved as 'value_tonnes'
+        self['value_tonnes'] = self['value']
```
chahank, line 382: why not 'tonnes_per_year'?

climada/entity/exposures/crop_production.py
```diff
@@ -371,76 +367,86 @@ def set_to_usd(self, input_dir=INPUT_DIR, yearrange=YEARS_FAO):
...
-            if country != 'No country':
-                if country == 'Other country':
-                    price = 0
-                else:
-                    idx_price = np.where((np.asarray(fao_country) == country) &
-                                         (np.asarray(fao_crops) == (CROP_NAME[self.crop])['fao']) &
-                                         (fao_year >= yearrange[0]) & (fao_year <= yearrange[1]))
-                    price = np.mean(fao_price[idx_price])
+        for country in list_countries:
+            idx_country = np.where(np.asarray(iso3alpha) == country)[0]
```
chahank, line 417: if the right-hand side is a list with a single element, you could also write
`[idx_country] = np.where(np.asarray(iso3alpha) == country)`

climada/entity/exposures/crop_production.py
```diff
@@ -466,9 +472,12 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
             [lon min, lat min, lon max, lat max]
         yearrange (array): year range for hazard set, f.i. (1976, 2005)
         unit (str): unit in which to return exposure (t/y or USD/y)
-        returns (str): returned output
-        'filename_list': returns list of filenames only, else returns also list of data
+        return_data (str): returned output
```
chahank, line 475: Isn't it a boolean then?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
+        histmean_prop = file.split('_')
+        crop_production = CropProduction()
+        crop_production.set_from_single_run(input_dir=input_dir, filename=filename,
+                                            hist_mean=hist_mean_dir, bbox=bbox,
+                                            yearrange=yearrange,
+                                            crop=((histmean_prop[2]).split('-'))[0],
+                                            irr=((histmean_prop[2]).split('-'))[1], unit=unit)
+        filename_list.append(('crop_production_' + ((histmean_prop[2]).split('-'))[0] + '-'
+                              + (((histmean_prop[2]).split('-'))[1]).split('.')[0] + '_'
+                              + str(yearrange[0]) + '-' + str(yearrange[1]) + '.hdf5'))
```
chahank, line 503: Can again here the code made be more readable by not using the hard indexing?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
+        output_list.append(crop_production)
+        crop_production.write_hdf5(os.path.join(output_dir, 'Exposure', filename_list[-1]))
```
chahank, line 505: I think it is better to above make a variable `filename` or `new_filename` or `expo_filename` or `filename_expo` which can then be appended to filenmae_list and used in the write_hdf5.

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
+    Parameters:
+        exp_firr (crop_production): exposure under full irrigation
+        exp_noirr (crop_production): exposure under no irrigation
+
+    Optional Parameters:
         input_dir (str): directory containing exposure input data
-        yearrange (array): the mean yield in this year range is used to normalize the exposure
-            data (default 2008-2018)
+        yearrange (array): the mean crop production in this year range is used to normalize
+            the exposure data (default 2008-2018)
```
chahank, line 523: Is the default yearrange chosen by you or related to the datasource?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
         unit (str): unit in which to return exposure (t/y or USD/y)
-        returns (str): returned output
-            'all': country list, ratio = FAO/ISIMIP, normalized exposures, yield per country
-            as documented by the FAO and calculated by the ISIMIP dataset
-            else: country list, ratio = FAO/ISIMIP, normalized exposures
-
+        return_data (str): returned output
+            True: returns country list, ratio = FAO/ISIMIP, normalized exposures, crop production
+            per country as documented by the FAO and calculated by the ISIMIP dataset
```
chahank, line 527: Is there somewhere a link to this FAO? What does FAO stand for?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
     """
 
-    # use the exposure in t/y to normalize with FAO yield values
-    if (exp_firr.value_unit == 'USD / y') and (exp_noirr.value_unit == 'USD / y'):
-        exp_firr.value = exp_firr.value_tonnes
-        exp_noirr.value = exp_noirr.value_tonnes
-    elif exp_firr.value_unit == 'USD / y':
+    # use the exposure in t/y to normalize with FAO crop production values
+    if exp_firr.value_unit == 'USD / y':
         exp_firr.value = exp_firr.value_tonnes
```
chahank, line 545: I am a bit confused. Wasn't `value_tonnes` used above for `t / y` ?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
         # if a country has no values in the exposure (e.g. Cyprus) the FAO average value is used
-        if exp_totyield[country] == 0:
-            exp_totyield[country] = fao_yield[country]
+        if exp_tot_production[country] == 0:
+            exp_tot_production[country] = fao_crop_production[country]
         # if a country has no fao value, the ratio is left being 1
-        elif fao_yield[country] != np.nan and fao_yield[country] != 0:
-            ratio[country] = fao_yield[country] / exp_totyield[country]
+        elif fao_crop_production[country] != np.nan and fao_crop_production[country] != 0:
+            ratio[country] = fao_crop_production[country] / exp_tot_production[country]
```
chahank, line 580: If the ratio should be one, why not set it exactly to one?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
-            data (default 2008-2018)
+        yearrange (array): the mean crop production in this year range is used to normalize
+            the exposure data (default 2008-2018)
         unit (str): unit in which to return exposure (t/y or USD/y)
-        returns (str): returned output
-            'all': lists containing data for each exposure file. Lists: crops, country list,
-            ratio = FAO/ISIMIP, normalized exposures, yield per country as documented by the
-            FAO and calculated by the ISIMIP dataset
-            else: lists containing data for each exposure file. Lists: crops, country list,
+        return_data (str): returned output
```
chahank, line 610: `return_data` should be a boolean no?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
-            countries, ratio, exp_firr2, exp_noirr2, fao_yield, \
-            exp_totyield = normalize_with_fao_cropyield(exp_firr, exp_noirr, input_dir=input_dir,
-                                                        yearrange=yearrange, unit=unit)
-            fao_yield_list.append(fao_yield)
-            exp_totyield_list.append(exp_totyield)
+    fao_cp_list = list()
+    exp_tot_cp_list = list()
+
+    for file_firr in filenames_firr:
+        str1, str2, crop_irr, years = file_firr.split('_')
```
chahank, line 643: What are `str1` and `str2` ?

climada/entity/exposures/crop_production.py
```diff
@@ -481,180 +490,201 @@ def init_full_exposure_set(input_dir=INPUT_DIR, filename=None, hist_mean_dir=HIS
...
-        filename_saveto = ('cropyield_isimip_' + ((item[2]).split('-'))[0] + '-'
-                           + (((item[2]).split('-'))[1]).split('.')[0] + '_'
-                           + str(yearrange[0]) + '-' + str(yearrange[1]) + '.hdf5')
-        filename_list.append(filename_saveto)
-        output_list.append(cropyield)
-        cropyield.write_hdf5(os.path.join(output_dir, 'Exposure', filename_saveto))
-
-    if returns == 'filename_list':
+    for file in filenames:
+        histmean_prop = file.split('_')
```
chahank, line 494: What does `histmean_prop` stand for?

climada/entity/exposures/crop_production.py
```diff
@@ -105,42 +92,45 @@
...
+                            bbox=BBOX, yearrange=(YEARCHUNKS['histsoc'])['yearrange'],
+                            cl_model=None, scenario='histsoc', crop=None, irr=None,
                             unit='USD', fn_str_var=FN_STR_VAR):
 
         """Wrapper to fill exposure from nc_dis file from ISIMIP
         Parameters:
             input_dir (string): path to input data directory
-            filename (string): name of the file to use
-            historic mean (array): historic mean yield per centroid
+            filename (string): name of the landuse data file to use,
```
chahank, line 123: General comment for the whole file. There seems to be quite a few methods that rely on a specific pattern for the filenames. This would imply that the current module only works for a very specific set of data files. This is suboptimal if in the future the module should be extended to use other data sources. It can work if there is a clear documentation on how to encode the data and how to name the files. Otherwise, the module is fixed for this one set of data files and should be documented accordingly.

climada/entity/exposures/test/test_crop_production.py
```diff
@@ -58,17 +58,18 @@ def test_set_to_usd(self):
         self.assertEqual(exp.value.shape, (1092,))
         self.assertEqual(exp.value_unit, 'USD / y')
         self.assertEqual(exp.crop, 'mai')
+        self.assertAlmostEqual(exp.value_tonnes[28], 1998.3634803238633)
         self.assertAlmostEqual(exp.value.max(), 51603897.28533253, places=5)
         self.assertAlmostEqual(exp.value.mean(), 907401.9933073953, places=5)
         self.assertEqual(exp.value.min(), 0.0)
         
     
     def test_set_to_usd_unnecessary(self):
```
chahank, line 67: Did you mean to delete test `test_set_to_usd_unnecessary` ? It looks sub-identical to `test_set_to_usd`

climada/entity/impact_funcs/relative_cropyield.py
```diff
@@ -14,30 +14,33 @@
...
-        self.name = 'Crop Potential ISIMIP'
+        self.name = 'Relative Cropyield ISIMIP'
         self.intensity_unit = ''
-        self.intensity = np.arange(0, 11)
-        self.mdr = (self.intensity - 1)
-        self.mdd = (self.intensity - 1)
+        #intensity = 0 when the crop production is equivalent to the historical mean
+        #intensity = -1 for a complete crop failure
+        #intensity = 1 for a crop production surplus of 100%
+        self.intensity = np.arange(-1, 10)
```
chahank, line 43: So the intensity range is from `-1 = failure`, to `10= 1000%` ? Any reason for the cap at `10`?

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+HAZ_TYPE = 'RC'
+"""Hazard type acronym for Relative Cropyield"""
+
+INT_DEF = 'Yearly Yield'
+
+BBOX = np.array([-180, -85, 180, 85])  # [Lon min, lat min, lon max, lat max]
+""""Default geographical bounding box of the global agricultural land extent"""
```
chahank, line 52: What do the coord. represents?

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+"""filename of ISIMIP output constant part"""
+
+
+class RelativeCropyield(Hazard):
+    """Agricultural climate risk: Relative Cropyield (relative to historical mean);
+    Each year corresponds to one hazard event;
+    Based on modelled crop yield, from ISIMIP (www.isimip.org, required input data).
+
+    Attributes:
+        crop_type (str): crop type (e.g. whe for wheat)
```
chahank, line 83: What are the acceptable values for `crop_type`?

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+        if pool:
+            self.pool = pool
+            LOGGER.info('Using %s CPUs.', self.pool.ncpus)
+        else:
+            self.pool = None
+
+        self.crop = ''
+        self.intensity_def = INT_DEF
+
+    def set_from_single_run(self, input_dir=None, filename=None, bbox=BBOX,
```
chahank, line 100: If this is specific to a particular file, the method name should reflect that.

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+                            soc=None, co2=None, crop=None, irr=None, fn_str_var=FN_STR_VAR):
+
+        """Wrapper to fill hazard from nc_dis file from ISIMIP
+        Parameters:
+            input_dir (string): path to input data directory
+            bbox (list of four floats): bounding box:
+                [lon min, lat min, lon max, lat max]
+            yearrange (int tuple): year range for hazard set, f.i. (1976, 2005)
+            ag_model (str): abbrev. agricultural model (only when input_dir is selected)
+                f.i. 'clm-crop', 'gepic','lpjml','pepic'
```
chahank, line 112: Same comment for `f.i.` as for the crop_production.py.

climada/entity/exposures/crop_production.py
carmensteinmann: It's chosen by us - should we specify something more about that in the docstring?

climada/entity/exposures/test/test_crop_production.py
chahank: I think the tests are a good start, but in the future a stronger coverage can be maybe obtained.

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+import cartopy
+import shapely.geometry
+from scipy import sparse
+import scipy.stats
+import h5py
+
+
+from climada.hazard.base import Hazard
+from climada.util import dates_times as dt
+from climada.util import coordinates
```
chahank, line 40: as coord

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+
+class RelativeCropyield(Hazard):
+    """Agricultural climate risk: Relative Cropyield (relative to historical mean);
+    Each year corresponds to one hazard event;
+    Based on modelled crop yield, from ISIMIP (www.isimip.org, required input data).
+
+    Attributes:
+        crop_type (str): crop type (e.g. whe for wheat)
+        intensity_def (str): intensity defined as:
+            'Yearly Yield' [t/(ha*y)], 'Relative Yield', or 'Percentile'
```
chahank, line 85: Please include all attributes in the docstring.

climada/entity/exposures/crop_production.py
carmensteinmann: Yes, in the very beginning of the file (line 83-90). We could maybe refer to that here?

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+    """
+
+    def __init__(self, pool=None):
+        """Empty constructor."""
+        Hazard.__init__(self, HAZ_TYPE)
+        if pool:
+            self.pool = pool
+            LOGGER.info('Using %s CPUs.', self.pool.ncpus)
```
chahank, line 93: Is pool used anywhere or is it for future use?

climada/entity/exposures/crop_production.py
carmensteinmann: Yes. So the unit of 'value_tonnes' is t/y and this is the exposure the normalisation has to be applied to (not to an exposure with a value in USD/y). So for the case that the exposure is given in USD/y, it is "reset" to t/y if that makes sense?

climada/hazard/relative_cropyield.py
sameberenz: It is not used currently.
Therefore I suggest that we replace

```
        if pool:
            self.pool = pool
            LOGGER.info('Using %s CPUs.', self.pool.ncpus)
        else:
            self.pool = None
```

With:
`self.pool = None`

climada/entity/exposures/crop_production.py
sameberenz: Precisely, maybe just make a small comment behind the line saying "temporary reset of value to t/y for normalization."

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+            event (int or str): event_id or event_name
+
+        Returns:
+            figure
+        """
+
+        if event is None:
+            event = self.event_name
+        else:
+            event = [str(n) for n in range(event[0], event[1] + 1)]
```
chahank, line 331: That is a quite obscure line of code. What is it suppose to do?

climada/entity/exposures/crop_production.py
sameberenz: yes just a short reference would be good guidance here

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+
+        self.centroids.set_meta_to_lat_lon()
+
+        len_lat = abs(self.centroids.lat[0] - self.centroids.lat[-1]) * (2.5 / 13.5)
+        len_lon = abs(self.centroids.lon[0] - self.centroids.lon[-1]) * (5 / 26)
```
chahank, line 336: What do the numbers represent?

climada/hazard/relative_cropyield.py
```diff
@@ -0,0 +1,710 @@
...
+
+    for his_file in his_file_list:
+        haz_his, filename, hist_mean = calc_his_haz(his_file, file_props, input_dir, bbox,
+                                                    yearrange_mean)
+        # save the historic mean depending on the crop-irrigation combination
+        # the idx keeps track of the row in which the hist_mean values are written per crop-irr to
+        # ensure that all files are assigned to the corresponding crop-irr combination
+        hist_mean_per_crop[(file_props[his_file])['crop_irr']]['value'][
+            hist_mean_per_crop[(file_props[his_file])['crop_irr']]['idx'], :] = hist_mean
+        hist_mean_per_crop[file_props[his_file]['crop_irr']]['idx'] += 1
```
chahank, line 479: Why add one to the mean?

sameberenz: It is not adding 1 to the mean but rather to a counting index 'idx' within that loop. this rather complex way forward ensures flexibilit for users of that method to put any combination of ISIMIP output files (netcdf) in a folder and let CLIMADA create all hazard sets from that. Users scared of outsourcing this complexity handling can create the hazard files one by one from the single netcdf files.

climada/test/test_cropyield_integr.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertAlmostEqual(exp.value[3], 0.0)
+        self.assertAlmostEqual(exp.value[1077], 405026.6857207429)
+        self.assertAlmostEqual(impact.imp_mat.data[3], -176102.5359452465 )
+        self.assertEqual(len(dif), 0)
+
+    def test_EU_nan(self):
+        """Test whether setting the zeros in exp.value to NaN changes the impact"""
+        bbox=[0, 42, 10, 52]
+        haz = RelativeCropyield()
+        haz.set_from_single_run(input_dir=INPUT_DIR, yearrange=(2001, 2005), bbox=bbox,
```
chahank, line 90: If the `haz.set_from_single_run` and `set_rel_yield_to_int` are expensive in computation time, look if you can avoid computing it separately for each test maybe. Again, please ask @emanuel-schmid for confirmation that this is a good idea.

climada/test/test_cropyield_integr.py
```diff
@@ -0,0 +1,165 @@
...
+        
+        # calculate mean hist_mean for each crop-irrigation
+        for crop_irr in crop_list:
+            mean = np.mean((hist_mean_per_crop[crop_irr])['value'], 0)
+            output_list.append(mean)
+        
+        self.assertEqual('whe-noirr', crop_list[0])
+        self.assertEqual(np.mean(hist_mean_per_crop['whe-noirr']['value']), np.mean(output_list[0]))
+        self.assertEqual(np.mean(hist_mean_per_crop['whe-noirr']['value'][:,1]), output_list[0][1])
+
```
chahank, line 160: Some methods seem to be missing a unit test such as `calc_fut_haz`

sameberenz: yes we excluded it because all parts of it are tested in the unit tests and we did not want to add more test files to the repositories

climada/test/test_cropyield_integr.py
sameberenz: it is not evry expensive especially because we use small test files. i would suggest to leave it as is.

climada/entity/exposures/crop_production.py
chahank: It is very confusing why if the unit is actually in USD/y, you just replace the value by value_tonnes. Is there somewhere a value_usd ?

climada/hazard/relative_cropyield.py
chahank: Actually, it is not a big deal and maybe should be kept for the future as it is now... It was just to know.

climada/hazard/relative_cropyield.py
chahank: Then you should use a loop with enumerate() to make it clear that you are just counting the elements in the loop, right?

climada/test/test_cropyield_integr.py
chahank: If it is very short fine for me!

climada/test/test_cropyield_integr.py
chahank: Fair enough. It is however a quite dangerous route to take as your test is now an interdependent function test.

climada/entity/exposures/crop_production.py
chahank: To be more precise, I find it confusing that you have both an attribute `value_tonnes` and an attribute `value_unit`. Here is my reasoning: either, you have some `value` and `value_unit` which allow you to store both 't/y' or 'usd/y', or then use `value_tonnes` and discard `value_unit`.

sameberenz: yes, i understand it is confusing. however, this is only a temporary thing happening within the normalization and never surfacing in the data. Carmen and me will figure out a way to make it less confusing for a reader of the code.

carmensteinmann: So our idea was that value_unit contains the unit of the exposure value (either t/y or USD/y if the method set_to_USD() has been applied).  If the latter is the case, the initial value (in t/y) gets overwritten, but we store it in 'tonnes_per_year'. Temporarily resetting the value in this function allowed to treat all exposures the same, regardless of whether set_to_USD() had been applied beforehand. In the end this reset is reverted to return the exposure in the unit it was given. Do you have an idea on how to make this better understandable?

climada/hazard/relative_cropyield.py
carmensteinmann: The problem is, that the loop is over the files in the given input directory and not over the different crop-irrigation combinations (and the mean is supposed to be saved per crop-irrigation combination). Because of the filename structure the crop-irr of the files actually alternate constantly (and for some model-combinations they only exist for certain crops). So this was a way to ensure that the historical mean is matched to the according crop-irr combination. I tried to explain it in the comment above those lines of code, but it's a bit confusing.

climada/entity/exposures/crop_production.py
chahank: I am sorry, but I really still do not understand what you are doing. So, the value of an exposure is defined in a certain unit, either t/y or USD/y. The unit value is stored in a variable `value_unit`. Hence, the variable for the value should not contain any reference to a value a be always in whatever value is given in `value_unit`. 

What does this sentence " If the latter is the case, the initial value (in t/y) gets overwritten, but we store it in 'tonnes_per_year'." mean? How can you convert the initial value into USD/y and then store it as t/y ?

If you have a function somewhere that requires all exposures to be in the same units, you could have for instance two variables, one `value_tonnes` and `value_in_unit`. The first is always set, and the second can be changed for whatever units you want. This has the disadvantage that the first might change without the second also being updated. But this could also be solved by making the second just a function of the first. However, this would required some computations everytime it is called...

Actually, the question is: what do you need the USD/y value for? Is it only for display?

sameberenz: Regarding your final question: we require exposure in USD/y for risk assessments interested in the monetary value of the impacted crops rather than just tonnes. This conversion is offered as a method in the exposure because crop prices are country specific (otherwise we could just convert the impact with a single factor per crop type).

Regarding your proposal:
What you would call `value_in_unit` is actually precisely our `value`. For CLIMADA to work with this exposure, it always needs a column named 'value' and one attribute 'value_unit'.

So now here we have to options for value_unit: 't/y' and 'USD/y'.

If value_unit='t/y', the numbers in 'value' represent crop production per grid cell in 't/y'.

If value_unit='USD/y', the numbers in 'value' represent crop production per grid cell in 'USD/y' AND there is an extra column named 'value_tonnes' storing the crop production per grid cell in 't/y'. Reason: this is required for normalization.

Now, in the normalization method, IF value_unit='USD/y', the column value is temporarily re-set to the values of 'value_tonnes' and then reconverted to 'USD/y' at the end of the method call. This is temporary and only within the method and we do it because we normalize crop production with the help of national crop production statistics that come in t/y and not in USD/y...

I hope that helps to clarify?

chahank: Yes, thanks a lot for your patience. Then maybe just rephrase the comment `# use the exposure in t/y to normalize with FAO crop production values` to make it clear that you temporarily reset the value and then put it back to USD/y. Maybe one could define a temp variable for this to be more clean, but not sure.

carmensteinmann: So value_unit contains only the unit and value contains the actual exposure. 
Initially every exposure is in t/y. Hence value contains the exposure values and value_unit is 't/y'.
However the user might want to use the exposure in USD/y and can then use the method set_to_USD. This first saves the initial value in 'tonnes_per_year' (I changed the name as you proposed but haven't pushed it yet) and then it overwrites value with the value in USD/y and sets value_unit to 'USD/y'. (I had messed up the order earlier.)
Now for the normalisation we need the exposure in t/y because we normalise it with cropprodution values of the FAO in t/y. So in case the exposure is now in USD/y we temporarily set it back to the initial t/y exposure (which is still saved in tonnes_per_year). 
The reason why we set the actual value in set_to_USD is that we can then use all Climada functions (such as plotting functions or impact calculation that always refer to exp.value). 
Sorry for the confusion, I hope I explained it a bit better now.

climada/entity/impact_funcs/relative_cropyield.py
carmensteinmann: Climada interpolates linearly after the last defined value, so we could also just define the impact function between -1 and 1. So the chosen value only defines e.g. the range plotted when visualising the impact function and therefore it might be good to keep a larger number than 1?

chahank: This should be specified somewhere. The range is fine, and you could recall that a linear interpolation after is used in CLIMADA impact function.

sameberenz: 👍 
it's good to have the range cover the common stretch of intensity values as the extrapolation is a bit slower. at least in CLIMADA on Matlab that was the case

PR #53, opened 2020-08-25
Feature: Centroids using distance to coast from NASA

mmyrte:
This goes out mostly to @tovogt - I changed `generate_nat_earth_centroids` so that one can generate a centroids instance that contains the resampled NASA distance to coast.

This would lead to a 7MiB file size for `data/system/NatEarth_Centroids_150as.hdf5`, which I'm not comitting, but I am relying on for my work. Just thought that if I can have that, others should, too. 

Should we, for flexibility's sake, remove the 360/150 arsec constraint and simply check that the resulting height/width are integers? In my opinion, it's sufficient to state in the docstring that the standard files are 360/150as.


climada/hazard/centroids/centr.py
```diff
@@ -821,13 +818,20 @@ def __deepcopy__(self, memo):
...
 
     Parameters:
         res_as (int): Resolution of file in arc-seconds. Default: 360.
+        path (str, optional): If set, write resulting hdf5 file here instead of
+            the default location.
+        dist_coast (bool): Should the distance to coast be read from a NASA
+            dataset? (see util.coordinates.dist_to_coast_nasa)
+            The 360as data include this distance by default, whereas the 150as
+            data by do not (thereby requiring the expensive computation to be
+            done locally).
```
tovogt, line 834: This is sentence does not accurately describe what this function does. It should be removed because it is misleading.

climada/hazard/centroids/centr.py
```diff
@@ -821,13 +818,20 @@ def __deepcopy__(self, memo):
...
-        files `NATEARTH_CENTROIDS_*AS`. These files are provided with CLIMADA
+        files in `NATEARTH_CENTROIDS`. These files are provided with CLIMADA
         so that this function should never be called!
 
     Parameters:
         res_as (int): Resolution of file in arc-seconds. Default: 360.
+        path (str, optional): If set, write resulting hdf5 file here instead of
+            the default location.
+        dist_coast (bool): Should the distance to coast be read from a NASA
+            dataset? (see util.coordinates.dist_to_coast_nasa)
```
tovogt, line 831: A parameter docstring shouldn't have the form of a question. Use instead: "If True, include distance to coast information in the output file. Default: False."

climada/hazard/centroids/centr.py
```diff
@@ -144,7 +142,7 @@ def equal(self, centr):
             and np.allclose(self.lon, centr.lon)
 
     @staticmethod
-    def from_base_grid(land=False, res_as=360):
+    def from_base_grid(land=False, res_as=360, base_file=None):
```
tovogt, line 145: Add docstring for new parameter.

climada/util/coordinates.py
```diff
@@ -413,9 +416,12 @@ def dist_to_coast_nasa(lat, lon, highres=False):
         os.remove(path_dwn)
         os.chdir(cwd)
 
-    intermediate_shape = None if highres else (3600, 1800)
+    if not highres and intermediate_shape is None:
+        intermediate_shape = (lat.size, lon.size)
```
tovogt, line 420: This fallback value doesn't make any sense. It will behave in an undesirable way for almost any possible lat/lon input.

climada/hazard/centroids/centr.py
```diff
@@ -821,13 +818,20 @@ def __deepcopy__(self, memo):
...
+        path (str, optional): If set, write resulting hdf5 file here instead of
+            the default location.
+        dist_coast (bool): Should the distance to coast be read from a NASA
+            dataset? (see util.coordinates.dist_to_coast_nasa)
+            The 360as data include this distance by default, whereas the 150as
+            data by do not (thereby requiring the expensive computation to be
+            done locally).
     """
     if res_as not in [150, 360]:
         raise ValueError("Only 150 and 360 arc-seconds are supported!")
```
tovogt, line 837: With the new `path` parameter, we only have to check if no `path` is given.

tovogt (commented):
Thanks, but note that I have just committed some major changes to `coordinates.py` and `centr.py` that affect your suggestions. Please have a look at the changes to adjust your PR accordingly.

climada/util/coordinates.py
mmyrte: You're right, of course. Should have been `np.unique(lat)`, if anything. Doesn't matter now.

tovogt:
Thanks, looks good!

PR #54, opened 2020-08-25
A netcdf writer for Impact class

mmyrte:
I added raster export capabilities to `Impact` - @emanuel-schmid is probably the one to have a look at this.

Note that I did try my darnedest to get CRS/projection information in there, but the xarray NetCDF adapter didn't allow me to write dimensionless variables. There are some small projects around on github that would support proper xarray-for-geodata writing, but the issue is minor enough to not get involved with those.


climada/engine/impact.py
```diff
@@ -882,9 +959,16 @@ def _exp_impact(self, exp_iimp, exposures, hazard, imp_fun, insure_flag):
 
     def _build_exp(self):
         eai_exp = Exposures()
+
+        # whacky bug: geom must be set first, otherwise NaNs are introduced
```
emanuel-schmid, line 963: wording: how is it whacky?
if it's a bug - shouldn't it be fixed instead?

climada/engine/impact.py
```diff
@@ -510,6 +514,79 @@ def write_sparse_csr(self, file_name):
         np.savez(file_name, data=self.imp_mat.data, indices=self.imp_mat.indices,
                  indptr=self.imp_mat.indptr, shape=self.imp_mat.shape)
 
+    def write_netcdf(self, path=None, attrs=None, with_imp_mat=True, **kwargs):
```
emanuel-schmid, line 517: To keep things simple I suggest to not use argument forwarding unless they always can be forwarded as a whole and untouched:
` rasterize(**kwargs)`
Otherwise one does not readily see what arguments are used.
In other words, I'd prefer to have the two additional arguments `event_id` and `drop_zero` explicit in the signature.

climada/engine/impact.py
```diff
@@ -510,6 +514,79 @@ def write_sparse_csr(self, file_name):
...
+
+        Parameters:
+            path (str, optional): Where to write the resulting netCDF to. If
+                None, the method stands in as a to_dataset
+            attrs (str, optional): Set as an attribute; Dataset.to_netcdf does
+                not handle dimensionless variables properly, so GDAL cannot
+                handle contained projection information. Still worthwhile for
+                human reference though.
+            with_imp_mat (bool): If true, read impact values from imp_mat
+                instead of eai_exp
```
emanuel-schmid, line 528: This is incomprehensible, especially without implementation knowledge but even when I look at the code it's not clear to me.

climada/engine/impact.py
```diff
@@ -510,6 +514,79 @@ def write_sparse_csr(self, file_name):
...
+        imp_ds = xr.DataArray(data, coords, dims, 'impact', attrs)
+
+        if path:
+            imp_ds.to_netcdf(path)
+
+        return imp_ds
+
+    def rasterize(self, event_id=None, meta=None, drop_zero=True):
+        """Rasterize this impact instance, returning a numpy ndarray with no
+        associated metadata. See to_dataarray for usage example.
```
emanuel-schmid, line 560: Couldn't this method be private?
What is `to_dataarray` referring to?
In any case, if documentation is "outsourced" the outer source must be linked to in the generated docs.

climada/engine/impact.py
```diff
@@ -510,6 +514,79 @@ def write_sparse_csr(self, file_name):
...
+
+        Parameters:
+            event_id (int, optional): If set, use imp_mat values for this event
+            meta (dict, optional): Should contain width, height, transform. If
+                not set, attempt to set it from self.exposures.meta
```
emanuel-schmid, line 565: I'd list the arguments explicitly rather than packing them in a dictionary.

climada/util/coordinates.py
```diff
@@ -932,6 +932,23 @@ def raster_to_meshgrid(transform, width, height):
     return np.meshgrid(np.arange(xmin + xres / 2, xmax, xres),
                        np.arange(ymin + yres / 2, ymax, yres))
 
+
+def unique_lat_lon_from_meta(meta):
```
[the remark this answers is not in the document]
mmyrte: `Centroids` already does some of these meta-to-lat-lon and vice versa conversions; I thought the meta construct was general enough to warrant this form. It'll make any calling code more verbose, but I understand your reasoning.

climada/util/coordinates.py
```diff
@@ -932,6 +932,23 @@ def raster_to_meshgrid(transform, width, height):
...
+    """Yields a tuple of unique latitudes and longitudes from a meta dict.
```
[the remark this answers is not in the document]
mmyrte: Did not know about yield; will change that. The `unique` stems from `np.unique` - how would you rename this?

climada/engine/impact.py
```diff
@@ -511,6 +514,79 @@ def write_sparse_csr(self, file_name):
         np.savez(file_name, data=self.imp_mat.data, indices=self.imp_mat.indices,
                  indptr=self.imp_mat.indptr, shape=self.imp_mat.shape)
 
+    def write_netcdf(self, path=None, attrs=None, with_imp_mat=True, **kwargs):
+        """Write this instance into a (rasterized) netcdf via a DataArray.
+
+        Parameters:
+            path (str, optional): Where to write the resulting netCDF to. If
```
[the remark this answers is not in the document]
mmyrte: I was thinking about implementing this, but thought that might be too obscure. Searching through the code for writers returns this method - coercion to DataArray, by contrast, is very unintuitive. Maybe it's sensible to place a reference to the possibility of writing to netcdf in the docstring.

climada/engine/impact.py
```diff
@@ -511,6 +514,79 @@ def write_sparse_csr(self, file_name):
...
+            coords['event_id'] = self.event_id
+            dims = ['event_id', 'latitude', 'longitude']
+        else:
+            data = self.rasterize(meta = self.exposures.meta, **kwargs)
+            dims = ['latitude', 'longitude']
+
+        if attrs is None:
+            attrs = {
+                'crs': 'EPSG:' + str(self.exposures.meta['crs'].to_epsg())
+            }
```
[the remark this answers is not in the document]
mmyrte: Will resolve this. Semi-related: I may have missed it - do we have a static typing convention yet?

climada/engine/impact.py
```diff
@@ -511,6 +514,79 @@ def write_sparse_csr(self, file_name):
...
                  indptr=self.imp_mat.indptr, shape=self.imp_mat.shape)
 
+    def write_netcdf(self, path=None, attrs=None, with_imp_mat=True, **kwargs):
+        """Write this instance into a (rasterized) netcdf via a DataArray.
+
+        Parameters:
+            path (str, optional): Where to write the resulting netCDF to. If
+                None, the method stands in as a to_dataset
+            attrs (str, optional): Set as an attribute; Dataset.to_netcdf does
+                not handle dimensionless variables properly, so GDAL cannot
```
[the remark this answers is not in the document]
mmyrte: Since I need to rename the method to `to_dataarray` or something like that anyways, I'll rework the docstring.

climada/engine/impact.py
mmyrte: I don't have the time to investigate this. It may be related to `Exposures` GDF inheritance, or it might even be a problem upstream. I could remove this, but it took me a whole while to find out, so documenting it like this seemed sensible. (And: whacky as in 'there is no discernible reason for this to happen'.)

climada/engine/impact.py
mmyrte: Yes, that is problematic. I'll translate it. To even have an imp_mat, one must set `save_mat` to true in Impact.calc, so I assumed that people would have knowledge. I suppose that argument will have to remain for now, but it could also be renamed for comprehension's sake.

climada/engine/impact.py
mmyrte: I did make it private at first, but thought that users might profit from the rasterization util. Will change that.

`to_dataarray` stems from a time where I did (as you had suggested) have a method that simply converted the instance to a dataarray. Sorry for my oversight.

How do you link to other docstrings? Is that in rst syntax?

climada/engine/impact.py
mmyrte: Will do.

climada/util/coordinates.py
```diff
@@ -932,6 +932,23 @@ def raster_to_meshgrid(transform, width, height):
...
+    """
+    width, height = meta['width'], meta['height']
+
+    min_lon, min_lat = meta['transform'] * (0,0)
+    max_lon, max_lat = meta['transform'] * (width, height)
+    
+    return (
+        np.linspace(min_lat, max_lat, height),
+        np.linspace(min_lon, max_lon, width),
+    )
```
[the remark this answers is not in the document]
mmyrte: This doesn't work. Multiplying with Affine matrices isn't commutative, and the linspace doesn't necessarily originate from 0, 0. The multiplication results in a tuple; unpacking it like I did is good for legibility. btw: Moving this chunk to Impact :)

climada/engine/impact.py
mmyrte: No, I've thought about this again. The meta dictionary format from rasterio is so pervasive, even in climada, that I'm sticking with this.

mmyrte:
I've addressed all your comments; thank you very much for helping! What do you think of it now?

emanuel-schmid:
The method `to_rasterized_dataarray` is not well integrated in the impact class.
The following code would fail:
```
imp = Impact()
imp.read_csv(impact_csv_file)
imp.to_rasterized_dataarray()
```
And imho it should pass.
It is not clear to me whether the impact class is the proper location for such a method.

PR #55, opened 2020-09-02
Feature: Centroids from GeoDataFrame

mmyrte:
@emanuel-schmid and @chahank 

This should be a relatively simple commit. I'm sorry about the superfluous commits; I had merged the `feature/impact_netcdf` branch, but reverted that. Please have a look at the diff instead. _If really necessary I can rework the history, i.e. delete the branch, rebranch, recommit._

The `climada/test/test_cent.py`  is not really an integration test, so in my opinion, it should be moved to `climada/hazard/centroids/test/`, do you agree?

### DIFF
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -34,8 +34,7 @@ from shapely.geometry.point import Point
 import climada.util.plot as u_plot
 from climada.util.constants import (DEF_CRS,
                                     ONE_LAT_KM,
-                                    NATEARTH_CENTROIDS_150AS,
-                                    NATEARTH_CENTROIDS_360AS)
+                                    NATEARTH_CENTROIDS)
 import climada.util.hdf5_handler as hdf5
 from climada.util.coordinates import (coord_on_land,
                                       dist_to_coast,
@@ -150,7 +149,7 @@ class Centroids():
             and np.allclose(self.lon, centr.lon)
 
     @staticmethod
-    def from_base_grid(land=False, res_as=360):
+    def from_base_grid(land=False, res_as=360, base_file=None):
         """Initialize from base grid data provided with CLIMADA
 
         Parameters:
@@ -158,13 +157,14 @@ class Centroids():
                 Default: False.
             res_as (int, optional): Base grid resolution in arc-seconds (one of
                 150, 360). Default: 360.
+            base_file (str, optional): If set, read this file instead of one
+                provided with climada.
         """
         centroids = Centroids()
-        base_file = NATEARTH_CENTROIDS_360AS
-        if res_as == 150:
-            base_file = NATEARTH_CENTROIDS_150AS
-        elif res_as != 360:
-            raise ValueError("No base grid for resolution %s", str(res_as))
+
+        if base_file is None:
+            base_file = NATEARTH_CENTROIDS[res_as]
+
         centroids.read_hdf5(base_file)
         if land:
             land_reg_ids = list(range(1, 1000))
@@ -840,15 +840,19 @@ class Centroids():
         return result
 
 
-def generate_nat_earth_centroids(res_as=360):
+def generate_nat_earth_centroids(res_as=360, path=None, dist_coast=False):
     """For reproducibility, this is the function that generates the centroids
-        files `NATEARTH_CENTROIDS_*AS`. These files are provided with CLIMADA
+        files in `NATEARTH_CENTROIDS`. These files are provided with CLIMADA
         so that this function should never be called!
 
     Parameters:
         res_as (int): Resolution of file in arc-seconds. Default: 360.
+        path (str, optional): If set, write resulting hdf5 file here instead of
+            the default location.
+        dist_coast (bool): If true, read distance from a NASA dataset
+            (see util.coordinates.dist_to_coast_nasa)
     """
-    if res_as not in [150, 360]:
+    if path is None and res_as not in [150, 360]:
         raise ValueError("Only 150 and 360 arc-seconds are supported!")
 
     res_deg = res_as / 3600
@@ -864,9 +868,10 @@ def generate_nat_earth_centroids(res_as=360):
     cen.lat = np.array([])
     cen.lon = np.array([])
 
-    path = NATEARTH_CENTROIDS_150AS
-    if res_as == 360:
-        path = NATEARTH_CENTROIDS_360AS
+    if path is None:
+        path = NATEARTH_CENTROIDS[res_as]
+
+    if dist_coast:
         cen.set_dist_coast(precomputed=True, signed=False)
         cen.dist_coast = np.float16(cen.dist_coast)
     cen.write_hdf5(path)
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -28,19 +28,25 @@ import h5py
 import pandas as pd
 from rasterio import Affine
 from rasterio.warp import Resampling
-from geopandas import GeoSeries
+import geopandas as gpd
 from shapely.geometry.point import Point
 
 import climada.util.plot as u_plot
-from climada.util.constants import DEF_CRS, ONE_LAT_KM, \
-                                   NATEARTH_CENTROIDS_150AS, \
-                                   NATEARTH_CENTROIDS_360AS
+from climada.util.constants import (DEF_CRS,
+                                    ONE_LAT_KM,
+                                    NATEARTH_CENTROIDS_150AS,
+                                    NATEARTH_CENTROIDS_360AS)
 import climada.util.hdf5_handler as hdf5
-from climada.util.coordinates import dist_to_coast, get_resolution, \
-                                     coord_on_land, pts_to_raster_meta, \
-                                     read_raster, read_vector, equal_crs, \
-                                     get_country_code, dist_to_coast_nasa, \
-                                     raster_to_meshgrid
+from climada.util.coordinates import (coord_on_land,
+                                      dist_to_coast,
+                                      dist_to_coast_nasa,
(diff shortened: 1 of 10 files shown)


climada/hazard/centroids/centr.py
```diff
@@ -172,6 +177,40 @@ def from_base_grid(land=False, res_as=360, base_file=None):
             centroids = centroids.select(reg_id=land_reg_ids)
         return centroids
 
+    @staticmethod
```
chahank, line 180: Why make it a @staticmethod?

climada/hazard/centroids/centr.py
```diff
@@ -172,6 +177,40 @@ def from_base_grid(land=False, res_as=360, base_file=None):
...
+
+        for col in gdf.columns:
+            if 'geom' in col:  # matches both 'geom' and 'geometry' columns
+                continue
+            val = gdf[col].to_numpy()
+            setattr(centroids, col, val)
+
+        if centroids.on_land.size == 0:
+            try:
+                centroids.on_land = ~np.isnan(centroids.region_id)
```
chahank, line 208: Consider whether `np.isfinite()` would be more appropriate than `~np.isnan()`

climada/test/test_centr.py
```diff
@@ -83,6 +85,21 @@ def test_base_grid(self):
...
+        )
+        gdf['elevation'] = np.random.rand(gdf.geometry.size)
+
+        centroids = Centroids.from_geodataframe(gdf)
+        centroids.check()
+
+        self.assertEqual(centroids.geometry.size, 45)
+        self.assertEqual(centroids.elevation.size, 45)
+        self.assertIsInstance(centroids.geometry.total_bounds, np.ndarray)
+
```
chahank, line 102: Maybe add a test for 
- when centrois.on_land.size == 0 (if this makes sense)
- Would it make sense to test the actual values in centroids.geometry and centroids.elevation instead of only there size?

climada/hazard/centroids/centr.py
mmyrte: This follows the `from_base_grid` example. I also find that accessing such a special constructor as `cent = Centroids.from_xy()` is intuitive to read...

climada/hazard/centroids/centr.py
mmyrte: I don't know how many people know that np.NaN is not finite - whereas the check for "not NaN" seems quite intuitive to me. I would leave it like this...

climada/test/test_centr.py
mmyrte: Since the test data does not set a region_id, this would have to take a synthetic region_id as input. I will write a test for the values in geometry, but not in elevation, since we're only copying a np.ndarray anyways; maybe checking the instance type would make more sense there.

climada/hazard/centroids/centr.py
chahank: I do not disagree. But, keeping the code homogeneous (i.e. not having different ways to initiate objects in CLIMADA for different types of objects) also has advantages, especially for not too python savvy people. Do you know if many other @Staticmethods except `from_base_grid` exist in CLIMADA?

climada/hazard/centroids/centr.py
chahank: But then any non-valid region_id that is not NaN will be used ... 

You can leave as is, I do not have a strong opinion.

climada/hazard/centroids/centr.py
mmyrte: -> Topic for coding convention meetings 🙂

climada/hazard/centroids/centr.py
mmyrte: I'll leave it as-is; this sort of problem should be prevented once we start type hinting or strong typing anyways.

climada/test/test_centr.py
mmyrte: Added a few test cases; closing...

climada/hazard/centroids/centr.py
```diff
@@ -172,6 +177,40 @@ def from_base_grid(land=False, res_as=360, base_file=None):
...
+                point features.
+        """
+        centroids = Centroids()
+
+        centroids.geometry = gdf.geometry
+        centroids.lat = gdf.geometry.y.to_numpy()
+        centroids.lon = gdf.geometry.x.to_numpy()
+
+        for col in gdf.columns:
+            if 'geom' in col:  # matches both 'geom' and 'geometry' columns
```
emanuel-schmid, line 201: if gdf itself has columns 'lat' or 'lon' the outcome will be unexpected. I'd rather exclude them in the loop.

emanuel-schmid, line 201: what's the point in not including any additional column that contains 'geom' in its name?

climada/hazard/centroids/centr.py
```diff
@@ -172,6 +177,40 @@ def from_base_grid(land=False, res_as=360, base_file=None):
...
+        >>> centroids = Centroids.from_geodataframe(gdf)
+
+        Parameters:
+            gdf (GeoDataFrame): Where the geometry column needs to consist of
+                point features.
+        """
+        centroids = Centroids()
+
+        centroids.geometry = gdf.geometry
+        centroids.lat = gdf.geometry.y.to_numpy()
```
emanuel-schmid, line 197: consider `to_numpy(copy=True)`

climada/hazard/centroids/centr.py
```diff
@@ -172,6 +177,40 @@ def from_base_grid(land=False, res_as=360, base_file=None):
...
+        centroids.lon = gdf.geometry.x.to_numpy()
+
+        for col in gdf.columns:
+            if 'geom' in col:  # matches both 'geom' and 'geometry' columns
+                continue
+            val = gdf[col].to_numpy()
+            setattr(centroids, col, val)
+
+        if centroids.on_land.size == 0:
```
emanuel-schmid, line 206: either add something in the description or skip this block

climada/test/test_centr.py
```diff
@@ -83,6 +85,24 @@ def test_base_grid(self):
...
+        centroids = Centroids.from_geodataframe(gdf)
+        centroids.check()
+
+        self.assertEqual(centroids.geometry.size, 45)
+        self.assertEqual(centroids.lon[0], 32.57)
+        self.assertEqual(centroids.lat[0], -25.95)
+        self.assertEqual(centroids.elevation.size, 45)
+        self.assertIsInstance(centroids.geometry, gpd.GeoSeries)
+        self.assertIsInstance(centroids.geometry.total_bounds, np.ndarray)
+
```
emanuel-schmid, line 105: the test does not handle any of the "special" columns ('geom', 'region_id', 'on_land')

emanuel-schmid:
> The `climada/test/test_cent.py` is not really an integration test, so in my opinion, it should be moved to `climada/hazard/centroids/test/`, do you agree?

Yes. That makes sense.

climada/hazard/centroids/centr.py
mmyrte: `geom` and `geometry` are synonymous in the "simple features" world of OGR; it's a convention. If you read a GDF from postgis for example, the geometry column is going to be called `geom`.

mmyrte: Good point, will implement.

climada/hazard/centroids/centr.py
mmyrte: AFAIK, the `copy=False` does not necessarily mean that the result is _not_ copied. I think I'll just trust the memory manager to decide sensibly. Maybe it even copies lazily as soon as the derived objects actually get manipulated?

climada/hazard/centroids/centr.py
mmyrte: Will add in the docstring.

climada/test/test_centr.py
mmyrte: Will think about some way to implement this. It'll have to be a synthetic region_id vector that includes some NaNs.

mmyrte: I didn't find a way to assert that the centroids.geom attribute does not exist, should I just use a try/except clause?

climada/hazard/centroids/centr.py
emanuel-schmid: well - then that should be reflected in line 196 `centroids.geometry = gdf.geom`
?

mmyrte: No, because a `geom` column that is set to be the GDF's geometry column can also be accessed through the `.geometry` property.

climada/hazard/centroids/centr.py
mmyrte: I haven't got the time to learn about such details of python memory management, but what you're describing sounds scary. I'm just going to put in the mandatory copy operation, even if it may be less efficient.

climada/hazard/centroids/centr.py
emanuel-schmid: I see. Then add some inline comment. And consider `col in ['geometry', 'geom']` instead of `'geom' in col`.

Or, since explicit is better then implicit (https://www.python.org/dev/peps/pep-0020/), and since the list of such names may be even bigger, add the name of the geometry column as an optional argument:
 `def from_geodataframe(gdf, geometry_col='geometry')`.

mmyrte: Thanks, can you have one last look and then hopefully merge?

mmyrte:
I reworded the docstring somewhat; it should be clear now to someone who knows geopandas a bit.

mmyrte:
Ok, I merged the branch with the added changes into develop; hope that's that.

PR #57, opened 2020-09-08
Add cfgrib to requirements

ThomasRoosli:
Add cfgrib from conda-forge to requirements for dealing with grib and bufr files. Also install ECMWFs eccodes.
Grip files are the standart file format for storing weather forecasts. The ECMWF produces eccodes to deal with this file format and suggest to use cfgrib in python (engine for xarray) to open grip files in python. That is why it is important for my work to somehow read grib files. Is cfgrib compatible with our environment? I was able to install it on windows with Anaconda without any problem.

### DIFF
--- a/requirements/env_climada.yml
+++ b/requirements/env_climada.yml
@@ -5,6 +5,7 @@ dependencies:
   - blas=1.0
   - bottleneck=1.3.2
   - cartopy=0.17.0
+  - conda-forge::cfgrib=0.9.7.7
   - curl=7.65.3
   - cython=0.29.7
   - dask=1.2.2


requirements/env_climada.yml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - blas=1.0
   - bottleneck=1.3.2
   - cartopy=0.17.0
+  - conda-forge::cfgrib=0.9.7.7
```
scem, line 8: just to keep in line with the other dependencies I'd move it to the `-pip` section and list `- cfgrib==0.9.7.7`
unless there is an issue with the pip installation

scem (changes requested):

requirements/env_climada.yml
ThomasRoosli: If I have understood the installation guide correctly (https://pypi.org/project/cfgrib/ : "alternatively, if you install the binary dependencies yourself, you can install the Python package from PyPI") The installation of cfgrib via pip only works if every user installs the binary dependencies (eccodes) themselves. Is this the prefered option?

Issue #56, opened 2020-09-07
Resources: Remove/replace large binary files


tovogt:
This is extracted from the discussion in #48 (sorry for retagging everyone). The lengthy part about large files became off topic, so I decided to move it here.

The repository size is currently around **~580 MiB** (using `git count-objects -vH`). Cloning this repository therefore takes considerable amounts of time. However, an analysis of large binary blobs in the git history suggests that we can easily save more than 80% of the space by rewriting git history (using https://github.com/newren/git-filter-repo).

### List of largest files in repository's history
For reference, the list of the 100 largest files in the repository's history (command line from https://stackoverflow.com/questions/10622179/how-to-find-identify-large-commits-in-git-history):
<details><summary>Click to expand the list of 100 largest files in the repository's history</summary>
<p>

```$ git rev-list --objects --all | git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' | sed -n 's/^blob //p' | sort --numeric-sort --key=2 | cut -c 1-12,41- | $(command -v gnumfmt || echo numfmt) --field=2 --to=iec-i --suffix=B --padding=7 --round=nearest | tail -100
Object hash   Size   Path
------------------------------------------------------------------------------------
afdf9d250244  1,5MiB data/supplychain/USA_services_XXX
7b0bde40f2d8  1,5MiB doc/tutorial/climada_entity_Exposures.ipynb
7b5cadab5cfa  1,6MiB doc/tutorial/climada_entity_Exposures.ipynb
c49fc7111e0d  1,6MiB doc/tutorial/climada_entity_Exposures.ipynb
71a04b0332ed  1,6MiB doc/tutorial/climada_entity_Exposures.ipynb
7771f3263d57  1,6MiB doc/tutorial/climada_entity_Exposures.ipynb
2f7e0ca73836  1,6MiB script/tutorial/1_main_climada.ipynb
607362854358  1,6MiB doc/tutorial/climada_entity_LitPop.ipynb
c90a0f3284d9  1,6MiB data/supplychain/FRA_agriculture_XXX
7a2a1295d6b6  1,6MiB doc/tutorial/climada_entity_LitPop.ipynb
358001e6d875  1,6MiB doc/tutorial/climada_entity_LitPop.ipynb
805f29c38ebc  1,6MiB doc/tutorial/climada_entity_LitPop.ipynb
7a9e5af40035  1,6MiB doc/tutorial/climada_entity_LitPop.ipynb
[...]

PR #59, opened 2020-09-09
Feature/isimip lowflow

sameberenz:
This feature branch improves the following module:  [climada/hazard/low_flow.py (diff)](https://github.com/CLIMADA-project/climada_python/compare/develop...feature/isimip_lowflow?expand=1#diff-dd6d9360dfc2a9f65c547f87fc22a33e) and the associated [unit test](https://github.com/CLIMADA-project/climada_python/compare/develop...feature/isimip_lowflow?expand=1#diff-0264ff7f62a1064262a41888d137fe47).

The development on the feature branch mainly expanded the unit tests and optimize the quite ressource intensive hazard generation to use less memory.

Furthermore, the demo file _data/demo/h08_gfdl-esm2m_ewembi_historical_histsoc_co2_dis_global_daily_DEMO_FR_2001_2005.nc_  is removed as requested in issue #56.

No other part of the repository is affected.

Please note that there is no tutorial for this module, yet. But it will be added in a future feature branch.


climada/hazard/low_flow.py
```diff
@@ -554,9 +692,18 @@ def _read_and_combine_nc(yearrange, input_dir, gh_model, cl_model,
     data.dis.values[data.dis.values < 0] = 0
     return data
 
-
 def _read_single_nc(filename, yearrange, bbox):
-    """import data from single nc file, return as xarray"""
+    """Import data from single nc file, return as xarray
+
+    Parameters:
+        filename (str or Path): full path of input netcdf file
```
emanuel-schmid, line 699: `filename (str or pathlib.Path)` ?

climada/hazard/low_flow.py
```diff
@@ -579,31 +726,109 @@ def _read_single_nc(filename, yearrange, bbox):
...
+        return data.mean(dim='time')
+    if fun[0] == 'p':
+        return data.reduce(np.nanpercentile, dim='time', q=percentile)
+    return None
+
+def _split_bbox(bbox, width=BBOX_WIDTH):
+    """split bounding box into squares, return new set of bounding boxes
+
+    Parameters:
+        bbox (list): geographical bounding box in the form:
```
emanuel-schmid, line 753: `(list of float)` ?

emanuel-schmid: since the size is four - why not a tuple?

climada/hazard/low_flow.py
```diff
@@ -579,31 +726,109 @@ def _read_single_nc(filename, yearrange, bbox):
...
+        [int(idc) for idc in np.arange(np.ceil(bbox[0]+width-1),
+                                       np.floor(bbox[2]-width+1), width)] + [bbox[2]]
+    lats = [bbox[1]] + \
+        [int(idc) for idc in np.arange(np.ceil(bbox[1]+width-1),
+                                       np.floor(bbox[3]-width+1), width)] + [bbox[3]]
+    bbox_list = list()
+    for ilon, _ in enumerate(lons[:-1]):
+        for ilat, _ in enumerate(lats[:-1]):
+            bbox_list.append([lons[ilon], lats[ilat], lons[ilon+1], lats[ilat+1]])
+    return bbox_list
```
emanuel-schmid, line 775: it's alright to get the 4 values as one argument, but within the method, they should have names.
`lon_min, lat_min, lon_max, lat_max = bbox`

climada/hazard/low_flow.py
```diff
@@ -629,26 +859,44 @@ def _days_below_threshold_per_month(data, threshold_grid, min_days_per_month, ke
...
+    Returns:
+        data (GeoDataFrame)."""
+
+    dataf = data.to_dataframe()
+    dataf.reset_index(inplace=True)
+    dataf = dataf.dropna()
+    dataf['iter_ev'] = np.ones(len(dataf), bool)
+    dataf['cons_id'] = np.zeros(len(dataf), int) - 1
+    # dataf['cluster_id'] = np.zeros(len(dataf), int)
+    # dataf['clus_id'] = np.zeros(len(dataf), int) - 1
```
emanuel-schmid, line 877: in or out?

climada/hazard/low_flow.py
```diff
@@ -203,70 +207,126 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
+                    self._intensity_one_cluster(tree_centr, cl_id,
+                                                res_centr, num_centr))
+            return sparse.csr_matrix(intensity_list)
+
+        for idx, stp in enumerate(stps[0:-1]):
+            intensity_list = []
+            if not idx:
+                stp_range = uni_ev[0:stps[1]]
+            else:
+                stp_range = uni_ev[stp:stps[idx+1]]
```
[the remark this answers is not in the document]
emanuel-schmid: why I prefer it that way: initializing an object just once and never change it beats initializing an object and afterwards manipulate it. If there is no good reason to mutate an object one shouldn't do it.

Apart from that it's nice when you have all the stuff you need defined at the same indentation level.

emanuel-schmid (commented):
Not a thorough review at all, just had a quick look and wanted to add a few comments.

climada/hazard/low_flow.py
```diff
@@ -203,70 +207,126 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
+            for cl_id in stp_range:
+                intensity_list.append(
+                    self._intensity_one_cluster(tree_centr, cl_id,
+                                                res_centr, num_centr))
+                if not idx:
+                    intensity_mat = sparse.csr_matrix(intensity_list)
+                else:
+                    intensity_mat = sparse.vstack((intensity_mat,
+                                               sparse.csr_matrix(intensity_list)))
```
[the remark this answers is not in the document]
sameberenz: Yes, Zélie put it inside the loop because this optimizes memory usage when run for large files (e.g., complete future 2006 to 2099)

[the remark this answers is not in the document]
sameberenz: c.f. commit https://github.com/CLIMADA-project/climada_python/commit/2461259bbf4b29825bb93ed93161829e2d9f7c7a

climada/hazard/low_flow.py
```diff
@@ -48,79 +47,67 @@
...
+YEARCHUNKS['rcp60'] = YEARCHUNKS['rcp26']
+"""future year chunks"""
 
-REFERENCE_YEARRANGE = [1971, 2005]
+REFERENCE_YEARRANGE = (1971, 2005)
+"""default year range used to compute threshold (base line reference)"""
 
-TARGET_YEARRANGE = [2001, 2005]
+TARGET_YEARRANGE = (2001, 2005)
+"""default year range of hazard"""
```
chahank, line 78: Which hazard?

sameberenz:
> Not a thorough review at all, just had a quick look and wanted to add a few comments.

Thanks, Emanuel! I have implemented all changes you suggested.

climada/hazard/low_flow.py
```diff
@@ -203,70 +203,119 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
             del haz_tmp
         if not keep_dis_data:
-            del self.data
+            self.data = None
         self.set_frequency(yearrange=yearrange)
         self.tag = TagHazard(haz_type=HAZ_TYPE, file_name=\
-                             FILENAME_NC % (gh_model, cl_model, "*", scenario, soc, \
-                                            fn_str_var, "*_*.nc"), \
+                             f'{gh_model}_{cl_model}_*_{scenario}_{soc}_{fn_str_var}_*.nc', \
                              description='yearrange: %i-%i (%s, %s), reference: %i-%i (%s, %s)' \
```
chahank, line 249: This part could probably also be updated to be more readable (line 237/249)

climada/hazard/low_flow.py
```diff
@@ -203,70 +203,119 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
                                  %(yearrange[0], yearrange[-1], scenario, soc, \
                                    yearrange_ref[0], yearrange_ref[-1], \
                                    scenario_ref, soc_ref)
                              )
 
+    def _intensity_loop(self, uni_ev, coord, res_centr, num_centr):
+        """Compute and intensity matrix. For each event, if more than one points of
```
chahank, line 256: "Compute and intensity matrix" is a confusing wording...

climada/hazard/low_flow.py
```diff
@@ -203,70 +203,119 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
+        data have the same coordinates, take the sum of days below threshold
+        of these points (duration as accumulated intensity).
+
+        Parameters:
+            uni_ev (list of str): list of unique cluster IDs
```
chahank, line 261: Sidenote: I think 'uniq' is clearer than 'uni'.

climada/hazard/low_flow.py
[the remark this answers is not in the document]
chahank: But wouldn't it be simpler to first create a tuple list of the stp_ranges, and then loop over these?

climada/hazard/low_flow.py
```diff
@@ -203,70 +203,119 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
+
+        Parameters:
+            uni_ev (list): list of unique cluster IDs
+        """
+        self.date = np.zeros(uni_ev.size, int)
+        self.date_start = np.zeros(uni_ev.size, int)
+        self.date_end = np.zeros(uni_ev.size, int)
+        for ev_idx, ev_id in enumerate(uni_ev):
+            # set event date to date of maximum intensity (ndays)
+            self.date[ev_idx] = self.data[self.data.cluster_id == ev_id]\
```
chahank, line 303: What happens if there is not matching ev_id for the ev_idx? Can this happen? Should maybe an error be raised then?

climada/hazard/low_flow.py
```diff
@@ -462,23 +553,30 @@ def _init_centroids(data_x, centr_res_factor=1):
 
 def unique_clusters(data):
     """identify unqiue clustes based on clusters in 3 dimensions and set unique
-    cluster_id"""
+    cluster_id
+
+    Parameters:
+        data (pandas.DataFrame): contains monthly gridded data of days below threshold
```
chahank, line 559: is data a good name here? data is very non-explicit. This comment probably applies to the whole file and to several methods.

climada/hazard/low_flow.py
```diff
@@ -571,39 +708,120 @@ def _read_single_nc(filename, yearrange, bbox):
...
+            lon_min, lat_min, lon_max, lat_max = bbox
+            data = data.sel(lon=slice(lon_min, lon_max), lat=slice(lat_max, lat_min),
                             time=slice(cftime.DatetimeNoLeap(yearrange[0], 1, 1),
                                        cftime.DatetimeNoLeap(yearrange[-1], 12, 31)))
         datetimeindex = data.indexes['time'].to_datetimeindex()
         data['time'] = datetimeindex
     return data
 
 
+def _xarray_reduce(data, fun=None, percentile=None):
```
chahank, line 720: Could this function be a future canditate for the climada.util module or is it too specific?

climada/hazard/low_flow.py
```diff
@@ -571,39 +708,120 @@ def _read_single_nc(filename, yearrange, bbox):
...
+        xarray
+    """
+    if fun == 'mean':
+        return data.mean(dim='time')
+    if fun[0] == 'p':
+        return data.reduce(np.nanpercentile, dim='time', q=percentile)
+    return None
+
+def _split_bbox(bbox, width=BBOX_WIDTH):
+    """split bounding box into squares, return new set of bounding boxes
```
chahank, line 741: Could this function be a candidate for climada.util in the future?

climada/hazard/low_flow.py
```diff
@@ -629,26 +852,42 @@ def _days_below_threshold_per_month(data, threshold_grid, min_days_per_month, ke
...
-    df.reset_index(inplace=True)
-    df = df.dropna()
-    df['iter_ev'] = np.ones(len(df), bool)
-    df['cons_id'] = np.zeros(len(df), int) - 1
-    # df['cluster_id'] = np.zeros(len(df), int)
-    # df['clus_id'] = np.zeros(len(df), int) - 1
-    df['dtime'] = df['time'].apply(lambda x: x.toordinal())
-    df['dt_month'] = df['time'].apply(lambda x: x.year * 12 + x.month)
-    return gpd.GeoDataFrame(df, geometry=[Point(x, y) for x, y in zip(df['lon'], df['lat'])])
+    """create GeoDataFrame from xarray with NaN values dropped
```
chahank, line 857: Candidate for climada.util?

climada/hazard/low_flow.py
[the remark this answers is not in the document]
sameberenz: Actually, you were right @emanuel-schmid! thanks :)
 I just checked with different values for `INTENSITY_STEP` and the if-else statement (`if not idx: intensity_mat = ....`) should be inside the outter for-loop, not the inner one. I have corrected for this and also took up the suggestion by @chahank in commit https://github.com/CLIMADA-project/climada_python/pull/59/commits/125abbb05f1bb8bca19e141347846c34305a71a4 .

climada/hazard/low_flow.py
sameberenz: will change comment to:
`"""arbitrary default, i.e. default year range of historical low flow hazard 2001-2005"""`

climada/hazard/low_flow.py
sameberenz: done: https://github.com/CLIMADA-project/climada_python/commit/9da58f7626072a15fa20ac1948af8d4d8ace8d19

climada/hazard/low_flow.py
sameberenz: yes, it is supposed to be "Compute intensity and populate intensity matrix."

climada/hazard/low_flow.py
sameberenz: This should not be possible, since `uni_ev` at this point only contains existing events, all filtering out of events happens earlier on.

climada/hazard/low_flow.py
sameberenz: you are right, i rename all `data` that are an DataFrame object to `data_df` and all `data` that are an xarray object to `data_xarray`

climada/hazard/low_flow.py
sameberenz: i guess it is a bit too specific

climada/hazard/low_flow.py
sameberenz: yes, but i will not ouch climada.util for this release

sameberenz: i will add a comment in the code

climada/hazard/low_flow.py
sameberenz: i will add a comment in the code

climada/hazard/low_flow.py
chahank: Hmmm not sure if this is really the optimal solution. What data is it? Data is a too generic name to mean anything. Is it houses? Hurricanes? apple juice flavors?

sameberenz: I called it data, because they contain several things.
As an xarray, it contains the variable "dis" + coordinates; which originally is "dis=discharge" as imported from the netcdf file. over time, it is however changing, so i could call it rather `dis_xarray` and later `days_below_thresh`? 

as a DataFrame, it contains various columns including days below threshold, date, lan, lot, cluster_ID, etc., basically all variables required to define and populate the hazard set. That's why I just called it `data`. Any suggestions how to call it different from `data_df`?

climada/hazard/low_flow.py
```diff
@@ -419,39 +478,73 @@ def _intensity_one_cluster(data, tree_centr, cluster_id, res_centr, num_centr):
...
+
+        lat_lon_uni, lat_lon_cpy = np.unique(temp_data[['lat', 'lon']].values,
+                                             return_inverse=True, axis=0)
+        index_uni = np.unique(lat_lon_cpy, axis=0)
         ind, _ = tree_centr.query_radius(lat_lon_uni, r=res_centr / 2, count_only=False,
                                          return_distance=True, sort_results=True)
         ind = np.array([ind_i[0] if ind_i.size else -1 for ind_i in ind])
         intensity_cl = _fill_intensity(num_centr, ind, index_uni, lat_lon_cpy,
                                        temp_data['ndays'].values)
         return intensity_cl
```
ThomasRoosli, line 527: I would leave two lines blank after the class definition

climada/hazard/low_flow.py
```diff
@@ -203,127 +203,189 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
-                intensity_list.append(
-                    self._intensity_one_cluster(self.data, tree_centr, cl_id,
-                                                res_centr, num_centr))
         self.tag = TagHazard(HAZ_TYPE)
         self.units = 'days'  # days below threshold
         self.centroids = centroids
 
         # Following values are defined for each event
-        self.event_id = np.sort(self.data.cluster_id.unique())
+        self.event_id = np.sort(self.data_df.cluster_id.unique())
```
ThomasRoosli, line 325: could you use uniq_ev? Should contain the same...

climada/hazard/low_flow.py
```diff
@@ -203,127 +203,189 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
-        self.intensity = sparse.lil_matrix(np.zeros((num_ev, num_centr)))
-        for idx, ev_intensity in enumerate(intensity_list):
-            self.intensity[idx] = ev_intensity
         self.intensity = self.intensity.tocsr()
         self.fraction = self.intensity.copy()
         self.fraction.data.fill(1.0)
 
     def identify_clusters(self, clus_thres_xy=None, clus_thresh_t=None, min_samples=None):
-        """call clustering and set events"""
+        """call clustering functions and set events in hazard
```
ThomasRoosli, line 342: As I understand you are not "setting events in hazard" but only identify the clusters inside the dataframe

climada/hazard/low_flow.py
```diff
@@ -203,127 +203,189 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
             centroids = centroids_import
         self.identify_clusters()
 
-        # sum "dis" (days per month below threshold) per pixel and cluster_id
-        # and write to hazard.intensiy
+        self.make_climada_hazard(centroids, min_intensity, min_number_cells,
+                                 yearrange, yearrange_ref, gh_model, cl_model,
+                                 scenario, scenario_ref, soc, soc_ref, fn_str_var, keep_dis_data)
+
+    def make_climada_hazard(self, centroids=None, min_intensity=1, min_number_cells=1,
```
ThomasRoosli, line 219: This function name could be more describtive "set_from_cluster"? or "make_hazard_from_cluster"?

climada/hazard/low_flow.py
```diff
@@ -357,11 +417,11 @@ def _df_clustering(data, cluster_vars, res_data, clus_thres_xy,
             cluster_id = cluster_id[x_y_cpy]
             data_iter[clus_id_var].values[temp] = cluster_id + data_iter[clus_id_var].max() + 1
 
-        data[clus_id_var].values[data['iter_ev'].values] = data_iter[clus_id_var].values
-        return data
+        data_df[clus_id_var].values[data_df['iter_ev'].values] = data_iter[clus_id_var].values
+        return data_df
 
     def filter_events(self, min_intensity=1, min_number_cells=1):
```
ThomasRoosli, line 423: Why use haz_tmp.date to select events to filter out? What if someone is interested in events in November of the year 5476? Why not create an array of True with the right dimension, set them to false if you want to filter them out and select the events by using the array of bools to index event_names.

ThomasRoosli (commented):
I read through the class LowFlow and made some suggestions, none of them are any kind of blocker, so fix what you agree with and disregard the rest.
To really review the clustering algorithms (meaning LowFlow class in full, not only this Pull Request), I would need more time and a working tutorial with data...

Further comment _centroids_resolution is a candidate for the Centroids class

climada/hazard/low_flow.py
chahank: Well, the data still represents something uniform, right? In your case, you say it is the hazard, which here is the low water flow? Then the name of the data variable should imo. reflect that. For instance, data_haz, or flow_data, or, df_haz, ... Would you agree? On another note, I think it is suboptimal if the content of a variable changes in 'spirit' during the code (inside the same method/class) without it being clear from other variable properties.

climada/hazard/low_flow.py
```diff
@@ -182,19 +175,26 @@ def set_from_nc(self, input_dir=None, centroids=None, countries=[], reg=None,
...
+                Threshold below which the grid is masked out. e.g.:
+                ('mean', 1.) --> grid cells with a mean discharge below 1 are ignored
+                ('percentile', .3) --> grid cells with a value of the computed percentile discharge
+                values below 0.3 are ignored. default: ('mean', 1}). Set to None for
+                no threshold.
+                Provide a list of tuples for multiple thresholds.
         raises:
             NameError
         """
         print('GETTING STARTED!')
```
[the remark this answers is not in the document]
chahank: Is it really necessary to have a 'GETTING STARTED!' ?

climada/hazard/low_flow.py
sameberenz: that's a good suggestion, will be changed to `set_intensity_from_clusters`

climada/hazard/low_flow.py
sameberenz: `hazard.select()` has limited input parameters that are possible, but I found a better way to filter using `self.orig` instead of `self.date`.

Issue #62, opened 2020-09-17
read_ibtracs_netcdf provides tracks with only missing values if estimate_missing is True


bguillod:
On the develop branch I found the following issue:

Calling read_ibtracs_netcdf with estimate_missing set to True return tracks with misleading data when both pressure and wind speed data are missing.

For instance:
```
all_tracks = TCTracks()
all_tracks.read_ibtracs_netcdf(provider='usa', basin='SI', year_range=(1850, 1851))
```
correctly issues a warning:
```
climada.hazard.tc_tracks - WARNING - No valid wind/pressure values found for 1851080S15062, 1851080S15063, 1851080S21060, 1851080S21420, 1851305S08089.
```

However:
```
all_tracks = TCTracks()
all_tracks.read_ibtracs_netcdf(provider='usa', basin='SI', year_range=(1850, 1851), estimate_missing=True)
```
does not issue any warning, but the tracks that are loaded contain data that don't look correct:
- max_sustained_wind and central_pressure are all -1
- radius_max_wind and radius_oci are all 0.

I think the issue is that _estimate_pressure and _estimate_vmax convert all NaNs to -1 and then the line

```
ibtracs_ds['valid_t'] &= ibtracs_ds.wind.notnull() & ibtracs_ds.pres.notnull()
```

does not filter those out.

Is this intended and if so what will be the impact on the wind field?
I would suggest that even if estimate_missing is True, tracks without any pressure or wind information MUST be skipped, but I might be wrong.

chahank:
Thanks @bguillod  for raising this issue! Could you please add labels to the issue?

bguillod:
@chahank I would like to but I am not able to. How should I do that? Or perhaps I can't because I'm not a member?

bguillod:
I guess best placed to answer is @tovogt

bguillod:
@tovogt Thanks! Currently I observe that if these tracks are kept, the intensity and all variables are empty in the TropCyclone hazard for that event. This leads to bugs in the plotting though. I suggest that when neither vmax nor pres is available then the track is entirely skipped, but I leave it up to you to decide.

My fix would be:
last line of _estimate_pressure, add:
```
cen_pres = np.where(cen_pres == -1, np.nan, cen_pres)
```
last line of _estimate_vmax, add:
```
v_max = np.where(v_max == -1, np.nan, v_max)
```
Hope this helps.

bguillod:
[...]

PR #63, opened 2020-09-17
Don't replace NaNs by -1 when estimating TC track parameters

tovogt:
This is a fix for #62. I took the opportunity to add docstrings and unit tests for all the parameter estimation functions.

I would suggest @bguillod as a reviewer because he's already familiar with the problem, but somehow I can't select him (?).

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -956,42 +956,104 @@ def _dist_since_lf(track):
     return dist_since_lf
 
 def _estimate_pressure(cen_pres, lat, lon, v_max):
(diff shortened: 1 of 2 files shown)


bguillod:
@tovogt Looks good. Perhaps I can't be reviewer because I'm not a member? If so, should I become one and what are the implications/duties associated with that?

chahank:
Not being a member might be the issue. We will discuss what it implies on our next developers' meeting and come back to you. Until then, @bguillod is it possible for you to participate in the review (i.e., read the commits and comment on them)?

PR #65, opened 2020-09-18
Feature/fix rasterio read netcdf

emanuel-schmid:


requirements/env_climada.yml
```diff
@@ -1,60 +1,45 @@
-name: climada_env
+name: rasterio
```
tovogt, line 1: This is by mistake, right?

emanuel-schmid:
```suggestion
name: climada_env
```

requirements/env_climada.yml
```diff
@@ -1,60 +1,45 @@
...
 channels:
+  - conda-forge
```
tovogt, line 3: Adding `conda-forge` to the `channels` list means in practice that almost all packages are installed from `conda-forge` instead of the default channel. Is this our intention? There is only a single package that requires `conda-forge`, namely `cfgrib` and that does not require this line since it comes with an explicit `conda-forge::` prefix.

emanuel-schmid: Agreed. But the channels configuration remains strange to me, for without explicit `conda-forge::` prefix, the cfgrib package cannot even be found.
```suggestion
```

setup.py
```diff
@@ -55,51 +55,43 @@ def package_files(directory):
...
+        'netcdf4==1.5.4',
+        'numba==0.51.2',
+        'numpy==1.19.1',
+        'matplotlib==3.2.2',
+        'matplotlib-base==3.2.2',
+        'pandas==1.0.5',
+        'pandas-datareader==0.8.1',
+        'pillow==7.2.0',
+        'pint==0.15',
+        'conda-forge::proj==7.0.0',
```
tovogt, line 78: For pip, this is not a valid project/version requirement specifier. But `proj` isn't a python package anyway. My suggestion is to simply remove this line.
```suggestion
```

setup.py
```diff
@@ -55,51 +55,43 @@ def package_files(directory):
...
-        'six==1.13.0', #
+        'h5py==2.10.0',
+        'haversine==2.3.0',
+        'nbconvert==5.6.1',
+        'nbformat==5.0.7',
+        'netcdf4==1.5.4',
+        'numba==0.51.2',
+        'numpy==1.19.1',
+        'matplotlib==3.2.2',
+        'matplotlib-base==3.2.2',
```
tovogt, line 73: There is no pip package of this name.
```suggestion
```

setup.py
```diff
@@ -55,51 +55,43 @@ def package_files(directory):
...
+        'pandas==1.0.5',
+        'pandas-datareader==0.8.1',
+        'pillow==7.2.0',
+        'pint==0.15',
+        'conda-forge::proj==7.0.0',
+        'pytables==3.6.1',
```
tovogt, line 79: In pip, this package is called `tables`:
```suggestion
        'tables==3.6.1',
```

tovogt (changes requested):
There are still some problems with the `setup.py`

tovogt:
In your latest commit you introduced new problems...

setup.py
emanuel-schmid: agreed

setup.py
emanuel-schmid: agreed

setup.py
emanuel-schmid: agreed

emanuel-schmid:
The setup.py file is still under construction. Currently climada is not yet installable through pip. Installation procedure is 
- setup conda environment
- import climada from local path

Hence `setup.py` has no effect for the default user. However, since it's present, it should also work.
To be fixed in the develop branch.

Issue #64, opened 2020-09-18
Read lon/lat from provider in read_ibtracs_netcdf instead of combined lon/lat


bguillod:
Currently in read_ibtracs_netcdf (tc_tracks.py) latitude and longitude are always the values created by IBTrACS (netCDF variables 'lat' and 'lon') by combining values from various providers. I suggest that, when 'provider' is given as an input, those values are instead read from the desired provider (ibtracs_ds[f'{provider}_lat'] and ibtracs_ds[f'{provider}_lon']). On the other hand, if provider is not given then I think it is fine to read from the combined IBTrACS data.

PR #69, opened 2020-09-22
TCTracks: read ibtracs lon/lat from specified provider

tovogt:
As suggested by @bguillod in #64, this PR makes sure that, in the IBTrACS reader routine, track coordinates are taken from the user-specified provider if applicable. When no provider is specified, the procedure remains unchanged. The PR comes with unit tests that verify this functionality.


mmyrte:
Nice, got to know f-strings through this! Looks good as far as I can tell, but of course, ibtracs is not at all my thing 🙂

bguillod:
Looks good to me, too.

PR #71, opened 2020-09-24
Hazard.select: accept general array-likes as date parameter

tovogt:
I ran into a problem where I wanted to filter hazard events by date and my filtering criterion was silently (!) ignored. This is because the `date` parameter in `Hazard.select` was ignored unless it was a `tuple` object. Passing a list `[date_ini, date_end]` did not work!

This pull request changes this behaviour and adds unit tests to make sure that valid inputs aren't ignored and invalid inputs raise exceptions. Furthermore, it updates the docstring of `Hazard.select`.

All unit tests on Jenkins are successful (see https://ied-wcr-jenkins.ethz.ch/job/climada_branches/job/feature%252Fhazard_select_date_invalid/).

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -527,23 +527,29 @@ class Hazard():
             raise var_err
 
     def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_frequency=False):
-        """Select events within provided date and/or (historical or synthetical)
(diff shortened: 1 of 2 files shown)


simonameiler (approved):
easily readable, clearly explained docstrings, neat as always.

tovogt:
Thanks! I think for such a small change one approval is enough.

PR #78, opened 2020-10-20
New random walk for TC tracks

bguillod:
This incorporates the new approach for the application of random walk on TC tracks. Please note:
- Static code analysis only approximately dealt with, but everything runs fine.
- I wasn't able to run the integration tests (`make integ_test`) since I get an error when I run this on my machine.
- I would propose to remove the `@jit(parallel=True)` label on `_one_rnd_walk` since it causes some warnings. This was already the case before I modified the code though.


bguillod:
Note I've fixed issue 77 as an additional commit.

climada/hazard/tc_tracks.py
```diff
@@ -666,9 +666,9 @@ def equal_timestep(self, time_step_h=1, land_params=False):
                                                       land_geom))
             self.data = new_data
 
-    def calc_random_walk(self, **kwargs):
+    def calc_random_walk(self, ens_size=9, **kwargs):
```
chahank, line 669: I would rename `ens_size` to `nb_synt_tracks` or analog to be clearer about what it stands for.

bguillod: Fine with me (although I find ens_size pretty clear: the size of the generated ensemble, common in climate science but certainly less in other fields), but that may imply breaking some earlier code such as:
`tc_track.calc_random_walk(ens_size=9)`
At the same time since all other input parameters changed this might not be an issue (perhaps even good so users realize the methodology changed), so I'm happy to change to `nb_synth_tracks`.

carmensteinmann (approved):

climada/hazard/tc_tracks.py
chahank: Very good. Any reason why to put default to 9? Maybe there should be no default number of tracks?

climada/hazard/tc_tracks_synth.py
```diff
@@ -137,25 +162,40 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
-        d_xy = np.sign(track.lat.values[0]) * d_xy
-
-        d_lat_lon = d_xy + np.expand_dims(xy_ini[:, i_ens], axis=1)
-
-        i_track.lon.values = i_track.lon.values + d_lat_lon[0, :]
-        i_track.lat.values = i_track.lat.values + d_lat_lon[1, :]
+        # angular perturbation
+        i_start_ang = 2 * ens_size + i_ens * n_seg
+        i_in_ang = (i_start_ang, i_start_ang + track.time.size - 1)
+        ang_pert = dt * np.degrees(max_ddir * (2 * rnd_vec[i_in_ang[0]:i_in_ang[1]] - 1))
```
chahank, line 176: Does it really make sense to multiply the angular perturbation by dt?

climada/hazard/tc_tracks_synth.py
```diff
@@ -137,25 +162,40 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+        new_lon[0] = i_track.lon.values[0] + xy_ini[0, i_ens]
+        new_lat[0] = i_track.lat.values[0] + xy_ini[1, i_ens]
+        for i in range(0, len(new_lon) - 1):
+            new_lon[i + 1], new_lat[i + 1] = \
+                _get_destination_points(new_lon[i], new_lat[i],
+                                        bearings[i] + ang_pert_cum[i],
+                                        trans_pert[i] * angular_dist[i])
+
+        i_track.lon.values = new_lon
+        i_track.lat.values = new_lat
```
chahank, line 198: This part of the code is very hard to read. Many obscure variables and hard-coded indices.

climada/hazard/tc_tracks_synth.py
chahank, line 176: Why define i_in_ang as tuple and after unpack it immediataly? Same in line 183.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+        lat (numpy.ndarray of shape (n,)): latitude coordinates, in decimal degrees
+
+    Returns:
+        bearing_angle (numpy.ndarray of shape (n-1,), in decimal degrees
+    """
+    lon, lat = map(np.radians, [lon, lat])
+    lat_1 = lat[:-1]
+    lon_1 = lon[:-1]
+    lat_2 = lat[1:]
+    lon_2 = lon[1:]
```
chahank, line 311:
```suggestion
    lat_2, *_, lat1 = lat
    lon_2, *_, lon_1 = lon
```

Not sure why lat/lon_1 are the last elements and lat/lon_2 the first

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+
+    Returns:
+        Angular distance (numpy.ndarray of shape (n-1,), in decimal degrees
+    """
+
+    lon, lat = map(np.radians, [lon, lat])
+    lat_1 = lat[:-1]
+    lat_2 = lat[1:]
+    delta_lon = lon[1:] - lon[:-1]
+    delta_lat = lat_2 - lat_1
```
chahank, line 341: c.f.308-311

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+                               np.cos(lat_1) * np.sin(lat_2) - \
+                               np.sin(lat_1) * np.cos(lat_2) * np.cos(delta_lon))
+    return np.degrees(earth_ang_fix)
+
+
+@jit
+def _get_angular_distance(lon, lat):
+    """Compute the angular distance of great circle paths defined by consecutive points
+
+    Uses the haversine formulat to get the great circle distance between each pair
```
chahank, line 326: formula

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+        Angular distance (numpy.ndarray of shape (n-1,), in decimal degrees
+    """
+
+    lon, lat = map(np.radians, [lon, lat])
+    lat_1 = lat[:-1]
+    lat_2 = lat[1:]
+    delta_lon = lon[1:] - lon[:-1]
+    delta_lat = lat_2 - lat_1
+    a = np.sin(delta_lat / 2.0) ** 2 + np.cos(lat_1) * np.cos(lat_2) * np.sin(delta_lon / 2.0) ** 2
+    c = np.degrees(2 * np.arctan2(np.sqrt(a), np.sqrt(1 - a)))
```
chahank, line 343: Is this equivalent to the formula from here: https://en.wikipedia.org/wiki/Haversine_formula ?

Why to you assume circles of radius 1?

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+def _h_ac(x, y, theta):
+    # https://stats.stackexchange.com/questions/48086/
+    #   algorithm-to-produce-autocorrelated-uniformly-distributed-number
+    # Note that the definition of Gamma is not very clear there, but the
+    # formulation below does what the text says.
+    # Tests indicated that this works, while the following fails:
+    # gamma = np.abs(np.mod(np.mod(theta, np.pi), np.pi / 2) - np.pi / 4)
+    gamma = np.abs(np.mod(theta, np.pi) - \
+                   np.floor((np.mod(theta, np.pi) / (np.pi / 2)) + 0.5) * np.pi / 2)
+    # np.abs(dat['theta2'] - np.floor((dat['theta2'] / (np.pi / 2)) + 0.5) * np.pi / 2)
```
chahank, line 253: Why commented out?

chahank (changes requested):
Overall, nice module! I would like to have a look at the main part again once the changes have been made.

climada/hazard/test/test_tc_tracks_synth.py
```diff
@@ -356,6 +356,27 @@ def test_wrong_decay_pass(self):
         self.assertTrue(np.array_equal(cp_ref, track_res.central_pressure[9:11]))
```
chahank, line 357: It would be useful to also have unit test for new methods _get_destinations_points, _get_angular_distance, _get_bearing_angle, _random_uniform_ac, _f_ac_, _h_ac, ...

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
     return ens_track
 
 
+def _random_uniform_ac(n_ts, autocorr, time_step_h):
+    """Generate a series of autocorrelated uniformly distributed random numbers.
+
+    This implements the algorithm described here to derive a uniformly distributed
+    series with specified autocorrelation (here at a lag of 1 hour):
+    https://stats.stackexchange.com/questions/48086/
+        algorithm-to-produce-autocorrelated-uniformly-distributed-number
```
emanuel-schmid, line 215: Why was this algorithm preferred to the one from the there mentioned paper [A method to generate autocorrelated uniform random numbers](https://www.tandfonline.com/doi/abs/10.1080/00949659308811469)?
As I understand it, it's only an approximation while the paper's method is seemingly a correct solution and looks like it would run much faster.
Apart from that, it has the benefit, that the reference to a paper is there to stay and its content cannot be removed or edited at any time in the future.

emanuel-schmid (changes requested):

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+        lon (numpy.ndarray of shape (n,)): longitude coordinates of start location,
+            in decimal degrees
+        lat (numpy.ndarray of shape (n,)): latitude coordinates of start location,
+            in decimal degrees
+        bearing (numpy.ndarray of shape (n,)): bearing (direction Northward, clockwise)
+        angular_distance (numpy.ndarray of shape (n,)): angular distance, in decimal degrees
+
+    Returns:
+        Angular distance (numpy.ndarray of shape (n,), in decimal degrees
+    """
```
tovogt, line 361: According to [CLIMADA's coding conventions](https://github.com/CLIMADA-project/climada_python/blob/develop/doc/guide/coding_conventions.rst#doc-strings), docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+
```
tovogt, line 362: "There's no blank line either before or after the docstring." ([PEP 257](https://www.python.org/dev/peps/pep-0257/))

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    Parameters:
+        lon (numpy.ndarray of shape (n,)): longitude coordinates of start location,
+            in decimal degrees
+        lat (numpy.ndarray of shape (n,)): latitude coordinates of start location,
+            in decimal degrees
+        bearing (numpy.ndarray of shape (n,)): bearing (direction Northward, clockwise)
+        angular_distance (numpy.ndarray of shape (n,)): angular distance, in decimal degrees
+
+    Returns:
+        Angular distance (numpy.ndarray of shape (n,), in decimal degrees
```
tovogt, line 360: Something is wrong with the docstring. The documented return value should be an ndarray, but the function returns a pair of ndarrays. Furthermore, the "Returns" part of the docstring says it returns an angular distance while the one-line-summary of the docstring says it returns coordinates of endpoints.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    # that case because angular distance will be 0.
+    earth_ang_fix = np.arctan2(np.sin(delta_lon) * np.cos(lat_2),
+                               np.cos(lat_1) * np.sin(lat_2) - \
+                               np.sin(lat_1) * np.cos(lat_2) * np.cos(delta_lon))
+    return np.degrees(earth_ang_fix)
+
+
+@jit
+def _get_angular_distance(lon, lat):
+    """Compute the angular distance of great circle paths defined by consecutive points
```
tovogt, line 324: There is a function `dist_approx` in `climada.util.coordinates` that computes approximate distances of points on earth. With the parameter `method="geosphere"`, it uses the Haversine formula. I suggest that you use that function instead of reimplementing the exact same functionality here.

climada/hazard/tc_tracks_synth.py
tovogt: Caution, the suggested change is not the same as the original code, as `lat_1` is supposed to be a slice of `lat` instead of a single element!

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    Uses the haversine formulat to get the great circle distance between each pair
+    of consecutive points, in decimal degrees.
+
+    Parameters:
+        lon (numpy.ndarray of shape (n,)): longitude coordinates, in decimal degrees
+        lat (numpy.ndarray of shape (n,)): latitude coordinates, in decimal degrees
+
+    Returns:
+        Angular distance (numpy.ndarray of shape (n-1,), in decimal degrees
+    """
```
tovogt, line 335: According to CLIMADA's coding conventions, docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    http://www.movable-type.co.uk/scripts/latlong.html
+    Here, the bearing of each pair of consecutive points is computed.
+
+    Parameters:
+        lon (numpy.ndarray of shape (n,)): longitude coordinates, in decimal degrees
+        lat (numpy.ndarray of shape (n,)): latitude coordinates, in decimal degrees
+
+    Returns:
+        bearing_angle (numpy.ndarray of shape (n-1,), in decimal degrees
+    """
```
tovogt, line 306: According to CLIMADA's coding conventions, docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    https://stats.stackexchange.com/questions/48086/algorithm-to-produce-autocorrelated-uniformly-distributed-number
+    The CDF of Y.
+
+    Parameters:
+        z (float): longitude coordinates, in decimal degrees
+        theta (float): arccos of the autocorrelation
+
+    Returns:
+        float (CDF at z)
+    """
```
tovogt, line 271: According to CLIMADA's coding conventions, docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    different temporal resolution (time_step_h), an hourly time series is generated
+    and resampled (using linear interpolation) to the target resolution.
+
+    Parameters:
+        n_ts (int): length of the series
+        autocorr: autocorrelation (between -1 and 1)
+
+    Returns:
+        numpy.ndarray of length n
+    """
```
tovogt, line 226: According to CLIMADA's coding conventions, docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks_synth.py
```diff
@@ -37,43 +37,50 @@
...
+            shift in decimal degree (up to +/-max_shift_ini for longitude and latitude).
+            Default: 0.75.
+        max_dspeed_rel (float, optional): amplitude of translation speed
+            perturbation in relative terms (e.g., 0.2 for +/-20%). Default: 0.3.
+        max_ddir (float, optional): amplitude of track direction (bearing angle) perturbation
+            per hour, in radians. Default: pi/180.
+        autocorr_dspeed (float, optional): autocorrelation of translation speed perturbation
+            at a lag of 1 hour. Default: 0.85.
+        autocorr_ddir (float, optional): autocorrelation of translational direction perturbation
+            at a lag of 1 hour. Default: 0.85.
```
tovogt, line 83: Since you are changing so much in this docstring anyways, maybe you can take the time to convert it to the NumPy style. According to CLIMADA's coding conventions, docstrings should follow the NumPy style. From one of the developer meetings, I remember that, while older code in CLIMADA is still allowed to contain google-style docstrings, new code is required to follow the new convention. Old docstrings should be converted to the new convention whenever they are touched.

climada/hazard/tc_tracks.py
tovogt: In my opinion, `nb_synth_tracks` is not very clear either. How about `ensemble_size`?

climada/hazard/tc_tracks_synth.py
```diff
@@ -82,26 +89,44 @@ def calc_random_walk(tracks,
...
 
-    random_vec = [np.random.uniform(size=ens_size * (2 + track.time.size))
-                  for track in tracks.data]
+    # ensure tracks have constant time steps
+    time_step_h = np.unique([np.unique(x['time_step']) for x in tracks.data])
+    if not np.allclose(time_step_h, time_step_h[0]):
+        LOGGER.error('Tracks have different temporal resolution. '
+                     'Please ensure constant time steps by applying equal_timestep beforehand')
+        raise ValueError('Tracks have different temporal resolution.')
+    time_step_h = time_step_h[0]
```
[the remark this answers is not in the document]
tovogt: Don't use `except` without specifying the type of error (in this case, `ValueError`).

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +206,170 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    # scale x to have magnitude [0,1]
+    x = (x + np.sqrt(3)) / (2 * np.sqrt(3))
+    # resample at target time step
+    x_ts = np.interp(np.arange(start=0, stop=n_ts_hourly_exact, step=time_step_h),
+                     np.arange(n_ts_hourly), x)
+    return x_ts
+
+
+@jit
+def _h_ac(x, y, theta):
```
tovogt, line 244: Missing docstring.

tovogt (changes requested):
Thanks for your PR!

I tried to focus on those parts of the code that haven't been commented on by the other reviewers already. Once you have addressed all of the mentioned issues, I will be happy to have another look at all changes.

In general, please make sure that your changes don't introduce any pylint messages.

climada/hazard/tc_tracks_synth.py
chahank: I suggested the change away from `ensemsble_size` because to me this is confusing wording. The method is calc_rnd_walks, so it computes random walks and not an ensemble. Also, I was always confused whether after applying the method, the set of track now contains `ens_size` number of tracks or `en_size + 1`. The answer is en_size + 1` (it contains the original tracks) which I find confusing. Why would the ensemble size be larger than `ens_size`? That is why I suggested a name which say explicitly how many tracks are generate by `calc_rnd_walks`. 

Any ideas how to resolve this? (it is not really critical, so it can also stay at ens_size).

tovogt:
Is the general methodology of `calc_random_walk` (the one proposed in this PR or any of the previous versions) discussed anywhere in the peer-reviewed literature? If so, we can point to it in the docstrings. If the exact method does not appear in the literature, but similar methods are, then it might be worth having a look at the wording to choose better variable and function names in CLIMADA's implementation.

chahank:
To the best of my knowledge, it is not published anywhere. Importantly, @bguillod significantly improved the algorithm with this pull request.

tovogt:
Okay, but it would still be useful to have a look at the existing literature on the topic to choose a good wording. For example, from reading Casson and Coles 2000 (https://doi.org/10.1111/1467-9876.00189), I would rather use the function name `perturb_trajectories` to describe what `calc_random_walk` does. At first, this makes clear that only existing tracks are perturbed and no new tracks are generated from more complex statistical analyses. Furthermore, only the trajectory is simulated, while all other physical variable (such as pressure and radius) remain untouched.

chahank:
That is an excellent point! However, should we rename such an 'old' CLIMADA method? I think we should be much more careful with naming in the future though (and pay attention to that in the review process).

tovogt:
You can safely rename the function by adding the old name as an alias for backwards compatibility. Users who use the old name should be warned that the name is deprecated and superseded by a new one.

climada/hazard/tc_tracks_synth.py
bguillod: Ok, I've converted to numpy style in commit e837bc88958cd756867fef5a7ef8d79571c8981c.

climada/hazard/tc_tracks_synth.py
bguillod: Also fixed in commit e837bc88958cd756867fef5a7ef8d79571c8981c
(@chahank I guess your comment here was a reply to something else so I didn't deal with it here, but likely above).

climada/hazard/tc_tracks_synth.py
bguillod: Forgot to remove comments to myself, now removed in e5411bd9dc72d90445c891bb64d455c5e0687d65

climada/hazard/tc_tracks_synth.py
bguillod: yes, e837bc88958cd756867fef5a7ef8d79571c8981c

climada/hazard/tc_tracks_synth.py
bguillod: yes e837bc88958cd756867fef5a7ef8d79571c8981c

climada/hazard/tc_tracks_synth.py
bguillod: I kept my code as is here.

climada/hazard/tc_tracks_synth.py
bguillod: Thanks for catching, defa80514a9a67f01471ec68174cbb35faafae2f.

climada/hazard/tc_tracks_synth.py
bguillod: e837bc88958cd756867fef5a7ef8d79571c8981c

climada/hazard/tc_tracks_synth.py
bguillod: I am using distance in radians, not in km, hence it looks like I'm using a radius of 1. But this is consistent throughout the code so all good. I preferred that to relying on an assumption about Earth's radius, also because it is then clear it assumed a sphere.

climada/hazard/tc_tracks_synth.py
bguillod: Yes, thanks, fixed in e837bc88958cd756867fef5a7ef8d79571c8981c

climada/hazard/tc_tracks_synth.py
bguillod: e837bc88958cd756867fef5a7ef8d79571c8981c

climada/hazard/tc_tracks_synth.py
bguillod: yes, fixed in e837bc88958cd756867fef5a7ef8d79571c8981c I believe

climada/hazard/test/test_tc_tracks_synth.py
bguillod: Actually there is in test_angle_funs_pass for the following:
- _get_bearing_angle
- _get_angular_distance
- _get_destination_points
But for the ones related to random uniform I must admit I won't find time...

climada/hazard/test/test_tc_tracks_synth.py
```diff
@@ -356,6 +356,27 @@ def test_wrong_decay_pass(self):
...
 
 class TestSynth(unittest.TestCase):
+    def test_angle_funs_pass(self):
+        """Test functions used by random walk code"""
+        bearings = tc_synth._get_bearing_angle([15, 20, 20], [0, 0, 5]).tolist()
```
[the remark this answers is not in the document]
bguillod: Well spoted, fixed in a58f1fa4490de7ea3799864824a13efdc805e0c6

climada/hazard/test/test_tc_tracks_synth.py
```diff
@@ -356,6 +356,27 @@ def test_wrong_decay_pass(self):
...
+        self.assertAlmostEqual(tc_synth._get_bearing_angle(np.array([15, 20]), np.array([0, 0]))[0], 90.0)
+        self.assertAlmostEqual(tc_synth._get_bearing_angle(np.array([20, 20]), np.array([0, 5]))[0], 0.0)
+        self.assertAlmostEqual(tc_synth._get_bearing_angle(np.array([0, 0.00001]), np.array([0, 0.00001]))[0], 45)
+        self.assertAlmostEqual(tc_synth._get_angular_distance(np.array([0, 0]), np.array([0, 0]))[0], 0.0)
+        self.assertAlmostEqual(tc_synth._get_angular_distance(np.array([0, 1]), np.array([0, 0]))[0], 1.0)
+        self.assertAlmostEqual(tc_synth._get_angular_distance(np.array([0, 0]), np.array([0, 1]))[0], 1.0)
+        self.assertAlmostEqual(tc_synth._get_angular_distance(np.array([0, 30]), np.array([0, 30]))[0], 41.40962211)
```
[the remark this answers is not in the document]
bguillod: Yep, 251e173ded1d7670290e52b37821ac61ed3a893e.

climada/hazard/test/test_tc_tracks_synth.py
```diff
@@ -356,6 +356,27 @@ def test_wrong_decay_pass(self):
...
+        self.assertAlmostEqual(tc_synth._get_angular_distance(np.array([0, 30]), np.array([0, 30]))[0], 41.40962211)
```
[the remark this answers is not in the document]
bguillod: Well spoted, removed in 7bdc73ba07a04fc506bd5f1d53cbf953cee8d31e

bguillod (commented):
@tovogt @chahank I think I've dealt with most requested changes and comment, just this one about renaming: I'm fine to rename to `calc_perturbed_trajectories`, then for backward-compatibility should it be:
```
def calc_random_walk(*args, **kwargs):
    calc_perturbed_trajectories(*args, **kwards)
```
Or how would you best do that?

climada/hazard/tc_tracks_synth.py
```diff
@@ -82,26 +89,44 @@ def calc_random_walk(tracks,
     """
     LOGGER.info('Computing %s synthetic tracks.', ens_size * tracks.size)
```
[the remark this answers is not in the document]
chahank: ens_size was renamed.

climada/hazard/tc_tracks_synth.py
chahank: But then you risk to go about 180/360degrees right?

climada/hazard/tc_tracks_synth.py
chahank: Yes, sorry, thanks @tovogt and @bguillod . Keep it as is. Maybe add a comment to explain why you define this. E.g. "Sequences between all points (0->1, 1->2,...n-1->n)"

climada/hazard/tc_tracks_synth.py
chahank: I think it is necessary to explain this in a short comment.

climada/hazard/tc_tracks_synth.py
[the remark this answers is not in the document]
bguillod: ah ok, yes I've renamed it everywhere.

climada/hazard/tc_tracks_synth.py
bguillod: Well theoretically yes, this would be undirected then. But if the valus of max_ddir are reasonable it shouldn't (dt is max, say, 6 (hours), so you'd have to set max_ddir to 180/6=30 degrees, far far more than the default of pi/180 (i.e. 1 degree).

climada/hazard/tc_tracks_synth.py
bguillod: Ok, I've done in 0edbeff33d5f4028788449a4245cfe8dcaa3f2f2.

climada/hazard/tc_tracks_synth.py
```diff
@@ -166,6 +241,219 @@ def _one_rnd_walk(track, ens_size, ens_amp0, ens_amp, max_angle, rnd_vec):
...
+    earth_ang_fix = np.arctan2(np.sin(delta_lon) * np.cos(lat_2),
+                               np.cos(lat_1) * np.sin(lat_2) - \
+                               np.sin(lat_1) * np.cos(lat_2) * np.cos(delta_lon))
+    return np.degrees(earth_ang_fix)
+
+
+@jit
+def _get_angular_distance(lon, lat):
+    """
+    Compute the angular distance of great circle paths defined by consecutive points
```
tovogt, line 400: It's true that `climada.util.coordinates.dist_approx` returns the distance in km, while here it's in degrees. But that's only a factor. I would modify `dist_approx` by adding an option to compute the angular or metric distance.

climada/hazard/tc_tracks_synth.py
bguillod: To me the term 'angular distance' is clear: it is expressing a distance on a sphere by its angle from the sphere center. So if this angle is in radians, it would happen to be the distance for a sphere if radius 1, but actually it is an angle, not a distance per se. If you can be more specific on what you would add there as a comment I can include that of course.

climada/hazard/tc_tracks.py
```diff
@@ -865,9 +865,9 @@ def equal_timestep(self, time_step_h=1, land_params=False):
                                                       land_geom))
             self.data = new_data
 
-    def calc_random_walk(self, **kwargs):
+    def calc_random_walk(self, nb_synth_tracks=9, **kwargs):
```
tovogt, line 868: This is the API function that has been available for several CLIMADA versions. I would rename it like this:
```
def calc_random_walk(self, *args, **kwargs):
    LOGGER.warn("The use of TCTracks.calc_random_walk is deprecated."
                "Use TCTracks.calc_perturbed_trajectories instead.")
    return self.calc_perturbed_trajectories(self, *args, **kwargs)

def calc_perturbed_trajectories(self, nb_synth_tracks=9, **kwargs):
    # ...
```

For the function in `climada.hazard.tc_tracks_synth`, I think we can rename without alias to `calc_perturbed_trajectories` because it's not really part of the official API.

climada/hazard/tc_tracks_synth.py
bguillod: I can do and started to, but `climada.util.coordinates.dist_approx ` compute distances between all pairs of points while here we only need those between consecutive points. So using that function would imply more computations than required, hence I would prefer not to change that. What do you think?

climada/hazard/tc_tracks.py
bguillod: Thanks, done in 6e24d67e8a0503417d59de7bf6460cad6d60bcac.

climada/hazard/tc_tracks_synth.py
tovogt: No extra computations are required if you replace the call
```
angular_dist = _get_angular_distance(lon, lat)
```
by
```
angular_dist = dist_approx(lat[:-1, None], lon[:-1, None], lat[1:, None], lon[1:, None],
                           method="geosphere", units="degree")[:, 0, 0]
```
The `units` parameter needs to be implemented and should default to `"km"` for backwards compatibility.

bguillod: Ok, did so in commits 0f02c64574d3acf39b8a68d6136336f3d711ab2e and 5e5623dbf2fd45f28ff529911aca21589c6cfb29

bguillod: Of course one could also think that functions `_get_bearing_angle` and `_get_destination_points` could be in `climada.util.coordinates`, but I would say they are too specific.

bguillod:
It looks like the test on Jenkins are not running for this PR, is this normal? There nothing I should do for these to be run, right?

climada/hazard/tc_tracks_synth.py
tovogt: Thanks! I added unit tests for the new `units` parameter in 569526105fe96fa8452604a3f3dcfb485f732f4d.

tovogt: All references to the now-removed function `_get_angular_distance` should be removed. Especially in the unit tests.

climada/hazard/tc_tracks_synth.py
```diff
@@ -37,43 +37,50 @@
...
+    resolution of the tracks. Note however that all tracks should be at the same
+    temporal resolution, which can be achieved using equal_timestep().
+    max_dspeed_rel and autocorr_dspeed control the spread along the track ('what distance
+    does the track run for'), while max_ddir and autocorr_ddir control the spread
+    perpendicular to the track movement ('how does the track diverge in direction').
+    max_dspeed_rel and max_ddir control the amplitude of perturbations at each track
+    timestep but perturbations may tend to compensate each other over time, leading to
+    a similar location at the end of the track, while autocorr_dspeed and autocorr_ddir
+    control how these perturbations persist in time and hence the amplitude of the
+    perturbations towards the end of the track.
```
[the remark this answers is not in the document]
chahank: Very nice!

climada/hazard/tc_tracks_synth.py
bguillod: Oh yes. Did in 256cc3711b8ac598633814a6c40904a4fe67d3e2

climada/hazard/tc_tracks_synth.py
```diff
@@ -37,43 +37,50 @@
...
+        max_shift_ini (float, optional): amplitude of max random starting point
+            shift in decimal degree (up to +/-max_shift_ini for longitude and latitude).
+            Default: 0.75.
+        max_dspeed_rel (float, optional): amplitude of translation speed
+            perturbation in relative terms (e.g., 0.2 for +/-20%). Default: 0.3.
+        max_ddir (float, optional): amplitude of track direction (bearing angle) perturbation
+            per hour, in radians. Default: pi/180.
+        autocorr_dspeed (float, optional): autocorrelation of translation speed perturbation
+            at a lag of 1 hour. Default: 0.85.
+        autocorr_ddir (float, optional): autocorrelation of translational direction perturbation
```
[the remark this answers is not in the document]
chahank: I am confused by this. Is this now an autocorrelation factor or max translational direction perturbation?

chahank:
Last comment: you could add to the general docstring a mathematical description of what you algorithm does. One can include latex into the docstring as

```
.. math::
    \sum_i x_i  = S
```

climada/hazard/tc_tracks_synth.py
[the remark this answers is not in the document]
bguillod: I've also renamed autocorr_ddir to autocorr_ddirection in e456bf7b0bfd5dcc84abc7214c8b17902a62511e. `max_ddirection` is the maximum amplitude of the perturbations in direction, in radian. `autocorr_ddirection` is the temporal autocorrelation of these perturbations.
Similarly, `max_dspeed_rel` is the maximum (relative) perturbation that can be applied to translational speed, and `autocorr_dspeed` is the temporal autocorrelation of these perturbations. Does it make sense?

bguillod:
> Last comment: you could add to the general docstring a mathematical description of what you algorithm does. One can include latex into the docstring as
> 
> ```
> .. math::
>     \sum_i x_i  = S
> ```

Where, you mean in `calc_perturbed_trajectories`? That would be nice, but for this one it is quite tricky to describe briefly in equations, isn't it?

chahank:
Maybe something along:
```
: math::
\begin{align*}
	(x_0, y_0) &= (\tilde x_0 + \tilde y_0) + (\delta x_0, \delta y_0) \\
	(x_n, y_n) &= (x_n, y_n) + v(\tilde l_n \cdot \delta l_n, \tilde\theta_n + \sum_{j=1}^n \delta \theta_j)\\
	\tilde l_n &= || \tilde x_n - \tilde x_{n-1}, \tilde y_n -\tilde y_{n-1}||_{deg} \\
	\tilde \theta_n &= \textrm{bearing}(\tilde x_n - \tilde x_{n-1}, \tilde y_n -\tilde y_{n-1}) \\
	\delta x_0, \delta y_0 & \in  [-max_shift_ini, max_shift_ini], \;\; \delta\theta \in dt \cdot [-max_ddirection,max_ddirection], \;\; \delta l \in [1-max_dspeed_rel, max_dspeed_rel]
\end{align*}
```

climada/hazard/tc_tracks_synth.py
emanuel-schmid: Agreed. Must not necessarily be part of this pull request. We can come back to it later.

emanuel-schmid (approved):

tovogt (approved):

bguillod:
@chahank I try to do the math:: thing but it doesn't display correclty if I type
```
help(calc_perturbed_trajectories)
```
in jupyter-notebook. How can I check what the output looks like?

chahank:
Did you put `: math::` ?

bguillod:
Yes, e.g.:
```
    Notes
    -----
    If the original track coordinates are (lon_orig, lat_orig), the algorithm is the following:
    .. math::

        lon(0) = lon_orig(0) + u_0*max\_shift\_ini\\
        lat[0] = lat_orig[0] + v_0*max\_shift\_ini
    where u0,v0 are random numbers from a uniform distribution between [-1, 1]
    Then for i=1:

    : math::
    \begin{align*}
        (x_0, y_0) &= (\tilde x_0 + \tilde y_0) + (\delta x_0, \delta y_0) \\
        (x_n, y_n) &= (x_n, y_n) + v(\tilde l_n \cdot \delta l_n, \tilde\theta_n + \sum_{j=1}^n \delta \theta_j)\\
        \tilde l_n &= || \tilde x_n - \tilde x_{n-1}, \tilde y_n -\tilde y_{n-1}||_{deg} \\
        \tilde \theta_n &= \textrm{bearing}(\tilde x_n - \tilde x_{n-1}, \tilde y_n -\tilde y_{n-1}) \\
        \delta x_0, \delta y_0 & \in  [-max_shift_ini, max_shift_ini], \;\; \delta\theta \in dt \cdot [-max_ddirection,max_ddirection], \;\; \delta l \in [1-max_dspeed_rel, max_dspeed_rel]
    \end{align*}
```
but it is displayed as it.

chahank:
Oh you are right, I meant ``.. math::`` ... well, I do not know then. I took the information from the numpydoc https://numpydoc.readthedocs.io/en/latest/format.html 

It should display correctly when interpreted with sphinx. Maybe this is not done in the jupyter notebook?

bguillod:
> It should display correctly when interpreted with sphinx. Maybe this is not done in the jupyter notebook?
That would be my guess and it looks like that's the case: https://github.com/jupyter/notebook/issues/2581
 Can I somehow test how it would look like with sphinx? Or should I ignore how it looks like and hope it will be fine?

chahank:
If it doesn't work, we can leave it as is. It would be a nice bonus, that's all.

bguillod:
Yes I guess most users use jupyter-notebook or ipython so to them that is not very convenient. So I leave it for now. @chahank Let me know if there anything else. I shouldn't worry that Jenkins doesn't display successful test, right? It looks like it didn't run recently for that PR somehow.

chahank:
The thing with the Jenkins test is weird. I can see one done 3 days ago that failed. For this, let's ask @emanuel-schmid  for confirmation that everything is ok. 

For the rest I have no more comments. Thanks again!

bguillod:
> The thing with the Jenkins test is weird. I can see one done 3 days ago that failed. For this, let's ask @emanuel-schmid for confirmation that everything is ok.

Yes and only 2 PR appear on Jenkins, the other ones are not displayed. Let's see. In any case thanks a lot for the useful comments!

emanuel-schmid:
> 
> 
> > The thing with the Jenkins test is weird. I can see one done 3 days ago that failed. For this, let's ask @emanuel-schmid for confirmation that everything is ok.
> 
> Yes and only 2 PR appear on Jenkins, the other ones are not displayed. Let's see. In any case thanks a lot for the useful comments!

Totally weird, yes. Haven't figured out what's up there. One reason that is imaginable is the disk space: we're slowly running out of it. Let's see what happens after we accept and purge obsolete branches.

PR #79, opened 2020-10-23
Feature: TCTracks to GeoDataFrame

mmyrte:
I was in need of a method to export my TCForecasts to a database, so I wrote this. I'm only neglecting one line of code in the tests that's really trivial (would return a tuple of DFs instead of one joint DF).

What do you think?

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -21,27 +21,33 @@ Define TCTracks: IBTracs reader and tracks manager.
 
 __all__ = ['CAT_NAMES', 'SAFFIR_SIM_CAT', 'TCTracks', 'set_category']
 
-import os
-import glob
-import shutil
-import logging
-import warnings
+# standard libraries
 import datetime as dt
+import glob
 import itertools
-import numpy as np
+import logging
+import os
+import shutil
+import warnings
+
+# additional libraries
+import cartopy.crs as ccrs
+import geopandas as gpd
 import matplotlib.cm as cm_mp
 from matplotlib.lines import Line2D
 from matplotlib.collections import LineCollection
 from matplotlib.colors import BoundaryNorm, ListedColormap
-import cartopy.crs as ccrs
-import pandas as pd
-import xarray as xr
-from sklearn.neighbors import DistanceMetric
 import netCDF4 as nc
 from numba import jit
+import numpy as np
+import pandas as pd
 import scipy.io.matlab as matlab
+from shapely.geometry import Point, LineString
+from sklearn.neighbors import DistanceMetric
 import statsmodels.api as sm
+import xarray as xr
 
+# climada dependencies
 from climada.util import ureg
 import climada.util.coordinates as coord_util
 from climada.util.constants import EARTH_RADIUS_KM, SYSTEM_DIR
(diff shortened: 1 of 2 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -799,6 +805,56 @@ def read_netcdf(self, folder_name):
...
+            gdf_long = gpd.GeoDataFrame({
+                'idx': idx,
+                'time': times,
+                'geometry': [Point(p[0], p[1]) for p in points],
+            }).set_index('idx')
+
+            if joint:
+                gdf = gdf.join(gdf_long)
+            else:
+                return gdf, gdf_long
```
emanuel-schmid, line 846: Unless there is a very good reason not to, the return value of a method/function should have a single (predefined) type - with the exception of `None` as the alternative, generic no-return type.\
In this particular case it's preferable to define two methods, one calling the other with a gdf and a tuple of gdfs respectively as return values.\
But consider whether it is not even preferable to leave the joining or splitting up again to the caller.

climada/hazard/tc_tracks.py
```diff
@@ -799,6 +805,56 @@ def read_netcdf(self, folder_name):
...
+        """Transform this TCTracks instance into a GeoDataFrame.
+
+        Parameters:
+            as_points (bool): If true, construct LineString or single Point
+                geometries. A point is created if a track has length one.
+                If set to false, the geometries are returned as points with an
+                additional timestamp column.
+            joint (bool): If using as_points, return geometries joined to
+                metadata. If set to False, return a tuple of two GeoDataFrames
+                that can be joined on their index.
```
emanuel-schmid, line 818: The meaning of parameters should ideally be independent from each other. This is not the case here. The meaning of `joint` depends on the value of `as_points`: if the latter is `True` it means something - otherwise not.

climada/hazard/tc_tracks.py
mmyrte: I did it like this because "unjoining" would be inefficient on large tracksets. Given that returning the tuple is possible with no overhead, I feel that this is acceptable.

emanuel-schmid: Inefficient in terms of performance or in terms of convenience of the caller?

climada/hazard/tc_tracks.py
mmyrte: I know that this is not optimal, but this again relates to the efficiency concerns above. I'm testing the entire ibtracs archive now to see how this performs at scale, and if the splitting afterward would indeed be as terrible as I think it is.

emanuel-schmid: Surprising! Do you have numbers?

mmyrte:
So, the difference is negligible for the 3680 tracks in IBTrACS, I'm changing it as you advised. Thanks!

```python
%%timeit
gdf_points_joint = ibtracs.to_geodataframe(as_points=True)
gdf_meta = gdf_points_joint.drop(['geometry', 'time'], axis=1).drop_duplicates()
gdf_points = gdf_points_joint[['time', 'geometry']]
```

> 5.05 s ± 93.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%%timeit
gdf_meta, gdf_points = ibtracs.to_geodataframe(as_points=True, joint=False)
```

> 4.82 s ± 71.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

climada/hazard/tc_tracks.py
mmyrte: In terms of performance. See my post below, I presumed wrongly 😀.

climada/hazard/tc_tracks.py
mmyrte: See comment below :)

climada/hazard/tc_tracks.py
```diff
@@ -799,6 +805,50 @@ def read_netcdf(self, folder_name):
             track.attrs['orig_event_flag'] = bool(track.orig_event_flag)
             self.data.append(track)
 
+    def to_geodataframe(self, as_points=False, joint=True):
```
emanuel-schmid, line 808:
```suggestion
    def to_geodataframe(self, as_points=False):
```

mmyrte: 🙈

Issue #74, opened 2020-10-12
Inconsistent value return for util.u_plot.geo_im_from_array()


chahank:
The plotting function `util.u_plot.geo_im_from_array(array_sub)` returns a numpy array of `mpl.geoaxes.GeoAxesSubplot` if `array_sub` is an iterable with two or more elements, otherwise returns a single instance of `mpl.geoaxes.GeoAxesSubplot`. 

This inconsistent behavious leads to a failure of for instance `hazard.plot_rp_intensity(return_periods=(25, 50, 100, 250))` if `return_period` is an iterable with one value, e.g. `return_period=[100]`. 

The failure is because the code tries to unpack the return value of `util.u_plot.geo_im_from_array(array_sub)` in the line `_, axis = u_plot.geo_im_from_array(inten_stats, self.centroids.coord,\
            colbar_name, title, smooth=smooth, axes=axis, **kwargs)`.

chahank:
Possible solution: always return a numpy array.

PR #81, opened 2020-10-27
Feature/fix plot rp intensity

chahank:
This is a simple fix for issue #74 . Now the function hazard.plot_rp_intensity returns the full set of subplots. There is no more unpacking to be done, and hence one can specify a single return period.


sameberenz:
Thanks for the fix. I tested it with a TC hazard and it worked.

sameberenz (approved):

PR #83, opened 2020-10-29
Feature/rc add scenario rcp26

sameberenz:
Before, RelativeCropyield could only be initiated for RCP6.0 and historical data (ISIMIP2b) as well as ISIMIP2a runs.
This feature enables the integration of RCP2.6, RCP8.0, as well as RCP2.6 extended (2100-2299) runs from ISIMIP2b as well.

Some workarounds were required to differentiate between the yearchunks for the two RCP2.6 file types.


carmensteinmann:
I don't have any remarks, except for the source code analysis showing some minor convention issues such as lines that are too long and wrong continued indentations.

carmensteinmann (approved):

PR #84, opened 2020-11-03
Improve performance of wind field computation

tovogt:
*Statement of the problem that this PR is trying to solve:* If the number of centroids is very large (e.g. if it is a global grid of comparably high resolution), the current wind field computation requires prohibitive amounts of memory. This is because, for each track, a dense windfield array is allocated covering all centroids - even those that are far away from the TC track: https://github.com/CLIMADA-project/climada_python/blob/6c27a6120815dde71fe0995ab5f8c963d33398bc/climada/hazard/trop_cyclone.py#L390

*Solution proposed by this PR:* We make sure that only those centroids in the direct vicinity of each TC track (the "reachable" centroids) are regarded in the wind field computation so that computational performance will only depend on the *resolution* of the centroids and not on the *total area covered* by the centroids:
https://github.com/CLIMADA-project/climada_python/blob/26ceaa3600fe693ecc961c7d312fd6eba8f0e049/climada/hazard/trop_cyclone.py#L409-L416

This does not change the API function `TropCyclone.set_from_track`, neither its structure nor its output, but only its overall performance. The changes do not affect performance of regional studies where centroids cover a comparably small area of the globe.

*Changes unrelated to the main topic of this PR:* While testing this functionality, I noticed that `TCTracks.equal_timestep` can take quite some time and decided to add a progress indicator for the convenience of CLIMADA users.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -669,9 +669,10 @@ class TCTracks():
             last_perc = 0
             new_data = []
             for track in self.data:
+                # progress indicator
                 perc = 100 * len(new_data) / len(self.data)
                 if perc - last_perc >= 10:
-                    LOGGER.info("Progress: %d%%", perc)
+                    LOGGER.debug("Progress: %d%%", perc)
                     last_perc = perc
                 new_data.append(
                     self._one_interp_data(track, time_step_h, land_geom))
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -515,7 +515,7 @@ class TCTracks():
                     hours[i_track, valid_idx[reference_idx]],)
                 datetimes = [reference_date + dt.timedelta(hours=int(step * i))
                              for i in range(nnodes)]
-            datetimes = np.array(datetimes)
+            datetimes = np.array(datetimes, dtype=np.datetime64)
 
             max_sustained_wind = tc_maxwind[i_track, valid_idx]
             max_sustained_wind_unit = 'kn'
@@ -666,10 +666,15 @@ class TCTracks():
                                       itertools.repeat(land_geom, self.size),
                                       chunksize=chunksize)
         else:
-            new_data = list()
+            last_perc = 0
+            new_data = []
             for track in self.data:
-                new_data.append(self._one_interp_data(track, time_step_h,
-                                                      land_geom))
+                perc = 100 * len(new_data) / len(self.data)
+                if perc - last_perc >= 10:
+                    LOGGER.info("Progress: %d%%", perc)
+                    last_perc = perc
+                new_data.append(
+                    self._one_interp_data(track, time_step_h, land_geom))
             self.data = new_data
 
     def calc_random_walk(self, **kwargs):
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -307,11 +307,12 @@ class TropCyclone(Hazard):
             raise ValueError
         ncentroids = centroids.coord.shape[0]
         coastal_centr = centroids.coord[coastal_idx]
-        windfields = compute_windfields(track, coastal_centr, mod_id)
+        windfields, reachable_centr_idx = compute_windfields(track, coastal_centr, mod_id)
+        reachable_coastal_centr_idx = coastal_idx[reachable_centr_idx]
         npositions = windfields.shape[0]
         intensity = np.zeros(ncentroids)
-        intensity[coastal_idx] = np.linalg.norm(windfields, axis=-1)\
-                                                .max(axis=0)
+        intensity[reachable_coastal_centr_idx] = np.linalg.norm(windfields, axis=-1)\
+                                                          .max(axis=0)
         intensity[intensity < self.intensity_thres] = 0
 
         new_haz = TropCyclone()
@@ -319,7 +320,7 @@ class TropCyclone(Hazard):
         new_haz.intensity = sparse.csr_matrix(intensity.reshape(1, -1))
         if store_windfields:
             wf_full = np.zeros((npositions, ncentroids, 2))
-            wf_full[:, coastal_idx, :] = windfields
+            wf_full[:, reachable_coastal_centr_idx, :] = windfields
             new_haz.windfields = [
                 sparse.csr_matrix(wf_full.reshape(npositions, -1))]
         new_haz.units = 'm/s'
@@ -371,13 +372,26 @@ class TropCyclone(Hazard):
 def compute_windfields(track, centroids, model):
     """Compute 1-minute sustained winds (in m/s) at 10 meters above ground
 
-    Parameters:
-        track (xr.Dataset): track infomation
-        centroids (2d np.array): each row is a centroid [lat, lon]
-        model (int): Holland model selection according to MODEL_VANG
+    In a first step, centroids within reach of the track are determined so that wind fields will
+    only be computed and returned for those centroids.
 
-    Returns:
-        np.array
+    Parameters
+    ----------
+    track : xr.Dataset
+        Track infomation.
+    centroids : 2d np.array
+        Each row is a centroid [lat, lon].
+        Centroids that are not within reach of the track are ignored.
+    model : int
+        Holland model selection according to MODEL_VANG.
+
+    Returns
+    -------
+    windfield : np.array of shape (npositions, nreachable, 2)
+        A directional wind field for each track position on those centroids within reach
+        of the TC track.
+    reachable_centr_idx : np.array of shape (nreachable,)
+        List of indices of input centroids within reach of the TC track.
     """
     # copies of track data
     # Note that max wind records are not used in the Holland wind field models!
@@ -385,13 +399,6 @@ def compute_windfields(track, centroids, model):
         track[ar].values.copy() for ar in ['lat', 'lon', 'time_step', 'radius_max_wind',
                                            'environmental_pressure', 'central_pressure']
     ]
-
-    ncentroids = centroids.shape[0]
-    npositions = t_lat.shape[0]
-    windfields = np.zeros((npositions, ncentroids, 2))
-
-    if t_lon.size < 2:
-        return windfields
 
     # never use longitudes at -180 degrees or below
     t_lon[t_lon <= -180] += 360
@@ -402,11 +409,20 @@ def compute_windfields(track, centroids, model):
 
     # restrict to centroids in rectangular bounding box around track
     track_centr_msk = _close_centroids(t_lat, t_lon, centroids)
-    track_centr_idx = track_centr_msk.nonzero()[0]
     track_centr = centroids[track_centr_msk]
 
+    nreachable = track_centr.shape[0]
+    npositions = t_lat.shape[0]
+    windfields = np.zeros((npositions, nreachable, 2))
+    [reachable_centr_idx] = track_centr_msk.nonzero()
+
+    # the wind field model requires at least two track positions because translational speed
+    # as well as the change in pressure are required
+    if t_lon.size < 2:
+        return windfields, reachable_centr_idx
+
     if track_centr.shape[0] == 0:
-        return windfields
+        return windfields, reachable_centr_idx
 
     # compute distances and vectors to all centroids
     d_centr, v_centr = [ar[0] for ar in dist_approx(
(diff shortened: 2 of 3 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -515,7 +515,7 @@ def _read_file_emanuel(self, path, hemisphere='S', rmw_corr=False):
                     hours[i_track, valid_idx[reference_idx]],)
                 datetimes = [reference_date + dt.timedelta(hours=int(step * i))
                              for i in range(nnodes)]
-            datetimes = np.array(datetimes)
+            datetimes = np.array(datetimes, dtype=np.datetime64)
```
tovogt, line 518: This is also unrelated to the main topic of this PR: I noticed along the way that this array's `dtype` would be automatically set to `object` (instead of the more natural `np.datetime64`) which might cause problems in the future.

climada/hazard/tc_tracks.py
```diff
@@ -666,10 +666,15 @@ def equal_timestep(self, time_step_h=1, land_params=False):
                                       itertools.repeat(land_geom, self.size),
                                       chunksize=chunksize)
         else:
-            new_data = list()
+            last_perc = 0
+            new_data = []
             for track in self.data:
-                new_data.append(self._one_interp_data(track, time_step_h,
-                                                      land_geom))
+                perc = 100 * len(new_data) / len(self.data)
```
chahank, line 672:
```suggestion
                 #progress indicator
                perc = 100 * len(new_data) / len(self.data)
```

climada/hazard/tc_tracks.py
```diff
@@ -666,10 +666,15 @@ def equal_timestep(self, time_step_h=1, land_params=False):
...
+                if perc - last_perc >= 10:
+                    LOGGER.info("Progress: %d%%", perc)
+                    last_perc = perc
+                new_data.append(
+                    self._one_interp_data(track, time_step_h, land_geom))
```
chahank, line 677: Not related to PR, but maybe use list comprehension here:

`new_data = [_one_interp_data(track, time_step_h, land_geom) for track in self.data]`

This would however remove the progress indicator. I am not 100% convinced by having by default so much logging, so I would be happy with the list comprehension.

climada/hazard/trop_cyclone.py
```diff
@@ -472,8 +485,8 @@ def compute_windfields(track, centroids, model):
     v_full = v_trans[1][1:, None, :] * v_trans_corr[1:, :, None] + v_ang
     v_full[np.isnan(v_full)] = 0
 
-    windfields[1:, track_centr_idx, :] = v_full
-    return windfields
+    windfields[1:, :, :] = v_full
+    return windfields, reachable_centroids
 
 def _close_centroids(t_lat, t_lon, centroids):
```
chahank, line 491: Should the bounding box size maybe be a variable?

climada/hazard/trop_cyclone.py
```diff
@@ -402,11 +408,18 @@ def compute_windfields(track, centroids, model):
 
     # restrict to centroids in rectangular bounding box around track
     track_centr_msk = _close_centroids(t_lat, t_lon, centroids)
-    track_centr_idx = track_centr_msk.nonzero()[0]
     track_centr = centroids[track_centr_msk]
 
+    nreachable = track_centr.shape[0]
+    npositions = t_lat.shape[0]
+    windfields = np.zeros((npositions, nreachable, 2))
+    reachable_centroids = track_centr_msk.nonzero()[0]
```
chahank, line 416: Isn't np.array.nonzero() deprecated? 

_Deprecated since version 0.24.0: Please use .to_numpy().nonzero() as a replacement._

chahank: Maybe also use `reachable_centroids_idx` to indicate that it is the indexes of the centroids and not the centroids themselves. Would have to adjust the variable downstream too.

chahank, line 416:
```suggestion
    [reachable_centroids_idx] = track_centr_msk.to_numpy().nonzero()
```

climada/hazard/trop_cyclone.py
```diff
@@ -402,11 +408,18 @@ def compute_windfields(track, centroids, model):
...
+
+    if t_lon.size < 2:
```
chahank, line 418: Why return if t_lon.size<2 ? MAybe add an inline comment.

climada/hazard/trop_cyclone.py
```diff
@@ -307,19 +307,19 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
             raise ValueError
         ncentroids = centroids.coord.shape[0]
         coastal_centr = centroids.coord[coastal_idx]
-        windfields = compute_windfields(track, coastal_centr, mod_id)
+        windfields, reachable_centroids = compute_windfields(track, coastal_centr, mod_id)
```
chahank, line 310: `reachable_coastal_centroids` or `reachable_coastal_centr`?

climada/hazard/trop_cyclone.py
```diff
@@ -307,19 +307,19 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
         npositions = windfields.shape[0]
         intensity = np.zeros(ncentroids)
-        intensity[coastal_idx] = np.linalg.norm(windfields, axis=-1)\
-                                                .max(axis=0)
+        intensity[coastal_idx[reachable_centroids]] = np.linalg.norm(windfields, axis=-1)\
```
chahank, line 313: It is a bit confusing to take indices of indices no?

chahank (commented):
Looks like the change we needed to speed-up the code. Did you already do a test of how much faster it is? Also, can one vary the size of the bounding box around the tracks?

Thanks for the swift work!

bguillod:
I am also facing performance issue on this so warmly welcome an improvement. Do you have any indication of the magnitude of the improvement, in memory and duration?
Also, I noticed that well parallelizing only one core seems to be active at a time. Did you notice that, too? Any idea why this is the case?

climada/hazard/trop_cyclone.py
tovogt: This is a Pandas warning and I have no idea how you end up having a Pandas Series at this point?! Can you give a code example?

climada/hazard/tc_tracks.py
tovogt: I think verbose output is helpful for debugging. What do you think about changing the logging level to `debug`?

tovogt:
I don't think that runtime improvement is a big thing. Of course, allocating a lot of memory takes time. But it's negligible compared to the time needed to do the main computations. This PR is mainly about the memory requirements which might be reduced by 95% in many cases since most TCs won't affect more than 5% of a global grid of centroids.

@bguillod How do you parallelize?

climada/hazard/trop_cyclone.py
chahank: Oh yeah, sorry got confused ^^. All good.

climada/hazard/trop_cyclone.py
tovogt: Since the bounding box is not introduced in this PR, I think changing it is beyond the scope of this PR. We should discuss this separately. The bounding box was there before to determine the centroids on which computations are performed. I only made sure that the memory allocation is restricted to those centroids for which computations will be performed.

chahank:
To make @tovogt clear: the computation time of the windfields scales with the number of tracks and the number of centroids. This PR fixes the problem for the particular situation when tracks do only affect a small subset of the whole centroid set. In this case you will get a significant speed-up due to less memory use.

tovogt:
Thanks a lot for your feedback so far! I adopted most of your proposed changes and commented on some.

mmyrte:
I'm sorry, I don't have the time to really go into the code, but this seems like a very nice amendment. I wanted to comment that my approach to this problem is to draw a coarse buffer around a track(set) (via geodataframe), fill that buffer with [h3 indices](https://h3geo.org/) and use those indices to fetch centroids with their distance to coast (all operations are very cheap). I then write those centroids back to my database, but you could also push that into some other data store. 

The same approach would work with a shapely vectorized contains on ordinary point features. I don't have the time to jot down more, I'm sorry. En guete!

bguillod:
@tovogt Here is how I parallelize (taken from the random walk example in the tutorial so I apply the same to this):
```
from pathos.pools import ProcessPool as Pool

pool = Pool(nodes=4)
tc_holland = TropCyclone(pool=pool)
tc_holland.set_from_tracks(tracks, centroids, ignore_distance_to_coast=True)
tc_holland.check()

pool.close()
pool.join()
```
This takes ages and first only one core looks active (from htop), later on all cores look inactive but these steps still aren't completed (it's in a function and the function still hasn't returned).

Am I doing anything wrong? I have 300'000 centroids, (0.05 degrees grid) so admittedly this takes a lot of resources, but still...

tovogt:
I didn't change anything related to the use of pools, so I guess that this is beyond the scope of this PR and we should discuss this in a different place. Furthermore, I can't give advice regarding the use of pathos since I never used it.

chahank:
@bguillod  Note that 300'000 centroids is a rather large amount, in particular if they are very close to each other. Hence, it is expected that this takes some time to compute. There might possibilities to speed this up, but not with this PR as @tovogt  said above.

bguillod:
Sure, sounds good.

bguillod:
If I'm not mistaken, centroids are filtered out not only by `_close_centroids` but also later on with `close_centr = (d_centr < CENTR_NODE_MAX_DIST_KM) & (d_centr > 1e-2)`, with the exception of v_trans (which is only filtered based on `_close_centroids`. Why not then further reduce the size of windfields to only use centroids defined by `close_centr`, which would further reduce the size of the array?

gabrielaznar:
Dear @bguillod, have you tried using the centroids as raster by using its "meta" attribute as explained here [tutorial hazard](https://github.com/CLIMADA-project/climada_python/blob/main/doc/tutorial/climada_hazard_Hazard.ipynb)? That would be something like that:

haz = Hazard('TC')
haz.centroids.set_raster_from_pix_bounds(xf_lat, xo_lon, d_lat, d_lon, n_lat, n_lon) # default crs used

tovogt:
@bguillod The reachable centroids that are defined in a first step include centroids in a rectangular bounding box around the whole track. It requires little computational effort to filter those out. In contrast to that, `close_centr` takes into consideration the actual geographical distance of each centroid to each of the eye positions along the track. So, for each eye position, centroids within a circular area around the eye are identified as close to that eye. That's why it's stored as a multi-dimensional mask. I don't see an obvious way how we could use that to save more memory. Maybe you can open a new PR for this?

bguillod:
@tovogt One could further remove all centroids for which all distances to the eye positions along the track are smaller than CENTR_NODE_MAX_DIST_KM, i.e. if the sum of close_centr along the centroids dimension is 0. I don't have time to look into this now, it was just an idea.
@gabrielaznar Not explicitly, but that wouldn't change anything for this holland wind model computation, would it?

bguillod:
Here is a way to further filter out centroids with `_close_centroids`, using an effective buffer around the track coordinates rather than a box - happy to open another feature branch for these if you deem necessary:
`dist_approx` takes a long time and memory when one has many centroids and track points. To reduce the number of centroid points, I propose to modify `_close_centroids` to create a buffer of `CENTR_NODE_MAX_DIST_DEG` degrees around the track and sub-select centroids for that:
```
# to be added to the file header
import geopandas as gpd
from shapely.geometry import LineString

# to replace content of `_close_centroids`
def _close_centroids(t_lat, t_lon, centroids):
    centr_lat, centr_lon = centroids[:, 0], centroids[:, 1]
    if (t_lon < -170).any() and (t_lon > 170).any():
        # crosses 180 degrees east/west -> use positive degrees east
        t_lon[t_lon < 0] += 360
        centr_lon[centr_lon < 0] += 360
    centr_gpd = gpd.GeoDataFrame(geometry=gpd.points_from_xy(centr_lon, centr_lat))
    track_bounds = LineString(np.c_[t_lon, t_lat]).buffer(distance=CENTR_NODE_MAX_DIST_DEG)
    cent_msk = centr_gpd.within(track_bounds)
    return cent_msk
```
**NOTE**: Running `_close_centroids` takes longer that way, but in my case this saves time when calling `dist_approx` later on in `compute_windfields` which makes that approach faster. It might not apply to cases with less data though, perhaps to be tested.

tovogt:
Following @bguillod's proposal, I modified the `_close_centroids` function. But I refrained from using geopandas or shapely since they are quite slow according to my experience. Using pure numpy functionality and rectangular instead of circular buffers around the track points, the buffer is not as tight as in @bguillod's proposal, but the improvement in memory requirement should be of the same order of magnitude while being considerably faster.

simonameiler (commented):
Thank you very much @tovogt - the changes you implemented fixes the problem of an inefficient wind field calculation for the case when TC tracks only affect a small area. (However, on a different but related matter, note that the problem scales with the number of tracks.)

emanuel-schmid (approved):

Issue #67, opened 2020-09-21
Impact class: add methods 'select' and 'append'?


sameberenz:
Hi everyone

Is there a any reasonable way to filter or combine CLIMADA [Impact](https://github.com/CLIMADA-project/climada_python/blob/develop/climada/engine/impact.py) instances at the moment?

My guess is that the Impact would require methods comparable to `Hazard.select()` and `Hazard.append()`, right?

Here is an example what I would like to do:
```

impact = Impact()
impact.calc(...)
impact_subset1 = impact.select(years=[2005,2006,2008,2020,2021])
impact_subset2 = impact.select(event_names=['Petra', 'Antoinette', 'Hugo', 'Sergej'])
```
and
`
impact_combined = impact_subset1.append(impact_subset2)`

When filtering or combining two impact sets, we would need in any case to recompute the values of **aai_agg**, **tot_value**, and **frequency**, right? Maybe filtering/combining the hazard sets and re-running `impact.calc()` in a loop is a valid alternative?

For my application, however, I want to filter and re-combine my impact instances several times during analysis without rerunning `impact.calc()` all the time. Therefore, filtering hazard set beforehand is not my favourite option.

Are there any suggestions other than writing such methods in `impact.py`?
Would it make sense to create a new module for "impact_manipulation"?

Thanks in advance for your inputs

ChrisFairless:
@sameberenz was right when he mentioned this could be relevant to other issues (being discussed on slack). it would be cool if we could design something that let you subset/reweight an impact set to answer questions such as
* what is the AAI of storms with < cat 3 winds that make landfall from the gulf of mexico?
* if climate change makes major hurricanes 10% more frequent, how does AAI change?
* what are the average TC losses in the five years following a billion-dollar event?
without the need to recalculate impacts


sometimes you would want to create a new Impact object (as in Sam's example), and sometimes you'd just want to reweight and recalculate statistics from an existing one (e.g. for large event sets with many reweightings)

sameberenz:
Weights as suggested by @ChrisFairless could work for filtering (set weight to 0 for events you don't want to consider, and 1 for the others).

For combining/appending `Impact` sets, a seperate method would still be required next to weights.

NB: A work-around to filter out events within an `Impact` set is setting frequency to 0, however, this is not what frequency is intended for.

davidnbresch:
as for combining impacts, you might have a brief look at the OLD MATLAB version too (for feature inspiration, not the code):

[...]

PR #89, opened 2020-11-16
Feature/impact select

chahank:
New method to select impacts based on event names, ids, date range, or on a subset of coord_exp (exposure coordinates). This requires the existence of the imp_mat matrix. All the impact values are recomputed for the selected events. 

Note that the select is an 'or', which is different from the select method in TropCyclone where it is implemented as an 'and'. I felt 'or' is more natural here, but this could be changed easily.

In response to request #67 .


bguillod:
Very nice feature, definitely a useful one! I just had a look out of interest and perhaps one question I'd have is: why you don't adjust frequencies. But how to do that is a tricky one, admittedly.
My starting point was: If I generate a TropCyclone hazard with 30 years of historical data, the frequency of each storm is 1/30. If I do the same but using only 10 years of historical data, then my frequency will be 1/10. Hence, when subsetting 10 years of my first impact object (with freq=1/30), I will end up with the same impacts by event as in my second object but the eai etc. will be roughly only one third of it.
However there might be cases where it makes sense to keep the original frequencies. And how would you adjust frequencies if you select not by a date range but by e.g. event IDs? Perhaps there could be an argument to determine whether frequencies should be scaled?
I'm sure others who have more experience dealing with the impact class will have some ideas here.

chahank:
Excellent point. I was starting to implement a frequency adjustement, but I ended up with the dilemna you presented above. There is one case when it can be done easily: when the frequency for each event is 1/N, N:total number of events. In any other case it should be the user's task to adjust the frequencies. I would also argue in your example that it is correct that for a subset of 10 years from a historical set of 30 years the frequencies remain identical. After all, the frequency of an event should not depend on our choice of binning right?

So in conclusion, it might be right and correct to actually never adjust the frequencies since taking a subset of the impact data does not imply a change in the hazard.

Concrete proposition: leave it as is, and add a note about the frequency (either in the docstring or as a warning or both).

chahank:
Extra comment: in my opinion, the test is too long. This is mainly due to the large number of events and centroids used in the default demo file.

```        
        # Read default entity values
        ent = Entity()
        ent.read_excel(ENT_DEMO_TODAY)
        ent.check()

        # Read default hazard file
        hazard = Hazard('TC')
```

emanuel-schmid:
The changes in doc/tutorial/climada_hazard_TropCyclone.ipynb are grotesquely big. Why?

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        # of length equal to the number of events (=nb_events), using the 
+        # selection sel_ev.
+        per_event_attr = self.get_per_event_attr(imp)
+        for (attr, value) in imp.__dict__.items():
+            if attr in per_event_attr:
+                if isinstance(value, np.ndarray) and value.ndim == 1 \
+                                                   and value.size > 0:
+                    setattr(imp, attr, value[sel_ev])
+                elif isinstance(value, sparse.csr_matrix):
+                    setattr(imp, attr, value[sel_ev, :][:, sel_exp])
```
emanuel-schmid, line 1089: `coord_exp` is only relevant for `sparse.csr_matrix` and not for `np.ndarray` then?
If that's correct it could perhaps be mentioned in the docstring.

chahank:
Hmmm the changes to the tutorial are supposed to be in the tutorial_update branch. It could be that I made the impact_select branch from the tutorial_update instead of from develop. 

The large changes to the tutorial are intended. This is part of the general overhaul of the tutorials.

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        if (dates, event_ids, event_names) == (None, None, None):
+            sel_ev = np.arange(nb_events)
+
+        # set all attributes that are 'per event', i.e. have a row or a column
+        # of length equal to the number of events (=nb_events), using the 
+        # selection sel_ev.
+        per_event_attr = self.get_per_event_attr(imp)
+        for (attr, value) in imp.__dict__.items():
+            if attr in per_event_attr:
+                if isinstance(value, np.ndarray) and value.ndim == 1 \
```
emanuel-schmid, line 1085: Or, perhaps a wee bit more directly:
```suggestion
        for attr in self.get_per_event_attr(imp):
            value = imp.__getattribute__(attr)
            if isinstance(value, np.ndarray) and value.ndim == 1 \
            ...

**Review by emanuel-schmid — CHANGES_REQUESTED** [review]
Most of the requested changes are just cosmetics.

**Comment by emanuel-schmid** [comment]
> 
> 
> Hmmm the changes to the tutorial are supposed to be in the tutorial_update branch. It could be that I made the impact_select branch from the tutorial_update instead of from develop.
> 
> I will redo this pull request with only the changes required for the introduction of the function impact_select (and related tests). Then I would make a new branch feature/impact_select_clean and delete the branch feature/impact_select. Is this fine so @emanuel-schmid ?
> 
> The large changes to the tutorial are intended. This is part of the general overhaul of the tutorials. C.f. branch tutorial_update.

Oh. Too sad. I was making review comments and missed the message. Can you keep this pull request for copy-paste purposes?

**Comment by chahank** [comment]
Sure. Then is it fine if I make a new branch + pull request as proposed above? I will then also directly include the changes proposed in this pull request.

File: climada/engine/impact.py [hunk 3]
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        nb_events = len(event_obj.event_id)
+        # select the attributes from hazard object with a length or a
+        # row-dimension equal to the number of event_ids
+        for attr, value in event_obj.__dict__.items():
+            try:
+                if len(value) == nb_events:
+                    per_event_attrs.append(attr)
+            except TypeError:
+                try:
+                    if nb_events in value.shape:
```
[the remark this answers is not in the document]
chahank: Yes, this is intended. Any of the array's dimension is fine. This is in case some future attributes are coded row-like instead of column-like.

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        nb_events = imp.event_id.size
+        nb_exp = len(imp.coord_exp)
+        
+        sel_exp = np.arange(nb_exp)
+        # filter exposures
+        if isinstance(coord_exp, np.ndarray):
+            sel_exp = [j
+                       for j, coord in enumerate(imp.coord_exp)
+                       if coord in coord_exp
+                      ]
```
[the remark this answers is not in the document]
chahank: But this is a slightly different case right? Now it only checks whether it is None, but does not make sure it is an array. Is this better?

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        # filter events by date
+        if dates is None:
+            sel_dt = np.zeros(nb_events, dtype=bool)
+        else:
+            sel_dt = np.ones(nb_events, dtype=bool)
+            date_ini, date_end = dates
+            if isinstance(date_ini, str):
+                date_ini = util_dt.str_to_date(date_ini)
+                date_end = util_dt.str_to_date(date_end)
+            sel_dt &= (date_ini <= imp.date) & (imp.date <= date_end)
```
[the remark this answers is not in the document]
chahank: Cool thanks!

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+            if not np.any(sel_dt):
+                LOGGER.info('No impact event in date range %s.', dates)
+                return None     
+        sel_dt = list(np.argwhere(sel_dt).reshape(-1))
```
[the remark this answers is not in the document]
chahank: Goot point. I copied the logic from a previous select method in climada. I didn't like it too much either, but thought if it is there, it might be fine. I will change it.

climada/engine/impact.py
chahank: It is actually relevant also for the recomputations of the values imp.at_event, imp.eai_exp, and imp.aai_agg. It also sets imp.coord_ex.

chahank:
I implemented all the relevant changes except the ones of type `if isinstance(event_names, list)` -> `if event_names is None`. Need feeback first.

climada/engine/impact.py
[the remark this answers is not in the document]
emanuel-schmid: Yes, this is different. But then it must not be an ndarray. The only occurrence of `coord_exp` is here and in context of the `in` operator.\
Hence it can be any kind of sequence: a string, a list, an array, a set, a dict, a generator, an iterator etc.\
If it's not such a thing the in operator will simply throw an exception.\
But unlike in the original version, it will not be ignored! Well, yes, I think that's preferable.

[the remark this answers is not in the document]
chahank: Ok, good point. It should indeed not be ignored if it is not an iterable.

climada/engine/impact.py
emanuel-schmid: Really? What I can see now is the influence on `imp.coord_exp` in lines 1104 and 1105, but I can't see how those are changed. Can you point it out?

climada/engine/impact.py
```diff
@@ -940,6 +946,172 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        freq_mat = imp.frequency.reshape(len(imp.frequency), 1)
+        # .A1 reduce 1d matrix to 1d array
+        imp.eai_exp = imp.imp_mat.multiply(freq_mat).sum(axis=0).A1 
+        imp.aai_agg = imp.eai_exp.sum()
+
+        if len(sel_exp) < nb_exp:
+            imp.coord_exp = imp.coord_exp[sel_exp]
+            # .A1 reduce 1d matrix to 1d array
+            imp.at_event = imp.imp_mat.sum(axis=1).A1
+            imp.tot_value = None
```
emanuel-schmid, line 1108: else these attributes are not set?

climada/engine/impact.py
emanuel-schmid: I can see `imp.at_event`, `imp.coord_exp` and `imp.tot_value` below, true. (Though not yet `imp.eai_exp` or `imp.aai_agg`?)
But still - if `value` is an array, couldn't it have an exposure dimension as well?

chahank: The `imp.eai_exp` and `imp.aai_agg` are computed from `imp_mat` which is set in line 1089 and which has indices select by `sel_exp`. Hence, if `sel_exp` is not None, the columns of `imp_mat` are modified, and then in turn `imp.eai_exp` and `imp.aai_agg` (in lines 1101, 1102).

chahank: Furthermore, yes,  if `value` is an array, it could have an exposure dimension as well. However, due to what the impact is, it is rather unlikely that this happens. So far, I decided not to do the same treatment as for event dimensions. I fear that if further values with exposure dimension are added to the impact object, it will probably require a rewriting of the select method to make sure that the other impact values are correctly recomputed. Thus, we could hide a potential future issue here. But maybe I should rethink about that a bit more.

chahank: I could rewrite the method `get_per_event_attr()` to be `get_per_attr(per_what)` and easily make the same logic for per event as for per events. Should I then place this new `get_per_attr(per_what)` function in util maybe?

emanuel-schmid: Why not make it a method of Impact?

emanuel-schmid: > 
> 
> ... However, due to what the impact is, it is rather unlikely that this happens. ...
Afaik "unlikely" is just the same as "granted" 😜

chahank: > Why not make it a method of Impact?

It could be used in all the `select` methods which appear/will appear in different CLIMADA classes.

chahank: > > ... However, due to what the impact is, it is rather unlikely that this happens. ...
> 
> Afaik "unlikely" is just the same as "granted" 😜

Yes, but this is part of the reflection. If it happens, then I suspect that the method would need some direct attention. But, I guess that the unit tests should take care of that.

emanuel-schmid: > 
> It could be used in all the `select` methods which appear/will appear in different CLIMADA classes.

Alright. In this case I suggest to call it by number, e.g.:
```
def get_attributes_with_matching_dimension(n: int):
    ...
    return list_of_attributes
```

emanuel-schmid: or even
```
def get_attributes_with_matching_dimension(*ns: int):
    ...
    return list_of_attributes # that have len(ns) dimensions or more, and all ns are individual part of shape
```

climada/engine/impact.py
chahank: They are not modified.

climada/engine/impact.py
chahank: So, I added the more general function to `util/select.py` (still need to write the test for that). 

I also slightly restructured the `impact.select` to use the more general attribute select function and added warnings when no events/exposures are found.

chahank: Ok, all done.

chahank:
@emanuel-schmid  Should I proceed with the merge?

emanuel-schmid:
transferred to PR #93

PR #90, opened 2020-11-17
Feature/select tc

chahank:
Add method to select from a TCTracks() object a subset of tracks intersecting with an exposure.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -27,9 +27,6 @@ import glob
 import itertools
 import logging
 import os
-import re
-import shutil
-import warnings
 import shutil
 import warnings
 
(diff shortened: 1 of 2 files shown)


simonameiler (commented):
A neat little feature that allows filtering for tracks that are in the vicinity of an exposure. This can be useful for TC track sets with a lot of storms that lie outside of the exposure of interest as a pre-processing step to the wind field calculation.

climada/hazard/tc_tracks.py
```diff
@@ -210,6 +207,46 @@ def subset(self, filterdict):
...
+        -------
+        filtered_tracks : climada.hazard.TCTracks()
+            TCTracks object with tracks from tc_tracks intersecting
+            the exposure whitin a buffer distance.
+
+        """
+
+        if buffer <= 0:
+            buffer = 1.0
+            LOGGER.warning(f"buffer={buffer} is invalid. buffer set to 1.0")
```
emanuel-schmid, line 236: Is there a reason for not just raising an exception if a value below zero is provided? 
```suggestion
        if buffer <= 0.0:
            raise ValueError(f"buffer={buffer} is invalid, must be above zero.")
```
What is the problem with a buffer of 0.0?

emanuel-schmid (changes requested):

climada/hazard/tc_tracks.py
chahank: This is based on the idea that better go to default values when obviously wrong values are given rather than ask for permission. A buffer at 0.0 is a bit of edge case where the whole geometry is collapsed to nothing instead of being actually a buffer of size 0. Hence, I do not allow for it.

emanuel-schmid: When obviously wrong values are given they should be rejected (without asking for permission). I'm opposed to auto-correction. It spoils the opportunity to fix errors early.

chahank:
I need this feature now. If more review is needed, please let me know.

PR #91, opened 2020-11-18
Feature/sig dig

chahank:
Add a math module to the util module. Included functions are: sig_dig which rounds numbers to n significant digits and money_unit which converts a list of numbers to a thousander unit of money ('K', 'M', 'Bn', 'Tn').


mmyrte:
Ok, I don't have the time to get into this, but I'm a bit weary of the confusion that could arise from having `math` and `util.math`. It's not really related to math functions, but to pretty printing...

manniepmkam:
Agree with @mmyrte that the name `climada.util.math` is confusing with the module `math`. Also, `money_unit` for me would be currencies (e.g. USD, GPB, CHF), a suggestion could be `money_abbrev`?

The rest of the codes looks good to me.

chahank:
I agree with both of you. Any ideas how to name the module? Because it is not only about printing, it is also about rounding. So maybe `number_representation` or `number_repr` or `value_repr` ?

And maybe instead of `money_unit` how about `value_to_monetary_unit` or `convert_to_monetary_unit` ?

manniepmkam:
I'm ok with 3 of the module names. Will this module include more functions in the future though?

I would prefer `value_to_monetary_unit`, sounds the most understandable for me (however I don't have any business/finance background).

chahank:
Any other comments? On the code itself (@manniepmkam please have a look at the reviewer's guidelines if you need too) ? On the tests?

climada/util/value_representation.py
```diff
@@ -0,0 +1,133 @@
...
+    -------
+    mon_val : np.ndarray
+        Array of values in monetary unit 
+    name : string
+        Monetary unit
+
+    """
+    
+    if not isinstance(values, list):
+        values = [values]
```
emanuel-schmid, line 99: The docstring says `ndarray` is a possible type too. That seems contradicting to just checking for a list.
In general, if multiple argument types are allowed, addressing them all in a `if isinstance(...)` block is probably a good idea. (Or none when duck-typing.)

climada/util/value_representation.py
```diff
@@ -0,0 +1,133 @@
...
+    Returns
+    -------
+    list
+        list of rounded floats
+
+    """
+    return np.vectorize(sig_dig)(iterable, n_sig_dig)
+
+
+def value_to_monetary_unit(values, n_sig_dig=None, abbreviations=None):
```
emanuel-schmid, line 68: There ain't no tests for `value_to_monetary_unit` - or are there?

climada/util/value_representation.py
```diff
@@ -0,0 +1,133 @@
...
+    mil_exp = 3 * math.floor(avg_exp/3)
+    
+    name = ''
+    thsder = int(10**mil_exp)
+    
+    try:
+        name = abbreviations[thsder] 
+    except KeyError: 
+        LOGGER.warning(f"Warning: The numbers are larger than \
+{list(abbreviations.keys())[-1]}")
```
emanuel-schmid, line 124:
```suggestion
        LOGGER.warning("Warning: The numbers are larger than %s", list(abbreviations.keys())[-1])
```
The LOGGER prefers lazy execution.

climada/util/value_representation.py
```diff
@@ -0,0 +1,133 @@
...
+    for val in values:
+        if val == 0:
+            exponents.append(0)
+            continue
+        exponents.append(math.log10(abs(val)))
+        
+    max_exp = max(exponents)
+    min_exp = min(exponents)
+    
+    avg_exp = math.floor((max_exp + min_exp) / 2) #rounded down
```
emanuel-schmid, line 114:
```suggestion
    avg_exp = math.floor((max_exp + min_exp) / 2)  # rounded down
```
PEP8

emanuel-schmid (changes requested):

climada/util/value_representation.py
chahank: Yes, forgot to rename.

emanuel-schmid (approved):

PR #92, opened 2020-11-19
Fix unit tests for new random walk (calc_perturbed_trajectories)

bguillod:
I forgot to edit the unit tests so these were not passed after renaming `calc_random_walk` to `calc_perturbed_trajectories`. This pull requests attempts to fix that.


bguillod:
@emanuel-schmid looks like you've fixed that already so that branch is probably to be deleted...

emanuel-schmid (approved):

Issue #86, opened 2020-11-06
Pool reset in climada.hazard.TropCyclone.set_from_tracks


bguillod:
The following code should allow to create an event set from TC tracks using parallel computation:
```
from pathos.pools import ProcessPool as Pool

pool = Pool(nodes=4)
tc_holland = TropCyclone(pool=pool)
tc_holland.set_from_tracks(tracks, centroids, ignore_distance_to_coast=True)
tc_holland.check()

pool.close()
pool.join()
```

However in `set_from_tracks`, `self.concatenate` calls `self.clear`, which does:
```
        for (var_name, var_val) in self.__dict__.items():
            if isinstance(var_val, np.ndarray) and var_val.ndim == 1:
                setattr(self, var_name, np.array([], dtype=var_val.dtype))
            elif isinstance(var_val, sparse.csr_matrix):
                setattr(self, var_name, sparse.csr_matrix(np.empty((0, 0))))
            else:
                setattr(self, var_name, var_val.__class__())
```

Hence, self.pool will be reset. If the number of nodes specified in the call to Pool above is different from the number of CPUs, a new pool will be created and will replace self.pool. The original pool is not removed, meaning that we now have two sets of pools. Also, a user might want to use tc_holland.pool somewhere and might be surprised it uses a different number of cores (and different nodes) than the pool (s)he setup in the first place, as was my case.

I can see two options to fix this:
1) In set_from_tracks, add the following lines around self.concatenate:
```
pool = self.pool
self.pool=None
self.concatenate(tc_haz)
self.pool=pool
```
2) In `clear` (hazard.base), replace the `else` statement by:
```
elif not isinstance(var_val, Pool):
```
to ensure a Pool is not modified, and add `Pool` in the header:
```
from pathos.pools import ProcessPool as Pool
```

In my case at least this seems to fix the issue. I am not sure which of both options are best. Also not sure if this should be part of a feature branch and/or who wants to implement but I think fixing this is worth it.

tovogt:
I would opt for changing `hazard.base.Hazard.clear`, as suggested, to not modify instances of `Pool`. This can be explained in the docstring by explicitly specifying that it "reinitializes data attributes (not the process Pool)". Can you create a PR?

PR #95, opened 2020-11-24
Prevent hazard.base.Hazard.clear from resetting a pool

bguillod:
This fixes issue #86: If a hazard attribute is a pathos pool, it won't be reset.

### DIFF
--- a/climada/hazard/test/test_base.py
+++ b/climada/hazard/test/test_base.py
@@ -30,6 +30,7 @@ from climada.hazard.centroids.centr import Centroids
 import climada.util.dates_times as u_dt
 from climada.util.constants import HAZ_TEMPLATE_XLS, HAZ_DEMO_FL
 from climada.util.coordinates import equal_crs
+from pathos.pools import ProcessPool as Pool
 
 DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
 HAZ_TEST_MAT = os.path.join(DATA_DIR, 'atl_prob_no_name.mat')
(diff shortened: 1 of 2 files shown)


tovogt (changes requested):
Thanks Benoit! Would you please add a small regression test for this?

bguillod:
> Thanks Benoit! Would you please add a small regression test for this?

@tovogt Hmm yes I thought about adding a test but to be honest I really wasn't sure what it should be. Would the following be fine:

1. Test whether if a pool is attached to a hazard instance, when cleared the pool is unchanged.
2. Test whether a hazard instance without pool is correctly cleared.

Is that what you had in mind? My preference would be to do it in my use case, i.e. actually on a TropCyclone instance.

tovogt:
Since it's a change to the base class, the test should be for the base class as well. As there is no test for the `clear` function at all in `climada/hazard/test/test_base.py` it would probably make sense to write a new one that tests both, whether other properties are reset, while the `pool` property remains unchanged.

chahank (approved):
I agree with the comments by @tovogt .

bguillod:
@tovogt I've implemented some test, let me know if that's along the lines of what you had in mind.

climada/hazard/test/test_base.py
```diff
@@ -1104,6 +1105,35 @@ def test_vector_to_raster_pass(self):
...
+
+class TestClear(unittest.TestCase):
+    """Test clear method"""
+
+    def test_clear(self):
+        """Clear method clears everything"""
+        haz1 = Hazard('TC')
+        haz1.read_excel(HAZ_TEMPLATE_XLS)
+        haz1.clear()
+        haz1.centroids.size
```
tovogt, line 1117: This line should probably be removed.

climada/hazard/test/test_base.py
```diff
@@ -1104,6 +1105,35 @@ def test_vector_to_raster_pass(self):
...
+        self.assertEqual([val for key, val in haz1.tag.__dict__.items()], ['', '', ''])
+        self.assertEqual(haz1.units, '')
+        self.assertEqual(haz1.centroids.size, 0)
+        self.assertEqual(len(haz1.event_name), 0)
+        for attr in list(set(haz1.__dict__.keys()).difference({'tag', 'units', 'event_name', 'pool'})):
+            self.assertEqual(haz1.__dict__[attr].size, 0)
```
tovogt, line 1123: Avoid using the double-underscore notation as much as possible:
```suggestion
        for attr in vars(haz1).keys():
            if attr not in ['tag', 'units', 'event_name', 'pool']:
                self.assertEqual(getattr(haz1, attr).size, 0)
```

climada/hazard/test/test_base.py
```diff
@@ -1104,6 +1105,35 @@ def test_vector_to_raster_pass(self):
...
+class TestClear(unittest.TestCase):
+    """Test clear method"""
+
+    def test_clear(self):
+        """Clear method clears everything"""
+        haz1 = Hazard('TC')
+        haz1.read_excel(HAZ_TEMPLATE_XLS)
+        haz1.clear()
+        haz1.centroids.size
+        self.assertEqual([val for key, val in haz1.tag.__dict__.items()], ['', '', ''])
```
tovogt, line 1118:
```suggestion
        self.assertEqual(vars(haz1.tag).values(), ['', '', ''])
```

climada/hazard/test/test_base.py
```diff
@@ -1104,6 +1105,35 @@ def test_vector_to_raster_pass(self):
...
         self.assertTrue(haz_fl.fraction.max() <= 0.5 / 2)
 
+
+class TestClear(unittest.TestCase):
+    """Test clear method"""
+
+    def test_clear(self):
+        """Clear method clears everything"""
+        haz1 = Hazard('TC')
+        haz1.read_excel(HAZ_TEMPLATE_XLS)
```
tovogt, line 1115: Manually set some more attributes (such as the `units` attribute) to make sure that they are actually touched by `clear`. 
```suggestion
        haz1.read_excel(HAZ_TEMPLATE_XLS)
        haz1.units = "m"
        haz1.foo = np.arange(10)
```

tovogt (changes requested):
Great! Just some minor remarks left :+1:

climada/hazard/test/test_base.py
bguillod: On that one: haz1.foo will be reset to an empty numpy array. Are we fine with that or do we want all non-standard attributes of the hazard object to be removed entirely from the object?

tovogt: Since the `clear` function from the base class is used in all derived classes, it's currently the easiest way to support additional attributes, because we don't currently define which attributes are considered "standard". So, for this PR, I think we should stick with this behavior. In the future, we might be defining something like "standard" attributes for derived Hazard classes etc.

bguillod:
Ok so from my perspective this is done now.

PR #97, opened 2020-11-26
Feature: TC surge from bathtub model

tovogt:
Due to problems with the `elevation` package under Windows (issue #9), the old surge module has been removed from the `main` and `develop` branches and archived under [archive/feature/tcsurge](https://github.com/CLIMADA-project/climada_python/tree/archive/feature/tcsurge). A completely new TC surge module is in the pipeline in the `feature/tc_surge_geoclaw` branch. However, due to it's heavy computational needs, there is still some interest in the old simplified approach, most recently by @QinhanZhu and @arunranain who managed to remove the dependency on the `elevation` package from the old module.

I added some references from the literature, implemented unit tests and modified the tutorial notebook. As discussed in #9, users are now required to provide their own DEM data set as a raster file. Two freely available data sets are mentioned in the tutorial notebook. I think, the module is now ready to be merged back into the develop branch.

*Naming:* I called the old model a "bathtub" model to make clear that it's a very simplified approach. The term "bathtub model" is often used in the storm surge literature to refer to a model that ignores all hydrodynamics and instead basically subtracts a digital elevation model from an assumed surge height.

@chahank: Please feel free to assign more/other reviewers.


climada/hazard/centroids/centr.py
```diff
@@ -574,16 +575,29 @@ def set_area_approx(self, min_resol=1.0e-8):
             LOGGER.error('Pixel area of points can not be computed.')
             raise ValueError
 
-    def set_dist_coast(self, signed=False, precomputed=False, scheduler=None):
-        """Set dist_coast attribute for every pixel or point. Distance to
-        coast is computed in meters.
+    def set_elevation(self, topo_path):
```
tovogt, line 578: This function might come in handy in the future. It sets the `elevation` property from a given elevation raster file. Currently, it's only used in the tutorial notebook for illustration purposes.

climada/hazard/centroids/centr.py
```diff
@@ -574,16 +575,29 @@ def set_area_approx(self, min_resol=1.0e-8):
...
+        """Set dist_coast attribute for every pixel or point in meters.
+
+        Parameters
+        ----------
+        signed : bool
+            If True, use signed distances (positive off shore and negative on land). Default: False.
+        precomputed : bool
+            If True, use precomputed distances (from NASA). Default: False.
+        scheduler : str
+            Used for dask map_partitions. "threads", "synchronous" or "processes"
```
tovogt, line 600: I changed this docstring format to NumPy style while I was at it.

chahank:
Were these changes discussed with @QinhanZhu and @arunranain ?

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,179 @@
...
+            Surge decays according to distance from coast. Default: True.
+        add_sea_level_rise : float, optional
+            Sea level rise effect in meters to be added to surge height.
+        """
+        centroids = copy.deepcopy(wind_haz.centroids)
+
+        if not centroids.coord.size:
+            centroids.set_meta_to_lat_lon()
+
+        # Select wind-affected centroids which are inside INLAND_MAX_DIST_KM and |lat| < 61
```
Evelyn-M, line 80: detail: var is called MAX_DIST_COAST, not INLAND_MAX_DIST_KM

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # elevation at coastal centroids, exclude high centroids
+        coastal_centroids_h = coord_util.read_raster_sample(
+            topo_path, centroids.lat[coastal_msk], centroids.lon[coastal_msk])
+        elevation_msk = ((coastal_centroids_h >= 0)
+                         & (coastal_centroids_h <= MAX_ELEVATION + add_sea_level_rise))
+        coastal_msk[coastal_msk] = elevation_msk
+        coastal_centroids_h = coastal_centroids_h[elevation_msk]
+        coastal_idx = coastal_msk.nonzero()[0]
+
```
Evelyn-M, line 96: This part I find very hard to follow. 
Could one add maybe a bit more explanation? Like:
`coastal_msk` is a boolean matrix indexing all coastal centroids, 
`elevation_msk` is a boolean matrix indexing all coastal centroids that are additionally within 0 and MAX_elevation + s.l.rise
Explaining the indexing / expansion and reassigning of `coastal_msk` in line 89

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,179 @@
...
+        #   In: 29th Conference on Hurricanes and Tropical Meteorology, 10–14 May. Tucson, Arizona.
+        #   https://ams.confex.com/ams/pdfpapers/168806.pdf
+        inten_surge.data = 0.1023 * np.fmax(inten_surge.data - 26.8224, 0) + 1.8288
+
+        if inland_decay:
+            # add decay according to distance from coast
+            dist_coast_km = np.abs(centroids.dist_coast[coastal_idx]) / 1000
+            coastal_centroids_h += DECAY_RATE_M_KM * dist_coast_km
+        coastal_centroids_h -= add_sea_level_rise
+        inten_surge[:,coastal_idx] -= coastal_centroids_h[None,:]
```
Evelyn-M: Detail: I am not 100% sure what we said on this kind of notation, but I have vague memories that Chahan just added sth to the style guide regarding avoiding this.

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # discard negative surge height values
+        inten_surge.data = np.fmax(inten_surge.data, 0)
+        inten_surge.eliminate_zeros()
+
+        # get fraction of centroid cells on land according to given DEM
+        fract_surge = sp.csr_matrix(_fraction_on_land(centroids, topo_path))
+        fract_surge = sp.csr_matrix(np.ones((inten_surge.shape[0], 1))) * fract_surge
```
Evelyn-M, line 127: Would be great if the purpose of `_fraction_on_land()` could be motivated a bit more (mostly in docstring below, perhaps, but also add why this is still necessary here). 
Do I understand this correctly that, due to resolution issues of the centroids, despite masking those that are "coastal" above, some of the coastal ones very close to the coast-line are actually partially in the sea? And the (higher-resolution) DEM helps identify this fact and correct for it by setting the `fraction`?

Evelyn-M (commented):
Thanks a lot for the work, I just have a few, mostly conceptual / readability comments

tovogt:
Thanks for the valuable feedback @Evelyn-M! I will comment directly to each of your points.

> Were these changes discussed with @QinhanZhu and @arunranain ?

They gave me access to their modified version and communicated that they don't plan to continue working on it to transform it into an update to the CLIMADA core code.

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,179 @@
...
+
+MAX_DIST_COAST = 50
+"""Maximum inland distance of the centroids in km"""
+
+MAX_ELEVATION = 10
+"""Maximum elevation of the centroids in m"""
+
+DECAY_RATE_M_KM = 0.2
+"""Decay rate of surge in meters per each km"""
+
```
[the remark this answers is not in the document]
tovogt: Well spotted. I decided to have the decay rate as a user parameter in my latest commit.

However, I think that `MAX_DIST_COAST` and `MAX_ELEVATION` shouldn't be considered modeling choices. They are chosen so high, that they don't interfere with the model mechanics. Their purpose is to exclude non-relevant centroids early-on for performance reasons. Independent of any surge model, there should never be any surge for centroids lying that high or that far from the coast. Including them as keyword arguments doesn't improve usability. We have to remember that too many function arguments can overwhelm and confuse users.

climada/hazard/tc_surge_bathtub.py
tovogt: It is generally considered bad practice to put information in comments that can be easily deferred from the code itself. The fact that `coastal_msk` and `elevation_msk` are boolean arrays is such an information, as well as the constraint values `0` and `MAX_ELEVATION + slr`.

However, I decided to add a comment in my latest commit. Maybe that helps.

climada/hazard/tc_surge_bathtub.py
tovogt: This is not mere "notation". How we write this does affect the performance of the code. The in-place operations `+=`, `-=`, `*=` and so forth are very efficient and should be used as much as possible.

climada/hazard/tc_surge_bathtub.py
chahank: Well, but if the code is not self-evident, then something is sub-optimal. Either the variable name makes it clear that it is a boolean, or the context or a comment.

climada/hazard/tc_surge_bathtub.py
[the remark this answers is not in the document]
chahank: Does this also apply to very flat places like e.g. the Netherlands?

climada/hazard/tc_surge_bathtub.py
tovogt: I added some more explanation in the docstring of `_fraction_on_land`.

climada/hazard/tc_surge_bathtub.py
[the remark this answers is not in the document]
tovogt: Yes, because of the decay. Even with a low decay rate and a very low-lying point (5-6 meters below sea level), the surge will be gone after 50 km. However, the Netherlands are never affected by tropical cyclones, so it doesn't really make sense to go into the details.

climada/hazard/tc_surge_bathtub.py
tovogt: Yes, "IF the code is not self-evident". That's exactly what I said: Don't put stuff in the comments that is self-evident.

Evelyn-M: yes, ok, I agree. the first instances of coastal_msk and elevation_msk can be figured out relatively easily, so if commenting on that is bad practice, so be it.
But as said, with line 89 `coastal_msk[coastal_msk] = elevation_msk`i really struggle and don't find it very self-evident what the expansion and re-assigning does.

climada/hazard/tc_surge_bathtub.py
Evelyn-M: Ok, good to know! @chahank , side note to your comment for the climada coding convention today, then - 
with saying
> avoid complex numpy expressions, e.g. `a = b + c * d` vs. `a = c * d; a += b` 
you mean the first should be discouraged, the second encouraged, hence? Then that was just an understanding issue from my side!

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,192 @@
...
+                       & (np.abs(centroids.lat) < 61))
+
+        # Load elevation at coastal centroids
+        coastal_centroids_h = coord_util.read_raster_sample(
+            topo_path, centroids.lat[coastal_msk], centroids.lon[coastal_msk])
+
+        # Update selected coastal centroids to exclude high-lying locations
+        elevation_msk = ((coastal_centroids_h >= 0)
+                         & (coastal_centroids_h <= MAX_ELEVATION + add_sea_level_rise))
+        coastal_msk[coastal_msk] = elevation_msk
```
tovogt, line 91: The expression `coastal_msk[coastal_msk] = elevation_msk` looks a bit strange, but it's exactly doing what the comment says: it updates the selected coastal centroids to exclude high-lying locations. If you think this still needs clarification, please say how we would clarify this. For me, it's just the usual way how to update the True entries of a numpy mask.

climada/hazard/tc_surge_bathtub.py
tovogt: Please continue the discussion in the updated code: https://github.com/CLIMADA-project/climada_python/pull/97/files#r531497507

climada/hazard/tc_surge_bathtub.py
chahank: Yes exactly. Maybe the guide should be rephrased then.

climada/hazard/tc_surge_bathtub.py
chahank: I am also confused by this way of writing it. Why is it not coastal_mask = elevation_msk or coastal_mask = (coastal_msk & elevation_mask) ?

chahank:
> Thanks for the valuable feedback @Evelyn-M! I will comment directly on each of your points.
> 
> > Were these changes discussed with @QinhanZhu and @arunranain ?
> 
> They gave me access to their modified version and communicated that they don't plan to continue working on a transformation into an update for the CLIMADA core.

Is the new implementation compatible with the way they used the TC_surge module for the Vietnam study? Because they are also modifying the TC_surge module at the moment to make it possible to use any DEM model. The crux is the clipping part. I just want to be sure that not two competing version of the same module are developed here.

climada/hazard/tc_surge_bathtub.py
tovogt: The `coastal_centroids_h` information is only obtained for the previously selected centroids. So we only update the previously selected centroids and not those that have been discarded in the previous step. Your proposed alternatives simply don't do the same thing.
```suggestion
        # Update selected coastal centroids to exclude high-lying locations
        elevation_msk = ((coastal_centroids_h >= 0)
                         & (coastal_centroids_h <= MAX_ELEVATION + add_sea_level_rise))
        # Only update the previously selected centroids (for which elevation information has been obtained)
        coastal_msk[coastal_msk] = elevation_msk
```

tovogt:
> Is the new implementation compatible with the way they used the TC_surge module for the Vietnam study? Because they are also modifying the TC_surge module at the moment to make it possible to use any DEM model. The crux is the clipping part. I just want to be sure that not two competing version of the same module are developed here.

As I said, they gave me access to their code and no, they are not modifying it at the moment - they actually stopped working on it. If I'm not mistaken, my PR is 100% compatible with what they did and it's supposed to reproduced the exact behavior of the legacy `tc_surge` module with a somewhat polished API and improved implementation.

chahank:
OK great. Just wanted to be sure. But could you please double check with @QinhanZhu as I talked to him recently and (probably I misunderstood) he seemed to imply to wanting to change something about the clipping of DEM in TC_surge.

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,192 @@
...
+        if not centroids.coord.size:
+            centroids.set_meta_to_lat_lon()
+
+        # Select wind-affected centroids which are inside MAX_DIST_COAST and |lat| < 61
+        if not centroids.dist_coast.size or np.all(centroids.dist_coast >= 0):
+            centroids.set_dist_coast(signed=True, precomputed=True)
+        coastal_msk = (((wind_haz.intensity > 0).sum(axis=0).A1 > 0)
+                       & (centroids.dist_coast < 0)
+                       & (centroids.dist_coast > -MAX_DIST_COAST * 1000)
+                       & (np.abs(centroids.lat) < 61))
```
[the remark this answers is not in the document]
tovogt: I added a constant for this.

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,193 @@
...
```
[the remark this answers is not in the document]
tovogt: Thanks, correct!

climada/hazard/tc_surge_bathtub.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    bounds = np.array(centroids.total_bounds)
+    shape = [0, 0]
+    if centroids.meta:
+        shape = centroids.shape
+        cen_trans = centroids.meta['transform']
+    else:
+        shape[0], shape[1], cen_trans = coord_util.pts_to_raster_meta(
+            bounds, min(coord_util.get_resolution(centroids.lat, centroids.lon)))
+    bounds += np.array([-.05, -.05, .05, .05])
```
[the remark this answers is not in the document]
tovogt: I just left it like it was in the legacy module. However, it's an arbitrary buffer value and, in my last commit, I now decided to change it to a more generic value based on the resolution.

tovogt:
Thanks @emanuel-schmid, I addressed all of your points in my latest commit.

PR #99, opened 2020-11-27
river_flood: replace large demo files with smaller ones

ingajsa:


climada/hazard/test/test_flood.py
```diff
@@ -78,18 +78,6 @@ def test_exact_area_selection_country(self):
...
-        testCentr, isos, natIDs = RiverFlood._select_exact_area(reg=['SWA'])
-
-        self.assertEqual(testCentr.shape, (877, 976))
-        self.assertAlmostEqual(np.min(testCentr.lat), -0.68767620000001, 4)
-        self.assertAlmostEqual(np.max(testCentr.lat), 38.43726119999998, 4)
-        self.assertAlmostEqual(np.min(testCentr.lon), 60.52061519999998, 4)
-        self.assertAlmostEqual(np.max(testCentr.lon), 101.1455501999999, 4)
-        self.assertAlmostEqual(testCentr.lon[10000], 98.27055479999999, 4)
-        self.assertAlmostEqual(testCentr.lat[10000], 11.47897099999998, 4)
-
```
tovogt, line 92: Why did you decide to remove this? Isn't this a valid test case?

tovogt (commented):
Looks good to me. I was only wondering why you removed the `test_exact_area_selection_region` test case. Maybe you can comment on that?

chahank:
@simonameiler Could you please have a look at the tutorial and see if it is compatible with the guidelines we set in the tutorial update sub-group?

simonameiler (commented):
This review concerns the **tutorial (climada_hazard_RiverFlood.ipynb)** only: 
Overall, the tutorial is well-structured and comprehensive. The data structure is explained, access to data and basic background information are given. 

I suggest adding a table of contents before the actual tutorial starts. Consider for example the following tutorial as template: [https://github.com/CLIMADA-project/climada_python/blob/tutorial_update/doc/tutorial/climada_hazard_Hazard.ipynb](url) 

Also, note that there is a typo/piece of strikethrough text in the second cell on ISIMIP2a.

emanuel-schmid:
Looks good to me.\
Cosmetic suggestion: I'd remove the Tags from the tutorial notebook. None of the other notebooks seem to have them and thus it stays out of line.

emanuel-schmid (commented):
Why are there Tag editors in the tutorial?

chahank:
What are Tag editors?

climada/hazard/test/test_flood.py
ingajsa: It is a valid test case, but this function refers to the selection of an entire region, so it is no longer possible with a test file covering only two countries.

ingajsa: I further don't think that the functionality of choosing an entire region instead of single countries is so important that it is worth having a larger test-file in there, that's why I decided to simply remove it instead of cutting an entire region as a test file.

tovogt: Maybe there's a misunderstanding what this function tests, because I think it doesn't rely on the test-file at all. It only loads grid points (not the values, only the coordinates) and checks whether the right grid has been loaded and truncated.

ingajsa: Yes, you are right. I will bring it back in.

ingajsa: Yes, you are of course right. Thanks, for pointing this out.

ingajsa:
> This review concerns the **tutorial (climada_hazard_RiverFlood.ipynb)** only:
> Overall, the tutorial is well-structured and comprehensive. The data structure is explained, access to data and basic background information are given.
> 
> I suggest adding a table of contents before the actual tutorial starts. Consider for example the following tutorial as template: [https://github.com/CLIMADA-project/climada_python/blob/tutorial_update/doc/tutorial/climada_hazard_Hazard.ipynb](url)
> 
> Also, note that there is a typo/piece of strikethrough text in the second cell on ISIMIP2a.

Thank you for the suggestions, I included them in the tutorial

ingajsa:
> Looks good to me.

> Cosmetic suggestion: I'd remove the Tags from the tutorial notebook. None of the other notebooks seem to have them and thus it stays out of line.

@emanuel-schmid  they are removed

ingajsa:
@chahank I think it is ready for the merge now. Shall I do it, or will you do it?

chahank:
If you have addressed all questions/remarks you can do it yourself. Maybe wait one day to see if one of the reviewers has something to add.

tovogt (approved):

emanuel-schmid (approved):

simonameiler (approved):

Issue #94, opened 2020-11-24
to_geodataframe method in TCTracks does not export point data when as_points=True


bguillod:
I was testing the new `to_geodataframe` method of TCTracks and I was surprised that when setting `as_points=True` the point data (max_sustained_wind, radius_max_wind, etc) are not in the dataframe.
My assumption was that:

1. If as_points is False, then only the tracks attributes are in the dataframe (as is now).
2. If as_points is True, all point properties are provided.

Should I create a pull request with these changes? What are others' opinion on this?

chahank:
I have no idea what the intended purpose of this attribute is, but it seems reasonable. A better person to answer this question might be @ThomasRoosli. From my point of view go ahead.

bguillod:
Ok, I’ve already coded that for myself anyway so I will put that in a branch and make a pull request next week, unless someone has something against it. It nicely allows to export tracks with properties, e.g. as a shapefile. And if someone doesn’t want these variables s/he can remove them from the data frame afterwards easily anyway.

ThomasRoosli:
I only know that it is a feature added by @mmyrte and that the as_points functionality was discussed in PR #79. From my side you can go ahead.

mmyrte:
I'm sorry I was absent for this - was busy with my thesis, among other things. I'm closing this issue, although I have some comments on PR #101; you're using an `apply` where I used a vectorised form.

PR #101, opened 2020-12-02
Feature/tc tracks to geodataframe data

bguillod:
This PR addresses issue #94:
When exporting TC tracks to geopandas dataframe (function `to_geodataframe` in TCTracks), variable data are also present in the geodataframe if `as_points` is set to `True`. Previously, only tracks attributes were saved in the dataframe. Backward compatibility is fine, because users can then drop columns they don't need from the dataframe.


climada/hazard/tc_tracks.py
```diff
@@ -1016,10 +1016,10 @@ def to_geodataframe(self, as_points=False):
...
 
         Parameters:
-            as_points (bool): If true, construct LineString or single Point
-                geometries. A point is created if a track has length one.
-                If set to false, the geometries are returned as points with an
-                additional timestamp column.
+            as_points : bool, optional
+                If False (default), one row per track with a LineString as geometry (or Point
+                geometry for tracks of length one) and all track attributes as dataframe columns.
+                If True, one row per track point, with variable values in addition to attributes.
```
chahank, line 1022: Could you clarify what 'variable values' are? Which variables? Also, what attributes? Do you mean it in the pythonic sense?

climada/hazard/tc_tracks.py
```diff
@@ -1029,22 +1029,11 @@ def to_geodataframe(self, as_points=False):
...
-                'time': times,
-                'geometry': [Point(p[0], p[1]) for p in points],
-            }).set_index('idx')
-
-            gdf = gdf.join(gdf_long)
+            gdf_long = pd.concat([track.to_dataframe().assign(idx=i) for i, track in enumerate(self.data)])
+            gdf_long['geometry'] = gdf_long.apply(lambda x: Point(x['lon'],x['lat']), axis=1)
+            gdf_long = gdf_long.drop(columns=['lon', 'lat'])
+            gdf_long = gpd.GeoDataFrame(gdf_long.reset_index().set_index('idx'), geometry='geometry')
+            gdf = gdf_long.join(gdf)
```
chahank, line 1036: Why does the join work here? Do the dataframes made from the tracks not have much less rows than the one made point the track points?

climada/hazard/tc_tracks.py
bguillod: I meant in the xarray sense. Attributes are unique for an xarray object i.e. for a track, e.g. track id, name, category, etc. Variables are for each time step, e.g. max sustained wind, central pressure, etc.
Hence, attributes can be set for whole tracks or to individual points (same for all points within a given track), while variables cannot be given to the line, but only to the points.

climada/hazard/tc_tracks.py
bguillod: > Do the dataframes made from the tracks not have much less rows than the one made point the track points?
Yes, exactly. So for the point case, you want to provide the track attributes for each point, hence the join.

climada/hazard/tc_tracks.py
chahank: Ok. This should be explained better in the docstring. Currently it is very confusing.

climada/hazard/tc_tracks.py
chahank: Ok, sounds correct. I probably misunderstood how join is working.

chahank:
Please make the changes to the docstring and then make an integration test to be sure that no other parts are broken since this function is used in other parts of the code. If it passes, please merge.

climada/hazard/tc_tracks.py
bguillod: Ok I've improved it, better now?

bguillod:
> Please make the changes to the docstring and then make an integration test to be sure that no other parts are broken since this function is used in other parts of the code. If it passes, please merge.

Actually, this function is not used anywhere else in the code besides in unit test. Hence we don't need an integration test, right?

chahank:
It is in fact used in the recently added tracks_in_exp.

climada/hazard/tc_tracks.py
chahank: Excellent!

mmyrte:
Thanks @bguillod for the expansion of the functionality - my constrained version tried to vectorize as much as possible, at the cost of dropping variables that may or may not be present in the original data.

Just for future reference, if we ever notice that this function is eating up too many cycles or too much memory (which I doubt, unless someone is working with >>1e5 tracks), we might optimise it by avoiding the `xr.Datasets.to_dataframe` calls and instead pre-allocating memory for numpy arrays. The xy-coords can be cast without copying using `shapely.geometry.asShape`, as [indicated here](https://shapely.readthedocs.io/en/stable/manual.html#numpy-and-python-arrays). Also, the anonymous function call in the apply statement likely introduces some overhead versus an enumeration calling `shapely.geometry.Point`.

PR #102, opened 2020-12-02
TCTracks: interpolate with arbitrary resolution

tovogt:
The function `TCTracks.equal_timestep` is currently limited to hourly resolution time steps. I don't think that we have to support arbitrary small time steps, but minute resolution might be useful for some case studies. Since supporting minute resolution is a very easy change and 100% backward compatible, I quickly created a PR for this (with unit test).

While I was at it, I updated the docstrings to NumPy format.

EDIT: Many sub-hour offsets (like 1/360 or 5/60) work well already without this PR. However, this PR now introduces support for arbitrary resolutions.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -887,15 +887,18 @@ class TCTracks():
             LOGGER.info("Progress: 100%")
 
     def equal_timestep(self, time_step_h=1, land_params=False):
-        """Generate interpolated track values to time steps of min_time_step.
-        Parameters:
-            time_step_h (float, optional): time step in hours to which to
-                interpolate. Default: 1.
-            land_params (bool, optional): compute on_land and dist_since_lf at
-                each node. Default: False.
+        """Generate interpolated track values to time steps of time_step_h.
+
+        Parameters
+        ----------
+        time_step_h : float or int, optional
+            Temporal resolution in hours (positive, may be non-integer-valued). Default: 1.
+        land_params : bool, optional
+            If True, recompute `on_land` and `dist_since_lf` at each node. Default: False.
         """
-        LOGGER.info('Interpolating %s tracks to %sh time steps.', self.size,
-                    time_step_h)
+        if time_step_h <= 0:
+            raise ValueError("time_step_h is not a positive number: %s", time_step_h)
+        LOGGER.info('Interpolating %s tracks to %sh time steps.', self.size, time_step_h)
 
         if land_params:
             extent = self.get_extent()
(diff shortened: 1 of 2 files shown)


bguillod:
It already works if you put time_step_h to 1/12, for instance. What exactly is the improvement in this PR besides from rounding time_step_h to the minute?

tovogt:
Ah, I didn't notice that. In my setup, I wanted to use something like `time_step_h=0.016667`, but it then raised a `ValueError: Invalid frequency: 0.016667H`. But you are right, many other offsets (like 1/360 or 5/60) worked well already before.

I now updated this PR to avoid rounding to minute resolution and even support arbitrary resolutions. Please have a look.

climada/hazard/tc_tracks.py
```diff
@@ -1112,7 +1120,7 @@ def _one_interp_data(track, time_step_h, land_geom=None):
                 # crosses 180 degrees east/west -> use positive degrees east
                 lon[lon < 0] += 360
 
-            time_step = '{}H'.format(time_step_h)
+            time_step = pd.tseries.frequencies.to_offset(pd.Timedelta(hours=time_step_h))
```
tovogt, line 1123: This is the only non-trivial change...

bguillod:
Ah ok, I had never realized it was not always working. That looks good to me then!

climada/hazard/tc_tracks.py
```diff
@@ -872,15 +872,16 @@ def read_simulations_storm(self, path, years=None):
...
+        Parameters
+        ----------
+        time_step_h : float or int, optional
+            Temporal resolution in hours (may be non-integer-valued). Default: 1.
+        land_params : bool, optional
+            If True, recompute `on_land` and `dist_since_lf` at each node. Default: False.
         """
-        LOGGER.info('Interpolating %s tracks to %sh time steps.', self.size,
-                    time_step_h)
+        LOGGER.info('Interpolating %s tracks to %sh time steps.', self.size, time_step_h)
```
chahank, line 885:
```suggestion
    if time_step_h <= 0:
        raise ValueError('time_step_h ist not a positive number')
```

chahank:
Looks good to me. Just one question: are you now confident that it works for generic float values? Also, I know it is out of your code changes, but you could add an error if time_step_h is not a positive number.

tovogt:
Thanks for the suggestion, I included a test for positivity. And yes, I'm confident that it works for generic float values, I even added a unit test for some "generic" float values. Moreover, I tested it once with hundreds of randomly generated float values to be sure.

Issue #100, opened 2020-12-01
Centroids: Why does transform need to contain upper left corner?


tovogt:
There are two places in `centr.py` where users are informed that the raster transform associated to a Centroids object needs to contain the upper left corner:
https://github.com/CLIMADA-project/climada_python/blob/dcef12ab075796baa80fa20d7cb392ab79e1c572/climada/hazard/centroids/centr.py#L86-L89
https://github.com/CLIMADA-project/climada_python/blob/dcef12ab075796baa80fa20d7cb392ab79e1c572/climada/hazard/centroids/centr.py#L133-L135
If not, the `check` method will fail. Can somebody explain why this is enforced?

Unfortunately, the function `from_base_grid` will generate a raster that does NOT contain the upper left corner:
https://github.com/CLIMADA-project/climada_python/blob/dcef12ab075796baa80fa20d7cb392ab79e1c572/climada/hazard/centroids/centr.py#L158-L159
I myself was not aware of this rule when implementing it:
https://github.com/CLIMADA-project/climada_python/blob/dcef12ab075796baa80fa20d7cb392ab79e1c572/climada/hazard/centroids/centr.py#L924
However, I'm reluctant to change the base grid (since it's a binary blob in the repository). Is it really necessary for the raster transform to have this property? Can't we go with arbitrary transforms?

chahank:
To be honest, I have no idea. We should ask someone that maybe worked on this code: @gabrielaznar @ThomasRoosli @sameberenz ?

sameberenz:
Unfortunately, I am not familiar with the centroids module from a developer's perspective and never encountered any problems initiating centroids.

tovogt:
The check was added in 979b316748524ea3b9ee80d14a053ce88653e0c7. To me, it seems like this was for the convenience of the developers so that they don't have to think twice when doing stuff like plotting or indexing (especially when dealing with affine transforms).

The Exposure class contains a similar assumption:
https://github.com/CLIMADA-project/climada_python/blob/f1fd1c7006b619a86cd1ead774276a145c223a5c/climada/entity/exposures/base.py#L89-L92
https://github.com/CLIMADA-project/climada_python/blob/f1fd1c7006b619a86cd1ead774276a145c223a5c/climada/entity/exposures/base.py#L378-L383

[...]

PR #104, opened 2020-12-03
Centroids: transform orientations, esp. base grid

tovogt:
`Centroids.check` raises a `ValueError` when the raster's affine transform is not oriented up-down in y-direction. Before this PR, it would complain that the "upper left corner" has to be contained in the transform. With this PR the output message is improved to be more clear and the x-orientation is checked, as well.

Until we find the reason for this check (see #100), this PR proposes to modify`Centroids.get_closest_point` to work independent from transform orientation (following a suggestion by @ThomasRoosli). Furthermore, the `Centroids.from_base_grid` is changed to produce a `Centroids` with negative y-orientation. Finally, this PR implements unit tests for these changes. (Plus some additional code cleanup along the way.)

Edit: The PR has been changed to leave basic behavior as it used to be. It now mainly fixes the problems with `get_closest_point` and `from_base_grid`.


chahank:
I am not sure to really understand the issue. Is there a good reason to allow not having the top left corner in?

climada/hazard/centroids/centr.py
```diff
@@ -140,9 +142,9 @@ def check(self):
                 LOGGER.error('Missing meta information: width, height,'
                              'crs or transform')
                 raise ValueError
-            if self.meta['transform'][4] > 0:
-                LOGGER.error('Meta does not contain upper left corner data.')
-                raise ValueError
+            if self.meta['transform'][4] > 0 or self.meta['transform'][0] < 0:
+                LOGGER.warning('Meta does not contain upper left corner data.'
+                               'Some functionality might not work as expected.')
```
chahank, line 147: This warning is way too fuzzy and will most probably lead to issues in the future.

climada/hazard/centroids/centr.py
```diff
@@ -22,20 +22,19 @@
...
+
+import geopandas as gpd
 import h5py
+import numpy as np
 import pandas as pd
-from rasterio import Affine
+import rasterio
 from rasterio.warp import Resampling
-import geopandas as gpd
+import scipy.sparse as sp
```
chahank, line 32: Probably better to keep is as `from scipy import sparse` as this seems to me the prevalent way in the current code base.

climada/hazard/centroids/centr.py
tovogt: So do you suggest to raise an error or do you suggest do modify the output/explanation?

climada/hazard/centroids/centr.py
```diff
@@ -140,9 +142,9 @@ def check(self):
                 LOGGER.error('Missing meta information: width, height,'
                              'crs or transform')
                 raise ValueError
-            if self.meta['transform'][4] > 0:
-                LOGGER.error('Meta does not contain upper left corner data.')
-                raise ValueError
+            if self.meta['transform'][4] > 0 or self.meta['transform'][0] < 0:
```
chahank, line 145: What are `self.meta['transform'][4]` and `self.meta['transform'][0]` ? Maybe defining variables for these makes is understandable.

climada/hazard/centroids/centr.py
tovogt: Agreed and changed!

climada/hazard/centroids/centr.py
chahank: The minimal solution is to provide a more detailed explanation, but is sub-optimal as the actual problem is not solved: why was there this constraint in the first place and do we need it?

Ideally, we want to know whether we need an error or whether the warning can be removed. I do not know the answer, however.

climada/hazard/centroids/centr.py
tovogt: Agreed and changed!

climada/hazard/centroids/centr.py
tovogt: Okay, then let's wait for answers in #100 before making decisions about this PR.

tovogt:
Okay, @chahank: I modified the PR to be more along the lines of what @ThomasRoosli proposed in #100. Namely, we leave the basic behavior as is, improve the output messages a bit, add more unit tests, and (most importantly) make sure that the base grid passes the `check()`.

chahank:
Cool thanks! I feel it is safer this way. It might be that we have to wait for the return of @gabrielaznar for a complete picture though.

climada/hazard/centroids/centr.py
```diff
@@ -187,6 +194,21 @@ def from_base_grid(land=False, res_as=360, base_file=None):
...
+                centroids.meta['transform'] = rasterio.Affine(xres, xshear, xoff, yshear,
+                                                              -yres, yoff + (shape[0] - 1) * yres)
+                # flip y-axis in data arrays
+                for name in ["region_id", "dist_coast"]:
+                    if not hasattr(centroids, name):
+                        continue
+                    data = getattr(centroids, name)
+                    if data.size == 0:
+                        continue
+                    setattr(centroids, name, np.flipud(data.reshape(shape)).reshape(-1))
```
ThomasRoosli, line 211: Would it make sense to use centroids.check() before returning centroids? That would warn the user, if they assign a unknown file using the parameter base_file, and the resulting centroids are somehow not in order.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -70,21 +70,25 @@ def test_centroid_pass(self):
...
             extent=(-39, -34.7, -55.5, -53.6)  # south georgia island
         ).size
-
         self.assertEqual(count_sgi, 296)
 
+        # test negative latitudinal orientation by testing that northern hemisphere (Russia)
+        # is listed before southern hemisphere (South Africa)
+        russia_max_idx = (centroids.region_id == 643).nonzero()[0].max()
+        safrica_min_idx = (centroids.region_id == 710).nonzero()[0].min()
+        self.assertTrue(russia_max_idx < safrica_min_idx)
```
ThomasRoosli, line 90: Smart way of testing the (newly) required behaviour, together with the negative/zero diff in line 75 should work fine.

climada/hazard/centroids/test/test_vec_ras.py
```diff
@@ -457,15 +460,36 @@ def test_size_pass(self):
...
+            ])
+            test_data[:,[1,3]] *= y_sign
+            for x_in, y_in, x_out, y_out, idx_out in test_data:
+                x, y, idx = centr_ras.get_closest_point(x_in, y_in)
+                self.assertEqual(x, x_out)
+                self.assertEqual(y, y_out)
+                self.assertEqual(idx, idx_out)
+                self.assertEqual(centr_ras.lon[idx], x)
+                self.assertEqual(centr_ras.lat[idx], y)
+
```
ThomasRoosli, line 486: Good idea, to test this functionality also for the implemented changes

ThomasRoosli (approved):
First of all, sorry for the delayed review. Thanks for making the changes. I think this way we are playing it safe, and at the sametime do not have to change the binary basefile. Thanks for implementing it like that. I approve, but probably the develop branch has to be merged in first.

climada/hazard/centroids/centr.py
tovogt: Yes, good idea! I added a check.

tovogt:
Okay, thanks! This seems to be ready to go. The merge will be done after our meeting on Wednesday.

PR #105, opened 2020-12-16
tests: fix tests failing due to changes worldbank input data for GDP …

sameberenz:
…and add comments to support future fixes when input data changes again

### DIFF
--- a/climada/util/test/test_finance.py
+++ b/climada/util/test/test_finance.py
@@ -88,6 +88,8 @@ class TestWBData(unittest.TestCase):
 
     def test_gdp_sxm_2010_pass(self):
         """Test gdp function Sint Maarten."""
+        # If World Bank input data changes, make sure to set ref_year to a year where
+        # no data is available so that the next available data point has to be selected.
         ref_year = 2010
         with self.assertLogs('climada.util.finance', level='INFO') as cm:
             res_year, res_val = gdp('SXM', ref_year)
--- a/climada/util/test/test_finance.py
+++ b/climada/util/test/test_finance.py
@@ -97,7 +97,7 @@ class TestWBData(unittest.TestCase):
         # GDP and years with data available might change if worldbank input
         # data changes, check magnitude and adjust ref_val and/or ref_year
         # if test fails:
-        self.assertIn('GDP SXM %i: %1.3e' %(2011, ref_val), cm.output[0])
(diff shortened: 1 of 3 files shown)


climada/util/test/test_finance.py
```diff
@@ -86,17 +86,20 @@ def test_ne_gdp_aia_2012_pass(self):
...
 
-        ref_val = 3.658e+08
-        ref_year = 2014
-        self.assertIn('GDP SXM 2014: 3.658e+08', cm.output[0])
+        ref_val = 936089385.47486 # reference GDP value
+        ref_year = 2011 # nearest year with data available (might change)
+        # GDP and years with data available might change if worldbank input
+        # data changes, check magnitude and adjust ref_val and/or ref_year
+        # if test fails:
+        self.assertIn('GDP SXM %i: %1.3e' %(2011, ref_val), cm.output[0])
```
tovogt, line 100:
```suggestion
        self.assertIn('GDP SXM %i: %1.3e' % (ref_year, ref_val), cm.output[0])
```

climada/util/test/test_finance.py
```diff
@@ -86,17 +86,20 @@ def test_ne_gdp_aia_2012_pass(self):
         self.assertEqual(res_year, ref_year)
         self.assertEqual(res_val, ref_val)
 
-    def test_gdp_sxm_2012_pass(self):
+    def test_gdp_sxm_2010_pass(self):
         """Test gdp function Sint Maarten."""
-        ref_year = 2012
+        ref_year = 2010
```
tovogt, line 91:
```suggestion
        # If World Bank input data changes, make sure to set ref_year to a year where
        # no data is available so that the next available data point has to be selected.
        ref_year = 2010
```

tovogt (changes requested):
Thanks, the comments will definitely be very helpful in the future!

For the SXM-test, I felt that we could still be a bit more verbose. Would you please have a look at my proposals?

climada/util/test/test_finance.py
tovogt: From my understanding, the ref_year is chosen on purpose in such a way that it's a year where World Bank data is missing in order to test the functionality of selecting the next available data point.

sameberenz: yes, that's correct. the additional commit you suggest points that out and is really helpful, thanks

PR #110, opened 2021-01-04
Add multi-bar plot util function

chahank:
This is a small util function to make bar plots with multiple bars per element.

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
     except KeyError:
         units = '°'
     return crs_epsg, units
+
+
+def bar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True):
+    """Draws a bar plot with multiple bars per data point.
+    https://stackoverflow.com/questions/14270391/python-matplotlib-multiple-bars
+
+    Parameters
+    ----------
+    ax : matplotlib.pyplot.axis
+        The axis we want to draw our plot on.
+
+    data: dictionary
+        A dictionary containing the data we want to plot. Keys are the names of the
+        data, the items is a list of the values.
+
+        Example:
+        data = {
+            "x":[1,2,3],
+            "y":[1,2,3],
+            "z":[1,2,3],
+        }
+
+    colors : array-like, optional
+        A list of colors which are used for the bars. If None, the colors
+        will be the standard matplotlib color cyle. (default: None)
+
+    total_width : float, optional, default: 0.8
+        The width of a bar group. 0.8 means that 80% of the x-axis is covered
+        by bars and 20% will be spaces between the bars.
+
+    single_width: float, optional, default: 1
+        The relative width of a single bar within a group. 1 means the bars
+        will touch eachother within a group, values less than 1 will make
+        these bars thinner.
+
+    legend: bool, optional, default: True
+        If this is set to true, a legend will be added to the axis.
+    """
+
+    # Check if colors where provided, otherwhise use the default color cycle
+    if colors is None:
+        colors = plt.rcParams['axes.prop_cycle'].by_key()['color']
+
+    # Number of bars per group
+    n_bars = len(data)
+
+    # The width of a single bar
+    bar_width = total_width / n_bars
+
+    # List containing handles for the drawn bars, used for the legend
+    bars = []
+
+    # Iterate over all data
+    for i, (name, values) in enumerate(data.items()):
+        # The offset in x direction of that bar
+        x_offset = (i - n_bars / 2) * bar_width + bar_width / 2
+
+        # Draw a bar for every value of that type
+        for x, y in enumerate(values):
+            bar = ax.bar(x + x_offset, y, width=bar_width * single_width, color=colors[i % len(colors)])
+
+        # Add a handle to the last drawn bar, which we'll need for the legend
+        bars.append(bar[0])
+
+    # Draw legend if we need
+    if legend:
+        ax.legend(bars, data.keys())
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -474,7 +474,7 @@ def get_transformation(crs_in):
     return crs_epsg, units
 
 
-def bar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True):
+def multibar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True, xticklabels=None):
     """Draws a bar plot with multiple bars per data point.
     https://stackoverflow.com/questions/14270391/python-matplotlib-multiple-bars
 
@@ -489,10 +489,13 @@ def bar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True
 
         Example:
         data = {
-            "x":[1,2,3],
-            "y":[1,2,3],
-            "z":[1,2,3],
+            "x": [1, 2, 3],
+            "y": [1, 2, 3],
+            "z": [1, 2, 3],
         }
+        fig, ax = plt.subplots()
+        multibar_plot(ax, data, xticklabels=["a", "b", "c"])
+        
 
     colors : array-like, optional
         A list of colors which are used for the bars. If None, the colors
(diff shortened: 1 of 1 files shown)


climada/util/plot.py
```diff
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
     except KeyError:
         units = '°'
     return crs_epsg, units
+
+
+def bar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True):
```
Evelyn-M, line 477: perhaps rather name the function `multibar` or `multibar_plot`? So that it's obviously an addition to the normal `plt.bar()`

climada/util/plot.py
```diff
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
...
+    """Draws a bar plot with multiple bars per data point.
+    https://stackoverflow.com/questions/14270391/python-matplotlib-multiple-bars
+
+    Parameters
+    ----------
+    ax : matplotlib.pyplot.axis
+        The axis we want to draw our plot on.
+
```
Evelyn-M, line 485: Perhaps not so easy to get that one needs to define this like `ax = plt.subplot()` beforehand. 
How about merging this into a working example, and include a proper example section in the doc. Something like:

```
Example
--------
data = {
            "var1":[1,2,3],
            "var2":[1,2,3],
            "var3":[1,2,3],
        }

ax = plt.subplot()
bar_plot(ax, data)
```

I think it's useful to also provide in the example how to include x-axis labels then, as probably this would be a frequent operation and it's not 100% straightforward (e.g. `ax.set_xticklabels( ('one', 'two', 'three') )`  would add all those names only to the first group. one needs to explicitly set the positions now with set_xticklabels(), cf answer 2 in your stackoverflow link)

climada/util/plot.py
```diff
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
...
+    data: dictionary
+        A dictionary containing the data we want to plot. Keys are the names of the
+        data, the items is a list of the values.
+
```
Evelyn-M, line 489: Could this be extended to work with any kind of iterable?
I'd guess normally one would already have a set of variables such as var1 = [1,2,3], var2 = [1,2,3], etc.
Not that it's hard to convert (var1, var2, var3) into a dictionary, but would be neat if this needn't be done by one's-self

climada/util/plot.py
```diff
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
     except KeyError:
         units = '°'
     return crs_epsg, units
+
+
+def bar_plot(ax, data, colors=None, total_width=0.8, single_width=1, legend=True):
+    """Draws a bar plot with multiple bars per data point.
+    https://stackoverflow.com/questions/14270391/python-matplotlib-multiple-bars
```
Evelyn-M, line 479: Just a detail, this works only for horizontal bar plots, right? (since below, x and y dimensions are explicitly treated.

climada/util/plot.py
```diff
@@ -472,3 +472,70 @@ def get_transformation(crs_in):
...
+        # Draw a bar for every value of that type
+        for x, y in enumerate(values):
+            bar = ax.bar(x + x_offset, y, width=bar_width * single_width, color=colors[i % len(colors)])
+
+        # Add a handle to the last drawn bar, which we'll need for the legend
+        bars.append(bar[0])
+
+    # Draw legend if we need
+    if legend:
+        ax.legend(bars, data.keys())
```
Evelyn-M, line 541: Last but not least, why is the ax.legend method wrapped into this function, but not for example the `ax.subplot, ax.title, ax.set_xticklabels`?

Evelyn-M (changes requested):

climada/util/plot.py
chahank: I added the xticklabels as optional argument.

climada/util/plot.py
chahank: I am confused here. The idea is to do bar plots where for each xticks there are multiple grouped bars. To have clear labeling, the sub-elements of each group are passed as dictionnary items.

climada/util/plot.py
chahank: Added xticklabels as optional parameter and completed the example.

climada/util/plot.py
chahank: No particular reason, except that for the most casual user it might be a bit of an effort to have correct legends. This does for me not apply to the title. But I do not have a strong opinion about keeping it or not.

The `ax.suplot` point I do not understand. The function works with an axis object...

climada/util/plot.py
Evelyn-M: i mean, usually you probably don't have a dictionary, but a set of pre-defined variables. Of course they could first be converted to a dictionary, and then passed into the function. But maybe the function could do this internally, so the user can either provide such a dictionary, or a set of variables.

Also, the variable names or keys in the current case are just relevant for labelling the legend. There is no option at the moment to pass group-labels (aka the xticks).

climada/util/plot.py
Evelyn-M: my bad, forget the ax.suplot, that was a typo

climada/util/plot.py
chahank: Well, you mean a two-dimensional list or np.array? But then how to make the legends if you do not have the variable series names?

The xticks were just added (see last commit).

climada/util/plot.py
chahank: Do you think it would be useful to also have the option to make it y-bars instead of x-bars? I could add it.

Evelyn-M: I think it would be a nice feature if it's not to much effort, yes

chahank: Done :D.

climada/util/plot.py
Evelyn-M: xticks: yep, i see. 

no, i just mean the case where you dont have a dictionary {"var1" = [1,2,3]; "var2" = [1,2,3]}, but a set of variables (var1 = [1,2,3], var2 = [1,2,3]) which you already defined somewhere on the way. instead of taking your variables and converting them manually to a dictionary first, maybe the function could check for the input types (e.g. is it already a dict --> all good, is it a set of variables --> make one out of those with keys and values)

chahank: Well... I still do not understand, but more convenience is certainly worth it. 

So, what you propose is to allow the user to give dictionnary or any other nested iterable? In your example you replaced the dictionary by a tuple of list? So possible entries would be:

`{'var1': [1,2,3], 'var2':[1,2,3]}`
`([1,2,3], [1,2,3])`
`[[1,2,3], [1,2,3]]`

Or would you like to have an arbitrary number of keyword arguments?

`multibar(var1=[1,2,3], var2=[1,2,3], var3=[1,2,3], ticklabel=['a', 'b', 'c']`

I am not sure how I could implement the latter together with the dictionary, which in my opinion is cleaner.

Evelyn-M: Ideally, i would like both this to work:
`multibar({'var1': [1,2,3], 'var2':[1,2,3]}, ticklabel=['a', 'b', 'c'])`
and this:
`multibar((var1, var2), ticklabel=['a', 'b', 'c'])` (where var1 and var2 have been defined previously somewhere).

The latter one (`multibar(var1=[1,2,3], var2=[1,2,3], var3=[1,2,3], ticklabel=['a', 'b', 'c'])` with an arbitrary amount of keyword arguments would also be possible, but is probably over-engineered (cf. (https://treyhunner.com/2018/04/keyword-arguments-in-python/#Capturing_arbitrary_keyword_arguments) )

chahank: Well, so you are proposing to have lists of lists, tuples of lists, or whatever 2d arrays. This is fine, but then there would be no legend, which in my opinion makes the graphic not readable. In your example, one should simply write

`multibar({'var1': var1, 'var2': var2}, ticklabel=['a','b','c'])`

which in my opinion is not more complex than 

`multibar((var1, var2), ticklabel=['a', 'b', 'c'])`.

chahank: What do you think? 

Also, the multi keyword argument route can of course be done, but together with the dictionary, it seems confusing to me (both technically and user side).

Evelyn-M: Right. I wouldn't do the dictionary _and_ the multi keyword argument. that's indeed to confusing.
I guess you're right, in case one already has variables, it's probably easy enough to manually make a dict via `multibar({'var1': var1, 'var2': var2}, ticklabel=['a','b','c'])`

So let's leave it as is

chahank: Very well, thanks a lot for the great review :D !

PR #111, opened 2021-01-05
Feature/uncertainty

chahank:
This pull request is not meant to end in a  merge directly but is to obtain a review of the code before it gets too large. Feedback on the general structure, the included and planned features, the computational efficiency are appreciated.

For a very brief tutorial see [climada_python/doc/tutorial/climada_engine_uncertainty.ipynb ](https://github.com/CLIMADA-project/climada_python/blob/feature/uncertainty/doc/tutorial/climada_engine_uncertainty.ipynb)


climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+            'names' : self.param_labels,
+            'bounds' : [[0, 1]]*len(self.param_labels)
+            }
+        self.problem = problem
+        sobol_params = saltelli.sample(problem, self.n_samples,
+                                       calc_second_order=calc_second_order)
+        return sobol_params
+
+
+    def calc_impact_sobol_sensitivity(self,
```
aleeciu, line 359: Why not a single kwarg, e.g. ooi (outcomes of interest) like ooi={calc_eai_exp=bool, calc_at_event=bool, rp=list} where only those of interest are listed (with all possible outcomes being all those supported by Impact().calc, which the user should be aware of) and then we assign e.g. self.calc_eai_exp or self.calc_at_event based on those we have. These self.... will also dictate what calculations we save (as it is done already).

climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+        df_aai_freq['aai_agg'] = aai_agg_list
+        self.aai_freq = df_aai_freq
+
+        if calc_eai_exp:
+            df_eai_exp = pd.DataFrame(eai_exp_list)
+            self.eai_exp = df_eai_exp
+
+        if calc_at_event:
+            df_at_event = pd.DataFrame(at_event_list)
+            self.at_event = df_at_event
```
aleeciu, line 538: at the end of this func we could aggregate all resulting dfs in a dict or something. Now it seems sobol indexes are only calculated based on self.aai_freq (line 427); we could instead iterate first over the dfs and then over the variables in each df.

climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+            rp=[5, 10, 20, 50, 100, 250]
+
+        aai_agg_list = []
+        freq_curve_list = []
+        if calc_eai_exp:
+            eai_exp_list = []
+        if calc_at_event:
+            at_event_list = []
+
+        self.rp = rp
```
aleeciu, line 503: I think this is a repetation; they are aleady specified at lines 414-416, before calling calc_impact_distribution

climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+            self.haz = haz_unc
+
+        self.params = pd.DataFrame()
+        self.problem = {}
+        self.aai_freq = pd.DataFrame()
+        self.eai_exp = pd.DataFrame()
+        self.at_event = pd.DataFrame()
+
+
+    def plot_uncertainty(self):
```
aleeciu, line 201: would be cool to have something that produces a plot like the one at page two:
https://safe-insurance.uk/EGU2020_display_Noacco_et_al_13_04_20.pdf

aleeciu (commented):
very cool work - I just made few suggestions of improvement

climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+    Uncertainty variable
+
+    An uncertainty variable requires a single or multi-parameter function.
+    The parameters must follow a given distribution.
+
+    Examples
+    --------
+
+    Categorical variable function: LitPop exposures with m,n exponents in [0,5]
+        def unc_var_cat(m, n):
```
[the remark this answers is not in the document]
aleeciu: impact, exposure, and hazard...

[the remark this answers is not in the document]
chahank: Is a good idea! I was thinking of first making a robust base module, and then add some helper functions later on.

climada/engine/uncertainty.py
chahank: Good idea. I will see how it works out.

climada/engine/uncertainty.py
chahank: Yes indeed. I would suggest to do that once the Module has reached some maturity in its content/form. Maybe you could then do that :D ?

chahank:
Thanks for the good and useful feedback!

climada/engine/uncertainty.py
aleeciu: Indeed, we could also come up with a list of stuff and distribute work - although I think you already coded the core and more, already..

chahank: Well, here is my plan before coming to some of the graphics:

- Separate the class into an Uncertainty class, a Sensitivity sub-class and a Robustness sub-class.
- Make it possible to use different kind of sampling and different kinds of sensitivity analysis (not only Sobol)
- Make everything as smooth/fast as possible

climada/engine/uncertainty.py
```diff
@@ -0,0 +1,592 @@
...
+        df_params = pd.DataFrame(sobol_uniform_sample, columns=self.param_labels)
+        for param in list(df_params):
+            df_params[param] = df_params[param].apply(
+                self.distr_dict[param].ppf
+                )
+        self.params = df_params
+
+
+
+    def _make_uniform_sobol_sample(self, calc_second_order):
```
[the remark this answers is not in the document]
chahank: O renamed it `_make_uniform_base_sample`. Note that now the sampling has an argument `method` which allows the user to choose any of the saLib sampling methods.

climada/engine/uncertainty.py
chahank: This is intended. My current idea is that the workflow would be:

1. define the UncVar()
2. make the sample 
3. compute the impact/cost benefit distributions
4. compute the sensitivity indices

aleeciu: Yes but it seems to me that the lines
`        self.rp = rp
        self.calc_eai_exp = calc_eai_exp
        self.calc_at_event = calc_at_event
`
are both in _calc_impact_sobol_sensitivity_ and _calc_impact_distribution_, with the latter being called once and inside the former; so those assignments are called twice. It is true though that, in principle, the second function could be used as standalone depending on the application, so it may be good to leave it like that.

chahank: Yes,  I agree, that is a good summary. I will do the refactoring and then we can decide whether this is a good idea or not.

PR #114, opened 2021-01-07
Feature/config conventions

emanuel-schmid:
This is about CLIMADA configuration.
It introduces 
- a method to read configuration values comfortably
- superseding default configuration by local files, only where required
- distinction of configuration value types: integers, floats, strings and pathes

In the mean time os.path was removed from the whole code base and replaced by pathlib.Path.

Please review quickly for this is needed in the coming convention workshop on January 18. ;)
Many thanks in advance


Evelyn-M (approved):
Since 90% is related to the pathlib-change anyways, looks all good.
The config class is honestly a bit obscure to me, so i cannot review in detail, but from the use of the implemented methods in other modules, it looks sensible & like what we've discussed in the meetings.

climada.conf
```diff
@@ -0,0 +1,27 @@
+{
+    "_comment": "this is a climada configuration file meant to supersede the default configuration in climada/conf during test",
+    "test_directory": "./climada",
```
chahank, line 3: Why is the test_directory the climada base directory? or what does line 3 mean?

climada/__init__.py
```diff
@@ -18,7 +18,89 @@
...
 climada init
 """
+from shutil import copyfile
+from pathlib import Path
 
-from .util.config import CONFIG, setup_conf_user, setup_logging
-setup_conf_user()
-setup_logging(CONFIG['global']['log_level'])
+from .util.config import CONFIG, setup_logging
+from .util.constants import *
```
chahank, line 25: Is this * import to be temporary or will constants remain in the util? I thought constant should be defined in the config file right?

chahank (approved):

PR #115, opened 2021-01-08
Improve TC wind performance with sinusoidal projection

tovogt:
As discussed already in a video call, we can improve performance of TC wind field computation by returning to the sinusoidal distance computations (that were default in a legacy version of CLIMADA). The loss in accuracy is very small (compare the values in the tests).

### DIFF
--- a/climada/hazard/test/test_trop_cyclone.py
+++ b/climada/hazard/test/test_trop_cyclone.py
@@ -46,59 +46,64 @@ class TestReader(unittest.TestCase):
 
     def test_set_one_pass(self):
         """Test _tc_from_track function."""
+        intensity_idx = [0, 1, 2,  3,  80, 100, 120, 200, 220, 250, 260, 295]
+        intensity_values = {
+            "geosphere": [25.76590964, 27.08333002, 28.46008202, 25.70445069, 31.45216632,
+                          36.45564037, 21.22679800, 28.22022122, 32.92315537, 31.60115745, 0,
+                          40.62433745],
+            "equirect": [25.76575952, 27.08314419, 28.45984342, 25.70460801, 31.45171899,
+                         36.45003013, 21.22671287, 28.22016474, 32.92054837, 31.59980326, 0,
+                         40.62068325]
+        }
+        # the values for the two metrics should agree up to first digit at least
+        for i, val in enumerate(intensity_values["geosphere"]):
+            self.assertAlmostEqual(intensity_values["equirect"][i], val, 1)
+
         tc_track = TCTracks()
         tc_track.read_processed_ibtracs_csv(TEST_TRACK)
         tc_track.equal_timestep()
         tc_track.data = tc_track.data[:1]
-        tc_haz = TropCyclone()
-        tc_haz.set_from_tracks(tc_track, centroids=CENTR_TEST_BRB, model='H08',
-                               store_windfields=True)
 
-        self.assertEqual(tc_haz.tag.haz_type, 'TC')
-        self.assertEqual(tc_haz.tag.description, '')
-        self.assertEqual(tc_haz.tag.file_name, 'Name: 1951239N12334')
-        self.assertEqual(tc_haz.units, 'm/s')
-        self.assertEqual(tc_haz.centroids.size, 296)
-        self.assertEqual(tc_haz.event_id.size, 1)
-        self.assertEqual(tc_haz.date.size, 1)
-        self.assertEqual(dt.datetime.fromordinal(tc_haz.date[0]).year, 1951)
-        self.assertEqual(dt.datetime.fromordinal(tc_haz.date[0]).month, 8)
-        self.assertEqual(dt.datetime.fromordinal(tc_haz.date[0]).day, 27)
-        self.assertEqual(tc_haz.event_id[0], 1)
-        self.assertEqual(tc_haz.event_name, ['1951239N12334'])
-        self.assertTrue(np.array_equal(tc_haz.frequency, np.array([1])))
-        self.assertTrue(isinstance(tc_haz.fraction, sparse.csr.csr_matrix))
-        self.assertEqual(tc_haz.fraction.shape, (1, 296))
-        self.assertEqual(tc_haz.fraction[0, 100], 1)
-        self.assertEqual(tc_haz.fraction[0, 260], 0)
-        self.assertEqual(tc_haz.fraction.nonzero()[0].size, 280)
+        for metric in ["equirect", "geosphere"]:
+            tc_haz = TropCyclone()
+            tc_haz.set_from_tracks(tc_track, centroids=CENTR_TEST_BRB, model='H08',
+                                   store_windfields=True, metric=metric)
 
-        self.assertTrue(isinstance(tc_haz.intensity, sparse.csr.csr_matrix))
-        self.assertEqual(tc_haz.intensity.shape, (1, 296))
-        self.assertEqual(np.nonzero(tc_haz.intensity)[0].size, 280)
+            self.assertEqual(tc_haz.tag.haz_type, 'TC')
+            self.assertEqual(tc_haz.tag.description, '')
+            self.assertEqual(tc_haz.tag.file_name, 'Name: 1951239N12334')
+            self.assertEqual(tc_haz.units, 'm/s')
+            self.assertEqual(tc_haz.centroids.size, 296)
+            self.assertEqual(tc_haz.event_id.size, 1)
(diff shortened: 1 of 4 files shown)


climada/hazard/test/test_trop_cyclone.py
```diff
@@ -46,59 +46,64 @@ class TestReader(unittest.TestCase):
...
         """Test _tc_from_track function."""
+        intensity_idx = [0, 1, 2,  3,  80, 100, 120, 200, 220, 250, 260, 295]
+        intensity_values = {
+            "geosphere": [25.76590964, 27.08333002, 28.46008202, 25.70445069, 31.45216632,
+                          36.45564037, 21.22679800, 28.22022122, 32.92315537, 31.60115745, 0,
+                          40.62433745],
+            "equirect": [25.76575952, 27.08314419, 28.45984342, 25.70460801, 31.45171899,
+                         36.45003013, 21.22671287, 28.22016474, 32.92054837, 31.59980326, 0,
+                         40.62068325]
+        }
```
tovogt, line 57: Compare these values to get an idea of the effective loss in accuracy for the two distance metrics.

tovogt, line 57: Compare these values to get an idea of the effective loss in accuracy for the two distance metrics.

climada/hazard/trop_cyclone.py
```diff
@@ -102,26 +102,37 @@ def __init__(self, pool=None):
 
     def set_from_tracks(self, tracks, centroids=None, description='',
                         model='H08', ignore_distance_to_coast=False,
-                        store_windfields=False):
+                        store_windfields=False, metric="equirect"):
```
tovogt, line 105: We are going back to "equirect" as the default metric because it's sufficient for 99.99% of the users and 2-3 times faster.

climada/hazard/trop_cyclone.py
```diff
@@ -410,12 +438,10 @@ def compute_windfields(track, centroids, model):
...
-    # never use longitudes at -180 degrees or below
-    t_lon[t_lon <= -180] += 360
-
-    # only use longitudes above 180, if 180 degree border is crossed
-    if t_lon.min() > 180:
-        t_lon -= 360
+    # normalize longitude values
+    mid_lon = 0.5 * sum(lon_bounds(t_lon))
+    lon_normalize(t_lon, center=mid_lon)
+    lon_normalize(centroids[:,1], center=mid_lon)
```
tovogt, line 444: The normalization of longitudinal values around some "central" longitude (see https://en.wikipedia.org/wiki/Sinusoidal_projection) is very important for the sinusoidal distance computations. It was not necessary at all for the spherical geodesic distances. However, the loss in performance is really negligible according to some performance tests I did.

climada/hazard/trop_cyclone.py
```diff
@@ -515,25 +542,14 @@ def _close_centroids(t_lat, t_lon, centroids, buffer=CENTR_NODE_MAX_DIST_DEG):
...
-    msk_lon[msk_lon_crosses] |= msk_lon_upper[msk_lon_crosses]
-    msk_lon[msk_lon_within] &= msk_lon_upper[msk_lon_within]
-    return (msk_lat & msk_lon).any(axis=0)
-
-def _vtrans(t_lat, t_lon, t_tstep):
+    msk = (eye_bounds[:,None,1] < centr_lat[None])
+    msk &= (centr_lat[None] < eye_bounds[:,None,3])
+    msk &= (eye_bounds[:,None,0] < centr_lon[None])
+    msk &= (centr_lon[None] < eye_bounds[:,None,2])
+    return msk.any(axis=0)
```
tovogt, line 550: Because we are normalizing all longitudinal values, we don't have to consider the antimeridian as a special case here.

climada/util/coordinates.py
```diff
@@ -122,13 +122,16 @@ def lon_normalize(lon, center=0.0):
     lon : np.array
         Longitudinal coordinates
     center : float, optional
-        Central longitude value to use instead of 0.
+        Central longitude value to use instead of 0. If None, the central longitude is determined
+        automatically.
```
tovogt, line 126: This is a very small, cheap change that might come in handy in the future (and is already useful for some of my usecases).

climada/util/coordinates.py
```diff
@@ -256,17 +290,17 @@ def dist_approx(lat1, lon1, lat2, lon2, log=False, normalize=True,
...
+            lon_normalize(lon1, center=mid_lon)
+            lon_normalize(lon2, center=mid_lon)
+        vtan = np.stack([lat2[:, None] - lat1[:, :, None],
+                         lon2[:, None] - lon1[:, :, None]], axis=-1)
+        fact1 = np.heaviside(vtan[..., 1] - 180, 0)
+        fact2 = np.heaviside(-vtan[..., 1] - 180, 0)
+        vtan[..., 1] -= (fact1 - fact2) * 360
+        vtan[..., 1] *= np.cos(np.radians(lat1[:, :, None]))
+        vtan *= unit_factor
+        dist = np.sqrt(np.einsum("...l,...l->...", vtan, vtan))
```
tovogt, line 303: Curiously, `numpy.sqrt(numpy.einsum(...))` is much faster than `numpy.linalg.norm(...)`.

climada/util/test/test_coordinates.py
```diff
@@ -117,10 +121,11 @@ def test_dist_approx_pass(self):
...
             # lat1, lon1, lat2, lon2, dist, dist_sph
-            [45.5, -32.2, 14, 56, 7709.827814738594, 8758.34146833],
-            [45.5, 147.8, 14, -124, 7709.827814738594, 8758.34146833],
-            [45.5, 507.8, 14, -124, 7709.827814738594, 8758.34146833],
-            [45.5, -212.2, 14, -124, 7709.827814738594, 8758.34146833],
+            [45.5, -32.1, 14, 56, 7702.88906574, 8750.64119051],
+            [45.5, 147.8, 14, -124, 7709.82781473, 8758.34146833],
+            [45.5, 507.9, 14, -124, 7702.88906574, 8750.64119051],
+            [45.5, -212.2, 14, -124, 7709.82781473, 8758.34146833],
+            [-3, -130.1, 4, -30.5, 11079.7217421, 11087.0352544],
```
tovogt, line 128: There was too much artificial symmetry in the test case (all pairs of points heaving exactly the same distance to each other), so I changed the values slightly.

chahank (changes requested):
All good, I would just add parts of the discussions in the docstrings directly for future use.

climada/hazard/trop_cyclone.py
tovogt: I added a remark to the docstring of `_close_centroids` for future reference.

climada/hazard/trop_cyclone.py
tovogt: This is now mentioned in the comment above and in the docstring of `_close_centroids`.

climada/util/coordinates.py
tovogt: That's a scalar product in the last axis, continue reading here for the syntax: https://numpy.org/doc/stable/reference/generated/numpy.einsum.html

climada/hazard/trop_cyclone.py
```diff
@@ -102,26 +102,37 @@ def __init__(self, pool=None):
...
+        ignore_distance_to_coast : boolean, optional
+            If True, centroids far from coast are not ignored. Default: False.
+        store_windfields : boolean, optional
+            If True, the Hazard object gets a list `windfields` of sparse matrices. For each track,
+            the full velocity vectors at each centroid and track position are stored in a sparse
+            matrix of shape (npositions,  ncentroids * 2) that can be reshaped to a full ndarray
+            of shape (npositions, ncentroids, 2). Default: False.
+        metric : str, optional
+            Specify an approximation method to use for earth distances:
+            * "equirect": Distance according to sinusoidal projection. Fast, but inaccurate for
```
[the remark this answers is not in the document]
tovogt: This is not relevant, since both methods are able to deal with tracks crossing the antimeridian. This was an issue with the legacy implementations in CLIMADA.

tovogt:
@chahank Thanks for the review! I think I addressed all of your points by fixing or commenting.

tovogt:
@ChrisFairless I remember that you were in the call we had about this topic, that's why I added you as a reviewer. Do you have anything else to add to @chahank's comments and remarks, or is this ready to be merged?

tovogt:
Merging this for new PR feature/tc_wind_memory

PR #121, opened 2021-01-25
Improve TropCyclone memory efficiency for large centroid sets

tovogt:
When computing wind fields from TCTracks and a Centroids object, this PR improves memory efficiency by filtering out centroids early on that are not within reach of any of the TC tracks.

Previously to this PR, filtering was done for coastal centroids, but the reachability was only checked on a track-by-track basis. This would allocate a lot of unnecessary memory when there are many centroids that are never reached by any of the TC tracks.

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -34,7 +34,7 @@ import contextily as ctx
 from climada.entity.tag import Tag
 import climada.util.hdf5_handler as hdf5
 from climada.util.constants import ONE_LAT_KM, DEF_CRS
-import climada.util.coordinates as co
+import climada.util.coordinates as u_coord
 from climada.util.interpolation import interpol_index
 import climada.util.plot as u_plot
 
@@ -199,7 +199,7 @@ class Exposures(GeoDataFrame):
         """
         LOGGER.info('Matching %s exposures with %s centroids.',
                     str(self.shape[0]), str(hazard.centroids.size))
-        if not co.equal_crs(self.crs, hazard.centroids.crs):
+        if not u_coord.equal_crs(self.crs, hazard.centroids.crs):
             LOGGER.error('Set hazard and exposure to same CRS first!')
             raise ValueError
         if hazard.centroids.meta:
@@ -229,7 +229,7 @@ class Exposures(GeoDataFrame):
             scheduler (str): used for dask map_partitions. “threads”,
                 “synchronous” or “processes”
         """
-        co.set_df_geometry_points(self, scheduler)
+        u_coord.set_df_geometry_points(self, scheduler)
 
     def set_lat_lon(self):
         """Set latitude and longitude attributes from geometry attribute."""
@@ -259,9 +259,9 @@ class Exposures(GeoDataFrame):
         self.__init__()
         self.tag = Tag()
         self.tag.file_name = file_name
-        meta, value = co.read_raster(file_name, [band], src_crs, window,
-                                     geometry, dst_crs, transform, width,
-                                     height, resampling)
+        meta, value = u_coord.read_raster(file_name, [band], src_crs, window,
+                                          geometry, dst_crs, transform, width,
+                                          height, resampling)
         ulx, xres, _, uly, _, yres = meta['transform'].to_gdal()
         lrx = ulx + meta['width'] * xres
         lry = uly + meta['height'] * yres
@@ -382,8 +382,7 @@ class Exposures(GeoDataFrame):
                 LOGGER.error('Points are not ordered according to meta raster.')
                 raise ValueError
         else:
-            raster, meta = co.points_to_raster(self, ['value'], res, raster_res,
-                                               scheduler)
+            raster, meta = u_coord.points_to_raster(self, ['value'], res, raster_res, scheduler)
             raster = raster.reshape((meta['height'], meta['width']))
         # save tiff
         if save_tiff is not None:
@@ -561,10 +560,10 @@ class Exposures(GeoDataFrame):
             if self.longitude.values[0] > self.longitude.values[-1]:
                 LOGGER.error('Points are not ordered according to meta raster.')
                 raise ValueError
-            co.write_raster(file_name, raster, self.meta)
+            u_coord.write_raster(file_name, raster, self.meta)
         else:
-            raster, meta = co.points_to_raster(self, [value_name], scheduler=scheduler)
-            co.write_raster(file_name, raster, meta)
+            raster, meta = u_coord.points_to_raster(self, [value_name], scheduler=scheduler)
+            u_coord.write_raster(file_name, raster, meta)
 
 def add_sea(exposures, sea_res):
     """Add sea to geometry's surroundings with given resolution. region_id
@@ -593,7 +592,7 @@ def add_sea(exposures, sea_res):
 
     lon_mgrid, lat_mgrid = np.meshgrid(lon_arr, lat_arr)
     lon_mgrid, lat_mgrid = lon_mgrid.ravel(), lat_mgrid.ravel()
-    on_land = ~co.coord_on_land(lat_mgrid, lon_mgrid)
+    on_land = ~u_coord.coord_on_land(lat_mgrid, lon_mgrid)
 
     sea_exp = Exposures()
     sea_exp['latitude'] = lat_mgrid[on_land]
--- a/climada/entity/exposures/crop_production.py
+++ b/climada/entity/exposures/crop_production.py
@@ -27,9 +27,10 @@ import pandas as pd
 import h5py
 from matplotlib import pyplot as plt
 from iso3166 import countries as iso_cntry
+
 from climada.entity.exposures.base import Exposures
 from climada.entity.tag import Tag
-import climada.util.coordinates as coord
+import climada.util.coordinates as u_coord
 from climada.util.constants import DEF_CRS
 from climada.util.coordinates import pts_to_raster_meta, get_resolution
 from climada import CONFIG
@@ -223,7 +224,7 @@ class CropProduction(Exposures):
         lon, lat = np.meshgrid(data.lon.values, data.lat.values)
         self['latitude'] = lat.flatten()
         self['longitude'] = lon.flatten()
-        self['region_id'] = coord.get_country_code(self.latitude, self.longitude)
+        self['region_id'] = u_coord.get_country_code(self.latitude, self.longitude)
 
         # The indeces of the yearrange to be extracted are determined
         time_idx = (int(yearrange[0] - yearchunk['startyear']),
@@ -486,7 +487,7 @@ class CropProduction(Exposures):
         fao['year'] = fao['file'].Year.values
         fao['price'] = fao['file'].Value.values
 
-        fao_country = coord.country_faocode2iso(getattr(fao['file'], 'Area Code').values)
+        fao_country = u_coord.country_faocode2iso(getattr(fao['file'], 'Area Code').values)
 
         # create a list of the countries contained in the exposure
         iso3alpha = list()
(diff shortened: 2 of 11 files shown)


climada/entity/exposures/crop_production.py
```diff
@@ -27,9 +27,10 @@
 import h5py
 from matplotlib import pyplot as plt
 from iso3166 import countries as iso_cntry
+
 from climada.entity.exposures.base import Exposures
 from climada.entity.tag import Tag
-import climada.util.coordinates as coord
+import climada.util.coordinates as co
```
tovogt, line 33: Since it also affected `trop_cyclone.py`, I used this PR to unify the shorthand for `climada.util.coordinates` everywhere in the whole project. The shorthand `co` is taken from the base modules in the hazard and exposures subpackages.

climada/hazard/tc_surge_bathtub.py
```diff
@@ -29,7 +29,7 @@
 import scipy.sparse as sp
 
 from climada.hazard.base import Hazard
-import climada.util.coordinates as coord_util
+import climada.util.coordinates as co
```
tovogt, line 32: Since it also affected trop_cyclone.py, I used this PR to unify the shorthand for climada.util.coordinates everywhere in the whole project. The shorthand co is taken from the base modules in the hazard and exposures subpackages.

climada/hazard/tc_tracks.py
```diff
@@ -49,7 +49,7 @@
 
 # climada dependencies
 from climada.util import ureg
-import climada.util.coordinates as coord_util
+import climada.util.coordinates as co
```
tovogt, line 52: Since it also affected trop_cyclone.py, I used this PR to unify the shorthand for climada.util.coordinates everywhere in the whole project. The shorthand co is taken from the base modules in the hazard and exposures subpackages.

climada/hazard/trop_cyclone.py
```diff
@@ -37,7 +37,7 @@
 from climada.hazard.tc_clim_change import get_knutson_criterion, calc_scale_knutson
 from climada.hazard.centroids.centr import Centroids
 from climada.util import ureg
-from climada.util.coordinates import dist_approx, lon_bounds, lon_normalize
+import climada.util.coordinates as co
```
tovogt, line 40: I used this PR to unify the shorthand for climada.util.coordinates everywhere in the whole project. The shorthand co is taken from the base modules in the hazard and exposures subpackages.

climada/hazard/trop_cyclone.py
```diff
@@ -151,7 +151,17 @@ def set_from_tracks(self, tracks, centroids=None, description='',
...
-        LOGGER.info('Mapping %s tracks to %s centroids.', str(tracks.size),
+        # Restrict to coastal centroids within reach of any of the tracks
+        t_bounds = tracks.get_bounds(deg_buffer=CENTR_NODE_MAX_DIST_DEG)
+        t_mid_lon = 0.5 * (t_bounds[0] + t_bounds[2])
+        coastal_centroids = centroids.coord[coastal_idx]
+        co.lon_normalize(coastal_centroids[:, 1], center=t_mid_lon)
+        coastal_idx = coastal_idx[((t_bounds[0] <= coastal_centroids[:, 1])
+                                   & (coastal_centroids[:, 1] <= t_bounds[2])
+                                   & (t_bounds[1] <= coastal_centroids[:, 0])
+                                   & (coastal_centroids[:, 0] <= t_bounds[3]))]
```
tovogt, line 162: I did some performance testing and found these lines to introduce no significant slow-down in cases where it is redundant, e. g. if there is only a single track or in cases where the centroids are already chosen tightly around the track set.

climada/hazard/trop_cyclone.py
```diff
@@ -334,14 +346,17 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
         npositions = windfields.shape[0]
-        intensity = np.zeros(ncentroids)
-        intensity[reachable_coastal_centr_idx] = np.linalg.norm(windfields, axis=-1)\
-                                                          .max(axis=0)
+
+        intensity = np.linalg.norm(windfields, axis=-1).max(axis=0)
         intensity[intensity < self.intensity_thres] = 0
+        intensity = sparse.csr_matrix(
+            (intensity, reachable_coastal_centr_idx, [0, intensity.size]),
+            shape=(1, ncentroids))
```
tovogt, line 354: Instead of first allocating a temporary dense array of size `ncentroids`, we directly write the data to a sparse array. Usually this doesn't make a big difference because `ncentroids` is not that huge and garbage collection is really fast. But I think it's still more elegant since we already have the non-sparse indices (`reachable_coastal_centr_idx`) anyway.

climada/hazard/trop_cyclone.py
```diff
@@ -546,14 +561,14 @@ def _close_centroids(t_lat, t_lon, centroids, buffer=CENTR_NODE_MAX_DIST_DEG):
...
-    msk = (eye_bounds[:,None,1] < centr_lat[None])
-    msk &= (centr_lat[None] < eye_bounds[:,None,3])
-    msk &= (eye_bounds[:,None,0] < centr_lon[None])
-    msk &= (centr_lon[None] < eye_bounds[:,None,2])
-    return msk.any(axis=0)
+    mask = ((eye_bounds[:, None, 1] < centr_lat[None])
+            & (centr_lat[None] < eye_bounds[:, None, 3])
+            & (eye_bounds[:, None, 0] < centr_lon[None])
+            & (centr_lon[None] < eye_bounds[:, None, 2]))
+    return mask.any(axis=0)
```
tovogt, line 571: This is just for readability and a little code linting.

climada/entity/exposures/crop_production.py
chahank: I would suggest to go the other way around. 'co' is very undescriptive. Since it is not a very often used module such as numpy or  pandas, this would improve readability. What do you think?

climada/hazard/trop_cyclone.py
chahank: Could you expand the t_bounds to avoid the hard-coded indexing which is always very obscure to the reader? I actually do not know how to interpret them, hence my example below is probably incorrect.

```
[tr_lat_min, tr_lat_max, tr_lon_min, tr_lon_max] = tracks.get_bounds(deg_buffer=CENTR_NODE_MAX_DIST_DEG)
```

climada/hazard/trop_cyclone.py
chahank: If I am not mistaken, the variable `intensity` changes type between line 350 and 352. Having a variable change type during a code is I think a bad habit.

climada/hazard/trop_cyclone.py
chahank: Maybe a short comment to explain the mask could be useful. The hard coded indices are hard to read.

chahank (changes requested):

chahank:
Thanks for the upgrade! This will come in handy when trying to go for high resolution world-wide hazard footprints.

climada/entity/exposures/base.py
```diff
@@ -34,7 +34,7 @@
 from climada.entity.tag import Tag
 import climada.util.hdf5_handler as hdf5
 from climada.util.constants import ONE_LAT_KM, DEF_CRS
-import climada.util.coordinates as co
+import climada.util.coordinates as u_coord
```
tovogt, line 37: Following another convention I found in some of CLIMADA's modules, I switched to the naming scheme with a prefix `u_` for submodules of `climada.util` which is short while being recognizable. What do you think? Would you rather go for longer names, like `util_coord`?

climada/hazard/trop_cyclone.py
```diff
@@ -151,7 +151,18 @@ def set_from_tracks(self, tracks, centroids=None, description='',
             coastal_idx = ((centroids.dist_coast < INLAND_MAX_DIST_KM * 1000)
                            & (np.abs(centroids.lat) < 61)).nonzero()[0]
 
-        LOGGER.info('Mapping %s tracks to %s centroids.', str(tracks.size),
+        # Restrict to coastal centroids within reach of any of the tracks
+        t_lon_min, t_lat_min, t_lon_max, t_lat_max = tracks.get_bounds(
+            deg_buffer=CENTR_NODE_MAX_DIST_DEG)
```
tovogt, line 156: I generally agree that more explicit naming and more variables are beneficial. But then we should change the pylint directive that restricts the number of local variables in a function:
https://github.com/CLIMADA-project/climada_python/blob/dc82cc86c026bbd90c40733c62c9474d4088dad2/.pylintrc#L402-L403
The same goes for the maximum number of lines per module:
https://github.com/CLIMADA-project/climada_python/blob/dc82cc86c026bbd90c40733c62c9474d4088dad2/.pylintrc#L218-L219

climada/hazard/trop_cyclone.py
```diff
@@ -334,14 +347,17 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
-        intensity = np.zeros(ncentroids)
-        intensity[reachable_coastal_centr_idx] = np.linalg.norm(windfields, axis=-1)\
-                                                          .max(axis=0)
+
+        intensity = np.linalg.norm(windfields, axis=-1).max(axis=0)
         intensity[intensity < self.intensity_thres] = 0
+        intensity_sparse = sparse.csr_matrix(
+            (intensity, reachable_coastal_centr_idx, [0, intensity.size]),
+            shape=(1, ncentroids))
+        intensity_sparse.eliminate_zeros()
```
tovogt, line 356: As mentioned above, this introduces another local variable, which is against the spirit of the pylint-directive that restricts the number of local variables in a function.

climada/hazard/trop_cyclone.py
```diff
@@ -545,15 +561,14 @@ def _close_centroids(t_lat, t_lon, centroids, buffer=CENTR_NODE_MAX_DIST_DEG):
...
-    msk &= (eye_bounds[:,None,0] < centr_lon[None])
-    msk &= (centr_lon[None] < eye_bounds[:,None,2])
-    return msk.any(axis=0)
+    # check for each track position which centroids are within buffer, uses NumPy's broadcasting
+    mask = ((t_lat[:, None] - buffer < centr_lat[None])
+            & (centr_lat[None] < t_lat[:, None] + buffer)
+            & (t_lon[:, None] - buffer < centr_lon[None])
+            & (centr_lon[None] < t_lon[:, None] + buffer))
+    # for each centroid, check whether it is in the buffer for any of the track positions
+    return mask.any(axis=0)
```
tovogt, line 571: Added comments to clarify what is happening here.

tovogt (commented):
Thanks for your comments, I think I addressed all of them (see my comments in the code).

I feel like we have some conflicting coding guidelines if we don't want tuples with hard-coded indices and don't want variables to be reused for other purposes (with other types) and still request a maximum number of local variables in the pylintrc.

climada/entity/exposures/base.py
chahank: u_coord looks good to me. It is I think quickly clear what it is. Especially if we take the habit to use 'u_' to denote methods imported from `climada.util`.

climada/hazard/trop_cyclone.py
chahank: Good point. Let's take that up Wednesday at the developer's meeting quickly?

climada/hazard/trop_cyclone.py
chahank: Indeed. But here I would argue that it is really needed. Variables changing types is worse than pylint complaining about too many variables imho.

PR #124, opened 2021-01-26
Feature/knutson big tc

chahank:
This pull request is to fix a problem where `TropCyclone.set_climate_scenario_knu` would crash when a TC with a large number of tracks and centroids is used. 

Basically (as explained in a comment in the code), multiplying the intensity or frequency, which are sparse matrices, by a constant using `*=` is both slow and memory hungry. 

I solved it by detecting sparse matrices and then
a) replace `*=cst` by `csr_matrix.multiply(cst)
b) chunk the multiplication in small bits (very fast and memory-efficient in my testing)

### DIFF
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -405,7 +405,20 @@ class TropCyclone(Hazard):
                 change = chg['change'] * scale
             if select.any():
                 new_val = getattr(haz_cc, chg['variable'])
-                new_val[select] *= change
+                #For sparse matrices *= is ineficient (slow, uses large memory)
+                #Instead use multiply and chunking
+                if isinstance(new_val, sparse.csr_matrix):
+                    chunk_size = 10
+                    n_chunks = int(len(select) / chunk_size)
+                    chunks = np.arange(0, n_chunks, chunk_size)
+                    chunks = np.append(chunks, n_chunks)
+                    for start, stop in zip(chunks[:-1], chunks[1:]):
+                        new_val[select[start:stop]] = \
+                            new_val[select[start:stop]].multiply(change)
+                        print(start, stop)
+                else:
+                    new_val[select] *= change
+                setattr(haz_cc, chg['variable'], new_val)
                 setattr(haz_cc, chg['variable'], new_val)
         return haz_cc
 
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -412,7 +412,6 @@ class TropCyclone(Hazard):
                 else:
                     new_val[select] *= change
                 setattr(haz_cc, chg['variable'], new_val)
-                setattr(haz_cc, chg['variable'], new_val)
         return haz_cc
 
 def compute_windfields(track, centroids, model, metric="equirect"):
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -408,13 +408,7 @@ class TropCyclone(Hazard):
                 #For sparse matrices *= is ineficient (slow, uses large memory)
                 #Instead use multiply and chunking
                 if isinstance(new_val, sparse.csr_matrix):
-                    chunk_size = 10
-                    n_chunks = int(len(select) / chunk_size)
-                    chunks = np.arange(0, n_chunks, chunk_size)
-                    chunks = np.append(chunks, n_chunks)
-                    for start, stop in zip(chunks[:-1], chunks[1:]):
-                        new_val[select[start:stop]] = \
-                            new_val[select[start:stop]].multiply(change)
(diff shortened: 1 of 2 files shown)


climada/hazard/trop_cyclone.py
```diff
@@ -405,7 +405,19 @@ def _apply_criterion(self, criterion, scale):
...
+                    n_chunks = int(len(select) / chunk_size)
+                    chunks = np.arange(0, n_chunks, chunk_size)
+                    chunks = np.append(chunks, n_chunks)
+                    for start, stop in zip(chunks[:-1], chunks[1:]):
+                        new_val[select[start:stop]] = \
+                            new_val[select[start:stop]].multiply(change)
+                else:
+                    new_val[select] *= change
+                setattr(haz_cc, chg['variable'], new_val)
                 setattr(haz_cc, chg['variable'], new_val)
```
tovogt, line 421: This line is duplicated now, looks like a mistake to me.

climada/hazard/trop_cyclone.py
```diff
@@ -405,7 +405,19 @@ def _apply_criterion(self, criterion, scale):
...
+                #For sparse matrices *= is ineficient (slow, uses large memory)
+                #Instead use multiply and chunking
+                if isinstance(new_val, sparse.csr_matrix):
+                    chunk_size = 10
+                    n_chunks = int(len(select) / chunk_size)
+                    chunks = np.arange(0, n_chunks, chunk_size)
+                    chunks = np.append(chunks, n_chunks)
+                    for start, stop in zip(chunks[:-1], chunks[1:]):
+                        new_val[select[start:stop]] = \
+                            new_val[select[start:stop]].multiply(change)
```
tovogt, line 417: First of all: Didn't you attend my talk about avoiding for-loops? Okay, just joking. :D

This code doesn't do what you think it does. I pushed new test cases to your branch that demonstrate this.

Let me quickly explain, what's going wrong: The matrix `select[start:stop]` is a binary matrix of size 10 (or smaller). With a dense NumPy array, this would raise an IndexError:
```
>>> import numpy as np
>>> import scipy.sparse as sparse
>>> select_start_stop = [True, False, True]
>>> new_val = np.arange(24).reshape(6,4)
>>> new_val
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11],
       [12, 13, 14, 15],
       [16, 17, 18, 19],
       [20, 21, 22, 23]])
>>> new_val[select_start_stop]
  File "<stdin>", line 1, in <module>
IndexError: boolean index did not match indexed array along dimension 0; dimension is 6 but corresponding boolean dimension is 3
```
However, with SciPy's sparse matrices, there is no error. The mask is assumed to be for the beginning of the axis and `False` is assumed as a value for the missing entries:
```
>>> sparse.csr_matrix(new_val)[select_start_stop].toarrray()
array([[ 0,  1,  2,  3],
       [ 8,  9, 10, 11]])
```
So, you are only ever indexing (and modifying) the first 10 rows of the sparse matrix which is definitely not what you intend, right?

Here is my proposed change:
```suggestion
                    new_val = new_val.multiply((change * select  + ~select)[:, None]).tocsr()
```

climada/hazard/trop_cyclone.py
```diff
@@ -405,7 +405,19 @@ def _apply_criterion(self, criterion, scale):
                 change = chg['change'] * scale
             if select.any():
                 new_val = getattr(haz_cc, chg['variable'])
-                new_val[select] *= change
+                #For sparse matrices *= is ineficient (slow, uses large memory)
+                #Instead use multiply and chunking
```
tovogt, line 409:
```suggestion
                # 1d-masks like `select` are inefficient for indexing sparse matrices since
                # they are broadcasted densely in the second dimension
```

climada/hazard/trop_cyclone.py
tovogt: I have another proposal that uses matrix multiplication which is even faster:
```suggestion
                    new_val = sparse.diags(np.where(select, change, 1)).dot(new_val)
```

chahank: Aaaa yeah of course. Doing stuff late when tired is not a good idea :D. Somehow, although I immediately thought of the diagonal matrix multiplication, in the night I was convinced that the multiplication with a diagonal matrix would not give the right result. Thanks!

tovogt (approved):
Looks good to me now!

chahank:
I am just testing with an actual big TC to be 100% sure it does not blow the memory again. Then I would merge.

tovogt:
> I am just testing with an actual big TC to be 100% sure it does not blow the memory again. Then I would merge.

I already tested with all global IBTrACS events 1950-2019 at 360 arc-seconds (I have those precomputed), but another test from your side wouldn't be amiss.

chahank:
It works fine now!

Issue #129, opened 2021-02-01
Incorrect centroids assignment in Exposures.assign_centroids()


ingajsa:
An incorrect centroid exposure assignment potentially occurs in Exposures.assign_centroids()  when hazard centroids and exposure coordinates do not perfectly match, in particular when the exposure is located at the same latitude as the hazard, but not at the same longitude.

The commands in line 210 - 213 do not correctly exclude exposures located outside the raster, but potentially assign them to other centroids within the raster.

As this function is a core CLIMADA - function I would recommend everyone to check whether the work is affected by this error. I will set up a fix branch for this very soon.

Cheers,

Inga

chahank:
Thanks for bringing this up!  Since this is a very core function of CLIMADA, we should proceed with care. Could you please give a detailed account of the problem here and discuss what you think is the solution?

chahank:
I suggests to take the occasion to have a more general discussion about what the `assign_centroids()` method should do and how it is thought  to be used. 

Let me start by noting two points:

- If the hazard centroids are a raster, should exposure points outside the raster be considered? (current default is no, but there is an error for points close to the raster as mentioned by Inga above)
- If the hazard centroids are points, how far away from the centroids should exposure points be considered? There is currently a `threshold` that can be set. However, when `impact.calc` is called, `assign_centroids` is called always with the default. Should the user be notified about this? Or can we expect a user to know this and apply `assign_centroids` manually before `impact.calc`?

tovogt:
@chahank There is actually a serious bug in the `assign_centroids` function. You are hijacking this issue to start a much broader discussion that will hamper a quick solution (hot fix) to the original bug this is about. I think that will be very harmful in the end.

chahank:
Fair enough. But please document clearly what the serious bug is. This is so far not clear at all.

tovogt:
Exposure points that lie outside of the hazard raster in longitudinal, but not in latitudinal directions, are assigned to completely wrong hazard raster points. I added unit tests (6e2b29e67f96a5dc3692d82dd6e1add0132abdc6) that reproduce this problem:
https://github.com/CLIMADA-project/climada_python/blob/6e2b29e67f96a5dc3692d82dd6e1add0132abdc6/climada/entity/exposures/test/test_base.py#L90-L129
[...]

PR #130, opened 2021-02-01
Fix assign_centroids for raster hazards

tovogt:
As noted in #129, the centroids assignment in `Exposure.assign_centroids` is flawed for raster hazards in three ways:

1. Exposure points that lie outside of the hazard raster in longitudinal, but not in latitudinal directions, are assigned to completely wrong hazard raster points. In other words, exposure points to the (longitudinal) left or right of the hazard raster are affected.
2. The hazard raster is assumed to be nodal instead of cell-centered.
3. The assignment is not by nearest-neighbor, but by next-neighbor-to-the-left (floor instead of round).

This PR is a hot fix to make sure that the current behavior is consistent. A broader discussion of the `assign_centroids` function should remain unaffected by this. We can still change the general behavior later on.

I commited the regression tests before the fix, so you can follow the problem in the test results: https://ied-wcr-jenkins.ethz.ch/job/climada_branches/job/feature%252Ffix_assign_centroids/

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -203,13 +203,13 @@ class Exposures(GeoDataFrame):
             LOGGER.error('Set hazard and exposure to same CRS first!')
             raise ValueError
         if hazard.centroids.meta:
-            x_i = ((self.longitude.values - hazard.centroids.meta['transform'][2])
-                   / hazard.centroids.meta['transform'][0]).astype(int)
-            y_i = ((self.latitude.values - hazard.centroids.meta['transform'][5])
-                   / hazard.centroids.meta['transform'][4]).astype(int)
+            xres, _, xmin, _, yres, ymin = hazard.centroids.meta['transform'][:6]
+            xmin, ymin = xmin + 0.5 * xres, ymin + 0.5 * yres
+            x_i = np.round((self.longitude.values - xmin) / xres).astype(int)
+            y_i = np.round((self.latitude.values - ymin) / yres).astype(int)
             assigned = y_i * hazard.centroids.meta['width'] + x_i
-            assigned[assigned < 0] = -1
-            assigned[assigned >= hazard.centroids.size] = -1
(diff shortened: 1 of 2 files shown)


chahank:
Thanks, it seems to fix the issue. I however have one doubt about the following case:

(0,0)       (1,0)

(1,0)        (1,1)

Then, `xmin = 0.5, ymin=0.5`. Now, the point (0.6,0.6) should be associated with (1,1) as this is the closest point. However, `x_i = round((0.6-0.5)/1) = 0` and `y_i = round((0.6-0.5)/1) = 0`.

In the test setting, this would for instance mean the the point `(-19, 6.7)` should associated with `20`.

ingajsa:
> Thanks, it seems to fix the issue. I however have one doubt about the following case:
> 
> (0,0) (1,0)
> 
> (1,0) (1,1)
> 
> Then, `xmin = 0.5, ymin=0.5`. Now, the point (0.6,0.6) should be associated with (1,1) as this is the closest point. However, `x_i = round((0.6-0.5)/1) = 0` and `y_i = round((0.6-0.5)/1) = 0`.
> 
> In the test setting, this would for instance mean the the point `(-19, 6.7)` should associated with `20`.

Why would xmin then be 0.5 in this case and not 0?

chahank:
Because of `xmin, ymin = xmin + 0.5 * xres, ymin + 0.5 * yres`

tovogt:
Maybe there is a misunderstanding about cell-centered vs. nodal grids here. In CLIMADA, always follow the GDAL-convention to have cell-centered grids, see, e.g. this example:
https://github.com/CLIMADA-project/climada_python/blob/7b132f9a58cfb48c5d38d09b0d1e3c8d3c0ddc84/climada/entity/exposures/base.py#L265-L269

ingajsa:
> Because of `xmin, ymin = xmin + 0.5 * xres, ymin + 0.5 * yres`

Ah, sorry I considered the xmin, ymin coming out of the centroids.meta

PR #133, opened 2021-02-04
Make assign_centroids faster if some coordinates match exactly

chahank:
This is a small addition to the `assign_centroids` method of `exposure`. This is to address the case, which in my experience is rather common, that the exposure coordinates are a proper subset of the centroid coordinates. In this case, one can assign the centroids by a direct mapping, which is much faster than computing the distances. This becomes particularly important for very large centroid sets. 

For example, consider that someone defined a global hazard set for TC's worldwide. The number of centroids can easily be 10^6 or more. The task is then to compute the impact for a given country. As long as the exposure is defined on the same set of points, it is possible to just look at the intersection of coordinates between the centroids and the exposure. 

Please feel free to comment if you think this is unnecessary. Ideally, for large sets, one should probably rather work with rasters.


tovogt:
It's a good idea to improve performance in that direction. However, your implementation breaks the functionality and will produce wrong results if the exposure coordinates are a permutation of the hazard coordinates, see https://github.com/CLIMADA-project/climada_python/pull/133/commits/8cdd202007d6b2059481fdff1a84612a66f63f4e. 

I liked your idea, however, and extended it to generally look for exact matches of coordinates before using the assignement by nearest neighbor: https://github.com/CLIMADA-project/climada_python/pull/133/commits/78f0ac5d09a2131067591a947a0a488d7d619390

chahank:
That is super clever! Great improvement, and well spotted with the permutations. I will make one or two tests with my data just to be sure. Thanks.

chahank:
Ok, I am happy with it :D . But I think we have to wait for merging (until Wednesday at least) as the exposure class is being refactored. This will imply changes in how the values of the Exposure are accessed which will impact this method.

PR #136, opened 2021-02-05
Feature/storm europe include forecast functionality

ThomasRoosli:
With this pull request it is possible to read in windstorms from two new sources. COSMO model forecasts (used by MeteoSwiss among others), ICON model forecasts (used by DWD among others). These are implemented as two new functions in climada.hazard.StormEurope
The weather forecasts of DWD are freely available on their opendata plattfrom. The functionality to download weather forecasts for wind but also for other parameters is included as util functions


climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            stacked2 = ncdf.sel(time=event_date.strftime('%Y-%m-%d')
+                                ).groupby('date'
+                                          ).max().stack(intensity=('y_1',
+                                                                   'x_1'))
+            considered_dates = np.datetime64(event_date)
+        else:
+            time_covered_step = ncdf['time'].diff('time')
+            time_covered_day = time_covered_step.groupby('date').sum()
+            # forecast run should cover at least 18 hours of a day
+            days_to_consider = time_covered_day > np.timedelta64(18,'h')
```
Evelyn-M, line 248: Just a detail and you might disagree; but `days_to_consider` is a bit confusing in relation to the `considered_dates`. Could it make sense to name this here `considered_dates_bool` (if I understand it and that's what it is)?

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            time_covered_step = ncdf['time'].diff('time')
+            time_covered_day = time_covered_step.groupby('date').sum()
+            # forecast run should cover at least 18 hours of a day
+            days_to_consider = time_covered_day > np.timedelta64(18,'h') 
+            stacked2 = ncdf.groupby('date'
+                                    ).max().sel(date=days_to_consider
+                                                ).stack(intensity=('y_1',
+                                                                   'x_1'))
+            unique_dates = ncdf['time'].groupby('date').min()
+            considered_dates = unique_dates.sel(date=days_to_consider).values
```
Evelyn-M, line 254: necessary to create another var `unique_dates` here? or could this be directly extracted from `stacked2`

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
         ncdf.close()
         return new_haz
 
+
+    def read_cosmoe_file(self, fp_file, run_date, event_date=None,
```
Evelyn-M, line 210: the function doesn't just read a cosmo-e file, right? it reads a file AND sets a hazard from this file.
I'd perhaps suggest that you either refactor the actual reading part (until line 259), or rename the function into `set_storm_from_cosmoe` (or sth of the sort)

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+                                        model_name=model_name,
+                                        download_dir=grib_dir)
+
+        # create centroids
+        nc_centroids_file = download_icon_centroids_file(model_name, grib_dir)
+        self.centroids = self._centroids_from_nc(nc_centroids_file)
+
+        # read intensity from files
+        for ind_i, file_i in enumerate(file_names):
+            gripfile_path_i = os.path.join(file_i[:-4])
```
Evelyn-M, line 334: I think this should be replaced by a pathlib instance.

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            with open(file_i, 'rb') as source, open(gripfile_path_i, 'wb') as dest:
+                dest.write(bz2.decompress(source.read()))
+            ds_i = xr.open_dataset(gripfile_path_i, engine='cfgrib')
+            # os.remove(gripfile_path_i)
```
Evelyn-M, line 338: del line?

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            stacked2 = stacked.sel(valid_time=event_date.strftime('%Y-%m-%d')).groupby('date').max()
+            considered_dates = np.datetime64(event_date)
+        else:
+            time_covered_step = stacked['valid_time'].diff('valid_time')
+            time_covered_day = time_covered_step.groupby('date').sum()
+            # forecast run should cover at least 18 hours of a day
+            days_to_consider = time_covered_day > np.timedelta64(18,'h') 
+            stacked2 = stacked.groupby('date').max().sel(date=days_to_consider)
+            unique_dates = stacked['valid_time'].groupby('date').min()
+            considered_dates = unique_dates.sel(date=days_to_consider).values
```
Evelyn-M, line 360: similar comment as above

climada/hazard/test/test_storm_europe.py
```diff
@@ -141,11 +144,59 @@ def test_generate_prob_storms(self):
...
+        self.assertEqual(haz.event_name[-1], '2018-01-03_ens21')
+        self.assertIsInstance(haz.intensity,
+                              sparse.csr.csr_matrix)
+        self.assertIsInstance(haz.fraction,
+                              sparse.csr.csr_matrix)
+        self.assertEqual(haz.intensity.shape, (21, 25))
+        self.assertAlmostEqual(haz.intensity.max(), 36.426735,places=3)
+        self.assertEqual(haz.fraction.shape, (21, 25))
+    
+    def test_icon_read(self):
```
Evelyn-M, line 176: could you test for the appropriate warning if the hours are not 00 or 12?

(not sure one can test for logger output. else, would it make sense to actually raise an exception in the function?)
Or do the download functions that are called actually already take care of this?

climada/util/dwd_icon_loader.py
```diff
@@ -0,0 +1,258 @@
...
+Software Foundation, version 3.
+
+CLIMADA is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
+
+---
```
Evelyn-M, line 17: Just as a general thing; should we create an entirely new module for the dwd stuff? 

Or should we make a module that accumulates ALL kinds of file download functions? (e.g., i might have similar url-creation functions for OSM data queries, etc etc., so I might eventually put them in a downloader module, too.

climada/util/test/test_dwd_icon.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+        Path(file_name_i).touch()
+        delete_icon_grib(dt.datetime(1908, 2, 2),
+                         max_lead_time=1,
+                         download_dir=SYSTEM_DIR
+                         )
+        self.assertFalse(Path(file_name_i).exists())
+
+
+# Execute Tests
```
Evelyn-M, line 115: Nice tests ;)!

Evelyn-M (commented):
Hello Thomas ;)
very extensive work that you did! let me know if any of my comments are unclear!

climada/hazard/storm_europe.py
ChrisFairless: good point - we could also think about renaming the existing `read_footprints` and `_read_one_nc` methods so it's clear they're for WISC. (but how much would that break backward compatibility?)

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
         ncdf.close()
         return new_haz
 
+
+    def read_cosmoe_file(self, fp_file, run_date, event_date=None,
+                         model_name='COSMO-2E', description=None):
+        """Clear instance and read MeteoSwiss cosmoe weather forecast
+        footprint netcdf files into it. One event is one full day in UCT.
```
ChrisFairless, line 213:
```suggestion
        footprint netcdf files into it. One event is one full day in UTC.
```

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+        """
+        self.clear()
+        # create centroids
+        self.centroids = self._centroids_from_nc(fp_file)
+
+        # read intensity from file
+        ncdf = xr.open_dataset(fp_file)
+        ncdf = ncdf.assign_coords(date=('time',ncdf["time"].dt.floor("D")))
+
+        if event_date:
```
ChrisFairless, line 236: it could be worth adding a check that the `event_date` is covered by the input file plus an informative error. right now when it fails the code gives ~100 lines of error message.

climada/hazard/storm_europe.py
ChrisFairless: agree with this. also i'd suggest changing the inequality to `>=` (and the same later for ICON)

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            days_to_consider = time_covered_day > np.timedelta64(18,'h') 
+            stacked2 = ncdf.groupby('date'
+                                    ).max().sel(date=days_to_consider
+                                                ).stack(intensity=('y_1',
+                                                                   'x_1'))
+            unique_dates = ncdf['time'].groupby('date').min()
+            considered_dates = unique_dates.sel(date=days_to_consider).values
+
+
+        stacked2 = stacked2.stack(event=('date', 'epsd_1'))
```
ChrisFairless, line 257: i've got a bit confused by what we're calling an event. am i right in saying that (for the COSMO-E forecast), we're considering each _day_ of the forecast as one event, and then each event has 21 instances (i.e. ensemble members)?

if i've understood that right, it might be worth calling the 'event' dimension something else here, 'date_ensemble' or something...

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+        stacked2 = stacked2.where(stacked2.VMAX_10M > self.intensity_thres)
+        stacked2 = stacked2.fillna(0)
+
+        # fill in values from netCDF
+        self.intensity = sparse.csr_matrix(stacked2.VMAX_10M.T)
+        self.event_id = np.arange(stacked2.date.size)+1
```
ChrisFairless, line 263: should `event_id` be the same length as `event_name`? (i.e. IDs repeated for each date?)
i could look this up but instead i'm asking you here, sorry!

ChrisFairless: in general i've not worked with the code enough to know what dimensions these attributes should be - maybe you can explain over a zoom?

climada/hazard/storm_europe.py
ChrisFairless: edit: i see the method documentation explains this 😛
though it might still be worth changing the name of the dimension here

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+            )
+        # close netcdf file
+        ncdf.close()
+        self.check()
+
+    def read_icon_grib(self, run_date, event_date=None,
+                       model_name='icon-eu-eps', description=None,
+                       grib_dir=None, delete_raw_data=True):
+        """Clear instance and download and read dwd icon weather forecast
+        footprints into it. One event is one full day in UCT. Current setup
```
ChrisFairless, line 301:
```suggestion
        footprints into it. One event is one full day in UTC. Current setup
```

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+
+        Parameters:
+            fp_file (str): string directing to one netcdf file
+            run_date (datetime): The starting timepoint of the forecast run
+                of the cosmo model
+            event_date (datetime, optional): one day within the forecast
+                period, only this day (00H-24H) will be included in the hazard
+            model_name (str,optional): select the name of the icon model to
+                be downloaded. Must match the url string
+                (see _download_icon_grib for further info)
```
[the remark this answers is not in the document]
ChrisFairless: yeah - as far as i can tell the `model_name` parameter is only used to set the description (when `description` isn't provided). nothing is downloaded.

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+                       grib_dir=None, delete_raw_data=True):
+        """Clear instance and download and read dwd icon weather forecast
+        footprints into it. One event is one full day in UCT. Current setup
+        works for runs starting at 00H and 12H. Otherwise the aggregation is
+        inaccurate, because of the given file structure with 1-hour, 3-hour
+        and 6-hour maxima provided.
+
+        Parameters:
+            run_date (datetime): The starting timepoint of the forecast run
+                of the icon model
```
ChrisFairless, line 308: I had a look at the download URLs - it seemed that `run_date` has to correspond to the latest model run available for whichever hour (00, 06, 12, 18) you're requesting (and that the runs become available >3 hr after their timestamp). Is it worth mentioning this in the documentation?

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+        self.orig = np.ones_like(self.event_id)*False
+        self.orig[(stacked2.epsd_1 == 0).values] = True
+        self.date = np.repeat(
+            np.array(datetime64_to_ordinal(considered_dates)),
+            ncdf.epsd_1.size
+            )
+        self.event_name = [date_i + '_ens' + str(ens_i)
+                           for date_i, ens_i in zip(date_to_str(self.date),
+                                                    stacked2.epsd_1.values+1)
+                           ]
```
ChrisFairless, line 279: does this make `event_name` an array of length `min([len(self.date), len(stacked2.epsd_1)])`?
i'd imagine we want all combinations of date and ensemble number?

climada/hazard/storm_europe.py
ChrisFairless: update: most of my comments also apply to the DWD ICON methods, but i'm not going to point it all out twice! except this one apparently!

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+        # delete generated .grib2 and .4cc40.idx files
+        for ind_i, file_i in enumerate(file_names):
+            gripfile_path_i = os.path.join(file_i[:-4])
+            os.remove(gripfile_path_i)
+            idxfile_path_i = gripfile_path_i + '.4cc40.idx'
+            os.remove(idxfile_path_i)
+
+        if delete_raw_data:
+            #delete downloaded .bz2 files
+            delete_icon_grib(run_date, model_name=model_name)
```
ChrisFairless, line 410:
```suggestion
            delete_icon_grib(run_date, model_name=model_name, download_dir=download_dir)
```

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+
+
+        # delete generated .grib2 and .4cc40.idx files
+        for ind_i, file_i in enumerate(file_names):
+            gripfile_path_i = os.path.join(file_i[:-4])
+            os.remove(gripfile_path_i)
+            idxfile_path_i = gripfile_path_i + '.4cc40.idx'
+            os.remove(idxfile_path_i)
+
+        if delete_raw_data:
```
ChrisFairless, line 408: does the downloaded `icon_grid_0028_R02B07_N02.nc` count as raw data that should be deleted?

climada/hazard/test/test_storm_europe.py
```diff
@@ -141,11 +144,59 @@ def test_generate_prob_storms(self):
         self.assertEqual(storms_prob.size, 60)
         self.assertTrue(np.allclose((1 / storms_prob.frequency).astype(int), 330))
         self.assertAlmostEqual(storms.frequency.sum(),
-                               storms_prob.frequency.sum())
+                                storms_prob.frequency.sum())
```
ChrisFairless, line 147:
```suggestion
                               storms_prob.frequency.sum())
```

climada/hazard/test/test_storm_europe.py
```diff
@@ -141,11 +144,59 @@ def test_generate_prob_storms(self):
...
         self.assertEqual(storms_prob.centroids.size, 3054)
         self.assertIsInstance(storms_prob.intensity,
                               sparse.csr.csr_matrix)
+        
+    def test_cosmoe_read(self):
+        """test reading from cosmo-e netcdf"""
+        haz = StormEurope()
+        haz.read_cosmoe_file(DATA_DIR.joinpath('storm_europe_cosmoe_forecast_vmax_testfile.nc'),
+                              run_date=dt.datetime(2018,1,1),
+                              event_date=dt.datetime(2018,1,3))
```
ChrisFairless, line 158:
```suggestion
                             run_date=dt.datetime(2018,1,1),
                             event_date=dt.datetime(2018,1,3))
```

climada/util/dwd_icon_loader.py
ChrisFairless: yeah, i don't know if utils is the place to put it, if it's only used in the StormEurope module
i quite like the idea of a module for download functions
though deciding this shouldn't stop us from merging the PR

ChrisFairless: though this might be a Big Conversation since it's different to `tc_tracks`' way of handling data

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+
+        # create centroids
+        nc_centroids_file = download_icon_centroids_file(model_name, grib_dir)
+        self.centroids = self._centroids_from_nc(nc_centroids_file)
+
+        # read intensity from files
+        for ind_i, file_i in enumerate(file_names):
+            gripfile_path_i = os.path.join(file_i[:-4])
+            with open(file_i, 'rb') as source, open(gripfile_path_i, 'wb') as dest:
+                dest.write(bz2.decompress(source.read()))
```
ChrisFairless, line 336: Not very important, but I'd suggest moving the file decompression into the utils, and decompressing when you download (like you do with the centroids).

climada/hazard/storm_europe.py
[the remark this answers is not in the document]
ChrisFairless: also also, worth doing a find-replace for `_download_icon_grib` > `download_icon_grib`

test_data_api.py
```diff
@@ -82,6 +86,22 @@ def test_ecmwf_tc_bufr(self):
         fcast = TCForecast.fetch_bufr_ftp()
         [f.close() for f in fcast]
 
+    def test_icon_forecast_download(self):
+        """Test availability of DWD icon forecast."""
+        run_date = dt.datetime.today().replace(hour=0,
+                                               minute=0,
+                                               second=0,
+                                               microsecond=0)
```
ChrisFairless, line 94: given the lag between the ICON model run and its availability on the DWD site ... will this test fail between midnight and ~3:30am?

ChrisFairless: ughh and in time zones that are ahead of germany?? what time zone is jenkins in?

ChrisFairless:
this looks really good! i've left a bunch of comments and suggestions :)

i got a bit confused with the dimensions of StormEurope object attributes, since I don't know what they're used for. ignore any comments that grossly misunderstand them, or we can have a call.

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+
+    def read_cosmoe_file(self, fp_file, run_date, event_date=None,
+                         model_name='COSMO-2E', description=None):
+        """Clear instance and read MeteoSwiss cosmoe weather forecast
+        footprint netcdf files into it. One event is one full day in UCT.
+        Works for COSMO-1E and COSMO-2E
+
+        Parameters:
+            fp_file (str): string directing to one netcdf file
+            run_date (datetime): The starting timepoint of the forecast run
```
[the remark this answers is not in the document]
ThomasRoosli: Yes, that could be clearer. I renamed the variable run_datetime.

climada/hazard/storm_europe.py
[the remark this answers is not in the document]
ThomasRoosli: Yes, this part was replaced in the new documentation

climada/hazard/storm_europe.py
ThomasRoosli: I implemented both of your points.

climada/hazard/storm_europe.py
ThomasRoosli: done

climada/hazard/storm_europe.py
ThomasRoosli: Yes, implemented

climada/hazard/storm_europe.py
ThomasRoosli: I would prefer to not rename the read_footprints function for backwards compatibility. That is why I would name the new functions consistently (at least consistently within this class). Is that ok?

climada/hazard/storm_europe.py
ThomasRoosli: Done

climada/hazard/storm_europe.py
ThomasRoosli: Done

climada/hazard/test/test_storm_europe.py
ThomasRoosli: I added such a test, good idea.

climada/util/dwd_icon_loader.py
ThomasRoosli: I hope this functionality can be used outside of StormEurope, e.g. for Heat or Rainfall, that is why i put it in util. I created a new module, because there is very specific functions for this particular provider as the files/filenames change each day (would be similar with OSM probably). For the download of a few static files, I would not recommend building a specific module. Is it ok to merge this PR like this?

climada/util/test/test_dwd_icon.py
ThomasRoosli: Thanks :)

climada/hazard/storm_europe.py
ThomasRoosli: yes adapted in both places

climada/hazard/storm_europe.py
ThomasRoosli: I changed the name as suggested.

climada/hazard/storm_europe.py
ThomasRoosli: done

climada/hazard/storm_europe.py
ThomasRoosli: adjusted in another commit

climada/hazard/test/test_storm_europe.py
ThomasRoosli: reformated

climada/hazard/test/test_storm_europe.py
ThomasRoosli: reformated

climada/hazard/storm_europe.py
ThomasRoosli: I would prefer leaving it there. When delete_raw_data is set to False, and this function is run every day, an archive is build automatically, and there it makes sense to keep the compressed version.

test_data_api.py
ThomasRoosli: I was able to adapt it, so the test can run at any time.

climada/util/dwd_icon_loader.py
Evelyn-M: I don't have a strong opinion on this. Slight preference for a general download / reader module. Naming a very explicit module (such as the `dwd_icon_loader`) makes it of course easier to infer the specific functions that can be found inside as opposed to having to look through a `data_download` module and see if your case is covered. On the other hand, having lots of modules (`loader_x`, `loader_y`, `download_file_z`) etc. make the util folder quite crowded..

climada/hazard/storm_europe.py
```diff
@@ -199,6 +206,209 @@ def _read_one_nc(self, file_name, centroids):
...
+        # create centroids
+        self.centroids = self._centroids_from_nc(fp_file)
+
+        # read intensity from file
+        ncdf = xr.open_dataset(fp_file)
+        ncdf = ncdf.assign_coords(date=('time',ncdf["time"].dt.floor("D")))
+
+        if event_date:
+            # stacked = np.max(ncdf.sel(time=event_date.strftime('%Y-%m-%d')).VMAX_10M,
+            #                  axis=0).stack(intensity=('y_1', 'x_1'))
```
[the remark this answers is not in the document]
Evelyn-M: now this is really just a detail remaining - i'm not suuuper happy with the name `stacked2` ;). but up to you, fine to merge like this.

climada/hazard/storm_europe.py
ChrisFairless: yeah, let's not mess with backward compatibility then.
since the method will be called via `StormEurope.read_cosmoe_file()` i dont think the name is too unclear, but something like evelyn's suggestion would make it super obvious what it's doing

climada/util/dwd_icon_loader.py
ChrisFairless: ahhh, that makes sense. i'm fine with merging as it is.

maybe extend the comment at the start of the file saying it's currently used for the DWD wind forecast but will easily adapt to work with other variables.

climada/hazard/storm_europe.py
ChrisFairless: i'm fine with that 👍

climada/hazard/storm_europe.py
```diff
@@ -234,27 +240,30 @@ def read_cosmoe_file(self, fp_file, run_date, event_date=None,
...
-                                ).groupby('date'
-                                          ).max().stack(intensity=('y_1',
-                                                                   'x_1'))
+            try:
+                stacked2 = ncdf.sel(time=event_date.strftime('%Y-%m-%d')
+                                    ).groupby('date'
+                                              ).max().stack(intensity=('y_1',
+                                                                       'x_1'))
+            except KeyError:
+                raise ValueError('The selected event_date {} is not contained'
```
ChrisFairless, line 249:
```suggestion
                raise ValueError('Extraction of date and coordinates failed. '
                                 'This is most likely because '
                                 'the selected event_date {} is not contained'
```
This is really picky, but there are other, less likely, reasons you could get an error here (e.g. MeteoSwiss changes the coordinate names), so I'm suggesting a longer message.

climada/hazard/storm_europe.py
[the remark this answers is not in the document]
ThomasRoosli: It is a good point, I wanted to be consistent with the naming in read_icon_grib, so the parallels can be understood more easily

climada/hazard/storm_europe.py
ThomasRoosli: I implemented this, good suggestion.

climada/util/dwd_icon_loader.py
```diff
@@ -84,14 +85,33 @@ def download_icon_grib(run_date,
 
         # download file if it does not exist already
         if not bz2_pathfile_i.exists():
-            download_file(url + file_name_i,
-                          download_dir=download_dir)
+            try:
+                download_file(url + file_name_i,
+                              download_dir=download_dir)
+            except ValueError as err:
```
ChrisFairless, line 91: nice checks 👍

climada/hazard/storm_europe.py
ThomasRoosli: Yes, this was fixed in download_icon_grib

climada/hazard/storm_europe.py
ThomasRoosli: No, my reasoning was: as the daily forecasts, will change everyday and `delete_raw_data` basically gives a choice if these daily forecasts should be archived. The `icon_grid...` file on the other hand is used in every call of this function and is saved permanently for later reuse.

test_data_api.py
```diff
@@ -88,13 +88,14 @@ def test_ecmwf_tc_bufr(self):
...
-        run_date = dt.datetime.today().replace(hour=0,
-                                               minute=0,
-                                               second=0,
-                                               microsecond=0)
-        icon_file = download_icon_grib(run_date,max_lead_time=1)
+        run_datetime = dt.datetime.utcnow() - dt.timedelta(hours=5)
+        run_datetime = run_datetime.replace(hour=run_datetime.hour//12*12,
+                                            minute=0,
+                                            second=0,
+                                            microsecond=0)
```
ChrisFairless, line 95: nice!

climada/hazard/storm_europe.py
ThomasRoosli: Done

climada/util/dwd_icon_loader.py
ThomasRoosli: Comment is extended

climada/hazard/storm_europe.py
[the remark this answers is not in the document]
ThomasRoosli: I changed both functions, now there is only one variable called `stacked` in both functions.

ThomasRoosli:
I think, this is now ready to merge. (I will not merge until after the next release involving the refactoring of exposure).

Issue #134, opened 2021-02-05
read_ibtracs_netcdf: why round first time to hours?


tovogt:
In TCTracks.read_ibtracs_netcdf, the first time record is rounded to full hours:
https://github.com/CLIMADA-project/climada_python/blob/b4e84e44c60a6e56a7d7bbc747ffe2da7265ce3c/climada/hazard/tc_tracks.py#L414
 Why is this necessary? I remember that I once stumbled upon a problem with `equal_timestep` if the first time wasn't rounded to full hours, but I don't have time right now to verify this.

This issue is for discussing whether the rounding is necessary, harmless, avoidable or should at least be better documented.

bguillod:
Do we know how often (how many tracks) experience first time steps that are not at a round hour?

tovogt:
Regarding your question: Only 22 of the 13469 tracks in IBTrACS are affected by this.

Furthermore, I found the time to check this out: The problem is that `equal_timestep` produces NaNs for data sets that don't start at full hours. Here is an example (after manually removing the rounding in `read_ibtracs_netcdf`):
```
>>> tr = TCTracks()
>>> tr.read_ibtracs_netcdf(storm_id="1998335S12138")
>>> tr.data[0].central_pressure[:5].to_series()
time
1998-11-30 23:30:00    1007.0
1998-12-01 00:00:00    1006.0
1998-12-01 03:00:00    1005.0
1998-12-01 05:30:00    1004.0
1998-12-01 06:00:00    1004.0
Name: central_pressure, dtype: float32
>>> tr.equal_timestep(time_step_h=1)
>>> tr.data[0].central_pressure[:5].to_series()
time
1998-11-30 23:00:00            NaN
1998-12-01 00:00:00    1006.000000
1998-12-01 01:00:00    1005.666667
1998-12-01 02:00:00    1005.333333
1998-12-01 03:00:00    1005.000000
Name: central_pressure, dtype: float64
```
Note that the time steps are not aligned with the start time of the data set, but they are always aligned to 00:00:00 of the first day. So, if you specify `time_step_h=0.75`, the first time step will be at `23:15:00` because the resample bin from `23:15:00` until `00:00:00` is the first resample bin that overlaps the time index. In this case, this produces a NaN value at `23:00:00` which might lead to unwanted problems later on.

I think that this is actually a problem of the `equal_timestep` function and we shouldn't fix this kind of problem in the `read_ibtracs_netcdf` function.

[...]

PR #137, opened 2021-02-08
TCTracks.equal_timestep: restrict to times within original bounds

tovogt:
Following up on #123 about improving the `TCTracks.read_ibtracs_netcdf` function, we found a little issue (#134) in the `TCTracks.equal_timestep` function that affects tracks for which the first time step is not rounded to full hours. More precisely, all tracks are affected for which the first time step is not a multiple of `time_step_h`.

For example, if `time_step_h=1` and the first time step of the track is `06:30:00`, then the output of `equal_timestep` will be a track that starts at `06:00:00` with NaN values. The following time steps (`07:00:00` and so on) will be okay.

Another, more general, example: If `time_step_h=0.75` and the first time step of the track is `02:00:00`, then the output of `equal_timestep` will be a track that starts at `01:30:00` with NaN values. The following time steps (`02:15:00`, `03:00:00` and so on) will be okay.

This PR restricts the output of `equal_timestep` to times within the time bounds of the original track so that no additional NaNs are produced.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -511,7 +511,6 @@ class TCTracks():
             st_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
 
             # set time_step in hours
-            track_ds.time.values[:1] = track_ds.time[:1].dt.floor('H')
             track_ds['time_step'] = xr.ones_like(track_ds.time, dtype=float)
             if track_ds.time.size > 1:
                 track_ds.time_step.values[1:] = (track_ds.time.diff(dim="date_time")
@@ -1293,6 +1292,9 @@ class TCTracks():
             track_int.attrs['category'] = set_category(
                 track_int.max_sustained_wind.values,
                 track_int.max_sustained_wind_unit)
+            # restrict to time steps within original bounds
+            track_int = track_int.sel(
+                time=(track.time[0] <= track_int.time) & (track_int.time <= track.time[-1]))
         else:
             LOGGER.warning('Track interpolation not done. '
                            'Not enough elements for %s', track.name)
--- a/climada/hazard/test/test_tc_tracks.py
+++ b/climada/hazard/test/test_tc_tracks.py
@@ -463,6 +463,7 @@ class TestFuncs(unittest.TestCase):
         self.assertEqual(np.max(tc_track.data[0].radius_max_wind), 0)
         self.assertEqual(np.min(tc_track.data[0].radius_max_wind), 0)
         self.assertEqual(tc_track.data[0].max_sustained_wind[21], 25)
+        self.assertTrue(np.isfinite(tc_track.data[0].central_pressure.values).all())
         self.assertAlmostEqual(tc_track.data[0].central_pressure.values[29], 1008, places=0)
         self.assertEqual(np.max(tc_track.data[0].environmental_pressure), 1010)
         self.assertEqual(np.min(tc_track.data[0].environmental_pressure), 1010)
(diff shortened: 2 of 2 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -511,7 +511,6 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
             st_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
 
             # set time_step in hours
-            track_ds.time.values[:1] = track_ds.time[:1].dt.floor('H')
```
tovogt, line 514: I was the one who originally added this line long ago, so I can guarantee that it was only a dirty fix for the behavior of equal_timestep. We can safely remove it now.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -463,6 +463,7 @@ def test_interp_track_pass(self):
         self.assertEqual(np.max(tc_track.data[0].radius_max_wind), 0)
         self.assertEqual(np.min(tc_track.data[0].radius_max_wind), 0)
         self.assertEqual(tc_track.data[0].max_sustained_wind[21], 25)
+        self.assertTrue(np.isfinite(tc_track.data[0].central_pressure.values).all())
```
tovogt, line 466: We want to guarantee that no additional NaNs are produced by `equal_timestep`.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -480,18 +481,23 @@ def test_interp_track_pass(self):
...
+                tc_track.equal_timestep(time_step_h=time_step_h)
+                np.testing.assert_array_equal(tc_track.data[0].time_step, time_step_h)
+                self.assertTrue(np.isfinite(tc_track.data[0].central_pressure.values).all())
 
         tc_track = tc.TCTracks()
         tc_track.read_processed_ibtracs_csv(TEST_TRACK)
         tc_track.equal_timestep(time_step_h=0.16667)
 
-        self.assertEqual(tc_track.data[0].time.size, 1333)
+        self.assertEqual(tc_track.data[0].time.size, 1332)
```
tovogt, line 497: The first time step used to be a NaN produced by `equal_timestep`. With this PR, the first time step is discarded because it's outside of the original time range.

climada/hazard/tc_tracks.py
```diff
@@ -1293,6 +1292,9 @@ def _one_interp_data(track, time_step_h, land_geom=None):
             track_int.attrs['category'] = set_category(
                 track_int.max_sustained_wind.values,
                 track_int.max_sustained_wind_unit)
+            # restrict to time steps within original bounds
+            track_int = track_int.sel(time=((track.time[0] <= track_int.time)
+                                            & (track_int.time <= track.time[-1])))
```
bguillod, line 1297: I'd find it more easy to read like that:
```suggestion
            track_int = track_int.sel(time=((track_int.time >= track.time[0])
                                            & (track_int.time <= track.time[-1])))
```
because you check whether `track_int.time` is within some bounds, but of course this has absolutely no impact on the result so you can keep as is, up to you.

bguillod (approved):
Looks perfect to me - thanks a lot.

climada/hazard/tc_tracks.py
tovogt: In Python, there is the shorthand notation `lower <= x <= upper` for scalar variables `x`. Unfortunately, there is no chaining of array comparisons in Python so far (see https://www.python.org/dev/peps/pep-0535/). That's why I'm always using the analogous ordering of expressions `(lower <= x_arr) & (x_arr <=upper)` for array variables `x_arr`.

I hope it's clearer when written in a single line of code:
https://github.com/CLIMADA-project/climada_python/blob/a33c91c54b867a5405798c06269a05033ed07a07/climada/hazard/tc_tracks.py#L1295-L1297

PR #138, opened 2021-02-09
Restructure xarray dataset in TCForecast tracks

manniepmkam:
This branch in starting from feature/bugfix_TCForecast_basins.

A minor amendment in the xarray dataset that holds the TC tracks from `TCForecast`. Data variables `environmental_pressure` and `radius_max_wind` were wrongly assigned to coordinates. Subsequently these two variables could not be interpolated using the `equal_timestep()` method. 

New amendments assigned the above two variables to the `xarray.Dataset.data_vars`. Details about the data structure of xarray please see[http://xarray.pydata.org/en/stable/generated/xarray.Dataset.html]( http://xarray.pydata.org/en/stable/generated/xarray.Dataset.html).


ThomasRoosli (changes requested):
I could not really verify the exact nature of the proposed changes as a test for this is missing. Please include the way `equal_timesteps` is accessing TCForecast into the test of TCForecast, so this bug does not reappear.

chahank:
If this pull request is ready to merge, please do so.

PR #142, opened 2021-02-10
Feature/refactor exposures

emanuel-schmid:
This pull request addresses the refactoring of the Exposures class.

* The new class design prefers composition over inheritance: instead of inheriting from `GeoDataFrame`, `Exposures` now has a `GeoDataFrame` attribute `gdf`. This enables climada to be less aware of internal changes in the `GeoDataFrame` implementation and thus facilitates updating the geopandas dependency.
* All accesses to `Exposures` objects have been changed throughout the code base to access the `gdf` attribute instead - unless they have not been noticed in the automated tests. 
* Additionally, a `concat` function has been added to the `exposures` module, replacing the `pandas.concat` function where they were applied to `Exposures` lists.
* Some minor bug fixes and code clearing was made along.


climada/entity/exposures/black_marble.py
```diff
@@ -36,6 +34,7 @@
 from climada.entity.exposures.base import Exposures, INDICATOR_IF
 from climada.util.constants import SYSTEM_DIR, DEF_CRS
 from climada.entity.exposures import nightlight as nl_utils
+from climada.entity.exposures import concat as concat_exp
```
tovogt, line 37: This PR is fine for me, apart from this detail: Calling a standalone function `concat` is kind of forcing you to use statements like `from xyz import concat as concat_xyz`. Instead, I would prefer to have a `@staticmethod` called `concat` added to the `Exposures` class, so that the API is always exposed as `Exposures.concat` - less headache about potential naming conflicts.

emanuel-schmid: Thanks for asking. There were two reasons for a function instead of a static method:
1) Analogy to pd.concat.
2) When we discussed static methods in the coding conventions meeting, we agreed on using them only with a good reason.

Your arguments convinced me that 1) is weak and there _is_ a good reason.
I'll change it.

tovogt: Thanks for your consideration! I don't like static methods either and remember that I also argued against static methods in the coding conventions meeting. I think that concatenation operations are among the few legitimate use cases of static methods.

Issue #139, opened 2021-02-10
u_coord.points_to_raster: bounds when crossing antimeridian


chahank:
For points sets that are distributed across the International Date Line, the raster will contain all longitudes (-180, 180) because of
```
# construct raster
xmin, ymin, xmax, ymax = (points_df.longitude.min(), points_df.latitude.min(),
                          points_df.longitude.max(), points_df.latitude.max())
```
As a consequence, using for example `impact.plot_raster_eai_exposure()` for the USA will contain all longitudes instead of only the USA.

Question: how does one make a raster which crosses the International Date Line with rasterio?

tovogt:
The way to go is to represent longitudes not in the interval [-180,180], but in some other interval of length 360, such as [0, 360].  For example, you could do something like this:
```
xmin, ymin, xmax, ymax = latlon_bounds(points_df.latitude.values, points_df.longitude.values)
x_mid = 0.5 * (xmin + xmax)
```
After that, you  have to renormalize all your x-coordinates (in `df_poly` and `points_df`) using `lon_normalize(..., center=x_mid)`.

bguillod:
We had the same issue in another context (with R and leaflet) and ended up creating a list containing either a single raster (if within -180 to +180) or two rasters (if crossing the +/-180, one e.g. between 160 and 180 and one between e.g. -180 and -170, if you raster goes from +160 to -170 through the antimeridian).
Either way, you have to make sure the functionalities using your object will get the correct coordinates or be able to deal with shifted coordinates. The tricky part is often to determine whether the range is -180 to +180 because it crosses the anti-meridian or whether your raster should cover the whole globe. But perhaps the reply from @tovogt above is sufficient for your needs.

tovogt:
Ideally, a function that gets longitudinal values as an input and requires them to be normalized to [-180, 180] should mention it in the docstring. Otherwise, functions should automatically normalize longitudinal input values according to their needs. We have the util-functions to do that, it's not that hard actually.

But you are right: At the moment, there are a lot of functions in CLIMADA that silently assume longitudinal values to be in [-180, 180] and will do crazy things for stuff crossing the antimeridian. I think, we should aim at improving all of those functions step by step so that they can handle those cases appropriately.

chahank:
Hmm does this mean one cannot solve this directly in the util function, but one hast to modify each plotting function since the data must be renomarlized?

tovogt:
[...]

PR #145, opened 2021-02-15
u_coord.points_to_raster: fix for antimeridian

tovogt:
Following the discussion in #139, this PR fixes `climada.util.coordinates.points_to_raster` to correctly handle points that lie in an area crossing the antimeridian.

While I'm at it, I simplified the code of `climada.util.coordinates.read_raster_bounds` which was more complicated than necessary when the given bounds cross the antimeridian.

Similarly, I fixed `Centroids.plot` as well as `Exposures.plot_raster` and `Exposures.plot_hexbin` to correctly handle plot data crossing the antimeridian.

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -32,6 +32,7 @@ from geopandas import GeoDataFrame
 import rasterio
 from rasterio.warp import Resampling
 import contextily as ctx
+import cartopy.crs as ccrs
 
 from climada.entity.tag import Tag
 import climada.util.hdf5_handler as u_hdf5
@@ -462,19 +463,22 @@ class Exposures():
             ras_tiff.close()
         # make plot
         crs_epsg, _ = u_plot.get_transformation(self.crs)
-        xmin, ymin, xmax, ymax = self.gdf.longitude.min(), self.gdf.latitude.min(), \
-        self.gdf.longitude.max(), self.gdf.latitude.max()
+        if isinstance(crs_epsg, ccrs.PlateCarree):
+            xmin, ymin, xmax, ymax = u_coord.latlon_bounds(
+                self.gdf.latitude.values, self.gdf.longitude.values)
+            mid_lon = 0.5 * (xmin + xmax)
+            crs_epsg = ccrs.PlateCarree(central_longitude=mid_lon)
+        else:
+            xmin, ymin, xmax, ymax = (self.gdf.longitude.min(), self.gdf.latitude.min(),
+                                      self.gdf.longitude.max(), self.gdf.latitude.max())
         if not axis:
             _, axis = u_plot.make_map(proj=crs_epsg)
         cbar_ax = make_axes_locatable(axis).append_axes('right', size="6.5%",
                                                         pad=0.1, axes_class=plt.Axes)
-        axis.set_extent([max(xmin, crs_epsg.x_limits[0]),
-                         min(xmax, crs_epsg.x_limits[1]),
-                         max(ymin, crs_epsg.y_limits[0]),
-                         min(ymax, crs_epsg.y_limits[1])], crs_epsg)
+        axis.set_extent((xmin, xmax, ymin, ymax), crs_epsg)
         u_plot.add_shapes(axis)
         imag = axis.imshow(raster_f(raster), **kwargs, origin='upper',
-                           extent=[xmin, xmax, ymin, ymax], transform=crs_epsg)
+                           extent=(xmin, xmax, ymin, ymax), transform=crs_epsg)
         plt.colorbar(imag, cax=cbar_ax, label=label)
         plt.draw()
         return axis
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -266,7 +266,7 @@ class Exposures():
             haz_coord = hazard.centroids.coord
 
             if np.array_equal(coord, haz_coord):
-                assigned = np.arange(self.shape[0])
+                assigned = np.arange(self.gdf.shape[0])
             else:
                 # pairs of floats can be sorted (lexicographically) in NumPy
                 coord_view = coord.view(dtype='float64,float64').reshape(-1)
@@ -462,23 +462,26 @@ class Exposures():
             ras_tiff.write(raster.astype(np.float32), 1)
             ras_tiff.close()
         # make plot
-        crs_epsg, _ = u_plot.get_transformation(self.crs)
-        if isinstance(crs_epsg, ccrs.PlateCarree):
+        proj_data, _ = u_plot.get_transformation(self.crs)
+        proj_plot = proj_data
+        if isinstance(proj_data, ccrs.PlateCarree):
+            # use different projections for plot and data to shift the central lon in the plot
             xmin, ymin, xmax, ymax = u_coord.latlon_bounds(
                 self.gdf.latitude.values, self.gdf.longitude.values)
-            mid_lon = 0.5 * (xmin + xmax)
-            crs_epsg = ccrs.PlateCarree(central_longitude=mid_lon)
+            proj_plot = ccrs.PlateCarree(central_longitude=0.5 * (xmin + xmax))
         else:
             xmin, ymin, xmax, ymax = (self.gdf.longitude.min(), self.gdf.latitude.min(),
                                       self.gdf.longitude.max(), self.gdf.latitude.max())
+
         if not axis:
-            _, axis = u_plot.make_map(proj=crs_epsg)
+            _, axis = u_plot.make_map(proj=proj_plot)
+
         cbar_ax = make_axes_locatable(axis).append_axes('right', size="6.5%",
                                                         pad=0.1, axes_class=plt.Axes)
-        axis.set_extent((xmin, xmax, ymin, ymax), crs_epsg)
+        axis.set_extent((xmin, xmax, ymin, ymax), crs=proj_data)
         u_plot.add_shapes(axis)
         imag = axis.imshow(raster_f(raster), **kwargs, origin='upper',
-                           extent=(xmin, xmax, ymin, ymax), transform=crs_epsg)
+                           extent=(xmin, xmax, ymin, ymax), transform=proj_data)
         plt.colorbar(imag, cax=cbar_ax, label=label)
         plt.draw()
         return axis
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -24,6 +24,7 @@ import copy
 import logging
 from pathlib import Path
 
+import cartopy.crs as ccrs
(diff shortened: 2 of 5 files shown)


climada/util/coordinates.py
```diff
@@ -1333,33 +1333,27 @@ def read_raster_bounds(path, bounds, res=None, bands=None):
...
+        crs = DEF_CRS if src.crs is None else src.crs
+        for iband, band in enumerate(bands):
+            rasterio.warp.reproject(
+                source=rasterio.band(src, band),
+                destination=data[iband],
+                src_transform=src.transform,
+                src_crs=src.crs,
+                dst_transform=transform,
+                dst_crs=crs,
+                resampling=resampling)
```
chahank, line 1356: This is the code that takes care of the anti-meridian right?

tovogt: This code is designed in a way that there is no need to have a special treatment of bounds that cross the anti-meridian.

chahank: Thanks a lot for taking up this issue! 

I tested the plotting function `impact().plot_raster_eai_exposure()` which is at the original motivation for issue #139. The plotting is not resolved. For the Fidji island, it is still the whole worl that is plotted (all longitudes due to the antimedirian).

tovogt: That's not the point of this PR. This PR is about fixing `points_to_raster` and simplifying `raster_read_bounds`.

chahank: But now the return value of `points_to_raster` is modified right? It potentially returns a raster which has  renormalized longitudes. This is not necessarily compatible with how currently other climada functions/method use it.

tovogt: Yes, that's true. We should check that this change doesn't break other functions that use `points_to_raster`. I did this now, and found that (luckily) only `Exposures.plot_raster` is affected and needs some adjustments. I will deal with it later, and come back to you with an update to this PR later this week.

chahank: Awesome! Thanks a lot for taking care of this. It would have taken me much more time and the outcome would not have been as clean.

climada/entity/exposures/base.py
```diff
@@ -462,8 +463,14 @@ def plot_raster(self, res=None, raster_res=None, save_tiff=None,
...
-        xmin, ymin, xmax, ymax = self.gdf.longitude.min(), self.gdf.latitude.min(), \
-        self.gdf.longitude.max(), self.gdf.latitude.max()
+        if isinstance(crs_epsg, ccrs.PlateCarree):
+            xmin, ymin, xmax, ymax = u_coord.latlon_bounds(
+                self.gdf.latitude.values, self.gdf.longitude.values)
+            mid_lon = 0.5 * (xmin + xmax)
+            crs_epsg = ccrs.PlateCarree(central_longitude=mid_lon)
+        else:
+            xmin, ymin, xmax, ymax = (self.gdf.longitude.min(), self.gdf.latitude.min(),
+                                      self.gdf.longitude.max(), self.gdf.latitude.max())
```
chahank, line 473: Coming back to the `impact().plot_raster_eai_exposure()`, shouldn't fix to `plot_raster` should fix the problem of plotting impacts for exposures crossing the antimeridian?

tovogt: The code you're referring to here was a quick sketch. I don't have time today to test it. But maybe you are willing to do the testing? And maybe you can report back whether it also fixes `plot_raster_eai_exposure` for you?

chahank: Unforuntaly the `plot_raster_eai_exposure` cannot deal with the change. For example I plotted the USA. Attached are the files before and after the change.

Imp_USA_year2020_rcp26

Imp_USA_year2020_rcp26

tovogt: The output is not really helpful. The code to reproduce would be helpful.

chahank: It is 
```
from climada.util.plot import make_map
fig, ax = make_map(num_sub=1, figsize=(9, 13))
imp.plot_raster_eai_exposure(axis=ax, label=xlabel)
```
where imp is an impact object with the USA as exposure.

To add to the above, not using a raster gives the correct map:

```
from climada.util.plot import make_map
fig, ax = make_map(num_sub=1, figsize=(9, 13))
imp.plot_hexbin_eai_exposure(axis=ax, label=xlabel)
```

Imp_USA_year2020_rcp26s

tovogt: Would you please share a fully working example?

chahank:
```
import numpy as np
import pandas as pd
from scipy import sparse as sparse
from climada.entity import Exposures, IFTropCyclone, ImpactFuncSet
from climada.engine import Impact

lon_list = np.append(np.arange(-170, -180.5, -1), np.arange(170, 180.5, 1))
lat_list = np.arange(-10, 10, 1)

df_exp = pd.DataFrame()
df_exp['latitude'] = np.repeat(lat_list, len(lon_list))
df_exp['longitude'] = np.repeat(lon_list, len(lat_list))
df_exp['value'] = 100
df_exp['if_TC'] = 1 
exp = Exposures(df_exp)
exp.set_geometry_points()
exp.check()

n_cen = len(exp.gdf.longitude) # number of centroids
n_ev = 10 # number of events
haz = Hazard('TC')
haz.centroids.set_lat_lon(exp.gdf.latitude, exp.gdf.longitude) # default crs used
haz.intensity = sparse.csr_matrix(np.random.random((n_ev, n_cen))*100)
haz.units = 'm'
haz.event_id = np.arange(n_ev, dtype=int)
haz.event_name = ['ev_12', 'ev_21', 'Maria', 'ev_35', 'Irma', 'ev_16', 'ev_15', 'Edgar', 'ev_1', 'ev_9']
haz.date = [721166, 734447, 734447, 734447, 721167, 721166, 721167, 721200, 721166, 721166]
haz.orig = np.zeros(n_ev, bool)
haz.frequency = np.ones(n_ev)/n_ev
haz.fraction = haz.intensity.copy()
haz.fraction.data.fill(1)
haz.check()
haz.centroids.plot()

exp.assign_centroids(haz)

impf = IFTropCyclone()
impf.set_emanuel_usa()
impf_set= ImpactFuncSet()
impf_set.append(impf)

imp = Impact()
imp.calc(exp, impf_set, haz)

imp.plot_hexbin_eai_exposure(pop_name=False)
imp.plot_raster_eai_exposure()

from climada.util.plot import make_map
fig, ax = make_map(num_sub=1, figsize=(9, 13))
imp.plot_raster_eai_exposure(axis=ax)
```

chahank: Now the plot with `imp.plot_raster_eai_exposure()` is completely false since the raster is now made with renormalized longitudes. On the other hand `imp.plot_hexbin_eai_exposure()` gives a correct plot (since it does not generate a raster), but shows all longitudes (i.e. does not deal with the antimeridian).

tovogt: Thanks for the example, this is very helpful!

I fixed `Centroids.plot`, `Exposures.plot_raster` and `Exposures.plot_hexbin` for data crossing the antimeridian. However, the version in the last two lines will never work. In that case, the axis is provided by the user, so the user has already specified the projection of the axis (including the central longitude). The projection of an axis cannot be changed after initialization.

climada/entity/exposures/base.py
```diff
@@ -265,7 +266,7 @@ def assign_centroids(self, hazard, method='NN', distance='haversine',
             haz_coord = hazard.centroids.coord
 
             if np.array_equal(coord, haz_coord):
-                assigned = np.arange(self.shape[0])
+                assigned = np.arange(self.gdf.shape[0])
```
tovogt, line 269: Totally unrelated with this PR, but an obvious bug that I stumbled upon occasionally.

climada/entity/exposures/base.py
```diff
@@ -461,20 +462,26 @@ def plot_raster(self, res=None, raster_res=None, save_tiff=None,
             ras_tiff.write(raster.astype(np.float32), 1)
             ras_tiff.close()
         # make plot
-        crs_epsg, _ = u_plot.get_transformation(self.crs)
-        xmin, ymin, xmax, ymax = self.gdf.longitude.min(), self.gdf.latitude.min(), \
-        self.gdf.longitude.max(), self.gdf.latitude.max()
+        proj_data, _ = u_plot.get_transformation(self.crs)
+        proj_plot = proj_data
```
tovogt, line 466: The problem with most of CLIMADA's plot functions is that they assume the same cartopy projection for the data as for the plot (axis). But the data might be given in a projection different from the axis projection. Here, I introduce two different variables for the two. Using this, the axis projection might shift the 0-longitude to a different location without any need to touch the data itself.

climada/hazard/centroids/centr.py
```diff
@@ -36,18 +37,7 @@
...
-                                      equal_crs,
-                                      get_country_code,
-                                      get_resolution,
-                                      pts_to_raster_meta,
-                                      raster_to_meshgrid,
-                                      read_raster,
-                                      read_raster_sample,
-                                      read_vector,
-                                      NE_CRS)
+import climada.util.coordinates as u_coord
```
tovogt, line 40: This is unrelated to this PR, but I felt like it improves code readability

climada/entity/exposures/base.py
chahank: :D I think I already commited that fix to develop.

tovogt: I don't see it in develop: https://github.com/CLIMADA-project/climada_python/blob/develop/climada/entity/exposures/base.py#L267-L269

chahank: Right, I added it to the latest pull request #147. Was maybe not the smartest.

climada/entity/exposures/base.py
chahank: Aaa now that makes sense!

climada/entity/exposures/base.py
chahank: Nice, that was quick! I guess we should now go through all the plotting functions. Or did you already check it?

climada/entity/exposures/base.py
tovogt: No, that's fine!

climada/entity/exposures/base.py
tovogt: There are so many plotting functions in CLIMADA that don't deal with the antimeridian correctly. Going through all of them would become a really huge task. The main idea behind this PR was the `points_to_raster` functionality and for that, I already checked all affected plotting functions. I suggest to merge the PR and deal with the remaining plotting functions later.

tovogt:
Continue the discussion for other plot functions in #153, please.

Issue #143, opened 2021-02-10
tests on Jenkins


aleeciu:
According to Jenkins, test_blackmarble.py is only crashing at Test2013.test_sint_maarten, but when I run the file in spyder with the develop branch I get way more errors, basically all tests within that file are crashing...

what am I missing? is Jenkins only running on main? I thought it was running on develop too...

ThomasRoosli:
For me all 15 tests in test_black_marble.py passed and returned ok. Does this relate to test files and the new CONFIG? maybe the solution in this issue helps you as well: #132 
(as a sidenote, on jenkins I only see one blackmarble test failing and it is climada.entity.exposures.test.test_black_marble.TestNightLight.test_cut_country_brb_2km_pass, seen on: https://ied-wcr-jenkins.ethz.ch/job/climada_branches/job/develop/350/)

aleeciu:
Hi thomas, I am referring to the integration _test_blackmarble.py_, not to the unit test _test_black_marble.py_; for which all 15 tests run fine for me too. Could you maybe try that too? (it takes a while, though, unless you already have downloaded the data)

Or maybe just try one of tests in the file, but a different one than Test2013.test_sint_maarten, which is the only which should actually crash according to https://ied-wcr-jenkins.ethz.ch/job/climada_ci_night/lastCompletedBuild/testReport/

emanuel-schmid:
The jenkins project climada_ci_night has the same environment as climada_branches. It's configured in the requirements/env_climada.yml file of the develop branch.

I have run blackmarble integration tests locally on windows with basically the same outcome as jenkins.

Can you post a typical error message from the 15 failed tests?

ThomasRoosli:
I have been running two tests of the `climada.test.test_blackmarble.py` file on windows. I can report the same problem as described by @aleeciu, all tests failed. The Error was always: `FileNotFoundError: [Errno 2] No such file or directory: 'C:\\shortpaths\\GitHub\\climada_python\\climada\\test\\results\\F182013.v4.tar'`. It occured somewhere here: `File "C:\shortpaths\GitHub\climada_python\climada\entity\exposures\nightlight.py", line 382, in load_nightlight_noaa`.
To me it seems that `SYSTEMS_DIR` in `climada.entity.exposures.nightlight.py` is somehow wrongly assigned. I hope this helps for finding the error. I will not have time to further look into that in the following days.

aleeciu:
[...]

Issue #98, opened 2020-11-26
Black Marble crashing with NASA data


aleeciu:
I am working on the BlackMarble tutorial, and a call as simple as:

`bm = BlackMarble()
bm.set_countries(['Switzerland'])`

gives the following error:
<img width="567" alt="Capture" src="https://user-images.githubusercontent.com/29018864/100377464-99edae00-3011-11eb-9a66-3546d628617e.PNG">

however, doing e.g.:
`bm = BlackMarble()
bm.set_countries(['Switzerland'], ref_year=2000)`

works, so I am pretty sure this has to do with the way the _Image_ package handles the NASA nighlights dataset. Does anybody encouter the same?

chahank:
Does it work for `ref_year=2016`?

aleeciu:
it crashes with any year that requries NASA data

chahank:
Sorry to ask, but didn't you say that `bm.set_countries(['Switzerland'], ref_year=2000]`  works?

aleeciu:
BlackMarble uses NOAA data untill 2013 and NASA data after 2013... it works with NOAA data

chahank:
Thanks for the clarification!

PR #146, opened 2021-02-15
Feature/fix bm

aleeciu:
A fix to some bugs in black marble:
- crash when loading NASA data as reported in #98. This was likely due to bugs with the pillow package reported here https://github.com/python-pillow/Pillow/issues/4237#issuecomment-570757948. I fixed that simply using the already existing function in LitPop, which uses gdal
- tests failure as reported in #143, due to paths mis-handling after the new config settings
- test failure due to rounding error when comapring expected and actual result

I also updated the tutorial.


climada/test/test_blackmarble.py
```diff
@@ -74,7 +74,7 @@ def test_sint_maarten_pass(self):
                       cm.output[0])
         self.assertIn("Processing country Sint Maarten.", cm.output[1])
         self.assertIn("Generating resolution of approx 0.2 km.", cm.output[2])
-        self.assertAlmostEqual(ent.gdf.value.sum(), 3.658e+08 * (4 + 1))
+        self.assertTrue(np.isclose(ent.gdf.value.sum(), 1.023e+09 * (4 + 1), 1))
```
tovogt, line 77: Are you sure that this is what you want? You set `rtol=1` in the `numpy.isclose` function. That means that the first value may deviate from the second by 100%. For example, `np.isclose(0, 1000, 1)` is True. Same for `np.isclose(2000, 1000, 1)`.

tovogt (changes requested):
Please look at my comments in the code.

climada/entity/exposures/nightlight.py
```diff
@@ -363,7 +364,7 @@ def load_nightlight_noaa(ref_year=2013, sat_name=None):
             for pre_i in np.arange(ini_pre, end_pre, -1):
                 url = NOAA_SITE + 'F' + str(pre_i) + str(ref_year) + '.v4.tar'
                 try:
-                    file_down = download_file(url)
+                    file_down = download_file(url, download_dir=SYSTEM_DIR)
```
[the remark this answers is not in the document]
aleeciu: That's what I thought too.. but it was not the case for me when I ran it, it was saving the downloaded files in my current directory

climada/test/test_blackmarble.py
aleeciu: Thanks, changed it ad also above which was there from before. Also changed the rest.

climada/entity/exposures/nightlight.py
```diff
@@ -223,12 +224,11 @@ def download_nl_files(req_files=np.ones(len(BM_FILENAMES),),
             'connection and whether filenames are still valid.') from exc
     return dwnl_path
 
-def load_nightlight_nasa(bounds, req_files, year):
```
aleeciu, line 226: This I noticed after running static code. This kwargs was not being used, even before the changes to this function.

tovogt: If you remove positional arguments, you break backwards compatibility for users. I'm not 100% familiar with this module, but I noticed that this change will even break code within CLIMADA:
https://github.com/CLIMADA-project/climada_python/blob/40a6b86bde05bcdbeddb908525deee07e9516f9e/climada/entity/exposures/black_marble.py#L296-L297

climada/test/test_blackmarble.py
```diff
@@ -199,7 +198,7 @@ def test_set_country_pass(self):
 # Execute Tests
 if __name__ == "__main__":
     TESTS = unittest.TestLoader().loadTestsFromTestCase(Test2013)
-    TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(Test1968))
-    TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(Test2012))
-    TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(BMFuncs))
+    # TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(Test1968))
+    # TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(Test2012))
+    # TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(BMFuncs))
```
tovogt, line 203: Why did you comment this out?

aleeciu: uncommented now, was trying failing test and did no want to run them all.. forgot about that

tovogt: You can run specific tests using a command line like this: `python -m unittest climada.test.test_blackmarble.BMFuncs`

climada/entity/exposures/nightlight.py
aleeciu: I see, but then it means we should drop that keyword from this function too... that kwards is not used within the function...

aleeciu: I can also leave it if it requires too many changes, but should be checked in the future I guess

tovogt: I can't judge because I'm not familiar with this module. If you're not sure whether code might break due to this change, you shouldn't change it or ask somebody who is familiar with it (e.g. the original authors). If you aren't able to solve this within this PR, you could leave a note `TODO: argument req_files is not used in this function`.

aleeciu: I brought it back

climada/entity/exposures/nightlight.py
```diff
@@ -228,7 +229,7 @@ def load_nightlight_nasa(bounds, req_files, year):
 
     Parameters:
         bounds (tuple): min_lon, min_lat, max_lon, max_lat
-        req_files (np.array): array with flags for NASA files needed
+        req_files (np.array): array with flags for NASA files needed
```
tovogt, line 232: remove trailing whitespace

aleeciu: done and added TODO

climada/entity/exposures/nightlight.py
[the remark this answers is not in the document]
tovogt: Maybe @emanuel-schmid can comment on this?

I'm not sure what exactly is happening here, but the rest of this PR is fine from my side.

tovogt (approved):
Thanks for the changes, this PR is fine from what I can judge.

aleeciu:
thanks @tovogt for the feedback

climada/entity/exposures/nightlight.py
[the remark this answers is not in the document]
Evelyn-M: yep, same from my side, all good, but would be nice if Emanuel could explain / help out on this one.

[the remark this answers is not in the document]
emanuel-schmid: The parameter `download_dir` of the `download_file` functions was introduced with the changes of the config settings.
(Before the changes, it used to download files into the working directory. Which was sometimes countered by undesired `os.chdir()` calls.)
The function's default directory is `CONFIG.local_data.save_dir`. The configuration's default value for this is `./results`.

`SYSTEM_DIR` is a constant, but it's not exactly hard coded. Its value is defined as `CONFIG.local_data.system`.

I hope these explanations help clarifying the behavior?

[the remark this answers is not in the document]
tovogt: > SYSTEM_DIR is a constant, but it's not exactly hard coded. Its value is defined as CONFIG.local_data.system.

Thanks! I think that is the answer to @Evelyn-M's original question and solves this confusion.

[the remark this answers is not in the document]
aleeciu: thanks Emanuel, yes, it was creating a ./results folder in the working directory and saving files therein. Problem is, the loading function called after was expecting the files to be in SYSTEM_DIR, so that was the problem.

So maybe that also explaines why the same did not crash on Jenkins? (see #143) Can it be that there it is CONFIG.local_data.save_dir == CONFIG.local_data.system ? (my guess)

PR #147, opened 2021-02-15
Feature/refactor crs assignment

emanuel-schmid:
This pull request addresses the updates that are necessary to deal with the changes from geopandas 3.0 to 3.1.
In particular, the crs attribute of GeoDataFrame and GeoSeries has the pyproj.crs.CRS type now.

Notably, the DEF_CRS object initiated in climada.constants is not a dict anymore but a string.
The dictionary that used to be obtained by calling fiona.crs.from_epsg was replaced by the mere 'EPSG:4326' string. The advantage is that both rasterio and pyproj corrrectly interprete this string as epsg 4326, while the dict from fiona was understood by rasterio alone.

After this pull request, both geopandas versions, 3.0 and 3.1, are compatible with climada.


tovogt (approved):
Looks fine to me.

Issue #148, opened 2021-02-16
 climada.util.coordinates.lon_normalize crashes tc_tracks.calc_perturbed_trajectories


chahank:
The changes to ` climada.util.coordinates.lon_normalize` and `tc_tracks.calc_perturbed_trajectories` are unstable. The following code now fails:

```
from climada.hazard import TCTracks
basin ='NA'
tc_tracks = TCTracks()
tc_tracks.read_ibtracs_netcdf(basin=basin, year_range=(1950, 2020)) 
tc_tracks.equal_timestep(time_step_h=1)
#Need to remove tracks with a single time step for calc_perturbed_trajectories
filter_only_one_time = [track for track in tc_tracks.data if track.time.size >= 2]
tc_tracks2= TCTracks()
tc_tracks2.data = filter_only_one_time
tc_tracks2.calc_perturbed_trajectories(nb_synth_tracks=10)
```

Error msg from `tc_tracks2.calc_perturbed_trajectories(nb_synth_tracks=10)`: climada.util.coordinates - WARNING - lon_normalize: killed before finishing

The `tc_tracks2` object is then empty.

Note that his bug was not present on the 25.01.2020.

bguillod:
First of all, does the following work?
```
tc_tracks.calc_perturbed_trajectories(nb_synth_tracks=10)
```
(it probably shouldn't).

My guess is that the call to `lon_normalize` comes from the following lines in `_one_rnd_walk`:
```
        angular_dist = climada.util.coordinates.dist_approx(i_track.lat.values[:-1, None],
                                                            i_track.lon.values[:-1, None],
                                                            i_track.lat.values[1:, None],
                                                            i_track.lon.values[1:, None],
                                                            method="geosphere",
                                                            units="degree")[:, 0, 0]
```
because `lon_normalize` is called in `dist_approx`.

Could you identify what changed in dist_approx and in lon_normalize since 25.01.2020 (I hope you actually mean 25.01.2021)? That should give you part of the answer, hopefully.

chahank:
`tc_tracks.calc_perturbed_trajectories(nb_synth_tracks=10)` does not work because some of the tracks have only one time step and then the method complains.

tovogt:
Since calc_perturbed_trajectories iterates over all tc tracks, this bug should be independent of the number of tracks in your `TCTracks` object. Would you please try to restrict your example to a single track that causes the error message?

chahank:
Ok, I will try to see.

[...]

Issue #151, opened 2021-02-17
Landfall decay issue in TC synthetic tracks


bguillod:
I have found an issue with the decay of wind speed and pressure at landfall in the `climada.hazard.tc_track_synth` module. I here describe the issue as well as a part of the solution, with some open questions remaining. It consists of two interrelated problems and I take the time to go through in detail below, because I consider this a critical bug that requires a robust solution - thanks in advance for going through the full description!

I would be glad if **CLIMADA's Tropical Cyclone experts can have a look and give their opinion** if they have any.

I am preparing a pull request to solve these issues, but would be glad to get feedback **ASAP** so I can include that and make some critical choices.

### General description
Here is a reproducible example:
```
from climada.hazard.tc_tracks import TCTracks

sid = '1997018S11059'
tracks = TCTracks()
tracks.read_ibtracs_netcdf(provider='usa', storm_id=sid, estimate_missing=True)

tracks.equal_timestep(3)
tracks.calc_perturbed_trajectories(nb_synth_tracks=10)
tracks.plot()
```
![image](https://user-images.githubusercontent.com/13904718/108193165-740a0e00-7115-11eb-9635-e7ab95ee7fa9.png)

Let me take the 3rd synthetic as it illustrates both issues: It makes a first landfall over Madagascar, then a 2nd landfall over Mozambique, and then goes back over the ocean, where it ends:
```
track_1 = TCTracks()
track_1.data = [tracks.get_track('1997018S11059_gen3')]
track_1.plot()
```
![image](https://user-images.githubusercontent.com/13904718/108193283-91d77300-7115-11eb-8c47-db43daef4352.png)

Let me first display a time series of the track, using the following functions:
```
import numpy as np
import pandas as pd
from matplotlib.dates import date2num
from climada.hazard.tc_tracks import TCTracks
import matplotlib.pyplot as plt

def get_lf_df(tr_df):
    lf_start = tr_df.iloc[np.where(np.diff(tr_df['on_land'].astype(float)) == 1)[0],:]['time']
    lf_end = tr_df.iloc[np.where(np.diff(tr_df['on_land'].astype(float)) == -1)[0],:]['time']
    if len(lf_end) < len(lf_start):
        lf_end = lf_end.append(max(tr_df['time']))
    lf_dat = pd.DataFrame({'start_lf':lf_start, 'end_lf': lf_end}).reset_index(drop=True)
    return lf_dat


[...]

PR #152, opened 2021-02-17
Feature/fix tc synth decay

bguillod:
This PR aims at solving the issues raised in issue #151:

- It applies a post-landfall correction even if the track ends over the ocean
- It applies pressure decay even if `central_pressure` at landfall is higher than `environmental_pressure` after the landfall.

In addition I added a few useful informations for the user:
- If it fails to apply wind or pressure decay, it issues a warning.
- It it uses only a few tracks (less than 10), it warns the user that the landfall decay it based only on those few historical tracks and recommends to use more tracks and then sub-select the desired tracks.
- If the calibrated landfall parameters would not lead to a decay but to an intensification of the Tropical Cyclone, it issues an error.


climada/hazard/tc_tracks_synth.py
```diff
@@ -452,6 +452,14 @@ def _calc_land_decay(hist_tracks, land_geom, s_rel=True, check_plot=False,
...
         v_rel (dict(category: A)), p_rel (dict(category: (S, B)))
     """
+    
+    if len(hist_tracks) < 10:
+        LOGGER.warning('Using only %s historical tracks to calculate '
+                       'landfall decay. Consider using a larger number of '
+                       'tracks from which you sub-select the tracks that '
+                       'are relevant to your application for a more robust '
+                       'calculation of landfall decay.', len(hist_tracks))
+
```
bguillod, line 462: Users might not realize that unlike for e.g. missing values estimations in read_ibtracs_netcdf, the landfall parameters are derived solely based on the historical tracks in the TCTracks object. Hence if less than 10 tracks are there, is warns the user. Is 10 a good threshold? Or what should be the criterion to warn the user?

climada/hazard/tc_tracks_synth.py
```diff
@@ -654,6 +662,20 @@ def _decay_calc_coeff(x_val, v_lf, p_lf):
...
+                         'decay values for wind speed.',
+                         climada.hazard.tc_tracks.CAT_NAMES[ss_scale - 2])
+            raise ValueError
+        elif p_rel[ss_scale][0] < 0 or p_rel[ss_scale][1] < 0:
+            LOGGER.error('Landfall pressure decay parameter is negative '
+                         'for category %s. Please check or use a larger '
+                         'number of events to obtain more sensible landfall '
+                         'decay values for pressure.',
+                         climada.hazard.tc_tracks.CAT_NAMES[ss_scale - 2])
+            raise ValueError
```
bguillod, line 678: This is to check that landfall parameters imply a decay of TC intensity. If they do not (i.e. they would enhance the TC strength), it issues an error.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
             continue
         if land_sea - sea_land == 1:
             continue
-        p_decay = _calc_decay_ps_value(track, p_landfall, land_sea - 1, s_rel)
-        p_decay = _decay_p_function(p_decay, p_rel[ss_scale_idx][1],
-                                    track.dist_since_lf[sea_land:land_sea].values)
-        # dont applay decay if it would decrease central pressure
-        p_decay[p_decay < 1] = track.central_pressure[sea_land:land_sea][p_decay < 1] / p_landfall
-        track.central_pressure[sea_land:land_sea] = p_landfall * p_decay
+        S = _calc_decay_ps_value(track, p_landfall, land_sea - 1, s_rel)
```
bguillod, line 751: I renamed to `S` because it is otherwise confusing.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
+        if S <= 1:
+            # central_pressure at start of landfall > env_pres after landfall:
+            # set central_pressure to environmental pressure during whole lf
+            track.central_pressure[sea_land:land_sea] = track.environmental_pressure[sea_land:land_sea]
```
bguillod, line 755: If S <= 1, a decay cannot be applied because it would decrease central pressure. Hence in that case central pressure is simply set to environmental pressure.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
+        else:
+            p_decay = _decay_p_function(S, p_rel[ss_scale_idx][1],
+                                        track.dist_since_lf[sea_land:land_sea].values)
+            # dont applay decay if it would decrease central pressure
+            if np.any(p_decay > 1):
+                LOGGER.warning('Landfall decay would decrease pressure for '
```
bguillod, line 761: If despite the check on `S` the decay would increase pressure drop (decrease pressure), it is not applied (as was before) but the user is also warned that this is the case.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
-        v_decay[v_decay > 1] = (track.max_sustained_wind[sea_land:land_sea][v_decay > 1]
-                                / v_landfall)
+        # dont apply decay if it would increase wind speeds
+        if np.any(v_decay > 1):
+            # should not happen unless v_rel is negative
+            LOGGER.warning('Landfall decay would increase wind speed for '
+                           'track id %s. Decay not applied, please check.',
+                           track.sid)
+            v_decay[v_decay > 1] = (track.max_sustained_wind[sea_land:land_sea][v_decay > 1]
+                                    / v_landfall)
```
bguillod, line 776: As for pressure, if wind speed decay would lead to an increase in wind speed it is not applied (as before), but now the user is also warned that this is the case.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
         track.max_sustained_wind[sea_land:land_sea] = v_landfall * v_decay
 
-        # correct values of sea between two landfalls
-        if land_sea < track.time.size and idx + 1 < sea_land_idx.size:
+        # correct values of sea after a landfall (until next landfall, if any)
+        if land_sea < track.time.size:
+            if idx + 1 < sea_land_idx.size:
```
bguillod, line 781: This new condition doesn't exclude cases where no further landfall occurs but the track ends over the ocean.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
+                # if there is a next landfall, correct until last point before
+                # reaching land again
+                end_cor = sea_land_idx[idx + 1]
+            else:
+                # if there is no further landfall, correct until the end of
+                # the track
+                end_cor = track.time.size
```
bguillod, line 788: The correction will be either until the next landfall, or if there is no further landfall, until the end of the track.

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
             r_diff = track.max_sustained_wind[land_sea].values - \
                      track.max_sustained_wind[land_sea - 1].values - rndn
-            track.max_sustained_wind[land_sea:sea_land_idx[idx + 1]] += - r_diff
+            track.max_sustained_wind[land_sea:end_cor] += - r_diff
+
+        # correct limits
+        np.warnings.filterwarnings('ignore')
+        cor_p = track.central_pressure.values > track.environmental_pressure.values
+        track.central_pressure[cor_p] = track.environmental_pressure[cor_p]
+        track.max_sustained_wind[track.max_sustained_wind < 0] = 0
```
bguillod, line 803: This code block was moved within the loop through landfalls, so that a non-first landfall already has sensible values to deal with. It helps solving the issue of no pressure correction after landfall.

climada/hazard/tc_tracks_synth.py
chahank: Well, the question should be investigated right? ... what is the error made from having 10 tracks? Does it depend on the geographical area?

climada/hazard/tc_tracks_synth.py
chahank: I think this is not really a good solution. Currently, the tc_track_synth easily throws errors and the user does not necessarily know what to do with it. Hence, it is better to write the code with extra assumptions...  But maybe first, could you please explain why a larger number of tracks would help?

climada/hazard/tc_tracks_synth.py
chahank: Could you use a more descriptive variable? Capital S is too short and non-descriptive.

climada/hazard/tc_tracks_synth.py
chahank: This is warning is maybe good, but the message is not really helpful. 'Decay not applied, please check' does not help a user to solve the problem. One needs to be more explicit here if a warning is made.

climada/hazard/tc_tracks_synth.py
chahank: Why can central pressure not decrease?

climada/hazard/tc_tracks_synth.py
```diff
@@ -726,37 +748,60 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
-        track.central_pressure[sea_land:land_sea] = p_landfall * p_decay
+        S = _calc_decay_ps_value(track, p_landfall, land_sea - 1, s_rel)
+        if S <= 1:
+            # central_pressure at start of landfall > env_pres after landfall:
+            # set central_pressure to environmental pressure during whole lf
+            track.central_pressure[sea_land:land_sea] = track.environmental_pressure[sea_land:land_sea]
+        else:
+            p_decay = _decay_p_function(S, p_rel[ss_scale_idx][1],
+                                        track.dist_since_lf[sea_land:land_sea].values)
+            # dont applay decay if it would decrease central pressure
```
chahank, line 759:
```suggestion
            # dont apply decay if it would decrease central pressure
```

climada/hazard/tc_tracks_synth.py
chahank: Should this be a warning or an error? What does 'Decay not applied, please check' mean? How should a user know what to do in this situation?

climada/hazard/tc_tracks_synth.py
chahank: Where is the correction for the second landfall (if there is one) applied?

chahank:
Thank you for the update Benoit! I posted several understanding questions, as well as some comments on the warning messages. The discussion on whether such changes are useful in CLIMADA will be done in #151 .

What is however clearly missing are tests for the updated functioning of the methods.

climada/hazard/tc_tracks_synth.py
bguillod: Adding this warning is actually not related to the issues this PR fixes, but I added it because I think it is useful to warn users here. Here is my reasoning: There are several cases in TCTracks where a fit is applied to the data. For instance, in `TCTracks.read_ibtracs_netcdf` with `estimate_missing=True`, a fit is applied to fill in missing values. However, in that case the fit has been calculated beforehand using the full IBTrACS data from 1980-2020 and the resulting parameter values are provided in a static way, e.g. in `estimate_pressure`:
https://github.com/CLIMADA-project/climada_python/blob/038cfc40ebb83c6f114557ed7576bbba022e9860/climada/hazard/tc_tracks.py#L1464
This ensures that the same "model" is used whichever input are passed to the function.

In the case of landfall decay for synthetic tracks, there is no pre-computed fitted parameters. Instead, the parameters are fitted from all the historical tracks that are passed to `calc_perturbed_trajectories`. Therefore, the parameters used to characterize landfall decay depend on the content of your TCTracks object when calling the function.

Personally, here is the approach I would propose:
1. Calculating landfall decay parameters using all tracks from 1980-2020 (for consistency with read_ibtracs_netcdf). However I would do that by basin.
2. Hard-code these values as a dictionary, and use these to apply landfall decay.

That being said, this is a separate topic that is not part of this PR, and investing time in this is not worthwhile in case the synthetic tracks model would be replaced by something more physically-based. Perhaps we could create a separate issue on this. In the meantime I propose this warning, although I'd suggest to strongly increase the threshold (there is one set of parameter per Saffir-Simpson category, which can only be calculated for tracks that make landfall at that category, so perhaps we could increase the threshold to 100 or something like this.

climada/hazard/tc_tracks_synth.py
bguillod: It is extremely unlikely that these parameters are negative. If they are, it can only be because you used a very low number of historical tracks as input and these by very very very bad luck increase intensity at landfall. The suggestion for improvements I suggested above (but which are in my view out of scope for this PR) would also avoid this issue. Hence in the meantime the proposition to raise an error here suggesting to use more historical tracks - should I change this to a warning?

bguillod: I agree however that the tc_track_synth throws too many error, which could be improved in a future PR.

climada/hazard/tc_tracks_synth.py
bguillod: I used S because it is how it is call in the model, see e.g.
https://github.com/CLIMADA-project/climada_python/blob/41a110a0255b5a4fdf786973dd24bcbfbea82988/climada/hazard/tc_tracks_synth.py#L442
Should I rename it to e.g. `S_param`, to make it clear it's a parameter?

climada/hazard/tc_tracks_synth.py
bguillod: When a TC makes landfall, its strength should decay, not increase, hence it would be unrealistic to increase pressure drop (=decrease pressure). However S<=1 can only happen if central pressure before landfall is already at the level of an environmental pressure, therefore it should simply stay to that level.

climada/hazard/tc_tracks_synth.py
bguillod: This actually cannot happen, because we checked that `S > 1` and `p_rel > 0`, but I still put that here in case future code changes so `p_rel` or `S` calculations would make it again possible for `p_decay` to be `>1`.

climada/hazard/tc_tracks_synth.py
bguillod: I considered specifying the track id is sufficient information. What information you find is missing in this warning?

climada/hazard/tc_tracks_synth.py
bguillod: This code block is within a for loop that loops through each landfall for that track:
https://github.com/CLIMADA-project/climada_python/blob/41a110a0255b5a4fdf786973dd24bcbfbea82988/climada/hazard/tc_tracks_synth.py#L718
hence `sea_land_idx[idx+1]` is the sea-to-land transition point of the next landfall.
The next loop iteration will then apply a new decay for the next landfall (and after that next landfall).

climada/hazard/tc_tracks_synth.py
chahank: OK, I see. I think your solution is valid and is also desired for the current implementation. 

Note that there will most probably never be a physics-based random sampling module. Doing such models is very expensive both in scientific knowledge, work time and computation time. This is also not our expertise nor is it the goal of CLIMADA. There are several research groups and commercial agents working on precisely this. What should be reflected upon is whether a better statistical model can be made. However, also this is actually a whole research subject on its own. The best is to have data APIs. It is not the aim, nor are there the capacities to make a full-fledged tropical cyclone statistical model.

Hence, doing the calibration is the best way. I agree that the per basin is also a good idea.

Concerning the threshold, why 100? Do you have a basis for this number (or an educated guess)? Why not 1000?

climada/hazard/tc_tracks_synth.py
```diff
@@ -452,6 +452,14 @@ def _calc_land_decay(hist_tracks, land_geom, s_rel=True, check_plot=False,
     Returns:
         v_rel (dict(category: A)), p_rel (dict(category: (S, B)))
     """
+    
+    if len(hist_tracks) < 10:
+        LOGGER.warning('Using only %s historical tracks to calculate '
+                       'landfall decay. Consider using a larger number of '
+                       'tracks from which you sub-select the tracks that '
+                       'are relevant to your application for a more robust '
+                       'calculation of landfall decay.', len(hist_tracks))
```
chahank, line 461:
```suggestion
        LOGGER.warning('For the calibration of the landfall decay '
                       ' it is recommended to provide at least 10 tracks, 
                       'but only %d historical tracks were provided. '
                       'For more robust calculation consider using '
                       ' a larger amount of tracks', len(hist_tracks))
```

climada/hazard/tc_tracks_synth.py
chahank: I reformulated your error message to be clearer for the user. Please note that for errors one should not use the Logger.

chahank: Question: Can you please provide a citation for the 'intensity at landfall cannot increase' and include it in the error message?

chahank, line 678:
```suggestion
            raise ValueError(The landfall pressure calibration '
                         'resulted in a pressure increase for category %s.'
                         'This behaviour is unphysical.'
                         'Please use a larger number of tracks',
                          climada.hazard.tc_tracks.CAT_NAMES[ss_scale - 2])
```

climada/hazard/tc_tracks_synth.py
```diff
@@ -654,6 +662,20 @@ def _decay_calc_coeff(x_val, v_lf, p_lf):
                          climada.hazard.tc_tracks.CAT_NAMES[close_scale - 2])
             v_rel[ss_scale] = v_rel[close_scale]
             p_rel[ss_scale] = p_rel[close_scale]
+        elif v_rel[ss_scale] < 0:
+            LOGGER.error('Landfall wind speed decay parameter is negative '
+                         'for category %s. Please check or use a larger '
+                         'number of events to obtain more sensible landfall '
+                         'decay values for wind speed.',
+                         climada.hazard.tc_tracks.CAT_NAMES[ss_scale - 2])
+            raise ValueError
```
chahank, line 671:
```suggestion
            raise ValueError('The landfall wind speed calibration '
                         'resulted in a windspeed increase for category %s.'
                         'This behaviour is unphysical.'
                         'Please use a larger number of tracks',
                         climada.hazard.tc_tracks.CAT_NAMES[ss_scale - 2])
```

climada/hazard/tc_tracks_synth.py
chahank: Idem for the 'landfall pressure cannot increase at landfall'?

climada/hazard/tc_tracks_synth.py
chahank: I see, that is unfortunate. Then it is better to leave it as is.

climada/hazard/tc_tracks_synth.py
chahank: Could you please provide a citation for the landfall strength decrease? Not that this is controversial, but we should differentiate in the code when assumptions are backed by external sources.

climada/hazard/tc_tracks_synth.py
chahank: Not sure if I like this, but then one should probably raise an error right? Or is it fine if the decay is not applied? Also, please update the message to something more useful. 'Please check' does not give me any actual information about the issue.

climada/hazard/tc_tracks_synth.py
chahank: I do not know why does landfafl decay increase? Why is the decay not applied? what should the user check?

chahank:
Thanks for the update, now that I understood the changes better I think they are useful!

What is missing are tests. Please add/modify tests for the introduced changes.

bguillod:
@chahank I've adjusted some things based on your comments. I'll now try to fix #148 as well, and add some tests, but feel free to have a look at the new changes on the landfall decay part.

bguillod:
Commits 8b342f80bceaae952aa69bfac5adbb4ad784207d and a2ccefaa574c608ffcd2a5fa1e9d137429a48628 provide a quick fix to issue #148.

climada/hazard/tc_tracks_synth.py
```diff
@@ -99,8 +143,25 @@ def calc_perturbed_trajectories(tracks,
...
+        Default: True.
     """
+    
+    # Remove tracks with a single point
+    single_point_ids = [track.attrs['sid'] for track in tracks.data if track.time.size < 2]
+    if len(single_point_ids) > 0:
+        LOGGER.info('%s storm events are discarded because they consist of a '
+                    'single point: %s.', len(single_point_ids),
+                    ", ".join(single_point_ids))
+        tracks.data = [track for track in tracks.data if track.time.size >= 2]
```
chahank, line 162: Discarding data without reason is not a good idea. Hence, I propose to simply ignore the single point events for the synthetic track generation. This is done here by saving the single point tracks in line 157, and appending them again to the full set in line 233.

```suggestion
    # Do no use tracks with a single time point for synthetic track generation
    single_point_tracks = [track for track in tracks.data if track.time.size <= 1]
    if single_point_tracks:
        LOGGER.info('%s storm events consist of a single time point. No synthetic'
                    'track is generated for these events: %s.', len(single_point_tracks),
                    ", ".join([track.attrs['sid'] for track in single_points_tracks]))
        tracks.data = [track for track in tracks.data if track.time.size >= 2]
```

climada/hazard/tc_tracks_synth.py
```diff
@@ -146,22 +207,28 @@ def calc_perturbed_trajectories(tracks,
...
+                try:
+                    v_rel, p_rel = _calc_land_decay(hist_tracks, land_geom,
+                                                    pool=tracks.pool)
+                    tracks.data = _apply_land_decay(tracks.data, v_rel, p_rel,
+                                                    land_geom, pool=tracks.pool)
+                except ValueError:
+                    LOGGER.info('No land decay coefficients could be applied.')
+            else:
+                LOGGER.error('No historical tracks contained. '
+                             'Historical tracks are needed for land decay.')
```
chahank, line 233:
```suggestion
        tracks.data.append(single_point_tracks)
```

climada/hazard/tc_tracks_synth.py
```diff
@@ -146,22 +207,28 @@ def calc_perturbed_trajectories(tracks,
...
```
chahank, line 231:
```suggestion
                LOGGER.error('No historical tracks found. Historical'
                             'tracks are needed for land decay calibration'
                             'if use_global_decay_params=True.')
```

climada/hazard/tc_tracks_synth.py
```diff
@@ -146,22 +207,28 @@ def calc_perturbed_trajectories(tracks,
...
+            # fit land decay coefficients based on historical tracks
+            hist_tracks = [track for track in tracks.data if track.orig_event_flag]
+            if hist_tracks:
+                try:
+                    v_rel, p_rel = _calc_land_decay(hist_tracks, land_geom,
+                                                    pool=tracks.pool)
+                    tracks.data = _apply_land_decay(tracks.data, v_rel, p_rel,
+                                                    land_geom, pool=tracks.pool)
+                except ValueError:
+                    LOGGER.info('No land decay coefficients could be applied.')
```
chahank, line 228: If you catch an error, you probably should also handle it... or at least the logging message must be very clear. Why could no land decay coefficients be applied? What does it mean?

climada/hazard/tc_tracks_synth.py
```diff
@@ -34,6 +34,49 @@
...
+    2: 0.0017219917301371703,
+    6: 0.002595834561794121,
+    1: 0.00012285815226960615,
+    5: 0.002644866334692554,
+    3: 0.002299182035879582,
+    4: 0.0025887346457309896,
+    7: 0.00397813058576188
+}
+"""Global landfall decay parameters for wind speed by TC category.
+Keys are TC categories with 1='TD', 2='TS', 3='Cat 1', ..., 7='Cat 5'.
```
chahank, line 47: Why not use the CLIMADA return values of -1, 0, ... , 5 ?

climada/hazard/tc_tracks_synth.py
```diff
@@ -34,6 +34,49 @@
...
+    2: (1.0194994980230085, 0.0030702654707023256),
+    6: (1.0825529416232729, 0.004157255859762298),
+    1: (1.0089315640011636, 0.0021694084866327895),
+    5: (1.0652603933624192, 0.0037557812788230143),
+    3: (1.0365767270424973, 0.0035347637567456768),
+    4: (1.0467348458061756, 0.004042906726642808),
+    7: (1.1092059075832368, 0.005485547796536179)
+}
+"""Global landfall decay parameters for pressure by TC category.
+Keys are TC categories with 1='TD', 2='TS', 3='Cat 1', ..., 7='Cat 5'.
```
chahank, line 69: Same as above: why not -1,...,5?

chahank (changes requested):
The latest updates are really useful, thank you @bguillod !

Before we can merge, some tests for the added features must be written. I think there needs some for at least
- discard_single_points = True/False
- decay = True/False
- use_global_decay_params = True/False
- ignore single point tracks
- original changes for multiple landfalls

bguillod:
Thanks @chahank I'll try to implement the tests.
In the meantime note that I now allow single point tracks to be perturbed (it's just the initial perturbation that is applied then). Do you want me to revert the changes to read_ibtracs_netcdf or should I keep that option of skipping single-point tracks?
I've also reformatted the code to use TC category as keys to dictionary consistently with `climada.hazard.tc_tracks.CAT_NAMES`, and use `climada.hazard.tc_tracks.set_category` instead of a manual calculation in the code.

climada/hazard/tc_tracks_synth.py
```diff
@@ -227,8 +234,9 @@ def calc_perturbed_trajectories(tracks,
                 except ValueError:
                     LOGGER.info('No land decay coefficients could be applied.')
             else:
-                LOGGER.error('No historical tracks contained. '
-                             'Historical tracks are needed for land decay.')
+                LOGGER.error('No historical tracks found. Historical'
+                             'tracks are needed for land decay calibration'
+                             'if use_global_decay_params=True.')
```
chahank, line 239: Please, do not use the LOGGER for errors. Raise a proper error with e.g. 

`raise ValueError('')`. 
```suggestion
               raise ValueError('No historical tracks found. Historical'
                             'tracks are needed for land decay calibration'
                             'if use_global_decay_params=True.')
```

climada/hazard/tc_tracks_synth.py
```diff
@@ -265,6 +273,8 @@ def _one_rnd_walk(track, nb_synth_tracks, max_shift_ini, max_dspeed_rel, max_ddi
     [dt] = np.unique(track['time_step'])
 
     ens_track.append(track)
+    cutoff_track_ids_ok = []
+    cutoff_track_ids_cat = []
```
chahank, line 277: What is the meaning of '_ok'  and '_cat'? These are quite confusing names.

climada/hazard/tc_tracks_synth.py
```diff
@@ -204,6 +196,21 @@ def calc_perturbed_trajectories(tracks,
...
+        LOGGER.warning('The following generated synthetic tracks moved beyond '
+                        'the range of [-70, 70] degrees latitude. Cut out '
+                        'at TC category >1: %s.',
+                        ', '.join(cutoff_track_ids_cat))
+    if len(cutoff_track_ids_ok) > 0:
+        LOGGER.info('The following generated synthetic tracks moved beyond '
+                    'the range of [-70, 70] degrees latitude. Cut out '
+                    'at TC category <= 1: %s.',
+                    ', '.join(cutoff_track_ids_ok))
+    new_ens = [x[0] for x in new_ens]
```
chahank, line 213: I am not really convinced that this is actually needed. In any case, when computing windfields, there is a hard cut at 60°. I think this is unecessary 'false' precision.

chahank: > Thanks @chahank I'll try to implement the tests.
> In the meantime note that I now allow single point tracks to be perturbed (it's just the initial perturbation that is applied then). Do you want me to revert the changes to read_ibtracs_netcdf or should I keep that option of skipping single-point tracks?
> I've also reformatted the code to use TC category as keys to dictionary consistently with `climada.hazard.tc_tracks.CAT_NAMES`, and use `climada.hazard.tc_tracks.set_category` instead of a manual calculation in the code.

We are getting there! I think it is fine to leave the changes to `read_ibtracs_netcdf`. The reformating for the TC categories is very helpful, thanks! 

Beyond the small points noted in the last changes, the only thing missing are the unit tests.

climada/hazard/tc_tracks_synth.py
```diff
@@ -99,15 +143,23 @@ def calc_perturbed_trajectories(tracks,
...
 
     if seed >= 0:
         np.random.seed(seed)
 
     # ensure tracks have constant time steps
-    time_step_h = np.unique([np.unique(x['time_step']) for x in tracks.data])
+    time_step_h = np.unique([np.unique(x['time_step']) for x in tracks.data if x.time.size > 1])
     if not np.allclose(time_step_h, time_step_h[0]):
         LOGGER.error('Tracks have different temporal resolution. '
                      'Please ensure constant time steps by applying equal_timestep beforehand')
```
chahank, line 165: ok, now I am bit confused about this line actually. Why would you use `np.allclose()`? What you expect is that `time_step_h` is a unique defined value (a float). As soon as it is a list, it means that there are two different timestep values. 

The following suggestion would replace line 162 - 167 (I was not able to include lines 166,167 in the the suggestion, so if you like it please remove these manually):
```suggestion
    try:
        [time_step_h] = np.unique([np.unique(x['time_step']) for x in tracks.data])
    except ValueError:
         raise ValueError('Not all tracks have the same temporal resolution. '
                     'Please ensure constant time steps by applying equal_timestep beforehand')
```

bguillod:
Thanks a lot @chahank for your patience and for the additional review! Highly appreciated!
Somehow I can't reply to your comments directly anymore so I answer your remaining points here:
- `cutoff_track_ids_ok` and `cutoff_track_ids_cat` are list of track IDs that were cut, the `_ok` is if they are cut with intensity of Category TD or TS, while the `_cat` is for TC categories (1-5). I could rename to `cutoff_track_ids_tc` and `cutoff_track_ids_ts`, for instance?
- I don't agree that letting the user know we have cut tracks is not needed, for advanced users it is useful information. I agree in terms of holland wind model only it's not relevant because of the 60 degrees cut there, however soon there will be a TC parametric insurance module that will use track data, not holland wind fields, so dealing with the tracks is relevant here for future purposes.
- I had to use `np.allclose(time_step_h, time_step_h[0])` because some small rounding approximations sometimes made it fail although values were essentially the same (with e.g. 0.00000001 difference). If you see a better solution I'm not against it but that was a pragmatic approach that does the job well. But your current proposition would fail in most cases.

chahank:
Concerning the added warnings, I still have the feeling that they are not particularly useful for a normal user. Several of these are probably better set as 'DEBUG' level instead of 'WARNING'. 

ALso to
> I don't agree that letting the user know we have cut tracks is not needed, for advanced users it is useful information
I did not suggest to not let the user know. What I suggested is to not issue a warning, but to explain it in the docstring. If there is an expert user that wants to have a deeper look, it is trivial to filter tracks to see if they end at category 1-5 or not.  

Currently, this method is generating to many warnings which are confusing to the normal user.

chahank:
For the 

> I had to use np.allclose(time_step_h, time_step_h[0]) because some small rounding approximations sometimes made it fail although values were essentially the same (with e.g. 0.00000001 difference). If you see a better solution I'm not against it but that was a pragmatic approach that does the job well. But your current proposition would fail in most cases.

I update my suggestion using a `np.round`. What do you think? Would this work well?

chahank:
> I could rename to cutoff_track_ids_tc and cutoff_track_ids_ts, for instance?
ts: tropical storm
tc: tropical cyclone?

Some are also td (tropical depression) right?

But yes, I think this is clearer than `_cat` `_ok`. Thanks!

bguillod:
@chahank I've added two commit to fix an issue I encountered with the identification of landfalls, which would ignore tracks that start over land. I've taken the opportunity to do that more consistently throughout the code and then I've adjusted the global landfall parameters using the new code. See 5d1f132 and d53596d, if you want to have a quick look that's be great.
I'll get back on your other points to complete this PR over the coming weeks. Thanks a lot!

bguillod (commented):
> Some are also td (tropical depression) right?

In fact the "_tc" ones would be TC cat 2-5, while the "_ts" ones would be TD, TS or TC cat 1...

climada/hazard/tc_tracks.py
```diff
@@ -488,6 +492,16 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
                            'have been found: %s%s', len(invalid_sids), ", ".join(invalid_sids[:5]),
                            ", ..." if len(invalid_sids) > 5  else ".")
             ibtracs_ds = ibtracs_ds.sel(storm=valid_storms_mask)
+            
+        if discard_single_points:
+            valid_storms_mask = ibtracs_ds.valid_t.sum(axis=1) > 1
```
tovogt, line 497: Thanks for consulting me about this change! Since single-point events are quite rare (~0.5% and mainly old tracks that never reached inhabited areas) when using `estimate_missing=True` and the default `provider`, I think it's safe to discard these events by default. Users who choose to take different parameters (e.g. different `provider`) will probably consider setting `discard_single_points=False` according to their needs.
 
Use dimension names for clarity:
```suggestion
            valid_storms_mask = ibtracs_ds.valid_t.sum(dim="date_time") > 1
```

climada/hazard/tc_tracks_synth.py
bguillod: That would likely work as well, yes. I'm not sure I understand why `np.allclose()` is not appropriate though (to me it sounds like exactly what's needed, namely checking that all values are very close to each other)? I'm fine with your suggestion, but would you mind just enlightening me on that one, so I can recall for next time? :-)

climada/hazard/tc_tracks_synth.py
bguillod: There could be several reasons why it failed. In a new commit I now raise an error, re-raising the error that was caught, too (with `from`) to help the user identifying the cause.

bguillod:
> Concerning the added warnings, I still have the feeling that they are not particularly useful for a normal user. Several of these are probably better set as 'DEBUG' level instead of 'WARNING'.
I've now changed logging levels for the cutoff tracks: debug for TC cat <= 1, and info for cat 2-5. Hence normal users won't see the ones that are fine and might be informed about the higher category ones.

climada/hazard/tc_tracks_synth.py
chahank: I agree that there is not much of a difference, but for me the suggestion is clearer in what this bit of code is doing. In the previous version (np.allclose), the purpose of the blob of code is harder to understand at one glance.

climada/hazard/tc_tracks_synth.py
chahank: A specific paper would be better as websites are not persistent.

climada/hazard/tc_tracks_synth.py
```diff
@@ -143,25 +195,47 @@ def calc_perturbed_trajectories(tracks,
                                  max_dspeed_rel, max_ddirection, rand)
                    for track, rand in zip(tracks.data, random_vec)]
 
+    cutoff_track_ids_tc = [x[1] for x in new_ens]
+    cutoff_track_ids_tc = sum(cutoff_track_ids_tc, [])
+    cutoff_track_ids_ts = [x[2] for x in new_ens]
+    cutoff_track_ids_ts = sum(cutoff_track_ids_ts, [])
+    if len(cutoff_track_ids_tc) > 0:
```
chahank, line 202: Thanks for updating the logging levels. 

One more question here: Suppose I generated tracks using this synthetic track generation method, and I get some tracks that where cut out at TC_category > 1. What should I do now? I could remove the track from the set, but this might screw up the frequency right? Otherwise I could rerun and hope to be lucky?

climada/hazard/tc_tracks_synth.py
```diff
@@ -452,6 +555,16 @@ def _calc_land_decay(hist_tracks, land_geom, s_rel=True, check_plot=False,
     Returns:
         v_rel (dict(category: A)), p_rel (dict(category: (S, B)))
     """
+    
+    if len(hist_tracks) < 100:
```
chahank, line 559: Is this still called when using the global decay parameters?

climada/hazard/tc_tracks_synth.py
```diff
@@ -707,56 +838,76 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
+        if S <= 1:
+            # central_pressure at start of landfall > env_pres after landfall:
+            # set central_pressure to environmental pressure during whole lf
+            track.central_pressure[sea_land:land_sea] = track.environmental_pressure[sea_land:land_sea]
+        else:
+            p_decay = _decay_p_function(S, p_rel[ss_scale][1],
+                                        track.dist_since_lf[sea_land:land_sea].values)
+            # dont apply decay if it would decrease central pressure
+            if np.any(p_decay < 1):
+                LOGGER.info('Landfall decay would decrease pressure for '
```
chahank, line 862: Maybe change to debug level?

climada/hazard/tc_tracks_synth.py
```diff
@@ -707,56 +838,76 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
...
+
+        v_decay = _decay_v_function(v_rel[ss_scale],
                                     track.dist_since_lf[sea_land:land_sea].values)
-        # dont applay decay if it would increas wind speeds
-        v_decay[v_decay > 1] = (track.max_sustained_wind[sea_land:land_sea][v_decay > 1]
-                                / v_landfall)
+        # dont apply decay if it would increase wind speeds
+        if np.any(v_decay > 1):
+            # should not happen unless v_rel is negative
+            LOGGER.info('Landfall decay would increase wind speed for '
```
chahank, line 876: debug level?

chahank:
@bguillod I think this pull request is close to be ready. It is a quite big change, thanks for the improvements! I now added only a few clarification comments. 

What is missing are test functions for ALL the added methods and the added kwargs. Without appropriate testing we cannot merge.

climada/hazard/tc_tracks_synth.py
mmyrte: No, see line 221 :)

chahank:
This pull request is still waiting for the test functions.

mmyrte:
@bguillod and I came to the conclusion that we won't be able to finish this in time for the release; sorry for that. It's on our to-do list though.

bguillod:
> This pull request is still waiting for the test functions.

@chahank As stated by Jan we did not find time to complete the tests for this PR in time. So I guess it's not going to be in this release, unfortunately. Sorry about that!
If you want to merge it anyway (the code works as we're using it) and we do a separate PR later this month to add appropriate tests, that is fine for us, but my understanding is you don't want code without full testing in the master and develop branch, in which case I assume the PR will only be merged for the next release.

chahank:
Thanks for the update @bguillod  and @mmyrte. It is for the next release then. Looking forward to it!

bguillod:
@chahank I believe I've now implemented unit tests for all changes. From my perspective the PR is now ready.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -283,6 +284,23 @@ def test_read_with_basin(self):
         for tr in tc_track.data:
             self.assertEqual(tr.basin[0], "EP")
             self.assertIn("WP", tr.basin)
+    
+    def test_discard_single_points(self):
+        """Check discard_single_points option"""
+        tc_track_singlept = tc.TCTracks()
+        passed = False
+        for year in range(1951, 1981):
```
chahank, line 292: This is a bit of a weird loop. It his to a)  take care of changing IBTrACS data and b) being fast?

chahank:
Thank you very much for adding extensive tests!

climada/hazard/test/test_tc_tracks.py
bguillod: Yes I agree it's weird, but exactly for the reasons you assume. Currently there is a single-point track in 1951 but if this changes then it would try further years. If none of all these 30 years contain single-point tracks at some point then the test could be updated by taking earlier years.

chahank: Why not let the loop run over more years then just to be sure?

chahank:
Please think about whether to extend the year range for the test discussed above and then you can proceed with the merging. Thanks again for the great work!

climada/hazard/test/test_tc_tracks.py
bguillod: Ok I'll start at 1863, which also has a unique points track, as well as 1865, 1866, 1867 and many others.

PR #160, opened 2021-03-05
util.config: add context manager to set logging level

chahank:
This adds a simple context manager to temporarily set all loggers (all CLIMADA, and also all other loggers) to a given level. This can come in handy since CLIMADA tends to produce a large number of logging messages in different modules which are not always needed for concrete applications.


tovogt:
Just a quick note that you can probably achieve this in a few lines of code by reseting only the root logger (this is taken from https://gist.github.com/simon-weber/7853144):
```
@contextmanager
def logging_disabled(highest_level=logging.CRITICAL):
    previous_level = logging.root.manager.disable
    logging.disable(highest_level)
    try:
        yield
    finally:
        logging.disable(previous_level)
```

climada/util/config.py
```diff
@@ -321,3 +321,33 @@ def _fetch_conf(directories, config_name):
     Path(Path.home(), '.config'),  # ~/.config directory
     Path.cwd(),  # current working directory
 ], CONFIG_NAME))
+
+
+class log_level:
+    """Context manager that sets ALL loggers to a given level."""
+
+    def __init__(self, level, climada_only=True):
```
chahank, line 329: I added the possibility to change the logging level of only CLIMADA loggers or of all loggers.

chahank:
Also, do you think it is a good idea to have the context manager placed in `util.config`?

climada/util/config.py
```diff
@@ -321,3 +321,33 @@ def _fetch_conf(directories, config_name):
...
+            for logger in self.loggers.values()
+            ] 
+
+    def __enter__(self):
+        for logger in self.loggers.values():
+            logger.setLevel(self.level)
+
+    def __exit__(self, exception_type, exception, traceback):
+        for logger, level in zip(self.loggers.values(), self.loglevels):
+            logger.setLevel(level)
```
tovogt, line 353: I suggest to have a kwarg for the logger prefix that defaults to an empty string, then filter by that name. Furthermore, the previous log-levels shouldn't be stored in a list because the order of dict keys/values might be non-persistent per the official specification.
```suggestion
    def __init__(self, level, name_prefix=""):
        self.level = level
        self.loggers = {
            name: (logger, logger.level)
            for name, logger in logging.root.manager.loggerDict.items()
            if isinstance(logger, logging.Logger) and name.startswith(name_prefix)
            }
        if name_prefix == "":
            self.loggers[""] = logging.getLogger()

    def __enter__(self):
        for logger, _ in self.loggers.values():
            logger.setLevel(self.level)

    def __exit__(self, exception_type, exception, traceback):
        for logger, previous_level in self.loggers.values():
            logger.setLevel(previous_level)
```

climada/util/test/test_config.py
```diff
@@ -61,6 +63,21 @@ def test_substitute(self):
...
+                logging.getLogger('climada').info('info')
+                logging.getLogger('climada').error('error')
+                self.assertEqual(cm.output, ['ERROR:climada:error'])
+        #Check other loggers are untouched
+        with self.assertLogs('matplotlib', level='DEBUG') as cm:
+            with log_level('ERROR', False):
+                logging.getLogger('climada').info('info')
+            logging.getLogger('matplotlib').debug('debug')
+            self.assertEqual(cm.output, ['DEBUG:matplotlib:debug'])
+
```
tovogt, line 79: Please adapt the tests accordingly if you would like to go with my `name_prefix` suggestion.

tovogt (changes requested):
Thanks, very nice!

Yes, we should probably move this to a different submodule. Since it doesn't fit into any of the existing submodules, I suggest to move it to the `__init__.py`. There you already find the `ureg` variable which is there for the same reason. Later, once we have a new submodule that fits better, we can still move it there and import it in `__init__.py` for backwards compatibility.

But if we now create a new submodule with only a single element, we might find later that a different structure (i.e. name of submodule) would have been better and backwards compatibility will be a pain.

climada/util/config.py
chahank: Good catch again! Should we allow only for prefix selection or generally any keyword present anywhere in the logger name?

climada/util/test/test_config.py
chahank: Where would you put the test if the module is in the `__init__.py`? Also, in which `__init__.py` file would you put it? The one from the `util` module?

climada/util/config.py
```diff
@@ -321,3 +321,38 @@ def _fetch_conf(directories, config_name):
...
+    
+    Examples
+    --------
+    Set ALL loggers temporarily to the level 'WARNING'
+    >>> with log_level(level='WARNING'):
+    >>>     ...
+        
+    Set the all climada loggers temporarily to the level 'ERROR'
+    >>> with log_level(level='ERROR'):
+    >>>     ...
```
tovogt, line 338: The `all climada loggers` part seems to be missing from this example.

climada/util/test/test_config.py
```diff
@@ -61,6 +63,21 @@ def test_substitute(self):
...
+    def test_set_logger(self):
+        #Check loggers are set to level
+        with self.assertLogs('climada', level='INFO') as cm:
+             with log_level('WARNING'):
+                logging.getLogger('climada').info('info')
+                logging.getLogger('climada').error('error')
+                self.assertEqual(cm.output, ['ERROR:climada:error'])
+        #Check if only climada loggers level change
+        with self.assertLogs('matplotlib', level='DEBUG') as cm:
+            with log_level('ERROR', name_prefix = 'climada'):
```
tovogt, line 75: code linting:
```suggestion
            with log_level('ERROR', name_prefix='climada'):
```

tovogt (changes requested):
Using a prefix is the most obvious use case because usually you will want to deactivate all loggers from a specific package. But if you want to go for the most general version, you can support regular expressions. Just matching for a substring is not a good option because it might lead to unwanted matches with subpackage names.

If you want to support regular expressions, you can stick to the `name_prefix` kwarg, but use `re.match` instead of `startswith` because `re.match` will also match at the beginning of the string. However, users can now choose to put `name_prefix=".*climada"` if they want to match anywhere in the name.

I suggest to move the code to the `climada/util/__init__py` and the tests to `climada/util/test/test_init.py`. The name of the test file doesn't really matter because it will never be imported. It can easily be changed later.

chahank:
Aaa no, regular expressions are painful. I will leave it a the prefix. I will move the files where you suggested. Maybe @emanuel-schmid can confirm this choice 
> I suggest to move the code to the `climada/util/__init__py` and the tests to `climada/util/test/test_init.py`

tovogt:
Well, there is code in `util.config` that affects the logging because there are parts of the configuration that affect the logging. So there is a reason why config-related logging code should be in `util.config`, but I see no reason why logging-related code that is not config-related should be in `util.config`.

chahank:
Fair point.

climada/util/__init__.py
```diff
@@ -26,3 +27,37 @@
...
+    name_prefix (default is "") to a given specified level.  
+    
+    Examples
+    --------
+    Set ALL loggers temporarily to the level 'WARNING'
+    >>> with log_level(level='WARNING'):
+    >>>     ...
+        
+    Set the all climada loggers temporarily to the level 'ERROR'
+    >>> with log_level(level='ERROR', name_prefix='cliamda'):
```
tovogt, line 42: typo:
```suggestion
    >>> with log_level(level='ERROR', name_prefix='climada'):
```

climada/util/config.py
```diff
@@ -321,3 +321,5 @@ def _fetch_conf(directories, config_name):
     Path(Path.home(), '.config'),  # ~/.config directory
     Path.cwd(),  # current working directory
 ], CONFIG_NAME))
+
+
```
tovogt, line 325: remove empty lines
```suggestion
```

tovogt (changes requested):
Remove the changes to `test_config.py`, since we moved the tests to a different file.

tovogt (approved):
Thanks, all good now!

PR #163, opened 2021-03-08
tc_surge_bathtub: fix tests for GDAL>=3.1

### LINKED ISSUE #156: test_fraction_on_land fails on develop in new environment
The test climada.hazard.test.test_tc_surge_bathtub.TestTCSurgeBathtub.test_fraction_on_land fails since commit f0adbe7920ff90ae68f9ebce5e99dd976c9bd7c7.

I assume, it is related to the updated rasterio and/or gdal versions, but I'm not 100% sure.

That the test fails is clear and the reasons are obvious. In the shore regions (fraction >0 <1) the values returned from `rasterio.warp.reproject` are different now. What is not clear is: are the new values any worse than the old ones? To me they look just as reasonable as the old ones, but I'm not familiar with the methods.

@tovogt: It would be great if you could have a look at this.

tovogt:
This will fix #156.

### DIFF
--- a/climada/hazard/test/test_tc_surge_bathtub.py
+++ b/climada/hazard/test/test_tc_surge_bathtub.py
@@ -44,7 +44,7 @@ class tmp_artifical_topo(object):
         res_deg : float
             Resolution in degrees
         """
-        lat = np.arange(bounds[1] + 0.5 * res_deg, bounds[3], res_deg)
+        lat = np.arange(bounds[3] - 0.5 * res_deg, bounds[1], -res_deg)
         lon = np.arange(bounds[0] + 0.5 * res_deg, bounds[2], res_deg)
         self.shape = (lat.size, lon.size)
         self.transform = rasterio.Affine(res_deg, 0, bounds[0], 0, -res_deg, bounds[3])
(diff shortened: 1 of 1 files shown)


climada/hazard/test/test_tc_surge_bathtub.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, bounds, res_deg):
         res_deg : float
             Resolution in degrees
         """
-        lat = np.arange(bounds[1] + 0.5 * res_deg, bounds[3], res_deg)
+        lat = np.arange(bounds[3] - 0.5 * res_deg, bounds[1], -res_deg)
```
tovogt, line 47: While I was at it, I corrected the latitudinal orientation of the temporary topo data.

climada/hazard/test/test_tc_surge_bathtub.py
```diff
@@ -69,9 +69,9 @@ def __enter__(self):
             'nodata': -32767.0,
         }
 
-        # In Windows, unlike Unix, the temporary file cannot be opened before it is closed
+        # In Windows, unlike Unix, the temporary file cannot be opened before it is closed.
         # Therefore it is closed right after creation and only the path/name is kept.
-        tmpfile = tempfile.NamedTemporaryFile()
+        tmpfile = tempfile.NamedTemporaryFile(delete=False)
```
tovogt, line 74: By default, the file is deleted when closed which is not what we want because we close it right away.

PR #164, opened 2021-03-08
Changing default parameter values for synthetic tracks generation

bguillod:
I would like to modify the default parameter values for the synthetic tracks.

Reason for this is an analysis over the North Atlantic, which highlights the current values to lead to unrealistic results. When I first wrote the new synthetic tracks algorithm, I was focusing on a few tracks that were of particular interest to me at the time. These happened to be relatively short tracks, leading to small perturbations by the end of the track (since the perturbations grow along the track). It looks like nobody has looked into this any further so here is how I defined the new default values.

I used all tracks from 1980-2019 over the North Atlantic basin as follows:
```
from climada.hazard.tc_tracks import TCTracks
tracks = TCTracks()
tracks.read_ibtracs_netcdf(provider='usa', basin='NA',
                           year_range=(1980,2019),
                           estimate_missing=True)
tracks.plot()
tracks.calc_perturbed_trajectories(nb_synth_tracks=15)
tracks.plot()
```

The spatial distribution of tracks looks as follows:
Historical data:
image
Event set:
image

So I tried a few alternatives and came using the parameter values proposed in this PR lead to:
image

I have also done a more detailed analysis looking at the frequency of tracks passing over 5x5 degrees boxes. The following shows the frequency in historical data (top), current default in the middle (center: difference in frequency, right: relative different in frequency), and new set of parameters (bottom):
image

The new one clearly is much closer than the current default (see also much smaller scale of differences). Finally, I did the same using individual points for TC of categories >=3 and >= 5, leading to the following two figures:
image
image
Again, much closer to the historical spatial distribution.

Note that the large relative values at the edges is because there are very few historical tracks there, hence values are very sensitive, so that is not a problem in my view.

I think this improvement should be included in the develop branch asap and possibly in the upcoming release, to avoid users playing around with an unrealistic set of parameter values.


bguillod:
I've allowed myself to also add a `legend` argument to `climada.hazard.tc_tracks.TCTracks.plot`. It sometimes come in handy to be able to deactivate legend display, e.g. when plotting several maps beside each other.

climada/hazard/tc_tracks_synth.py
```diff
@@ -39,9 +39,9 @@ def calc_perturbed_trajectories(tracks,
                                 nb_synth_tracks=9,
                                 max_shift_ini=0.75,
                                 max_dspeed_rel=0.3,
-                                max_ddirection=np.pi / 180,
+                                max_ddirection=np.pi / 360,
                                 autocorr_dspeed=0.85,
-                                autocorr_ddirection=0.85,
+                                autocorr_ddirection=0.5,
```
chahank, line 44: I do not see a problem with changing the default values in principles. I am not really convinced that the new defaults are much better (they do not hold up to a real scientific inquiry) though. What happens to the other basins? I did a similar analysis for all the basins, and some small changes in the values were needed for all of them. Hence, I suggest that if, as you say "I think this improvement should be included in the develop branch asap and possibly in the upcoming release, to avoid users playing around with an unrealistic set of parameter values.", then it is best to do this for all basins accordingly. It does not make much sense to argue for very fine-tuned values for NA, but not provide it for all basins. 

Here are the values I used (empty dics means I used the defaults prior to this pull request):

```
    params_basin = {
        'SA': {},
        'WP': {'max_dspeed_rel' : 0.2,
               'autocorr_ddirection' : 0.75},
        'EP': {'max_dspeed_rel' : 0.2,
               'autocorr_ddirection': 0.255},
        'NA': {'max_dspeed_rel' : 0.2, 
               'autocorr_ddirection' : 0.75, 
               'max_ddirection' : 3.1415/360}, 
        'SP': {'autocorr_ddirection' : 0.5, 
               'max_shift_ini' : 0.3},
        'SI': {},
        'NI': {'max_dspeed_rel'  : 0.2, 
               'autocorr_ddirection' : 0.25,
               'max_ddirection' : 3.1415/90},
        }
```

climada/hazard/tc_tracks.py
```diff
@@ -1119,13 +1119,16 @@ def extent(self):
         """Exact extent of trackset as tuple, no buffer."""
         return self.get_extent(deg_buffer=0.0)
 
-    def plot(self, axis=None, **kwargs):
+    def plot(self, axis=None, legend=True, **kwargs):
```
chahank, line 1122: I think it is fine that you added the possibility to turn off the legends, but it is not very clean to make it in an unrelated pull request.

chahank (changes requested):
Thanks for the suggestions! I would argue that it makes little sense to only fine tune for the NA basin. Thus, could you please make the same analysis for all ibtracks basins?

climada/hazard/tc_tracks_synth.py
bguillod: Ok so I've re-run the analysis for the global scale and here are the results:

**Frequency of all bypassing tracks:**
image

**Frequency of tracks Cat 3-5:**
image

Also, a comparison of the frequency (all tracks) grid cell by grid cell shows that the new parameters greatly reduce the deviation from historical frequencies, shown with standard and log axes (however please note that the grid cells with 0 in historical are not displayed in the log plot, but they are in the other one):
image
image

What do you think? Should change the parameters as I propose, or do you (or somebody else) wants to do a more detailed e.g. basin-specific analysis?

climada/hazard/tc_tracks.py
bguillod: You are right, do you want me to set that back? However I guess just this is not worth a pull request...

chahank: No no, just for the future ;)

climada/hazard/tc_tracks_synth.py
chahank: Thanks for the global analysis! I think that it is a great improvement to the first default parameters. Hence, please proceed and change the parameters. I am curious to see how they differ from the ones I used.

chahank: Also please do update the tests accordingly. Thanks!

bguillod:
Hi @chahank, I've adjusted the tests and they pass on my machine now. If you're fine with all of it you can approve the PR and merge. Thanks a lot!

climada/hazard/tc_tracks_synth.py
```diff
@@ -72,6 +72,10 @@ def calc_perturbed_trajectories(tracks,
     control how these perturbations persist in time and hence the amplitude of the
     perturbations towards the end of the track.
 
+    Note that the default parameter value have not been thoroughly calibrated, but they lead
+    to a reasonable distribution of tracks and their intensities over the North Atlantic
+    basin.
+
```
chahank, line 78: Given that you made an analysis worldwide, how about
```suggestion
    Note that the default parameter values have been only roughly calibrated so that
    the frequency of tracks in each 5x5degree box remains approximately constant.
    This is not an in-depth calibration and should be treated as such.
```

chahank:
Great, I just added one suggestion for the docstring to be a bit more explicit about the numbers.

PR #166, opened 2021-03-10
Fix to TCtracks.to_geodataframe

bguillod:
This pull requests improves the `climada.hazard.tc_tracks.TCTracks.to_geodataframe()` in two ways:

1. It ensures the track lines or points are within (-180, +180) degrees longitude.
2. If the new argument `split_lines_antimeridian` is `True` (default), tracks that cross the antimeridian are split into several lines, leading to a MULTILINE.

This ensures that when doing intersections with other features, tracks that cross the antimeridian are not thought by geopandas/shapely to cross the whole hemisphere from +180 to -180 degrees, for example.


climada/hazard/tc_tracks.py
```diff
@@ -1851,3 +1854,51 @@ def set_category(max_sus_wind, wind_unit='kn', saffir_scale=None):
...
+    lat = track.lat.values
+    # LineString only works with more than one lat/lon pair
+    if lon.size == 1:
+        return Point(lon, lat)
+    extent = u_coord.lon_bounds(lon)
+    if split_lines_antimeridian and extent[0] <= 180 <= extent[1]:
+        # in this case split into multilines
+        hem_west = lon < 0
+        # if all points on one side are +180 or -180, keep as a single line
+        if np.all(lon[hem_west] == -180) or np.all(lon[~hem_west] == 180):
```
mmyrte, line 1869: Seems like a strange condition, but of course an elegant solution if it's met!

mmyrte:
Seems to cover all the bases 🙂 
If performance should become a problem, then the iterator using `np.append()` may be worth looking at.

tovogt (changes requested):
I strongly discourage implementing the splitting yourself. Instead, I recommend to create your geodataframe and choose a CRS with a central longitude where no lines cross the antimeridian. Then, you can apply `shapely.ops.split` with the line from `(180, -90)` to `(180, 90)` to get your `MultiLineString`. After that, you change the CRS back to a CRS with central longitude 0. (I added suggested code in a commend below.)

Furthermore, you should add unit tests for your new feature.

bguillod:
> I strongly discourage implementing the splitting yourself. Instead, I recommend to create your geodataframe and choose a CRS with a central longitude where no lines cross the antimeridian. Then, you can apply `shapely.ops.split` with the line from `(180, -90)` to `(180, 90)` to get your `MultiLineString`. After that you change the CRS back to a CRS with central longitude 0.

But how do you do when you have global data then? I do not see how to choose a central longitude where no line crossed the anti-meridian. At least I was in a case where tracks covered, say, (-180, +210) without interruption, which is why I implemented this splitting myself. But if the method your propose works also for such cases then I don't have anything against it.

tovogt:
Okay, I thought that a central longitude around -160 should be save, but actually there is a handful of tracks in historical records that cross that line. But that's not a big thing, because you simply have to apply the splitting only to that part of the geodataframe where tracks come close to the antimeridian. Just take all tracks that have nodes in [160,200] and apply the splitting (as described in my post above) only to those with `mid_lon = 180`.

climada/hazard/tc_tracks.py
```diff
@@ -1237,17 +1241,16 @@ def to_geodataframe(self, as_points=False):
...
             gdf = gdf_long.join(gdf)
 
         else:
-            # LineString only works with more than one lat/lon pair
             gdf.geometry = gpd.GeoSeries([
-                LineString(np.c_[track.lon, track.lat]) if track.lon.size > 1
-                else Point(track.lon.data, track.lat.data)
+                _make_line_from_track(track, split_lines_antimeridian=split_lines_antimeridian)
                 for track in self.data
             ])
```
tovogt, line 1255: I suggest to remove `_make_line_from_track` and use `shapely.ops.split` instead:
```
            gdf.geometry = gpd.GeoSeries([
                LineString(np.c_[track.lon, track.lat]) if track.lon.size > 1
                else Point(track.lon.data, track.lat.data)
                for track in self.data
            ])
            gdf.crs = DEF_CRS
            if split_lines_antimeridian:
                t_mask = np.asarray([
                    (t.lon > 170).any() and (t.lon < -170).any() and t.lon.size > 1
                    for t in self.data])
                gdf.loc[t_mask, "geometry"] = gdf.geometry[t_mask] \
                    .to_crs({"proj": "longlat", "lon_wrap": 180}) \
                    .apply(lambda line: shapely.geometry.MultiLineString([
                        shapely.geometry.LineString([(x - 360, y) for x, y in segment.coords])
                        if any(x > 180 for x, y in segment.coords) else segment
                        for segment in shapely.ops.split(line, antimeridian).geoms
                    ]))
```

climada/hazard/tc_tracks.py
```diff
@@ -1851,3 +1854,64 @@ def set_category(max_sus_wind, wind_unit='kn', saffir_scale=None):
...
+        # if all points on one side are +180 or -180, keep as a single line
+        if np.all(lon[hem_west] == -180) or np.all(lon[~hem_west] == 180):
+            if np.all(lon[hem_west] == -180):
+                lon[hem_west] = 180
+            else:
+                lon[~hem_west] = -180
+            return LineString(np.c_[lon, lat])
+        # otherwise, split lines
+        split_idx = np.concatenate((np.array([0]),
+                                    np.where(np.diff(lon > 0))[0] + 1,
```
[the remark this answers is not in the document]
mmyrte: Both cases will not come to pass, since there is a check on the `extent` before this. Not even the (enormously odd) case of all values lying on the antimeridian would be met, since that is caught beforehand.

climada/hazard/tc_tracks.py
mmyrte: It's not equivalent though; it will definitely clean up those tracks that have orphaned points lying exactly on the "eastern half" of the dateline when moving westward. I agree that the `hem_west` should be replaced by a single condition for `hem_east`, of course.

climada/hazard/tc_tracks.py
mmyrte: The problem with the shapely ops is that you introduce unmergeable `MultiLineString`s on self-intersections of the underlying LineString - and since performance-wise, the difference is only small (@bguillod's version even wins out in my testing), I would propose to leave it as is and fix the `_make_line_from_track`.

tovogt: I don't understand the problem with self-intersections. Would you please add a unit test for that.

mmyrte: When a track intersects itself, the shapely operation will cut the `LineString` into a `MultiLineString` at the points of intersection. Benoit's homebrewed function avoids that and _only_ introduces breaks into the original line where it crosses the antimeridian. The issue is documented [here](https://github.com/Toblerity/Shapely/issues/572). The `shapely.ops.linemerge` does not help to stitch/meld the lines back together again.

In this concrete case an example looks like this:
image

And the `MultiLineString` coordinates around that track section that your code produces:

```
…
  (152.3000030517578, -23.200000762939453),
  (152.47250366210938, -23.279993057250977),
  (152.5104058915784, -23.285940228219452)],
 [(152.5104058915784, -23.285940228219452),
  (152.5702812971104, -23.29533517090472)],
 [(152.5702812971104, -23.29533517090472),
  (152.60000610351562, -23.299999237060547),
  (152.60571764022595, -23.292983514697355)],
 [(152.60571764022595, -23.292983514697355),
  (152.6875457763672, -23.19247055053711),
  (152.6999969482422, -23.100000381469727),
  (152.58509826660156, -23.185026168823242),
  (152.5104058915784, -23.285940228219452)],
 [(152.5104058915784, -23.285940228219452),
  (152.5, -23.299999237060547),
  (152.5702812971104, -23.29533517090472)],
 [(152.5702812971104, -23.29533517090472),
  (152.60571764022595, -23.292983514697355)],
 [(152.60571764022595, -23.292983514697355),
  (152.61257934570312, -23.29252815246582),
  (152.8000030517578, -23.299999237060547),
  …
  ```

mmyrte: I will write a unit test, let me see if we already have some tracks from that region in the repo.

tovogt: I see that this seems unexpected, but at the same time, it's not that bad: The docs to the shapely.ops.split function don't claim to preserve line strings. It just says that geometries are split at the desired locations and a `GeometryCollection` is returned. That is done successfully, also in your example. For plotting purposes and intersections with other geometries that will work just fine. I can't even think about a real usecase where this will be a problem. We could even use this workaround the shapely issue: https://github.com/Toblerity/Shapely/issues/1068#issuecomment-770296614

However, I understand that the result is unsatisfactory to you and would concede until the shapely issue is resolved. In this case, I insist that you add really rigorous testing. And I will probably have a few more review comments about the current implementation ...

tovogt:
I added some unit tests some of which currently fail. We need to fix the current implementation to make all tests pass before merging.

tovogt (approved):
Thanks, everything fine! I added a link to the shapely issue in the docstring for future reference.

mmyrte:
Vielen herzlichen Dank! I'm merging just in time for the release :)

bguillod:
Many thanks to both of you @mmyrte @tovogt for helping finalizing this PR!

PR #168, opened 2021-03-15
read_ibtracs_netcdf: fix for tracks crossing antimeridian

tovogt:
Usually, if an agency reports about a track that crosses the antimeridian, the longitude is always chosen positive within IBTrACS. However, it can happen that the TC crosses the antimeridian according to one agency, but not according to another. When mixing agency data, this can yield inconsistent sign changes in longitude. When using the `interpolate_missing` option, this will lead to wrong interpolated values.

According to my analysis, this only affects a single storm in IBTrACS with the default settings: [2013224N12220](http://ibtracs.unca.edu/index.php?name=v04r00-2013224N12220). However, you never know what might change in the future in IBTrACS. Furthermore, this PR also makes sure that no longitude values are less than -180 (which can happen for some agencies).

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -457,6 +457,19 @@ class TCTracks():
             preferred_idx = all_vals.notnull().any(dim="date_time").argmax(dim='agency')
             ibtracs_ds[tc_var] = all_vals.isel(agency=preferred_idx)
 
+            # Usually, if an agency reports about a track that crosses the antimeridian, the
+            # longitude is always chosen positive. However, it can happen, that the TC crosses the
+            # antimeridian according to one agency, but not according to another. When mixing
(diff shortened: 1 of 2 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -545,7 +563,7 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
             track_ds['poci'][:] = np.fmax(track_ds.poci, track_ds.pres)
 
             all_tracks.append(xr.Dataset({
-                'time_step': ('time', track_ds.time_step),
+                'time_step': ('time', track_ds.time_step.data),
```
tovogt, line 566: This produces a DeprecationWarning, so we correct this along the way.

bguillod (approved):
Looks good to me, thanks a lot!

PR #171, opened 2021-03-16
Feature/categorical plot

chahank:
This pull request adds a util function that can scatter plot categorical data, and a helper function to convert categorical data to numbered categories. For example, one could plot the largest TC storm category at each exposure point. 

The function is a copy of `util.plot.geo_scatter_from_array`, with an additional wrapper to convert the data to numbered categories with the new function `util.value_representation.val_to_cat` and a conversion of the continuous color map to a discrete one.


README.md
```diff
@@ -53,7 +53,7 @@ We use [SemVer](http://semver.org/) for versioning. For the versions available,
 
 Copyright (C) 2017 ETH Zurich, CLIMADA contributors listed in AUTHORS.
 
-CLIMADA is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, version 3.
+CLIMADA is free software: you can redistribute it and/or modify it under the terms of the GNU GENERAL PUBLIC LICENSE Version 3, 29 June 2007 as published by the Free Software Foundation, https://github.com/CLIMADA-project/climada_python/blob/main/LICENSE
```
tovogt, line 56: If you change the line above, you should also change the line below and also the link that comes at the very end.

chahank: Hmmm this change to the Readme was done by @emanuel-schmid in commit b945410 on the develop branch.

tovogt: Ah, this is a PR for the main branch, probably by mistake, right?

chahank: Yes, thanks, that is the problem! I will redo.

climada/util/plot.py
```diff
@@ -290,6 +292,98 @@ def geo_im_from_array(array_sub, coord, var_name, title,
 
     return axes
 
+def geo_scatter_categorical(array_sub, geo_coord, var_name, title,
```
chahank, line 295: This new method is copied from `geo_scatter_from_array`. The main differences are:

- Addition or argument `cat_name`
- Docstring
- Cmap definition in lines 335 - 339
- Definition of discrete (not continuous) cmap in lines 361 - 366
- Plot in line 369 adds `norm` and `cmap` arguements
- Add ticks label to colorbar in 377 -379

climada/util/plot.py
```diff
@@ -290,6 +292,98 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    num_im, list_arr = _get_collection_arrays(array_sub)
+    list_tit = to_list(num_im, title, 'title')
+    list_name = to_list(num_im, var_name, 'var_name')
+    list_coord = to_list(num_im, geo_coord, 'geo_coord')
+    
+    #default cmap
+    cmap_name = 'Accent'
+    if 'cmap' in kwargs:
+        cmap_name = kwargs['cmap']
+        del kwargs['cmap']
```
chahank, line 339: This change is needed because the `cmap` is now specified explicitly in line 369 so that the new discrete cmap can be used with its normalization factor. 

```
hex_bin = axis.scatter(coord[:, 1], coord[:, 0], c=valcat, norm=norm,
                               cmap=cmap, transform=proj, **kwargs)
```

climada/util/plot.py
tovogt: If it's a variation of that other function, then maybe it's easier to maintain the code if you don't create a new function but just add a new parameter to the existing function that allows to change its behavior.

chahank: That could be an idea indeed... at the same time, too many different behaviours for the same function are also hard to maintain right? Not sure what is best here.

zeliest: I also think it would make sense to not copy the code from the other function. Isn't it also possible to call the original function in the new function after creating the colormap and the category data? Or would it slow it to much?

chahank: Also a good idea. So there are three possible ways I see:

1. Keep as is, two separate methods (current implementation)
2. Add  a categorical  boolean argument to `geo_scatter_from_array` (suggested by @tovogt )
3. Make the `geo_scatter_categorical` method only build the colormap and then call `geo_scatter_array` (suggested by @zeliest)

An important design question that remains independently of the choices is:
- should the categories be shared among all subplots, or should each subplots have its own categories?
Example:
array1 = [1, 2, 1, 1]
array2 = [1, 2, 3, 1,  1]
array3 = ['a', 'b']
Shared categories: [1, 2, 3, 'a', 'b']
Categories per subplots: [1, 2] and [1, 2, 3] and ['a', 'b']

zeliest:
Screen Shot 2021-03-17 at 11 34 35

just a tiny comment, which is not specific to this plot: wouldn't it make sense to have only the coordinates on the bottom and right by default so that things are not on top of each other? I think we often just want to use the default and it looks a little messy like this. Or maybe it just happens to me?

chahank:
Thanks for the comment. If I am not mistaken, this is an issue with `cartopy=0.17` right @tovogt ?

tovogt:
Yes, @zeliest, please make sure you are using cartopy 0.18. The version of cartopy in the requirements file changed recently. The API of cartopy's gridlines feature changed from 0.17 to 0.18

climada/util/plot.py
zeliest: I think it might become too much information with different categories per subplots. And you could always make a separate plot right if you wanted different categories?

chahank: Yes, this is correct. So you recommend the shared categories approach? Also, which one of the three implementations do you favor?

zeliest: yes shared categories, and I guess I would personnaly do it how I suggested (3), but I cannot say it is the best way :)

chahank:
I modified the method according to proposal 3. The function `geo_scatter_categorical` still exists, and internally calls `geo_scatter_from_array`. 

While doing this refactoring, I also made sure that the categories are shared among all subplots (instead of being for a single category). To do so, the util function `val_to_cat` was updated to accept 1D and 2D arrays.

climada/util/plot.py
```diff
@@ -290,6 +293,79 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+            in corresponding geo_coord that are ploted in one subplot.
+        coord (2d np.array): (lat, lon) for each point in a row.
+            The same grid is used for all subplots.
+        var_name (str or list(str)): label to be shown in the colorbar. If one
+            provided, the same is used for all subplots. Otherwise provide as
+            many as subplots in array_sub.
+        cat_name (list(str)), optional: List of category names in same order
+            as the ordered categories obtain from np.unique(array_sub). All
+            subplots are assumed to have the same categories. 
+            Default is numbers (1,2,...)
```
tovogt, line 323: How are you supposed to know the names if you don't know the categories that are determined automatically within this function? Furthermore, if `cat_name` is not supplied, it doesn't seem to be very useful to fall back to numbers. Instead, the unique values should be used as category names.

And if you know the categories before, you shouldn't use this function to re-define the categories. They might differ from your original categories. For example, if you think you have the categories 0-5, but you somehow truncated your data and, in fact, only the categories 3 and 4 appear in the data you provide to this function, your category names for 0-5 won't be matched correctly.

tovogt (commented):
Some conceptual notes.

climada/util/value_representation.py
```diff
@@ -134,3 +134,44 @@ def value_to_monetary_unit(values, n_sig_dig=None, abbreviations=None):
...
+        Note: int and float are considered different numbers (1 != 1.0)
+    Returns
+    -------
+    valcat : np.array
+        List of the input values mapped onto categories.
+        The categories are deduced from the ordered input values.
+        
+    Example
+    -------
+    val_to_cat([1, 2, 1, 2, 2, 10]) = np.array([0, 1, 0, 1, 1, 2])
```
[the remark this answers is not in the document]
chahank: Excellent point, because I didn't know this function. Thanks!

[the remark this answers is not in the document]
chahank: However, np.unique flattens arrays, so one has to work a bit to apply it to a 2D array (i.e. reshape). I will remove this.

climada/util/value_representation.py
```diff
@@ -134,3 +134,44 @@ def value_to_monetary_unit(values, n_sig_dig=None, abbreviations=None):
...
+    val_to_cat([1, 'a', 'a']) = np.array([0, 1, 1])
+    val_to_cat([1, 1.0, 'a']) = np.array([0, 1, 2])
+    val_to_cat([[1, 1, 2], [1, 'a']]) = np.array([[0, 0, 1],
+                                                  [0, 2]])
```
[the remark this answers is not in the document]
chahank: I works only for square matrices.

climada/util/plot.py
chahank: Good points. One precision: it uses the unique valus as the category names, this is a bad Docstring. 

So, do you propose to not allow renaming the categories? One can always rename them afterwards from the axis handles (I would add this as example in the docstring since it is not obvious how too).

climada/util/plot.py
```diff
@@ -290,6 +293,79 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    # define the colormap
+    cmap = mpl.cm.get_cmap(cmap_name)       
+    # convert categories to numeric array [0, 1, ...]
+    array_sub_cat = u_vrpr.val_to_cat(array_sub) 
+    bounds = np.arange(-0.5, np.unique(array_sub_cat)[-1] + 1, 1) #np.unique sorts values
+    norm = mpl.colors.BoundaryNorm(bounds, cmap.N)
+    
+    #create the axes
+    kwargs['cmap'] = cmap
+    kwargs['norm'] = norm
```
tovogt, line 351:
```suggestion
    # convert categories to numeric array [0, 1, ...]
    array_sub  = np.array(array_sub)
    array_sub_unique, array_sub_cat = np.unique(array_sub, return_inverse=True)
    array_sub_cat = array_sub_cat.reshape(array_sub.shape)
    array_sub_n = array_sub_unique.size
    # define the colormap
    kwargs['cmap'] = mpl.cm.get_cmap(cmap_name, array_sub_n)
    kwargs['vmin'] = -0.5
    kwargs['vmax'] = array_sub_n - 0.5
    
    # create the axes
```

climada/util/plot.py
```diff
@@ -290,6 +293,79 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+                            shapes=shapes, axes=axes, **kwargs)
+    
+    #add colorbar labels
+    if not isinstance(axes, np.ndarray):
+        axes = [axes]
+    for ax in axes:
+        cbar = ax.collections[-1].colorbar
+        cbar.set_ticks(np.unique(array_sub_cat))
+        if cat_name:
+            cbar.set_ticklabels(cat_name)
```
tovogt, line 364:
```suggestion
        cbar = ax.collections[-1].colorbar
        cbar.set_ticks(np.arange(array_sub_n))
        cbar.set_ticklabels(cat_name if cat_name else array_sub_unique)
```

tovogt (commented):
Is there a reason why you choose `array_sub` instead of just `values`?

climada/util/plot.py
tovogt: I would rather have `cat_name` be a dict mapping from values to names.

climada/util/plot.py
tovogt: If the parameter `cat_name` is a dict, you would do: 
```
    cbar.set_ticklabels([cat_name[val] for val in array_sub_unique])
```

climada/util/plot.py
chahank: So you are using vmin and vmax instead of the boundary norm?

climada/util/plot.py
chahank: Yes. Now I tend to think that maybe not specifying the labels is better?

climada/util/plot.py
tovogt: Yes, it's more high-level.

chahank: The `array_sub` name is taken over from the other plotting functions.

climada/util/plot.py
tovogt: Maybe you are right. But if you want to write some kind of tutorial for how to do this, you can just as well remove the whole function and just add a tutorial on how to use `geo_scatter_from_array` or any other plotting function with discrete value ranges.

chahank: Hmmm I think this is a bit extreme. I mostly wrote this function because I want to use in the `uncertainty` module and tought I might not be the only one. Hence, I think this function is useful.

tovogt: From my perspective, this function is a wrapper around `geo_scatter_from_array` that applies a discrete scaling. However, I don't see, why we shouldn't implement it in a way so that it can be applied to an arbitrary plotting function that uses colormaps. Exlicitly repeating all the parameters of `geo_scatter_from_array` in the function header together with all the docstrings is quite some overhead.

chahank: @zeliest : what would be your preference? The possibility to specify `cat_name` or not?

chahank:
Ok, I implemented the suggestion from @tovogt . Thanks a lot for the enhancement!

I also updated the docstring to numpy (only for this one function so far).

I also changed slightly the implementation of `cat_name` dictionnary to tick labels by first converting all the keys of the dictionary to strings. This is because `np.unique` converts its output to strings if one of the input elements is a string. This way, the function can handle both numbers only, numbers and strings, and strings only.

climada/util/plot.py
chahank: Aaa now I understand your idea. But the potential issue I see is that for instance lets say you want to wrap around `geo_im_from_array`. I think this will not work well, as the latter makes an interpolation. 

This is why I did not go the route of a general wrapper function...

climada/util/plot.py
```diff
@@ -290,6 +293,103 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+        arguments for hexbin matplotlib function.
+
+    Returns
+    -------
+        cartopy.mpl.geoaxes.GeoAxesSubplot
+        
+    """
+    
+    #default cmap
+    cmap_name = 'Accent'
```
tovogt, line 357: Accent isn't a particular good choice if you don't happen to have exactly 8 categories, right? And it's really bad if you have more than 8 categories. Maybe the most generally useful solution would be to manually assemble a colormap containing the right number of colors.

climada/util/plot.py
```diff
@@ -290,6 +293,103 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    extend : str, optional
+        extend border colorbar with arrows. The default is 'neither'.
+        options: [ 'neither' | 'both' | 'min' | 'max' ]
+    proj : ccrs
+        coordinate reference system used in coordinates
+    shapes : bool, optional
+        Overlay Earth's countries coastlines to matplotlib.pyplot axis.
+        The default is True.
+    kwargs : optional
+        arguments for hexbin matplotlib function.
```
tovogt, line 348: Maybe we can include all parameters that are already defined in `geo_scatter_from_array` in `kwargs` and also remove the redundant docstrings.

climada/util/plot.py
```diff
@@ -290,6 +293,103 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    kwargs['vmax'] = array_sub_n - 0.5
+    
+    # #create the axes
+    axes = geo_scatter_from_array(array_sub_cat, geo_coord, var_name, title,
+                            pop_name=pop_name, buffer=buffer,
+                            extend=extend, proj=proj,
+                            shapes=shapes, axes=axes, **kwargs)
+    
+    #add colorbar labels
+    if not isinstance(cat_name, dict):
```
tovogt, line 380: What if `cat_name` is `None`?
```suggestion
    if cat_name is None:
        cat_name = array_sub_unique.astype(str)
    if not isinstance(cat_name, dict):
```

tovogt (commented):
Some more remarks.

climada/util/plot.py
tovogt: Here is how you would define a colormap from a list of colors (from https://matplotlib.org/stable/tutorials/colors/colormap-manipulation.html#creating-listed-colormaps):
```
cmap = ListedColormap(["darkorange", "gold", "lawngreen", "lightseagreen"])
```

chahank: Yes, you are right. However I prefer not to make colormaps myself, as this is not such an easy thing to do correctly (color blindness, black-white shades, good differentiation etc...). I will see whether I can make a good one using colorbrew.

climada/util/plot.py
chahank: I thought about that too. The annoying thing with this is that when one uses the helper overlay one sees only `kwargs` instead of the actual arguments.

climada/util/plot.py
tovogt: Yes, of course, you shouldn't reinvent the wheel but take the first `N` colors from an existing colormap.

climada/util/plot.py
tovogt: That doesn't justify the redundancy, in my opinion.

chahank: Hmmm I will think about it :D

climada/util/plot.py
chahank: Imho, it would be good to include `seaborn` into climada for the plotting.

climada/util/plot.py
tovogt: Or you can use something like docrep (https://docrep.readthedocs.io/en/latest/)

chahank: That looks great! Thanks

chahank: The questions is here: add another dependency to CLIMADA or not?

tovogt: `docrep` is a really light-weight dependency. It only depends on a single, trivial package (`six`, no version requirement).

chahank: ok, I will try it out for the current case

climada/util/plot.py
zeliest: I do think it makes sense to be able to specify it. I think that there are a lot of applications where the categories wouldn't directly be the intensity.

climada/util/plot.py
chahank: I made a try with docrep. It works, but I am not yet convinced whether this is a good addition to the CLIMADA code base. It is a bit more difficult to read for non-expert python developers and not that much is gained.

climada/util/plot.py
chahank: For the time being, I replaced the default with 'set3' which has 12 colors.

climada/util/plot.py
chahank: OK, then we can keep it. Note that intensity will almost never be a good candidate for a categoric plot, as it is usually a continuous variable in CLIMADA.

climada/util/plot.py
tovogt: Thanks for checking it out!

The use of similar docstring replacements is a standard approach in many Python libraries. See https://github.com/scipy/scipy/blob/master/scipy/_lib/doccer.py and https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/docstring.py

I think, a non-expert climada user/developer should look into the compiled API reference instead of the source code.

chahank:
So, I think that we reached a good level. @zeliest do you have some open comments?

climada/util/plot.py
```diff
@@ -31,17 +31,22 @@
...
+import matplotlib as mpl
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from shapely.geometry import box
 import cartopy.crs as ccrs
 from cartopy.io import shapereader
 from cartopy.mpl.gridliner import LONGITUDE_FORMATTER, LATITUDE_FORMATTER
 from rasterio.crs import CRS
 import requests
+from docrep import DocstringProcessor
+
```
tovogt, line 43:
```suggestion
```

climada/util/plot.py
```diff
@@ -138,37 +156,53 @@ def geo_bin_from_array(array_sub, geo_coord, var_name, title, pop_name=True,
...
+    kwargs : optional 
+        arguments for hexbin matplotlib function
+
+    Returns
+    -------
+    axes : cartopy.mpl.geoaxes.GeoAxesSubplot
+
+    Raises
+    ------
         ValueError
```
tovogt, line 203:
```suggestion
    ------
    ValueError
```

climada/util/plot.py
```diff
@@ -290,18 +335,95 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    ----------
+    cat_name : dict, optional
+        Categories name for the colorbar labels. 
+        Keys are all the unique values in array_cub, values are their labels.
+        The default is labels = unique values.
+    %(geo_scatter_from_array.parameters)s
+
+    Returns
+    -------
+        cartopy.mpl.geoaxes.GeoAxesSubplot
```
tovogt, line 369:
```suggestion
    cartopy.mpl.geoaxes.GeoAxesSubplot
```

climada/util/plot.py
```diff
@@ -462,11 +613,18 @@ def _get_borders(geo_coord, buffer=0, proj_limits=(-180, 180, -90, 90)):
...
+    Parameters
+    ----------
+    crs_in : str
+        Current crs
 
-    Returns:
-        ccrs.Projection, str
+    Returns
+    ------
+    crs_epsg, units: ccrs.Projection, str
```
tovogt, line 627:
```suggestion
    crs_epsg : ccrs.Projection
    units : str
```

chahank:
@emanuel-schmid Would you be fine with adding  `docrep` to the list of dependencies?

emanuel-schmid:
> @emanuel-schmid Would you be fine with adding  `docrep` to the list of dependencies?

Well, arh, I'm not convinced yet. These decorators are kind of hard to read - unless you're familiar with the package.\
How much would we gain in this pull request? How much would we possibly gain if we applied it all over climada?\
Me personally I wouldn't probably use it. I think in most cases I'd prefer redundancy in documentation over the obligation to check consistency. But maybe I haven't got the point of it.

chahank:
Ok, then I will remove it and just duplicate the docstring. I think also that this is simpler. It was worth a try.

climada/util/plot.py
chahank: I now use 'brg' which has 256 distinct colors. That should do it for the moment. If you have another suggestion for a super good set of colors (maybe @zeliest you can have a look whether some standard color palette could be defined for CLIMADA?) please let me know.

chahank:
This pull request is now ready to be merged from my side. Any more comments / issues?

climada/util/plot.py
```diff
@@ -290,18 +331,124 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    array_sub_n = array_sub_unique.size
+    
+    #default cmap
+    cmap_name = 'brg'
+    if 'cmap' in kwargs:
+        cmap_name = kwargs['cmap']
+        del kwargs['cmap'] 
+    
+    # define the discrete colormap
+    kwargs['cmap'] = mpl.cm.get_cmap(cmap_name, array_sub_n)
```
tovogt, line 410: The color map `brg` is not a qualitative color map and therefore not well-suited for the purpose of this plotting function. I suggest to use a color map with less colors and warn users that are about to abuse it.

```suggestion
    #default cmap
    cmap = mpl.colors.ListedColormap(plt.get_cmap('tab10').colors[:array_sub_n]) 
    if array_sub_n > cmap.N:
        LOGGER.warning("More than %d categories cannot be plotted accurately "
                       "using the default color map '%s'. You can manually specify "
                       "a color map using the `cmap` attribute.", cmap.N, cmap.name)
    if 'cmap' in kwargs:
        cmap = kwargs['cmap']
    
    # define the discrete colormap
    kwargs['cmap'] = mpl.cm.get_cmap(cmap, array_sub_n)
```

chahank: Fair enough. I had the same idea. Since we have the same opinion, that is perfect. Thanks!

climada/util/plot.py
```diff
@@ -400,14 +400,24 @@ def geo_scatter_categorical(array_sub, geo_coord, var_name, title,
     array_sub_cat = array_sub_cat.reshape(array_sub.shape)
     array_sub_n = array_sub_unique.size
     
-    #default cmap
-    cmap_name = 'brg'
+    #default qualitative cmap
+    CMAP = 'Dark2'
```
tovogt, line 404: Uppercase variable names are reserved for module-level constants.

climada/util/plot.py
```diff
@@ -290,18 +331,134 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    """
+    
+    # convert sorted categories to numeric array [0, 1, ...]
+    array_sub = np.array(array_sub)
+    array_sub_unique, array_sub_cat = np.unique(array_sub, return_inverse=True) #flattens array
+    array_sub_cat = array_sub_cat.reshape(array_sub.shape)
+    array_sub_n = array_sub_unique.size
+    
+    #default qualitative cmap
+    CMAP = 'Dark2'
```
tovogt, line 404: Uppercase variable names are reserved for module-level constants.

climada/util/plot.py
```diff
@@ -290,18 +331,134 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+        CMAP = kwargs['cmap']
+        del kwargs['cmap'] 
+        
+    cmap = mpl.colors.ListedColormap(plt.get_cmap(CMAP).colors[:array_sub_n]) 
+    if array_sub_n > cmap.N:
+        LOGGER.warning("More than %d categories cannot be plotted accurately "
+                       "using the default color map '%s'. You can manually "
+                       "specify a QUALITATIV color map using the `cmap` "
+                       "attribute. Matplotlibs default are "
+           "https://matplotlib.org/stable/tutorials/colors/colormaps.html",
```
tovogt, line 416:
```suggestion
                       "using the default colormap '%s'. You can manually "
                       "specify a qualitative colormap using the `cmap` "
                       "attribute. For Matplotlib's built-in colormaps, see "
           "https://matplotlib.org/stable/tutorials/colors/colormaps.html",
```

climada/util/plot.py
```diff
@@ -290,18 +331,134 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+                       cmap.N, cmap.name)
```
tovogt, line 417: Maybe you mean:
```suggestion
                       CMAP.N, CMAP.name)
```

climada/util/plot.py
```diff
@@ -290,18 +331,134 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    array_sub_unique, array_sub_cat = np.unique(array_sub, return_inverse=True) #flattens array
+    array_sub_cat = array_sub_cat.reshape(array_sub.shape)
+    array_sub_n = array_sub_unique.size
+    
+    #default qualitative cmap
+    CMAP = 'Dark2'
+        
+    if 'cmap' in kwargs:
+        CMAP = kwargs['cmap']
+        del kwargs['cmap']
```
tovogt, line 408:
```suggestion
```

climada/util/plot.py
chahank: True. Probably we should define color-shemes on the module-level at some point, but not in this pull request.

climada/util/plot.py
chahank: No, because it is taken from the variable `cmap = mpl.colors.ListedColormap(plt.get_cmap(cmap_name).colors[:array_sub_n]`.

chahank: `CMAP` is a string.

tovogt: No, CMAP is not necessarily a string if it is provided by the user using `kwargs`. However, `cmap.name` will always be `"from_list"` because all colormaps created using `ListedColormap` have that name.

chahank: Not always a string, but can be. But indeed `from_list` is not intended (I actually copied that from you ;) Thanks!

chahank: But I see I changed the order of the lines, which changes the behaviour.

chahank: I changed the lines a bit again so that the check is made not only for the default colormap, but also the user defined colormap.

climada/util/plot.py
```diff
@@ -290,18 +331,143 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+        ) 
+
+    #optional user defined color map (can be continuous)
+    if 'cmap' in kwargs:
+        cmap = kwargs['cmap']
+        if isinstance(cmap, str):
+            cmap_name = cmap
+            cmap = mpl.cm.get_cmap(cmap)
+        else:
+            cmap_name = 'defined by the user'
```
tovogt, line 417: Some cleanup:
```suggestion
    if 'cmap' in kwargs:
        # optional user defined colormap (can be continuous)
        cmap = kwargs['cmap']
        if isinstance(cmap, str):
            cmap_name = cmap
            cmap = mpl.cm.get_cmap(cmap)
        else:
            cmap_name = 'defined by the user'
    else:
      # default qualitative colormap        
      cmap_name = 'Dark2'
      cmap = mpl.colors.ListedColormap(
          plt.get_cmap(cmap_name).colors[:array_sub_n]
          ) 
```

climada/util/plot.py
```diff
@@ -290,18 +331,143 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    
+    if array_sub_n > cmap.N:
+        LOGGER.warning("More than %d categories cannot be plotted accurately "
+                       "using the colormap %s. Please specify "
+                       "another qualitative colormap using the `cmap` "
```
tovogt, line 422: "another" != "a different"
```suggestion
                       "a different qualitative colormap using the `cmap` "
```

chahank:
I think this ready to merge. @zeliest , any more comments?

tovogt (approved):
Looks good now.

PR #173, opened 2021-03-17
Feature/event impact2annual impact

carmensteinmann:
Hi Chahan, 
This is a first version of the event_yearset functions. I'll be writing unit tests in the coming days. 
Please let me know if parts of the code are unclear / can be structured in a better way.

Best, 
Carmen


chahank (changes requested):
Excellent addition to CLIMADA! The review will probably take some time as it proposes a new functionality for CLIMADA which requires some care.

This first shot is only a short review mainly concerning the docstrings. I think these can be directly updated. The actual code will be reviewed soonish.

chahank:
I add this comment here so that it does not get buried by the "outdating" mechanism of github:

Please update all the Docstrings to the Numpy default. See https://github.com/CLIMADA-project/climada_python/blob/develop/doc/guide/Guide_PythonDos-n-Donts.ipynb

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+
+    ais_eai = np.sum(impact_per_year)/nr_resampled_years
+    eis_eai = np.sum(eis.frequency*eis.at_event)
+    correction_factor = eis_eai/ais_eai
+    LOGGER.info("The correction factor amounts to %s", (correction_factor-1)*100)
+
+    return correction_factor
+
+def wrapper_multi_impact(list_impacts, nr_resampled_years):
```
chahank, line 250:
```suggestion
def wrapper_multi_impact(list_impacts, n_resampled_years):
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+    """Compute the total impact of several event impact sets in one annual impact set
+
+    INPUT:
+        list_impacts (list): list of impact class objects
+        nr_resampled_years (int): the target number of years the impact yearset shall
```
chahank, line 255:
```suggestion
        n_resampled_years (int): the target number of years the impact yearset shall
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+
+import climada.util.dates_times as u_dt
+
+LOGGER = logging.getLogger(__name__)
+
+def impact_yearset(eis, n_resampled_years=None, year_list=None, multiple_events=True,
+                   sampling_vect=None, nr_events_per_year=None):
+
+    """PURPOSE:
+      convert an event (per occurrence) impact set (eis) into an annual impact
```
chahank, line 29: By 'eis', do you simply mean an Impact() object? 

Currently this is not very clear. In CLIMADA, there is no concept of event impact set and annual impact set in the core engine. It is not a good idea to introduce a new nomenclature in the middle of the code.

Maybe something like this is clearer? 
```suggestion
      Create an annual impact set (ais) by sampling events for each year from an existing probabilistic event impact set (eis).
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+            (by default a list starting on the 01-01-0001 is generated)
+        multiple_events (boolean): True if the hazard causing the given impact can occur
+            several times per year (such as several tropical cyclones); False for hazards
+            that compute the impact on an annual scale (such as relative cropyield)
+        sampling_vect (array): the sampling vector, technical, see code (can be used to
+          re-create the exact same yearset). Needs to be obtained in a first
+          call, i.e. [ais, sampling_vect] = climada_eis2ais(...) and then
+          provided in subsequent calls(s) to obtain the exact same sampling
+          structure of yearset, i.e ais = climada_eis2ais(..., sampling_vect)
+        nr_events_per_year (array): amount of resampled events per year
```
chahank, line 47: Same as for `n_resampled_years`, `nr_events_per_year` -> `n_events_per_year`

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+
+    return sampling_vect, nr_events_per_year
+
+def calculate_correction_fac(impact_per_year, nr_resampled_years, eis):
+    """Apply a correction factor to ensure the expected annual impact (eai) of the annual
+    impact set(ais) amounts to the eai of the event impact set (eis)
+
+    INPUT:
+        impact_per_year (array): resampled annual impact set before applying the correction factor
+        nr_resampled_years (int): the target number of years the annual impact set contains
```
chahank, line 236: Why do you need this, isn't the number of sample years a variable that can be retrieved from eis? The same for impact_per_year?

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+        eis (impact class object): event impact set
+
+    OUTPUT:
+        correction_factor (int): the correction factor is calculated as eis_eai/ais_eai
+    """
+
+    ais_eai = np.sum(impact_per_year)/nr_resampled_years
+    eis_eai = np.sum(eis.frequency*eis.at_event)
+    correction_factor = eis_eai/ais_eai
+    LOGGER.info("The correction factor amounts to %s", (correction_factor-1)*100)
```
chahank, line 246: Why add this in the info for the Logger? How would that be useful for a normal user?

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+    OUTPUT:
+        ais_total (impact class object): combined annual impact set for all given event impact sets
+    """
+
+    ais_total = np.zeros(nr_resampled_years)
+    for impact in list_impacts:
+        ais_impact = impact_yearset(impact, nr_resampled_years)
+        ais_total += ais_impact
+
+    return ais_total
```
chahank, line 266:
```suggestion
    return sum([impact_yearset(impact, n_resampled_years) for impact in list_impacts])
```

climada/util/yearsets.py
chahank, line 250: Is this function really needed? It is a one-liner ...

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+                                                    size=nr_resampled_years)).astype('int')
+
+    #non_zero_years =  np.where(nr_events_per_year != 0)
+    tot_nr_events = sum(nr_events_per_year)
+
+    sampling_vect = sampling_uniform(tot_nr_events, nr_input_events)
+
+    return sampling_vect, nr_events_per_year
+
+def calculate_correction_fac(impact_per_year, nr_resampled_years, eis):
```
chahank, line 230: Very simple way to bias correct the sample, which might or might not be a good idea depending on the application. I think this ties in very well with the larger 'resampling' / ' emulator' discussion. We should keep it as an example for this discussion.

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+
+    idx = 0
+    for year in range(nr_resampled_years):
+        impact_per_year[year] = np.sum(impact_per_event[idx:(idx+nr_events[year])])
+        idx += nr_events[year]
+
+    return impact_per_year, nr_events, sampling_vect
+
+
+def sampling_uniform(tot_nr_events, nr_input_events):
```
chahank, line 181: If you would use scipy, I think you can make one single sampling function for both `uniform` and `poisson`, with in addition the possibility to use any other type of discrete sampling. This would make the code significantly shorter and more general. Pleaselet me know if you think it is a good idea and I can propose an implementation.

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+    ais.date = u_dt.str_to_date(year_list)
+    ais.at_event = impact_per_year / correction_factor
+    ais.frequency = np.ones(n_resampled_years)*sum(nr_events)/n_resampled_years
+
+
+    return ais, sampling_vect, nr_events
+
+def resample_single_annual_event(eis, nr_resampled_years, sampling_vect=None):
```
chahank, line 116: Why use the words `resample` if the method is only `sampling` ? idem for `resample_multiple_annual_events`.

chahank (changes requested):
This is a second round of comments which still focus on the code structure/style.

I suggest that you change everywhere `nr` to `n`. The former is simply not a standard notation, and it is unnecessarily longer imho.

climada/util/yearsets.py
carmensteinmann: Yes, that sounds good! Would be great if you could propose how to implement it :).

climada/util/yearsets.py
carmensteinmann: I'll note it down as a point for further discussion.

climada/util/yearsets.py
carmensteinmann: No, both refer to the output of the whole function: the target amount of (re-)sampled years (- maybe n_target_years would be a better name?) and the impact_per_year of the annual impact set

climada/util/yearsets.py
carmensteinmann: maybe not, I just started writing it as I was thinking about multi-hazard. I could imagine it will get more complex once these thoughts are fleshed out...

climada/util/yearsets.py
chahank: But it should be that `len(impact_per_year) == nr_resampled_years` right? This function here is only doing the bias correction, hence the number of years cannot be changed by this function. Thus, the number of years should also not be a necessary input variable.

climada/util/yearsets.py
chahank: Interresting! What do you mean it will get more complex?

climada/util/yearsets.py
```diff
@@ -0,0 +1,266 @@
...
+          date(i): the year i
+          impact(i): the sum of impact for year(i).
+          frequency(i): the annual frequency, =1
+      sampling_vect: the sampling vector, technical, see code (can be used to
+                     re-create the exact same yearset)
+      """
+
+    # OLD CODE - STILL NEEDED?
+    # if years are known:
+    # events = list(EIS.calc_impact_year_set().keys())
```
chahank, line 62: Why did you decide to not use the already existing method for computing the impact year set for a given set of events?

carmensteinmann: I think I wanted to use this for hazards that have 1 event per year (otherwise sampling on impact_year_set basis doesn't make much sense, right?). I guess we can delete it.

climada/util/yearsets.py
carmensteinmann: I think it gives an estimate on how the eai differs for the input event impact set and the newly generated one - maybe reformulating it (adding the information on how the correction factor is calculated) would make it clearer / more informative? The idea was also to add the possibility to apply a different kind of correction if the correction factor exceeds a certain threshold (see lines ~105).

climada/util/yearsets.py
```diff
@@ -161,50 +158,46 @@ def sample_events(tot_n_events, n_input_events):
         selected_events (array): uniformaly sampled events (length: len(tot_n_events))
       """
 
-    repetitions = np.ceil(tot_n_events/(n_input_events-1)).astype('int')
+    repetitions = np.ceil(tot_n_events/n_input_events).astype('int')
+    indeces = np.tile(np.arange(n_input_events), repetitions)
```
chahank, line 162:
```suggestion
    indices = np.tile(np.arange(n_input_events), repetitions)
```

climada/util/yearsets.py
```diff
@@ -25,31 +25,32 @@
...
-            are selected in the newly created annual_impacts and the number of events per year in the
-            annual_impacts, therefore the sampling_vect contains two arrays:
+        sampling_vect (dict): the sampling vector specifying which events contained in the
+            events_impacts are selected in the newly created annual_impacts and the number
+            of events per year in the annual_impacts, therefore the sampling_vect contains
+            two arrays:
                 selected_events (array): sampled events (len: total amount of sampled events)
-                events_per_year (array): events per resampled year 
-            The sampling_vector needs to be obtained in a first call, 
+                events_per_year (array): events per resampled year
```
chahank, line 43: Just a thought: with the restructuring and simplification of the module, does the name `resampling` still makes sense, or is `sampling` more appropriate?

climada/util/yearsets.py
```diff
@@ -72,7 +73,7 @@ def impact_yearset(event_impacts, sampled_years=None, sampling_vect=None):
 
     if not np.all(event_impacts.frequency == event_impacts.frequency[0]):
         LOGGER.warning("The frequencies of the single events in the given event_impacts"
-                       "differ among each other. Please beware that this will influence" 
+                       "differ among each other. Please beware that this will influence"
```
chahank, line 76: Could you say influence in what way?

climada/util/yearsets.py
```diff
@@ -25,31 +25,32 @@
 def impact_yearset(event_impacts, sampled_years=None, sampling_vect=None):
 
     """PURPOSE:
-      Create an annual_impacts object containing a probabilistic impact for each year 
-      in the sampled_years list (or a list generated with the length of given sampled_years) 
+      Create an annual_impacts object containing a probabilistic impact for each year
+      in the sampled_years list (or a list generated with the length of given sampled_years)
       by sampling events from the existing input event_impacts.
```
chahank, line 30: Maybe this would make sense?
```suggestion
      by sampling events from the existing input event_impacts with a Poisson distribution centered
      around n_events = sum(event_impacts.frequency).
```

climada/util/yearsets.py
carmensteinmann: You're right, I'll change all the "resampling" to "sampling"

climada/util/event_yearset.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # event_date = date_to_str(eis.date)
+    # event_year = [i.split('-', 1)[0] for i in event_date]
+
+    #n_years_his = np.ceil(1/min(eis.frequency)).astype('int')
+    # n_events = len(eis.event_id)
+
+
+    if not year_list:
+        year_list = [str(date) + '-01-01' for date in np.arange(1, nr_resampled_years+1).tolist()]
```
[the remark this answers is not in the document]
carmensteinmann: I later compute the date in ordinal format (when saving it). This step allows the user to give a list of years as input - can it be left as is or would you suggest changing it?

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+      centered around n_events per year (n_events = sum(event_impacts.frequency)).
+
+    Parameters:
+      event_impacts : impact object
+          impact object containing impacts per event
+    Optional parameters:
+        sampled_years : int or list
+            Either an integer specifying the amount of years to
+            be sampled or a list of years that shall be covered by the resulting annual_impacts
+            default: a 1000 year-long list starting in the year 0001
```
chahank, line 40:
```suggestion
            Either an integer specifying the number of years to
            be sampled (labelled [0001,...,sampled_years]) or a list 
            of years that shall be covered by the resulting annual_impacts.
            The default is a 1000 year-long list starting in the year 0001.
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+              selected_events (array) : sampled events (len: total amount of sampled events)
+              events_per_year (array) : events per sampled year
+          Can be used to re-create the exact same annual_impacts yearset
+      """
+
+
+    if not sampled_years and not sampling_vect:
+        sampled_years = np.arange(1, 1000+1).tolist()
+    elif not sampled_years:
+        sampled_years = np.arange(1, len(sampling_vect['selected_events'])+1).tolist()
```
chahank, line 69:
```suggestion
        sampled_years = np.arange(1, len(sampling_vect['selected_events']) + 1).tolist()
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Parameters:
+      event_impacts : impact object
+          impact object containing impacts per event
+    Optional parameters:
+        sampled_years : int or list
+            Either an integer specifying the amount of years to
+            be sampled or a list of years that shall be covered by the resulting annual_impacts
+            default: a 1000 year-long list starting in the year 0001
+        sampling_vect : dict
```
chahank, line 41:
```suggestion
        sampling_dict : dict
```

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+            in the annual_impacts, therefore the sampling_vect contains two arrays:
+                selected_events (array): sampled events (len: total amount of sampled events)
+                events_per_year (array): events per sampled year
+            The sampling_vector needs to be obtained in a first call,
+            i.e. [annual_impacts, sampling_vect] = climada_yearsets.impact_yearset(...)
+            and can then be provided in subsequent calls(s) to obtain the exact same sampling
+            (also for a different event_impacts object)
+        correction_fac : boolean
+            If True the resulting annual_impacts are scaled in so that the expected annual
+            impact (eai) of the annual_impacts = eai of the events_impacts
```
chahank, line 53: I think this docstring would benefit from a bit from clarity.

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+        sampled_years = np.arange(1, 1000+1).tolist()
+    elif not sampled_years:
+        sampled_years = np.arange(1, len(sampling_vect['selected_events'])+1).tolist()
+    elif isinstance(sampled_years, int):
+        sampled_years = np.arange(1, sampled_years+1).tolist()
+
+    if sampling_vect:
+        if len(sampled_years) != len(sampling_vect['events_per_year']):
+            raise ValueError("The number of sampled_years and the length of the list of "
+                             "events_per_year in the sampling vector must be equal.")
```
chahank, line 76: Would it not be more logical to infer `sampled_years` from `sampling_vect` ? I do not really see a possiblity to define these two separately. Then you do not need the error message, just a warning maybe that `sampled_years` was ignored.

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+        if sum(sampling_vect['events_per_year']) != len(sampling_vect['selected_events']):
+            raise ValueError("The sampling vector is faulty: the sum of selected events "
+                             "does not correspond to the number of selected events.")
+
+    year_list = [str(date) + '-01-01' for date in sampled_years]
```
chahank, line 81: This is never actually used right? Then it is probably better to remove it here and directly set
```
n_sampled_years = len(sampled_years)
annual_impacts.date = u_dt.str_to_date([str(year) + '-01-01' for year in sampled_years])
```

climada/util/event_yearset.py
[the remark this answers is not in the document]
chahank: It is good yes! I added a comment because the variable `year_list` is now never used and thus can be removed.

https://github.com/CLIMADA-project/climada_python/pull/173/files#r604697615

climada/util/yearsets.py
carmensteinmann: Yes, it can (line 69) - I'll add this case there

climada/util/yearsets.py
```diff
@@ -0,0 +1,244 @@
...
+    if not np.all(event_impacts.frequency == event_impacts.frequency[0]):
+        LOGGER.warning("The frequencies of the single events in the given event_impacts "
+                       "differ among each other. Please beware that this will influence "
+                       "the resulting annual_impacts as the events are sampled uniformaly "
+                       "and different frequencies are (not yet) taken into account.")
+
+
+    # INITIALISATION
+    # to do: add a flag to distinguish annual_impacts from other impact class objects
+    # artificially generate a generic  annual impact set (by sampling the event impact set)
```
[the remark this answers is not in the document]
carmensteinmann: That could be an option, e.g. by adding: annual_impacts.tag['annual_impacts object'] = True
What do you think?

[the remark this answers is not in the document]
chahank: Excellent idea!

climada/util/yearsets.py
```diff
@@ -0,0 +1,247 @@
...
+    annual_impacts = copy.deepcopy(event_impacts)
+    annual_impacts.event_id = [] # to indicate not by event any more
+    annual_impacts.event_id = np.arange(1, n_sampled_years+1)
+    annual_impacts.frequency = [] # init, see below
```
carmensteinmann, line 101: @chahank: does it make sense to have this line or is the overwriting of the frequency at a later stage sufficient?

climada/util/yearsets.py
```diff
@@ -0,0 +1,247 @@
...
+    annual_impacts.at_event = np.zeros([1, n_sampled_years+1])
+
+    #create sampling dictionary if not given as input
+    if not sampling_dict:
+        n_annual_events = np.sum(event_impacts.frequency)
```
carmensteinmann, line 106: @chahank I'm questioning this computation - would it not make more sense to calculate the average number of annual events as np.mean(1/event_impacts.frequency)?

climada/util/yearsets.py
chahank: It is not necessary at all and adds data handling, but might increase readability. You can choose to keep or remove.

climada/util/yearsets.py
chahank: I think your formula is not correct. Suppose I have one event which happens `1/100` years, thus `event_impacts.frequency = [1/100]`. Then `np.mean(1/event_impacts.frequency) = 100` instead of `1/100`.

climada/util/yearsets.py
```diff
@@ -0,0 +1,247 @@
...
+import numpy as np
+from numpy.random import default_rng
+
+import climada.util.dates_times as u_dt
+
+LOGGER = logging.getLogger(__name__)
+
+def impact_yearset(event_impacts, sampled_years=None, sampling_dict=None, correction_fac=True):
+
+    """Create an annual_impacts object containing a probabilistic impact for each year
```
zeliest, line 27: I think it could be nice to add a small sentence explaining how this is different to the expected annual impact, I guess the difference is already explained in the "impact for **each** year" but it still took me while to understand

climada/util/test/test_yearsets.py
```diff
@@ -0,0 +1,111 @@
...
+EVENT_IMPACTS = Impact()
+EVENT_IMPACTS.at_event = np.arange(10,110,10)
+EVENT_IMPACTS.frequency = np.array(np.ones(10)*0.2)
+
+SAMPLING_DICT = dict()
+SAMPLING_DICT = {'selected_events': np.tile(np.arange(0,10),2),
+                  'events_per_year': np.array(np.ones(10, dtype=int)*2)}
+
+N_SAMPLED_YEARS = 10
+YEAR_LIST = np.arange(2000, 2010).tolist()
```
zeliest, line 40: I guess it makes sense to write this directly as a list if you are going to turn it to a list after? list(range(2000, 2010))
there are a few of these below as well :)

climada/util/yearsets.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+
+    #sample number of events per year
+    if n_annual_events != 1:
+        events_per_year = np.round(np.random.poisson(lam=n_annual_events,
+                                                     size=n_sampled_years)).astype('int')
+    else:
+        events_per_year = np.ones(len(n_sampled_years))
+
+    tot_n_events = sum(events_per_year)
```
zeliest, line 156: I think it is better to use np.sum with numpy arrays in general :)

climada/util/test/test_yearsets.py
```diff
@@ -0,0 +1,111 @@
...
+        n_annual_events = 2
+        n_input_events = 10
+        sampling_dict = yearsets.create_sampling_dict(n_sampled_years, n_annual_events, n_input_events)
+        self.assertAlmostEqual(np.round(np.mean(sampling_dict['events_per_year'])), 2)
+        self.assertTrue(len(np.unique(list(collections.Counter(sampling_dict['selected_events']).values()))), 2)
+
+    def test_computing_annual_impacts(self):
+        """Test the calculation of annual impacts from a given sampling dictionary."""
+        impact_per_year = yearsets.compute_annual_impacts(EVENT_IMPACTS, SAMPLING_DICT)
+
```
chahank, line 97:
```suggestion
```

chahank:
Please perform a static code analysis with pylint to polish the code.

climada/util/yearsets.py
carmensteinmann: you're right, I got confused with the frequencies yesterday. I think the formula is correct as is :).

PR #174, opened 2021-03-18
Feature/scipy_16_update

emanuel-schmid:
the zoom `mode='constant'` with `cval=0.0` reproduces the former `mode='nearest'` results.
those are the default values of the respective parameters of `zoom`.

This pull request fixes the `climada.entity.exposures.test.test_black_marble.TestNightLight.test_cut_country_brb_2km_pass` failure that was introduced on February 8, 2020.


emanuel-schmid:
> 
> 
> Thanks for taking this up!
> 
> I'm not convinced that we want to change the boundary mode though. The `mode="nearest"` parameter was actually quite reasonable. I would rather change the test values and stick to the boundary mode.

👍 just as well.

PR #176, opened 2021-03-18
Bugfix/exposure.assign centroids/coord.dtype

### LINKED ISSUE #175: Exposure.assign_centroids assumes that hazard.centroids.coord is in 'float64'
In assign_centroids, the code assumes that the variable `hazard.centroids.coord` (a `numpy ndarray`) is of `dtype float64`. Now I fear, that if that is not the case there can either be an error or even a misinterpretation in this line:
`haz_coord_view = haz_coord.view(dtype='float64,float64').reshape(-1)`
[Link to code](https://github.com/CLIMADA-project/climada_python/blob/abfbe1b7124a8e49bda565261c9a358ce25d78ef/climada/entity/exposures/base.py#L312)

My solution would be make sure, that haz_coord is actually float64 beforehand, with:
`haz_coord = haz_coord.astype('float64')`

But I would appreciate if somebody who looked into that code recently could quickly confirm, if that is a good idea or not.

PS:
I just experienced the following error because `hazard.centroids.coord.dtype` was `float32`:
`File "C:\shortpaths\GitHub\climada_python\climada\entity\exposures\base.py", line 312, in assign_centroids
 haz_coord_view = haz_coord.view(dtype='float64,float64').reshape(-1)
ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.`

ThomasRoosli:
Fixes #175 
I changed a test so the hazard.centroids.lat and lon are in float32 which produced an error and then fixed the error as suggested in the issue.

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -301,8 +301,8 @@ class Exposures():
             assigned[(x_i < 0) | (x_i >= hazard.centroids.meta['width'])] = -1
             assigned[(y_i < 0) | (y_i >= hazard.centroids.meta['height'])] = -1
         else:
-            coord = np.stack([self.gdf.latitude.values, self.gdf.longitude.values], axis=1)
-            haz_coord = hazard.centroids.coord
+            coord = np.stack([self.gdf.latitude.values, self.gdf.longitude.values], axis=1).astype('float64')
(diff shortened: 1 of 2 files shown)


tovogt (approved):
Great, very quick and on point!

I just added a test to make sure that both float32 and float64 input data are tested.

PR #187, opened 2021-03-25
Feature/bushfire

samluethi:
full update of the bushfire module. Essentially the same request as two months ago.
Adjusted tests and alpha_shape so that it should pass the tests now.


samluethi:
See pull request #188

PR #201, opened 2021-04-08
Feature/country numeric2iso

manniepmkam:
I added a function `country_iso_numeric2alpha()`  in `util/coordinates.py` that is able to transform the iso 3166-1 numeric-3 to alpha-3 code (e.g. 578 to 'NOR'). 

2 more test functions for testing the `country_iso_numeric2alpha()` and `country_iso_alpha2numeric()`


climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
...
+    
+    Parameters
+    ----------
+    isos_numeric : int or list of int
+        ISO numeric-3 codes of countries (or single code).
+        
+    Returns
+    -------
+    isos : str or list of str
+        Will only return a list if the input is a list.
```
chahank, line 958: What are the isos returned? This is needed in the description.

climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
         nums.append(num)
     return nums[0] if return_int else nums
 
+def country_iso_numeric2alpha(isos_numeric):
+    """Convert ISO 3166-1 numeric-3codes to alpha-3
+    
+    Parameters
+    ----------
+    isos_numeric : int or list of int
```
chahank, line 952: why isos? Isn't it ISO?

climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
...
+    isos : str or list of str
+        Will only return a list if the input is a list.
+    """
+    return_str = isinstance(isos_numeric, int)
+    isos_numeric = [isos_numeric] if return_str else isos_numeric
+    old_iso = {
+        0: '',  # Ocean or fill_value
+        530: "ANT", # Netherlands Antilles: split up since 2010
+        891: "SCG", # Serbia and Montenegro: split up since 2006
+    }
```
chahank, line 966: There are several more, such as the URSS, Yougoslavia,... please update.

climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
...
+    isos = []
+    for iso_numeric in isos_numeric:
```
chahank, line 968: It is not very readable to use singular and plural in codes. Rather use the type. E.g. `for iso in iso_list`

climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
...
+        if iso_numeric in old_iso:
+            iso = old_iso[iso_numeric]
+        else:
+            iso = iso_cntry.get(iso_numeric).alpha3
```
chahank, line 972: A `try...except...` block might be better here, because if an iso3 that is neither old nor existing, the error message will be unclear as it will come from ` iso = old_iso[iso_numeric]`.

climada/util/coordinates.py
```diff
@@ -944,6 +944,35 @@ def country_iso_alpha2numeric(isos):
...
+        Will only return a list if the input is a list.
+    """
+    return_str = isinstance(isos_numeric, int)
+    isos_numeric = [isos_numeric] if return_str else isos_numeric
+    old_iso = {
+        0: '',  # Ocean or fill_value
+        530: "ANT", # Netherlands Antilles: split up since 2010
+        891: "SCG", # Serbia and Montenegro: split up since 2006
+    }
+    isos = []
```
chahank, line 967: Also here, use rather type `list` in name than plurals. This is quickly confusing otherwise.

tovogt:
@chahank Note that there already exists a function that does the reverse conversion `country_iso_alpha2numeric`. The code provided by this PR is just the exact analogous. All your remarks also apply the reverse conversion (or its current implementation). I suggest to have a single collection of obsolete ISO codes that we can use in both functions. Furthermore, I suggest to improve also the old `country_iso_alpha2numeric` if you decide to improve the new `country_iso_numeric2alpha`.

climada/util/coordinates.py
manniepmkam: Description has been added

climada/util/coordinates.py
manniepmkam: changed to `iso_numeric`.

climada/util/coordinates.py
manniepmkam: These has been changed too. Also the reverse functions.

climada/util/coordinates.py
manniepmkam: Name changed to `iso_list`.

climada/util/coordinates.py
manniepmkam: implemented the `try...except...`. Please have a look if that's the correct use.

climada/util/coordinates.py
manniepmkam: This list should have all the old iso cold? Also, the `old_iso` dictionary will be quite long, maybe worth to put them in the constant files?

tovogt:
If we want to go beyond the official ISO-3166-1 countries, we should consider using [pycountry](https://pypi.org/project/pycountry/) instead of [iso3166](https://github.com/deactivated/python-iso3166/issues/23). This would also provide access to the historic country codes in ISO-3166-3 two of which we currently handle manually. The package `pycountry` is based on the well-established Debian package [iso-codes](https://salsa.debian.org/iso-codes-team/iso-codes).

chahank:
I agree that having a single collection of obsolete ISO codes would be great. Whether we do it manually or use an external package depends on @emanuel-schmid I think. The disadvantage to use the pycountry is that this will be something to be maintained to some extent. The advantage is that it is much more complete and it is multi-language which might be an asset for CLIMADA to be used outside of the anglo-sphere and the direct research community.

tovogt:
Using `pycountry` definitely requires less maintenance than sticking to `iso3166` and maintaining our own database of ISO-3166-3.

climada/util/coordinates.py
chahank: Yes, this list should be correct. As @tovogt argued below, it is a good idea to either put them in the constant files, propose a pull request to include old iso3 to the pacakge `iso3166` or use `pycountry` .

climada/util/coordinates.py
```diff
@@ -915,34 +915,73 @@ def region2isos(regions):
...
+    iso_num = []
+    for iso in iso_list:
+        try:
+            if iso in old_iso:
+                num = old_iso[iso]
+            else:
+                num = int(iso_cntry.get(iso).numeric)
+            iso_num.append(num)
+        except ValueError as ver:
+            LOGGER.error('Unknown country ISO: %s', iso)
```
chahank, line 948: This is not needed.
```suggestion
```
The raise KeyError already provides the error message.

climada/util/coordinates.py
```diff
@@ -915,34 +915,73 @@ def region2isos(regions):
...
+    """Convert ISO 3166-1 numeric-3codes to alpha-3
+    
+    Parameters
+    ----------
+    iso_numeric : int or list of int
+        ISO numeric-3 codes of countries (or single code).
+        
+    Returns
+    -------
+    isos : str or list of str
```
chahank, line 962:
```suggestion
    iso_list : str or list of str
```

climada/util/coordinates.py
```diff
@@ -915,34 +915,73 @@ def region2isos(regions):
...
+    iso_list = []
+    for iso_num in iso_num_list:  
+        try:
+            if iso_num in old_iso:
+                iso = old_iso[iso_num]
+            else:
+                iso = iso_cntry.get(iso_num).alpha3
+            iso_list.append(iso)
+        except ValueError as ver:
+            LOGGER.error('Unknown country ISO: %s', iso)
```
chahank, line 982:
```suggestion
```

tovogt:
Here is how we could get rid of `iso3166` in favor of `pycountry`. This would also merge the functionality of this PR into one util function: #205

PR #202, opened 2021-04-08
Coerce user-supplied Centroid lat-lon data to np.arrays

ChrisFairless:
This is a small edit to the Centroid class to coerce `Centroid.lat` and `Centroid.lon` to `np.array` objects when they're directly supplied by the user.

Previously a `Centroid` class with user-supplied lat and lon arrays (via `Centroid.set_lat_lon`) preserves the classes of whatever lat/lon data is passed to it. This is fine in most cases, but if it's given a `pandas.Series` (for example, when setting centroids from the lat/lon columns of an `Exposure.gdf`), then the `TropCyclone.set_from_tracks()` method falls  over because it's unable to use `np.nonzero()` on a Series.

Code to reproduce the bug:
```
from climada.hazard import TropCyclone, TCTracks, Centroids
from climada.entity.exposures import LitPop

exp_litpop = LitPop()
exp_litpop.set_country('Puerto Rico', res_arcsec=300)

cent = Centroids()
cent.set_lat_lon(exp_litpop.gdf.latitude, exp_litpop.gdf.longitude)

track = TCTracks()
track.read_ibtracs_netcdf(storm_id='2017260N12310')

haz = TropCyclone()
haz.set_from_tracks(track, cent) # Error here
```

### DIFF
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -325,7 +325,7 @@ class Centroids():
             CRS. Default: DEF_CRS
         """
         self.__init__()
-        self.lat, self.lon, self.geometry = np.array(lat), np.array(lon), gpd.GeoSeries(crs=crs)
+        self.lat, self.lon, self.geometry = np.asarray(lat), np.asarray(lon), gpd.GeoSeries(crs=crs)
 
     def set_raster_file(self, file_name, band=[1], src_crs=None, window=False,
                         geometry=False, dst_crs=False, transform=None, width=None,
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -325,7 +325,7 @@ class Centroids():
             CRS. Default: DEF_CRS
         """
         self.__init__()
-        self.lat, self.lon, self.geometry = lat, lon, gpd.GeoSeries(crs=crs)
+        self.lat, self.lon, self.geometry = np.array(lat), np.array(lon), gpd.GeoSeries(crs=crs)
 
     def set_raster_file(self, file_name, band=[1], src_crs=None, window=False,
                         geometry=False, dst_crs=False, transform=None, width=None,


chahank:
Looks fine to me. The only potential issue I could see is if somewhere `lat1 + lat2` is used, which would behave differently for lists (appending) and numpy arrays (element-wise addition). Do all the tests pass?

climada/hazard/centroids/centr.py
```diff
@@ -325,7 +325,7 @@ def set_lat_lon(self, lat, lon, crs=DEF_CRS):
             CRS. Default: DEF_CRS
         """
         self.__init__()
-        self.lat, self.lon, self.geometry = lat, lon, gpd.GeoSeries(crs=crs)
+        self.lat, self.lon, self.geometry = np.array(lat), np.array(lon), gpd.GeoSeries(crs=crs)
```
tovogt, line 328: Note that `np.array(lat)` creates a copy if `lat` is an array already. Is this what we want? An alternative would be `np.asarray`.

ChrisFairless: that probably makes more sense!

ChrisFairless, line 328:
```suggestion
        self.lat, self.lon, self.geometry = np.asarray(lat), np.asarray(lon), gpd.GeoSeries(crs=crs)
```

ChrisFairless: updated

ChrisFairless:
> Looks fine to me. The only potential issue I could see is if somewhere `lat1 + lat2` is used, which would behave differently for lists (appending) and numpy arrays (element-wise addition). Do all the tests pass?

The tests all pass!
@chahank do you know where lat1 + lat2 might be? It sounds like it might be risky allowing the lat/lon attributes to take any type in this case, and maybe we want to spend some time (later) forcing consistency.

chahank:
I have no idea whether `lat1+lat2` exists somewhere in the code base. I agree that we should enforce type consistency here. I suppose that if all tests are passing, there is no problem to merge this.

ChrisFairless:
Who is in charge of merging a request? Is it me or a reviewer?

I ought to know this ... I wrote the guide.

chahank:
We have no fixed guidelines for that :D. You can do it!

ChrisFairless:
Aw yiss. Here we go.

PR #204, opened 2021-04-13
Update antimeridian handling util.plot.geo_scatter_from_array

chahank:
This pull request adds the antimeridian handling to the `util.plot.geo_scatter_from_array` plotting function. The code is copied from `util.geo_bin_from_array`.

As a consequence, the two above functions are identical up to the following lines:
In `geo_bin_from_array` we use `plt.hexbin()`:
```
        if 'gridsize' not in kwargs:
            kwargs['gridsize'] = min(int(array_im.size / 2), MAX_BINS)
        hex_bin = axis.hexbin(coord[:, 1], coord[:, 0], C=array_im,
                              transform=proj, **kwargs)
```
In `geo_scatter_from_array` we use `plt.scatter()`:
```
        scatter = axis.scatter(coord[:, 1], coord[:, 0], c=array_im,
                               transform=proj, **kwargs)
```

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -55,10 +55,10 @@ MAX_BINS = 2000
 """Maximum number of bins in geo_bin_from_array"""
 
 
-def geo_bin_from_array(array_sub, geo_coord, var_name, title, pop_name=True,
-                       buffer=BUFFER, extend='neither',
-                       proj=ccrs.PlateCarree(), axes=None, figsize=(9, 13),
-                       **kwargs):
+def geo_bin_from_array(array_sub, geo_coord, var_name, title,
+                       pop_name=True, buffer=BUFFER, extend='neither',
+                       proj=ccrs.PlateCarree(), shapes=True, axes=None,
+                       figsize=(9, 13), **kwargs):
     """Plot array values binned over input coordinates.
 
     Parameters
@@ -84,6 +84,9 @@ def geo_bin_from_array(array_sub, geo_coord, var_name, title, pop_name=True,
         [ 'neither' | 'both' | 'min' | 'max' ], by default 'neither'
     proj : ccrs, optional
         coordinate reference system of the given data, by default ccrs.PlateCarree()
+    shapes : bool, optional
+        Overlay Earth's countries coastlines to matplotlib.pyplot axis.
+        The default is True
     axes : Axes or ndarray(Axes), optional
         by default None
     figsize : tuple, optional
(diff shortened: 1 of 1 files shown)


tovogt:
If the functions are so similar, I think it's save to merge them into one function. The API isn't changed though: There is a new non-exposed (internal) function that is called from both `geo_scatter_from_array` and `geo_bin_from_array`. This reduces the amount of code while keeping the API perfectly untouched.

chahank:
Great! @zeliest actually did the same in a separate branch. Let us keep the best version. @zeliest , could you please have a look here too?

climada/util/plot.py
```diff
@@ -173,19 +127,20 @@ def geo_scatter_from_array(array_sub, geo_coord, var_name, title,
...
-        add names of the populated places, by default True
+        add names of the populated places, by default False
     buffer : float, optional
         border to add to coordinates, by default BUFFER
     extend : str, optional
         extend border colorbar with arrows.
         [ 'neither' | 'both' | 'min' | 'max' ], by default 'neither'
     proj : ccrs, optional
-        coordinate reference system used in coordinates, by default ccrs.PlateCarree()
+        coordinate reference system if the given data, by default ccrs.PlateCarree()
```
chahank, line 140:
```suggestion
        coordinate reference system of the given data, by default ccrs.PlateCarree()
```

climada/util/plot.py
```diff
@@ -210,36 +179,56 @@ def geo_scatter_from_array(array_sub, geo_coord, var_name, title,
...
         kwargs['cmap'] = 'Wistia'
+
     if axes is None:
-        _, axes = make_map(num_im, proj=proj, figsize=figsize)
-    axes_iter = axes
+        proj_plot = proj
+        if isinstance(proj, ccrs.PlateCarree):
+            # use different projections for plot and data to shift the central lon in the plot
+            xmin, xmax = u_coord.lon_bounds(np.concatenate([c[:, 1] for c in list_coord]))
+            proj_plot = ccrs.PlateCarree(central_longitude=0.5 * (xmin + xmax))
```
chahank, line 188: Would this code work if instead of `ccrs.PlateCarree` one uses another projection from `ccrs` that uses lat/lon?

tovogt: There is no other projection that has lat/lon as output type because that is the unique defining property of the "Equidistant Cylindrical (Plate Carrée)" projection: https://proj.org/operations/projections/eqc.html 

Note that, according to the docstring, the `proj` attribute specifies the CRS of the data. There is currently no way to specify the projection of the axes. To achieve that, the user has to provide an own axes object using the `axes` parameter.

What you might have intended with your question: If the user has data for Fiji in some CRS other than lat/lon (like UTM), and the user doesn't specify any `axes`, we will probably end up with a plot that doesn't have the "nice" central longitude. However, this is at the same time very hypothetical AND hard to deal with. I strongly suggest to leave this for a discussion in the CRS task force.

zeliest:
> Great! @zeliest actually did the same in a separate branch. Let us keep the best version. @zeliest , could you please have a look here too?
> 
> I also update in a commit the function `geo_scatter_categorical` to use the new `_plot_scattered_data` function.

Yes it was pretty much the exact same thing that I did, here is the branch that I am working on: https://github.com/CLIMADA-project/climada_python/pull/207. I merged this branch into it to only have your version of the scatter/hexbin plot.

tovogt:
Thanks for your work @zeliest !

Since #207 includes this PR, I propose to merge this one and then have a look at the additional changes introduced by #207

tovogt (approved):

PR #205, opened 2021-04-13
Use pycountry instead of iso3166 for country code lookup

tovogt:
Currently, CLIMADA uses the Python package `iso3166` to look up country codes according to the ISO 3166 standard. Unfortunately, that package doesn't cover historical countries that are covered in ISO 3166-3, but only the countries in ISO 3166-1. Furthermore, the package `iso3166` maintains its own database of country codes instead of building upon already existing and well-maintained databases of country codes like Debian's [iso-codes](https://salsa.debian.org/iso-codes-team/iso-codes).

Furtunately, there is an alternative to `iso3166` that is build on Debian's iso-codes: `pycountry`. That package is well-maintained and covers ISO 3166-1 as well as ISO 3166-3 (historic countries), so we don't have to manually deal with those (see #201).

This PR adds the requirement `pycountry` and removes `iso3166`. It makes sure that there are no places in CLIMADA left where `iso3166` is used and provides a util function `country_to_iso` that is able to replace all future use cases where users might want to have a functionality that was formerly provided by `iso3166`.

### DIFF
--- a/climada/util/constants.py
+++ b/climada/util/constants.py
@@ -162,8 +162,33 @@ ISIMIP_NATID_TO_ISO = [
     'TUN', 'TUR', 'TUV', 'TWN', 'TZA', 'UGA', 'UKR', 'URY', 'USA', 'UZB', 'VCT',
     'VEN', 'VGB', 'VIR', 'VNM', 'VUT', 'WLF', 'WSM', 'YEM', 'ZAF', 'ZMB', 'ZWE',
 ]
-"""ISO3166 alpha-3 codes of countries used in ISIMIP_GPWV3_NATID_150AS"""
+"""ISO 3166 alpha-3 codes of countries used in ISIMIP_GPWV3_NATID_150AS"""
 
+NONISO_REGIONS = [
+    # Dummy region for numeric 0 (or empty string), sometimes used for oceans
+    dict(name="", alpha_2="", alpha_3="", numeric="000"),
+    dict(name="Akrotiri", alpha_2="XA", alpha_3="XXA", numeric="901"),
+    dict(name="Baikonur", alpha_2="XB", alpha_3="XXB", numeric="902"),
+    dict(name="Bajo Nuevo Bank", alpha_2="XJ", alpha_3="XXJ", numeric="903"),
+    dict(name="Clipperton I.", alpha_2="XC", alpha_3="XXC", numeric="904"),
+    dict(name="Coral Sea Is.", alpha_2="XO", alpha_3="XXO", numeric="905"),
+    dict(name="Cyprus U.N. Buffer Zone", alpha_2="XU", alpha_3="XXU", numeric="906"),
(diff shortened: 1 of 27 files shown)


climada/util/constants.py
```diff
@@ -162,8 +162,33 @@
...
+    dict(name="Scarborough Reef", alpha_2="XS", alpha_3="XXS", numeric="912"),
+    dict(name="Serranilla Bank", alpha_2="XR", alpha_3="XXR", numeric="913"),
+    dict(name="Siachen Glacier", alpha_2="XH", alpha_3="XXH", numeric="914"),
+    dict(name="Somaliland", alpha_2="XM", alpha_3="XXM", numeric="915"),
+    dict(name="Spratly Is.", alpha_2="XP", alpha_3="XXP", numeric="916"),
+    dict(name="USNB Guantanamo Bay", alpha_2="XG", alpha_3="XXG", numeric="917"),
+]
+"""Geopolitical areas that are not listed in the ISO 3166 standard, but might be relevant when
+working, e.g. with Natural Earth shape files. The alpha-2, alpha-3 and numeric representations are
+unofficial and for internal use only."""
```
tovogt, line 191: These used to be listed in `util/coordinates.py`.

tovogt, line 191: These used to be listed in `util/coordinates.py`.

climada/util/coordinates.py
```diff
@@ -987,37 +1024,16 @@ def country_iso2natid(isos):
...
-    "Kosovo": 983,  # Same as iso3166 package
-    "N. Cyprus": 910,
-    "Norway": 578,  # Bug in Natural Earth
-    "Scarborough Reef": 912,
-    "Serranilla Bank": 913,
-    "Siachen Glacier": 914,
-    "Somaliland": 915,
-    "Spratly Is.": 916,
-    "USNB Guantanamo Bay": 917,
-}
```
tovogt, line 1012: Moved to `constants.py`.

chahank:
I will read this pull request in detail later, but one comment/question: shall we introduce a more consistent import of the util functions? In some previous discussion there was the suggestion to always import `util` modules as `from climada.util import files_handler as u_fh` instead of importing module functions directly.

tovogt:
I remember that we talked about that and I still think it makes sense, so I changed every import from util.coordinates accordingly. However, is this an official guideline already?

chahank:
It is 'official' in the sense that it is written in the developers guide yes.

tovogt:
For testing purposes, I added `pycountry` as a requirement to the develop branch and everything works as expected. Also, all tests pass with the changes introduced by this PR (see https://ied-wcr-jenkins.ethz.ch/job/climada_branches/job/feature%252Fpycountry_test/).

Since pycountry is well-maintained and doesn't introduce any extra-dependencies on its side and it replaces another CLIMADA dependency (iso3166), so that it doesn't increase the number of dependencies for CLIMADA, I think it's save to stick to this.

Furthermore, I merged #201 so that this PR can build on that.

PR #206, opened 2021-04-13
Improve speed of subsetting an Impact by centroids

ChrisFairless:
The existing method for subsetting an Impact by centroid was very slow when working with Impacts and selections that had large numbers of points.

The culprit was in `Impact._selected_exposures_idx()` which constructed a list of the indices of centroids to select using a list comprehension. I've changed this: it now converts all coordinates to pandas data frames, joins them, and records the indices of common values. It's a lot faster.

There may be an even better way, but this takes large calculations down from ~40 seconds to < 1 sec which is all I need.

Tests all pass 👍

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1093,6 +1093,7 @@ class Impact():
         else:
             coords_merged = pd.DataFrame(self.coord_exp).merge(pd.DataFrame(coord_exp), how='left', indicator=True)
             sel_exp = np.where(coords_merged['_merge'] == 'both')[0]
+            sel_exp = list(sel_exp)
             if not sel_exp:
                 LOGGER.warning("No exposure coordinates matches the selection.")
 
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1091,11 +1091,8 @@ class Impact():
         if coord_exp is None:
             sel_exp = []
         else:
-            sel_exp =  [
-                j
-                for j, coord in enumerate(self.coord_exp)
-                if coord in coord_exp
-                ]
+            coords_merged = pd.DataFrame(self.coord_exp).merge(pd.DataFrame(coord_exp), how='left', indicator=True)
+            sel_exp = np.where(coords_merged['_merge'] == 'both')[0]
             if not sel_exp:
                 LOGGER.warning("No exposure coordinates matches the selection.")
 
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1091,9 +1091,8 @@ class Impact():
         if coord_exp is None:
             sel_exp = []
         else:
-            coords_merged = pd.DataFrame(self.coord_exp).merge(pd.DataFrame(coord_exp), how='left', indicator=True)
-            sel_exp = np.where(coords_merged['_merge'] == 'both')[0]
-            sel_exp = list(sel_exp)
+            assigned_idx = u_coord.assign_coordinates(self.coord_exp, coord_exp, threshold=0)
+            sel_exp = list((assigned_idx >= 0).nonzero()[0])
             if not sel_exp:
                 LOGGER.warning("No exposure coordinates matches the selection.")
 
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1007,10 +1007,10 @@ class Impact():
         nb_exp = len(self.coord_exp)
 
         if self.imp_mat.shape != (nb_events, nb_exp):
-            raise ValueError("The impact matrix is missing or incomplete. " +
-                             "The eai_exp and aai_agg cannot be computed. " +
-                             "Please recompute impact.calc() with save_mat=True" +
-                             " before using impact.select()")
+            raise ValueError("The impact matrix is missing or incomplete. "
+                             "The eai_exp and aai_agg cannot be computed. "
+                             "Please recompute impact.calc() with save_mat=True "
+                             "before using impact.select()")
 
         if nb_events == nb_exp:
             LOGGER.warning("The number of events is equal to the number of "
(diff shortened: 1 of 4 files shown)


tovogt:
Nice! But I found that it's even faster (factor 1.5) to apply the logic in `Exposures.assign_centroids` also to this case. So, I decided to create a util function from `Exposures.assign_centroids` and use that for the purpose of speeding up the impact selection. It's not only about the speed-up, but I simply like reusing code in unexpected ways ;-)

climada/util/coordinates.py
```diff
@@ -891,6 +892,70 @@ def mapping_grid2flattened(col, row, matrix_shape):
...
+        assigned_idx[sort_assign_idx[exact_assign_idx]] = exact_assign_idx
+
+        # assign remaining coordsinates to their geographically nearest neighbor
+        if threshold > 0 and exact_assign_idx.size != coords_view.size:
+            not_assigned_idx_mask = (assigned_idx == -1)
+            assigned_idx[not_assigned_idx_mask] = u_interp.interpol_index(
+                coords_to_assign, coords[not_assigned_idx_mask],
+                method=method, distance=distance, threshold=threshold)
+    return assigned_idx
+
```
tovogt, line 958: This is more or less copy-pasted from the former `Exposure.assign_centroids` function. I applied some variable renaming, that's all.

climada/util/coordinates.py
```diff
@@ -891,6 +892,70 @@ def mapping_grid2flattened(col, row, matrix_shape):
...
 
+def assign_coordinates(coords, coords_to_assign, method="NN", distance="haversine", threshold=100):
+    """To each coordinate in `coords`, assign a matching coordinate in `coords_to_assign`
+
+    If there is no exact match for some entry, an attempt is made to assign the geographically
+    nearest neighbor. If the distance to the nearest neighbor exceeds `threshold`, the index `-1`
+    is assigned.
+
+    Currently, the nearest neighbor matching works with lat/lon coordinates only. However, you can
+    disable nearest neighbor matching by setting `threshold` to 0. In this case, only exactly
```
chahank, line 903:
```suggestion
    disable nearest neighbor matching by setting `threshold` to 0, in which case, only exactly
```

climada/util/coordinates.py
```diff
@@ -891,6 +892,70 @@ def mapping_grid2flattened(col, row, matrix_shape):
...
+        system as `coords`.
+    method : str, optional
+        Interpolation method to use for non-exact matching. Currently, "NN" (nearest neighbor)
+        is the only supported value, see `climada.util.interpolation.interpol_index`.
+    distance : str, optional
+        Distance to use for non-exact matching. Possible values are "haversine" and "approx", see
+        `climada.util.interpolation.interpol_index`. Default: "haversine"
+    threshold : float, optional
+        If the distance to the nearest neighbor exceeds `threshold`, the index `-1` is assigned.
+        Set `threshold` to 0, to disable nearest neighbor matching. Default: 100
```
chahank, line 923:
```suggestion
        Set `threshold` to 0 to disable nearest neighbor matching. Default: 100
```

climada/util/coordinates.py
```diff
@@ -891,6 +892,70 @@ def mapping_grid2flattened(col, row, matrix_shape):
...
+
+    Returns
+    -------
+    assigned_idx : np.array of size equal to the number of rows in `coords`
+        Index into `coords_to_assign`. Note that the value `-1` is used to indicate that no
```
chahank, line 928: Should we maybe use `np.Nan` instead of `-1` then?

climada/util/test/test_coordinates.py
```diff
@@ -295,7 +296,27 @@ def test_mapping_grid2flattened(self):
...
             mapping_grid2flattened(4, 7, matrix.shape)
-        
+
+    def test_assign_coordinates(self):
+        """Test assign_coordinates function"""
+        coords = np.array([(0.2, 2), (0, 0), (0, 2), (2.1, 3), (1, 1), (-1, 1)])
+        coords_to_assign = np.array([(2.1, 3), (0, 0), (0, 2), (0.9, 1.0)])
+        expected_results = [
+            # test with different thresholds
+            (100, [2, 1, 2, 0, 3, -1]),
```
chahank, line 306: I am a bit confused. Why is the distance from `(-1,1)` to `(0,0)` larger than `100` and `20`?

climada/util/coordinates.py
```diff
@@ -891,6 +892,70 @@ def mapping_grid2flattened(col, row, matrix_shape):
         raise ValueError
     return row * matrix_shape[1] + col
 
+def assign_coordinates(coords, coords_to_assign, method="NN", distance="haversine", threshold=100):
```
chahank, line 895: It is starting to getting an old question, sorry for repeating it again: do we handle the antimeridian correctly here or is it up to the code/user to first transform the data?

climada/util/coordinates.py
tovogt: `np.nan` is not a valid value for an integer array. It's only available for float data types.

climada/util/test/test_coordinates.py
tovogt: Because, we are in a lat-lon-coordinate system and the threshold distance is in km.

climada/engine/impact.py
```diff
@@ -1091,11 +1091,8 @@ def _selected_exposures_idx(self, coord_exp):
         if coord_exp is None:
             sel_exp = []
         else:
-            sel_exp =  [
-                j
-                for j, coord in enumerate(self.coord_exp)
-                if coord in coord_exp
-                ]
+            assigned_idx = u_coord.assign_coordinates(self.coord_exp, coord_exp, threshold=0)
+            sel_exp = list((assigned_idx >= 0).nonzero()[0])
```
chahank, line 1095: Is it really necessary to convert to a list? Keeping a numpy array would not work?

climada/entity/exposures/base.py
```diff
@@ -285,15 +284,19 @@ def assign_centroids(self, hazard, method='NN', distance='haversine',
...
+            Hazard to match (with raster or vector centroids).
+        method : str, optional
+            Interpolation method to use in case of vector centroids. Currently, "NN" (nearest
+            neighbor) is the only supported value, see `climada.util.interpolation.interpol_index`.
+        distance : str, optional
+            Distance to use in case of vector centroids. Possible values are "haversine" and
+            "approx", see `climada.util.interpolation.interpol_index`. Default: "haversine"
+        threshold : float
+            If the distance to the nearest neighbor exceeds `threshold`, the index `-1` is
+            assigned. Set `threshold` to 0, to disable nearest neighbor matching. Default: 100
```
chahank, line 299:
```suggestion
            assigned. Set `threshold` to 0, to disable nearest neighbor matching. Default: 100km
```

climada/util/test/test_coordinates.py
chahank: I suggested to add this to the defaults, i.e., 100km (see suggestions)

chahank:
Great improvement, thanks @tovogt !

climada/engine/impact.py
tovogt: I meant to stick to the previous behavior, but of course you are right. Fixed in latest commit.

climada/util/coordinates.py
tovogt: The exact matching will only work if all longitudinal values are in the same longitudinal range. However, the "haversine" distance is able to correctly find nearest neighbors independent of longitudinal range. I added a unit test for that and added a note in the docstring.

ChrisFairless:
This is brilliant, thanks @tovogt. I'm happy for this to be merged.

PR #207, opened 2021-04-15
Plots improvements

zeliest:


climada/util/plot.py
```diff
@@ -519,11 +519,11 @@ def add_populated_places(axis, extent, proj=ccrs.PlateCarree()):
         if ext_trans[2][0] < point.x <= ext_trans[0][0]:
             if ext_trans[0][1] < point.y <= ext_trans[1][1]:
                 place_name = _fix_place_names(rec.attributes['name'])
-                axis.plot(point.x, point.y, 'ko', markersize=7,
+                axis.plot(point.x, point.y, color='navy', marker='o', markersize=7,
                           transform=ccrs.PlateCarree(), markerfacecolor='None')
                 axis.text(point.x, point.y, place_name,
                           horizontalalignment='right', verticalalignment='bottom',
-                          transform=ccrs.PlateCarree(), fontsize=14)
+                          transform=ccrs.PlateCarree(), fontsize=14, color='navy')
```
zeliest, line 528: this is a very small detail, but I think it is easier to visualise like this

tovogt (approved):
Everything fine from my side.

zeliest:
There are still a few things that I wanted to work on in the plots, should I already merge this?

chahank:
If the future changes you plan do not interfere with the already present changes in this pull request you can go ahead and already merge. Then we have smaller pull requests. Thanks!

An additional  point that would be very useful: could you please add tests for the plots? It is enough to just test whether the plot is not throwing an error. See for instance the example in https://github.com/CLIMADA-project/climada_python/blob/main/climada/util/test/test_plot.py . Then there would also be a test for example for the added `crs_epsg` argument.

PR #208, opened 2021-04-15
Avoid some errors by coercing some Impact.select() input vars

ChrisFairless:
`Impact.select()` takes lists as input for its dates, event_ids and event_names parameters. If you instead pass a numpy array, the method fails at line 1031:
```
   if (dates, event_ids, event_names) != (None, None, None):
```
with a ValueError:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

It's not obvious to the user why this is. So I've added lines at the start of this method to coerce the input to a list. This will allow a bit more flexibility with inputs and when it fails it will be more obvious why.

Questions:
- Do we want to be even more explicit with type checking (e.g. make this a try/except)?
- Could there be any unexpected consequences using `tolist()`here?

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1021,6 +1021,13 @@ class Impact():
                            "method.")
             return None
 
+        if not isinstance(dates, list):
+            dates = list(dates)
+        if not isinstance(event_ids, list):
+            event_ids = list(event_ids)
+        if not isinstance(event_names, list):
+            event_names = list(event_names)
+
         if (dates, event_ids, event_names) != (None, None, None):
             sel_ev = self._selected_events_idx(event_ids, event_names, dates, nb_events)
         else:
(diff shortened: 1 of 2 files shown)


tovogt:
Let me quote from PEP 8: "Comparisons to singletons like None should always be done with is or is not, never the equality operators." (https://www.python.org/dev/peps/pep-0008/)

Here is an example:
```
class Test:
    def __eq__(self, other):
        return True

t = Test()
print(t is not None)
# output: True
print(t != None)
# output: False
```

Therefore, the line `if (dates, event_ids, event_names) != (None, None, None):` is problematic to begin with. However, there is more to say about that:

* There is no reason why we should restrict to arrays, tuples, sets, lists, Series or whatever. If we use NumPy functions everywhere, NumPy is able to do all the necessary conversions.
* Currently, if no events/exposures match the selection criteria, no filtering is applied at all. That's a serious bug, in my opinion!

I pushed a fix for these to points.

ChrisFairless:
Just had a look through your changes. They look great, and it's a better fix.
Happy to merge.

Issue #199, opened 2021-04-07
Impact Tutorial Failure


emanuel-schmid:
Error message:

notebook climada_engine_Impact.ipynb cell 26 failed with <class 'ValueError'> operands could not be broadcast together with shapes (9,) (3,)

From the non-markdown cells that's number 4.
Stack trace:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-fa45310dc440> in <module>
      6 print('num tracks hist:', ibtracks_na.size)
      7 #Add randomly generated tracks using the calc_random_walk method (2 per historical track)
----> 8 ibtracks_na.calc_random_walk(ens_size=1)
      9 print('num tracks hist+syn:', ibtracks_na.size)
     10 ibtracks_na.equal_timestep(0.5)  # Interpolation to make the track smooth c.f.

~\www\climada_python\climada\hazard\tc_tracks.py in calc_random_walk(self, **kwargs)
   1085         if kwargs.get('ens_size'):
   1086             kwargs['nb_synth_tracks'] = kwargs.pop('ens_size')
-> 1087         return self.calc_perturbed_trajectories(**kwargs)
   1088 
   1089     def calc_perturbed_trajectories(self, **kwargs):

~\www\climada_python\climada\hazard\tc_tracks.py in calc_perturbed_trajectories(self, **kwargs)
   1089     def calc_perturbed_trajectories(self, **kwargs):
   1090         """See function in `climada.hazard.tc_tracks_synth`."""
-> 1091         climada.hazard.tc_tracks_synth.calc_perturbed_trajectories(self, **kwargs)
   1092 
   1093     @property

~\www\climada_python\climada\hazard\tc_tracks_synth.py in calc_perturbed_trajectories(tracks, nb_synth_tracks, max_shift_ini, max_dspeed_rel, max_ddirection, autocorr_dspeed, autocorr_ddirection, seed, decay)
    111 
    112     # ensure tracks have constant time steps
--> 113     time_step_h = np.unique([np.unique(x['time_step']) for x in tracks.data])
    114     if not np.allclose(time_step_h, time_step_h[0]):
    115         LOGGER.error('Tracks have different temporal resolution. '

<__array_function__ internals> in unique(*args, **kwargs)

~\anaconda3\envs\climada_v2\lib\site-packages\numpy\lib\arraysetops.py in unique(ar, return_index, return_inverse, return_counts, axis)
[...]

PR #211, opened 2021-04-17
Bug fix: impacts tutorial (TC synthetic tracks)

bguillod:
This PR fixes issue #199.

However, note that the notebook `climada_engine_Impact.ipynb` should also be edited as follows in cell 4, since

1. `calc_random_walk` has been replaced with `calc_perturbed_trajectories`.
2. This method requires equal timesteps, hence `equal_timestep` must be called before and not after generating synthetic tracks.

```
from climada.hazard import TCTracks, TropCyclone, Centroids

# Load histrocial tropical cyclone tracks from ibtracs over the North Atlantic basin between 2010-2012
ibtracks_na = TCTracks()
ibtracks_na.read_ibtracs_netcdf(provider='usa', basin='NA', year_range=(2010, 2012), correct_pres=True)
print('num tracks hist:', ibtracks_na.size)

ibtracks_na.equal_timestep(0.5)  # Interpolation to make the track smooth and to allow applying calc_perturbed_trajectories
# Add randomly generated tracks using the calc_perturbed_trajectories method (1 per historical track)
ibtracks_na.calc_perturbed_trajectories(nb_synth_tracks=1)
print('num tracks hist+syn:', ibtracks_na.size)
```

I did not edit this because I'm not sure how to edit the notebook properly, so I leave that up to @emanuel-schmid to do so (feel free to do so within this PR).

@emanuel-schmid Please also note that I've run a string search for `calc_random_walk` on the climada code and found a usage in `./climada/test/test_tc.py`. However the whole function in which the call for `calc_random_walk` occurs is commented out. Is this file event needed and/or used anywhere? It might be worth checking.

### DIFF
--- a/climada/hazard/tc_tracks_synth.py
+++ b/climada/hazard/tc_tracks_synth.py
@@ -110,7 +110,7 @@ def calc_perturbed_trajectories(tracks,
         np.random.seed(seed)
 
     # ensure tracks have constant time steps
-    time_step_h = np.unique([np.unique(x['time_step']) for x in tracks.data])
+    time_step_h = np.unique(np.concatenate([np.unique(x['time_step']) for x in tracks.data]))
     if not np.allclose(time_step_h, time_step_h[0]):
         LOGGER.error('Tracks have different temporal resolution. '
                      'Please ensure constant time steps by applying equal_timestep beforehand')


emanuel-schmid:
Very well. Thanks a lot for the PR.\
I've just got the usual remark: do you think you could add a unit-test that covers this fix? Ideally, without the fix,  it should fail in just the same way as cell [4] in #199.

PR #213, opened 2021-04-20
Fix Impact's _build_exp_event to work with ids indexing from  > 0

ChrisFairless:
`Impact`'s `_build_exp_event` is used in a couple of the plotting methods. It takes an event ID as input and constructs an
exposure object based on the Impact's centroids.

This fixes how it deals with the event ID it's provided: previously it assumed the ID was also the integer position of
that event in the `Impact.event_id` array (i.e. that event IDs are always `[1, 2, 3, ...]`. The fix explicitly finds the event's position before subsetting the impact matrix.

I had a quick look through `Hazard` and `Impact` to see if this bug shows up anywhere else but didn't spot anything.

(No tests seem to test this method)

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -433,8 +433,7 @@ class Impact():
             return []
 
         if event_id not in self.event_id:
-            LOGGER.error('Event ID %s not found', str(event_id))
-            return []
+            raise ValueError(f'Event ID {event_id} not found')
 
         impact_at_events_exp = self._build_exp_event(event_id)
         axis = impact_at_events_exp.plot_basemap(mask, ignore_zero, pop_name,
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -915,7 +915,7 @@ class Impact():
         Parameters:
             event_id(int): id of the event
         """
-        ix = np.isin(self.event_id, event_id).nonzero()[0]
+       [[ix]] = np.argwhere(self.event_id == event_id)
         return Exposures(
             data={
                 'value': self.imp_mat[ix].toarray().ravel(),
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -915,7 +915,7 @@ class Impact():
         Parameters:
             event_id(int): id of the event
         """
-        [[ix]] = np.argwhere(self.event_id == event_id)
+        [[ix]] = (self.event_id == event_id).nonzero()
         return Exposures(
             data={
                 'value': self.imp_mat[ix].toarray().ravel(),
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -911,9 +911,10 @@ class Impact():
         Parameters:
             event_id(int): id of the event
         """
+        ix = np.where(self.event_id == event_id)[0][0]
         return Exposures(
             data={
-                'value': self.imp_mat[event_id - 1].toarray().ravel(),
+                'value': self.imp_mat[ix].toarray().ravel(),
                 'latitude': self.coord_exp[:, 0],
                 'longitude': self.coord_exp[:, 1],
             },
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -915,7 +915,7 @@ class Impact():
         Parameters:
             event_id(int): id of the event
         """
-       [[ix]] = np.argwhere(self.event_id == event_id)
+        [[ix]] = np.argwhere(self.event_id == event_id)
         return Exposures(
             data={
                 'value': self.imp_mat[ix].toarray().ravel(),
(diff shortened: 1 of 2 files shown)


climada/engine/impact.py
```diff
@@ -911,9 +915,10 @@ def _build_exp_event(self, event_id):
         Parameters:
             event_id(int): id of the event
         """
+        ix = np.isin(self.event_id, event_id).nonzero()[0]
```
tovogt, line 918: After this, `ix` will be a 1-d array, right? Fortunately, it has only a single entry, but it's still confusing. Furthermore, I find it even more confusing to use `np.isin` with a scalar. Why don't you use the following?
```suggestion
        [[ix]] = (self.event_id == event_id).nonzero()
```

chahank:
Good fix. Could you add a test function while you are at it?

climada/engine/impact.py
chahank: Just a general interest question: is `ix` a standard abreviation for index in python coding? I usually use `idx`. Probably we should just use what is more comon.

chahank:
Just a general interest question: is `ix` a standard abreviation for index in python coding? I usually use `idx`. Probably we should just use what is more comon.

climada/engine/impact.py
```diff
@@ -432,6 +432,10 @@ def plot_basemap_impact_exposure(self, event_id=1, mask=None, ignore_zero=True,
                          'instance with parameter save_mat=True')
             return []
 
+        if event_id not in self.event_id:
+            LOGGER.error('Event ID %s not found', str(event_id))
```
chahank, line 436: Do you want to raise an error for the user? Then do so properly. The `LOGGER.error` is mostly for debugging purposes. In this case this might be the wanted behaviour.

tovogt:
Whether to use `ix` or `idx` or `index` or just plain `i` is up to the context. All of them are common. In most contexts, it will even be a matter of taste. I wouldn't enforce the use of one over the other in a review. You will just be wasting time...

climada/engine/impact.py
tovogt: `argwhere` and `nonzero` are not equivalent, in general. They are equivalent in this particular case. However, I think that, in most use cases, you are interested in all indices of nonzero entries, most probably in a 1-d array. And that is what you get from `[ix] = (...).nonzero()` while it has to be the rather awkward `ix = np.argwhere(...)[:,0]`.

In my opinion, `argwhere` is quite useless and I never use it. But there might be relevant use cases somewhere out there. Still, I wouldn't tell anyone in a review to prefer `nonzero` over `argwhere` unless the result is wrong. But outside of a review and in general use, I would always recommend to stick to `nonzero` and only use `argwhere` if there is a palpable advantage.

chahank: Fair enough. I always found it weird to use `nonzero` to test True/False values.

tovogt: I understand that you find the naming not very intuitive, but this has been around forever. It's also mentioned in the docs: "A common use for nonzero is to find the indices of an array, where a condition is True" (https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html). Everybody who is working with NumPy for a while gets used to it. It's rather irritating NOT to use `nonzero` :smile:

ChrisFairless: @tovogt I'll use your suggestion, it's a good improvement. with From my R days, it still feels weird to convert a boolean array to a set of indices because they're equivalent for subsetting.

climada/engine/impact.py
ChrisFairless: good spot

climada/engine/impact.py
ChrisFairless: ack, committed chahan's suggestion instead 😩
there are like, four more commits now, sorry.

ChrisFairless:
@chahank I've added (very basic) tests for the methods in question, and I've updated the default testing dataset so that the tests would have failed before this PR.

I'm not so experienced in python testing, so please check I've done this right.

ChrisFairless:
@emanuel-schmid would you be able to activate this PR on Jenkins? Since it's got new tests it's probably worth running.

tovogt:
There is a function included with NumPy for doing the `self.assertTrue(np.array_equal(..., ...))` check: `np.testing.assert_array_equal`. This provides a really useful output if the assertion doesn't hold.

I replaced all occurrences in the tests.

ChrisFairless:
ahhh thanks!
so this will be ready to merge once we've checked everything's passing on jenkins

climada/engine/impact.py
```diff
@@ -432,6 +432,10 @@ def plot_basemap_impact_exposure(self, event_id=1, mask=None, ignore_zero=True,
...
+            return []
```
tovogt, line 437:
```suggestion
        if event_id not in self.event_id:
            raise ValueError(f'Event ID {event_id} not found')
```

PR #216, opened 2021-04-23
Change "if" to "impf"

zeliest:
I so far went through the .py scripts and the needed csv files. I still have to work on the notebooks. The tests seem to run fine but let me know if there are any files I might have missed. I also did not aim for backwards compatibility.


chahank:
Great, very efficient! It is hard to review as is, so I propose to first see wether this branch would pass the integration test (minus the tutorials and guides). @emanuel-schmid , could you please run the integration test on this branch?

zeliest:
I added some commits to the feature/if_to_impf branch but they somehow don't appear here, any idea why? otherwise I will just open a new pr

chahank:
Are you sure that you pushed all the changes to origin?

PR #217, opened 2021-04-23
Feature/update tc clim

chahank:
The goal of this pull request is to update the Knutson climate change function for tropical cyclone track ensembles. In the previous implementations, the frequency changes were applied directly to the frequency distributions, whereas the Knutson paper factors are meant to act on the cumulative distributions. 

For example, for the NorthPacific basin, the criteria are
Cat 0-5 : -34.5%
Cat 1-5 : -31.6%
Cat 3-5 : -16.9
Cat 4-5 : 1 (change not statistically significant)

The current algorithm multiplies the frequency of the tracks of all storms of category 3-5 by -16.9%, 1-2 by -31.6%, and all storms of category 0 by -34.5%.

The new implementation applies the changes one after the other. Here, storms of categories 4-5 are untouched. Frequencies of storms cat 3 are multiplied by a factor (math below) such that the cumulative frequencies of cat 3 to 5 change by -16.9%. Then, the frequencies of category 1-2 are multiply by a factor such that the new frequencies (with updated cat 3 freqs) from cat 1-5 change by -31.6%. And analogolously for cat 0.

### DIFF
--- a/climada/hazard/tc_clim_change.py
+++ b/climada/hazard/tc_clim_change.py
@@ -126,7 +126,7 @@ def get_knutson_criterion():
          'year': 2100, 'change': 1.033, 'variable': 'intensity'}
         ]
 
-    return na + ep + wp +sp + ni + si
+    return na + ep + wp + sp + ni + si
 
 def calc_scale_knutson(ref_year=2050, rcp_scenario=45):
     """
--- a/climada/hazard/tc_clim_change.py
+++ b/climada/hazard/tc_clim_change.py
@@ -39,10 +39,10 @@ def get_knutson_criterion():
     """
     criterion = list()
     # NA
-    tmp_chg = {'basin': 'NA', 'category': [0, 1, 3, 4, 5],
+    tmp_chg = {'basin': 'NA', 'category': [0, 1, 2, 3, 4, 5],
                'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
     criterion.append(tmp_chg)
-    tmp_chg = {'basin': 'NA', 'category': [1, 3, 4, 5],
+    tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NA', 'category': [3, 4, 5],
@@ -51,13 +51,13 @@ def get_knutson_criterion():
     tmp_chg = {'basin': 'NA', 'category': [4, 5],
                'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
     criterion.append(tmp_chg)
-    
+
     tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
     criterion.append(tmp_chg)
 
     # EP
-    tmp_chg = {'basin': 'EP', 'category': [0, 1, 3, 4, 5],
+    tmp_chg = {'basin': 'EP', 'category': [0, 1, 2, 3, 4, 5],
                'year': 2100, 'change': 1.163, 'variable': 'frequency', 'function': np.multiply}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'EP', 'category': [1, 2, 3, 4, 5],
--- a/climada/hazard/tc_clim_change.py
+++ b/climada/hazard/tc_clim_change.py
@@ -29,119 +29,124 @@ TOT_RADIATIVE_FORCE = SYSTEM_DIR.joinpath('rcp_db.xls')
 generated: 2018-07-04 10:47:59."""
 
 def get_knutson_criterion():
-    """Fill changes in TCs according to Knutson et al. 2015 Global projections
+    """
+    Fill changes in TCs according to Knutson et al. 2015 Global projections
     of intense tropical cyclone activity for the late twenty-first century from
     dynamical downscaling of CMIP5/RCP4.5 scenarios.
 
-    Returns:
-        list(dict) with items 'criteria' (dict with variable_name and list(possible values)),
-        'year' (int), 'change' (float), 'variable' (str), 'function' (np function)
+    Returns
+    -------
+    criterion : list(dict)
+        list of the criterion dictionary for frequency and intensity change
+        per basin, per category taken from the Table 3 in Knutson et al. 2015.
+        with items 'basin' (str), 'category' (list(int)), 'year' (int),
+        'change' (float), 'variable' ('intensity' or 'frequency')
     """
     criterion = list()
     # NA
     tmp_chg = {'basin': 'NA', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NA', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NA', 'category': [4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.045, 'variable': 'intensity'}
     criterion.append(tmp_chg)
 
     # EP
     tmp_chg = {'basin': 'EP', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.163, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1.163, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'EP', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.193, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1.193, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'EP', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1.837, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1.837, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'EP', 'category': [4, 5],
-               'year': 2100, 'change': 3.375, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 3.375, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     tmp_chg = {'basin': 'EP', 'category': [0],
-               'year': 2100, 'change': 1.082, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.082, 'variable': 'intensity'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'EP', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.078, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.078, 'variable': 'intensity'}
     criterion.append(tmp_chg)
 
     # WP
     tmp_chg = {'basin': 'WP', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.345, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.345, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'WP', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.316, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.316, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'WP', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1 - 0.169, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.169, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'WP', 'category': [4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     tmp_chg = {'basin': 'WP', 'category': [0],
-               'year': 2100, 'change': 1.074, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.074, 'variable': 'intensity'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'WP', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.055, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.055, 'variable': 'intensity'}
     criterion.append(tmp_chg)
 
     # NI
     tmp_chg = {'basin': 'NI', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NI', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.256, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1.256, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NI', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'NI', 'category': [4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     # SI
     tmp_chg = {'basin': 'SI', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.261, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.261, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SI', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.284, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.284, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SI', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SI', 'category': [4, 5],
-               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     tmp_chg = {'basin': 'SI', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1.033, 'variable': 'intensity', 'function': np.multiply}
+               'year': 2100, 'change': 1.033, 'variable': 'intensity'}
     criterion.append(tmp_chg)
 
     # SP
     tmp_chg = {'basin': 'SP', 'category': [0, 1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.366, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.366, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SP', 'category': [1, 2, 3, 4, 5],
-               'year': 2100, 'change': 1 - 0.406, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.406, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SP', 'category': [3, 4, 5],
-               'year': 2100, 'change': 1 - 0.506, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.506, 'variable': 'frequency'}
     criterion.append(tmp_chg)
     tmp_chg = {'basin': 'SP', 'category': [4, 5],
-               'year': 2100, 'change': 1 - 0.583, 'variable': 'frequency', 'function': np.multiply}
+               'year': 2100, 'change': 1 - 0.583, 'variable': 'frequency'}
     criterion.append(tmp_chg)
 
     return criterion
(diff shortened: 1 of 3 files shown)


chahank:
The multiplication factor for each stage is computed as 
```
(a - b) / c
```
where

a: sum over all current frequencies in this group of categories (0-5, 1-5, 3-5, 4-5) multiplied with the desired factor (34.5%,  -31.6%, -16.9%, 1)
b: sum over all current frequencies of the categories in the group above ( 1-5, 3-5, 4-5 or None)
c : sum over the current frequencies of the categories to be changed (0, 1-2, 3, 4-5)

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,69 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
-                if isinstance(new_val, sparse.csr_matrix):
-                    new_val = sparse.diags(np.where(select, change, 1)).dot(new_val)
-                else:
-                    new_val[select] *= change
-                setattr(haz_cc, chg['variable'], new_val)
-        return haz_cc
+        tc_cc = copy.deepcopy(self)
+        tc_cc_basin = np.array(tc_cc.basin)
+        
+        #criterion per basin
```
tovogt, line 401: 1. Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
2. https://www.python.org/dev/peps/pep-0008/#block-comments
```suggestion

        # criterion per basin
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,69 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+        for basin in np.unique(tc_cc_basin):
+        
+            basin_chg = [chg for chg in chg_int_freq if chg['basin'] == basin]
+            bas_sel = (tc_cc_basin == basin)
+
+            #Apply intensity change
```
tovogt, line 407: 1. Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
2. https://www.python.org/dev/peps/pep-0008/#block-comments
```suggestion

            basin_chg = [chg for chg in chg_int_freq if chg['basin'] == basin]
            bas_sel = (tc_cc_basin == basin)

            # Apply intensity change
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,69 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+                         if chg['variable'] =='intensity'
+                         ]
+            for chg in inten_chg:
+                sel_cat_chg = np.isin(tc_cc.category, chg['category']) & bas_sel
+                inten_scaling = 1 + (chg['change'] - 1) * scaling_rcp_year
+                tc_cc.intensity = sparse.diags(
+                    np.where(sel_cat_chg, inten_scaling, 1)
+                    ).dot(tc_cc.intensity)
+            
+            #Apply frequency change
```
tovogt, line 419: 1. Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
2. https://www.python.org/dev/peps/pep-0008/#block-comments
```suggestion

            # Apply frequency change
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,69 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+            freq_chg = [chg 
+                        for chg in basin_chg
+                        if chg['variable']=='frequency'
+                        ]
+            freq_chg.sort(reverse = False, key = lambda x: len(x['category']))
+            
+
+            #Iteratively scale frequencies for each category such that 
+            #cumulative frequencies are scaled according to Knuston criterion.
```
tovogt, line 428: 1. Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
2. https://www.python.org/dev/peps/pep-0008/#block-comments
```suggestion


            # Iteratively scale frequencies for each category such that 
            # cumulative frequencies are scaled according to Knuston criterion.
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,69 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+                    freq_scaling_cor = (
+                        (np.sum(self.frequency[sel_cat_all]) * freq_scaling
+                         - np.sum(tc_cc.frequency[sel_cat_larger]))
+                        / np.sum(self.frequency[sel_cat_chg])
+                    )
+                    tc_cc.frequency[sel_cat_chg] *= freq_scaling_cor
+                cat_larger_list += cat_chg_list
+                
+        return tc_cc
+
```
tovogt, line 452: Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
```suggestion

        return tc_cc

```

climada/hazard/tc_clim_change.py
```diff
@@ -39,77 +39,108 @@ def get_knutson_criterion():
...
+    tmp_chg = {'basin': 'NA', 'category': [1, 3, 4, 5],
+               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+    criterion.append(tmp_chg)
+    tmp_chg = {'basin': 'NA', 'category': [3, 4, 5],
+               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+    criterion.append(tmp_chg)
+    tmp_chg = {'basin': 'NA', 'category': [4, 5],
+               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+    criterion.append(tmp_chg)
+
```
tovogt, line 54: Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
```suggestion

```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -299,15 +299,29 @@ def test_apply_criterion_track(self):
...
-        tmp_chg = {'criteria': {'basin': ['NA'], 'category': [1, 2, 3, 4, 5]},
+        tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                    'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
         criterion.append(tmp_chg)
-        tmp_chg = {'criteria': {'basin': ['WP'], 'category': [1, 2, 3, 4, 5]},
+        tmp_chg = {'basin': 'WP', 'category': [1, 2, 3, 4, 5],
                    'year': 2100, 'change': 1.025, 'variable': 'intensity', 'function': np.multiply}
         criterion.append(tmp_chg)
-        tmp_chg = {'criteria': {'basin': ['WP'], 'category': [1, 2, 3, 4, 5]},
+
```
tovogt, line 308: Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
```suggestion

```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -299,15 +299,29 @@ def test_apply_criterion_track(self):
...
-        tmp_chg = {'criteria': {'basin': ['WP'], 'category': [1, 2, 3, 4, 5]},
+        tmp_chg = {'basin': 'WP', 'category': [1, 2, 3, 4, 5],
                    'year': 2100, 'change': 1.025, 'variable': 'intensity', 'function': np.multiply}
         criterion.append(tmp_chg)
-        tmp_chg = {'criteria': {'basin': ['WP'], 'category': [1, 2, 3, 4, 5]},
+        
+        tmp_chg = {'basin': 'WP', 'category': [1, 2, 3, 4, 5],
                    'year': 2100, 'change': 1.025, 'variable': 'frequency', 'function': np.multiply}
         criterion.append(tmp_chg)
+
```
tovogt, line 312: Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
```suggestion

```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -337,8 +351,12 @@ def test_two_criterion_track(self):
             np.allclose(tc.intensity[2, :].toarray() * 1.03375, tc_cc.intensity[2, :].toarray()))
         self.assertTrue(
             np.allclose(tc.intensity[3, :].toarray() * 1.01875, tc_cc.intensity[3, :].toarray()))
+
```
tovogt, line 354: Remove trailing white space (https://www.python.org/dev/peps/pep-0008/#other-recommendations)
```suggestion

```

tovogt (commented):
Some linting. Make sure to tell your editor to automatically remove trailing white space when saving files.

chahank:
> Some linting. Make sure to tell your editor to automatically remove trailing white space when saving files.

Thanks, good advice! I did not transfer correctly my settings reinstalling spyder.

chahank:
Linting fixed in fd5ecbc01095957752548007775a1dc76b86c9c4 .

tovogt:
Maybe it's a bit too hypothetical, but what if it's impossible to apply the criterion exactly? For an arbitrary non-trivial criterion, it should be possible to generate a hazard set where it can't be applied exactly. Your procedure would then yield negative frequencies.

Toy example: 
```
criterion = [
    {'basin': 'NA', 'category': [0, 1],
     'year': 2100, 'change': 1.1, 'variable': 'frequency', 'function': np.multiply},
    {'basin': 'NA', 'category': [1],
     'year': 2100, 'change': 3, 'variable': 'frequency', 'function': np.multiply},
]

tc = TropCyclone()
tc.frequency = np.ones(2)
tc.basin = ['NA', 'NA']
tc.category = np.array([0, 1])

tc_cc = tc._apply_knutson_criterion(criterion, 1)
print(tc_cc.frequency)
# output: [-0.8  3. ]
```

I would suggest to apply a weaker least-squares procedure where you minimize the mean square error of all specified criteria. Then you obtain the new frequency vector `f_new` from the original frequency vector `f_orig` by the following minimization procedure:
```
min_{f_new} sum_i [s_i * (f_new - f_orig * factor_i)]**2
```
The sum is over all specified criteria (indexed by `i`). The row vectors `s_i` are 1 if an event is of category compliant with criterion `i` and 0 else. By the first-order criterion of optimality, this minimization problem has a minimum at the solution of the linear system of equations `sum_i s_i^T s_i * f_new = sum_i s_i^T s_i * f_orig * factor_i`. Depending on the vectors `s_i` this might have zero, one or infinitely many solutions. So, for example, if there is only a criterion for category 5, but there is no event with category 5, then `s_i = 0` and this has infinitely many solutions.

tovogt:
You still have trailing white spaces in `tc_clim_change.py` and in the test file.

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +387,68 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+            # Apply frequency change
+            freq_chg = [chg
+                        for chg in basin_chg
+                        if chg['variable'] == 'frequency'
+                        ]
+            freq_chg.sort(reverse=False, key=lambda x: len(x['category']))
+
+
+            # Iteratively scale frequencies for each category such that
+            # cumulative frequencies are scaled according to Knuston criterion.
```
tovogt, line 427:
```suggestion
            # cumulative frequencies are scaled according to Knutson criterion.
```

climada/hazard/tc_clim_change.py
```diff
@@ -39,77 +39,108 @@ def get_knutson_criterion():
...
+               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+    criterion.append(tmp_chg)
+    
+    tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
     criterion.append(tmp_chg)
 
     # EP
-    tmp_chg = {'criteria': {'basin': ['EP'], 'category': [0]},
+    tmp_chg = {'basin': 'EP', 'category': [0, 1, 3, 4, 5],
```
tovogt, line 60:
```suggestion
    tmp_chg = {'basin': 'EP', 'category': [0, 1, 2, 3, 4, 5],
```

climada/hazard/tc_clim_change.py
```diff
@@ -39,77 +39,108 @@ def get_knutson_criterion():
     """
     criterion = list()
     # NA
-    tmp_chg = {'criteria': {'basin': ['NA'], 'category': [1, 2, 3, 4, 5]},
+    tmp_chg = {'basin': 'NA', 'category': [0, 1, 3, 4, 5],
```
tovogt, line 42:
```suggestion
    tmp_chg = {'basin': 'NA', 'category': [0, 1, 2, 3, 4, 5],
```

climada/hazard/tc_clim_change.py
```diff
@@ -39,77 +39,108 @@ def get_knutson_criterion():
...
+               'year': 2100, 'change': 1, 'variable': 'frequency', 'function': np.multiply}
+    criterion.append(tmp_chg)
+    tmp_chg = {'basin': 'NA', 'category': [1, 3, 4, 5],
```
tovogt, line 45:
```suggestion
    tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
```

tovogt (commented):
Apparently, you forgot to include category 2 in some of the criteria.

chahank:
> Maybe it's a bit too hypothetical, but what if it's impossible to apply the criterion exactly? For an arbitrary non-trivial criterion, it should be possible to generate a hazard set where the criterion can't be applied exactly. Your procedure would then yield negative frequencies.
> 
> Toy example:
> 
> ```
> criterion = [
>     {'basin': 'NA', 'category': [0, 1],
>      'year': 2100, 'change': 1.1, 'variable': 'frequency', 'function': np.multiply},
>     {'basin': 'NA', 'category': [1],
>      'year': 2100, 'change': 3, 'variable': 'frequency', 'function': np.multiply},
> ]
> 
> tc = TropCyclone()
> tc.frequency = np.ones(2)
> tc.basin = ['NA', 'NA']
> tc.category = np.array([0, 1])
> 
> tc_cc = tc._apply_knutson_criterion(criterion, 1)
> print(tc_cc.frequency)
> # output: [-0.8  3. ]
> ```
> 
> I would suggest to apply a weaker least-squares procedure where you minimize the mean square error of all specified criteria. Then you obtain the new frequency vector `f_new` from the original frequency vector `f_orig` by the following minimization procedure (for simplicity, let `f_orig` and `f_new` be column vectors):
> 
> ```
> min_{f_new} sum_i [s_i * (f_new - factor_i * f_orig)]**2
> ```
> 
> The sum is over all specified criteria (indexed by `i`). The row vectors `s_i` are 1 if an event is of category compliant with criterion `i` and 0 else. By the first-order criterion of optimality, this minimization problem has a minimum at the solution of the linear system of equations `sum_i s_i^T s_i * f_new = sum_i s_i^T s_i * factor_i * f_orig`. Depending on the vectors `s_i`, this might have zero, one or infinitely many solutions. So, for example, if there is only a criterion for category 5, but there is no event with category 5, then `s_i = 0` and this has infinitely many solutions.
> 
> **Edit:** This was just the first thing that came into my mind. It definitely has too many degrees of freedom to be explicitly solvable in general. Feel free to modify this in a way that makes it more stable, for example enforcing all events of one category to be multiplied by the same factor.

Thanks for the toy example! I was thinking of this too while writing this pull request. Another issue might occur when the prescription requires the category 5 to change by x%, categories 4,5 by y%, and 3,4,5 by z% but there is no category 4 storm. Then, requirement x and y cannot be fullfilled simoultaneously, and applying z is somehow ambiguous.

So far, I decided to ignore these pathological cases. I could apply a least-square minization as you suggested, or raise an error if the requirements cannot be applied without leading to negative frequencies, or if the requirements cannot be fulfilled to reverse to the previous implementation where frequencies are directly multiplied by the factors (instead of the cumulative frequencies). What do you think?

chahank:
The other issue is that in some cases people will try to apply the criterion to a single track. In this case, what makes the most sense? Apply the criterion to the track according to what?

Maybe, now that I think of it, the best way to proceed would rather be similar to other approaches in the TC module: compute the effective factors for each category for each basin for a reference period (say IBTrACS 1980 - 2020) and use these afterwards?

tovogt:
I think it's fine to have simplifying assumptions as long as we check that the simplified assumptions are satisfied and raise an exception if they aren't. Furthermore, we should make note of this in the docstrings.

As for applying this to single tracks: You wrote in your original post that this functions applies Knutson's statistics to an "ensemble" of tracks. I would update the docstring of `set_climate_scenario_knu` to reflect this. It should inform the user that this function applies an ensemble scaling of frequencies and intensities, assuming that the unscaled ensemble is representative of current climate conditions. Your checks (see paragraph above) should constitute a sufficient safeguard if the user tries to apply this to a single track.

chahank:
Hmmmm... This is definitely one possible solution. 

Does this mean you are rather against the idea of using a reference ensemble (1980-2020) to convert the cumulative Knutson factor into factors per category, which can then be applied to any ensemble easily, and also in principle to a single track or a small ensemble?

tovogt:
It seems to me like such a change is beyond the scope of this PR. I don't know who originally implemented `set_climate_scenario_knu` and what was the original intention. Neither do I know who are the actual or supposed users today. The main topic of this PR seems to be a bug in the original `set_climate_scenario_knu`. This PR doesn't change the fundamental functionality of `set_climate_scenario_knu`, but only corrects for the fact that frequency changes are given as cumulatives over TC categories 0-5, 1-5, 3-5 and 4-5 in the original reference (Knutson et al. 2015), which has been misinterpreted by the original authors, apparently.

You seem to propose a more fundamental change to the way `set_climate_scenario_knu` works. That's a different discussion that you should have with the original authors and/or the current user group. I'm neither an author nor a user of `set_climate_scenario_knu` (and have never been) and can only provide a review of your changes based on what the function did before and what is given in Knutson et al. 2015.

Having said that, from looking at Knutson et al. 2015, it seems to me like a fixed reference ensemble for 1980-2008 is part of that study. And if you want to have changes for each category according to Knutson et al. 2015, then you should also use the reference ensemble used in the study to determine those factors. Maybe you can directly ask Thomas Knutson and he might directly provide you with more detailed statistics. Or he might point you to more recent results. In any case, that clearly goes beyond the scope of this PR.

davidnbresch:
points understood, but I am very much in favour of this fix as proposed by Chahan. No need to go deeper here, as advanced users will anyway implement their own delta climate TC sets (e.g. based on Emanuel). And for the kind of broad brush use set_climate_scenario_knu is intened for, the fix is a good pragmatic choice.

chahank:
@tovogt and @davidnbresch Thank you both for your input. Since we all agree, I will proceed with this simple fix, with a an error if frequencies become negative as well as a more explicit docstring.

climada/hazard/tc_clim_change.py
```diff
@@ -29,88 +29,124 @@
...
+    tmp_chg = {'basin': 'NA', 'category': [3, 4, 5],
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
+    criterion.append(tmp_chg)
+    tmp_chg = {'basin': 'NA', 'category': [4, 5],
+               'year': 2100, 'change': 1, 'variable': 'frequency'}
+    criterion.append(tmp_chg)
+
+    tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
+               'year': 2100, 'change': 1.045, 'variable': 'intensity'}
     criterion.append(tmp_chg)
```
tovogt, line 62: Thanks for cleaning up a bit. While you're at it, can you maybe remove all the annoying append-operations?
```suggestion
    criterion = [
        {'basin': 'NA', 'category': [0, 1, 2, 3, 4, 5],
         'year': 2100, 'change': 1, 'variable': 'frequency'},
        {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
         'year': 2100, 'change': 1, 'variable': 'frequency'},
        {'basin': 'NA', 'category': [3, 4, 5],
         'year': 2100, 'change': 1, 'variable': 'frequency'},
        {'basin': 'NA', 'category': [4, 5],
         'year': 2100, 'change': 1, 'variable': 'frequency'},
        {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
         'year': 2100, 'change': 1.045, 'variable': 'intensity'},
         # and so on for all the criteria in the function ...
         ]
```

climada/hazard/tc_clim_change.py
```diff
@@ -119,17 +155,23 @@ def calc_scale_knutson(ref_year=2050, rcp_scenario=45):
...
+    ref_year : int, optional
+        year between 2000 ad 2100. Default: 2050
+    rcp_scenario: int, optional
+        26 for RCP 2.6, 45 for RCP 4.5. The default is 45
+        60 for RCP 6.0 and 85 for RCP 8.5.
+
+    Returns
+    -------
+        : float
+        factor to scale Knuston parameters to the give RCP and year
```
tovogt, line 172:
```suggestion
    Returns
    -------
    factor : float
        factor to scale Knutson parameters to the give RCP and year
```

climada/hazard/tc_clim_change.py
```diff
@@ -119,17 +155,23 @@ def calc_scale_knutson(ref_year=2050, rcp_scenario=45):
...
     """
+
```
tovogt, line 174: There is a convention not to put any blank lines before or after function docstrings:
http://www.pydocstyle.org/en/stable/error_codes.html (D202)
https://stackoverflow.com/questions/18914108/python-pep-blank-line-after-function-definition
https://pandas.pydata.org/docs/development/contributing_docstring.html#general-rules
https://www.python.org/dev/peps/pep-0257/
```suggestion
```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -37,7 +37,7 @@
 TEST_TRACK_SHORT = DATA_DIR.joinpath("trac_short_test.csv")
 
 CENTR_TEST_BRB = Centroids()
-CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
+#CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
```
tovogt, line 40: Is this a mistake?

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -257,7 +257,7 @@ class TestClimateSce(unittest.TestCase):
     def test_apply_criterion_track(self):
         """Test _apply_criterion function."""
         criterion = list()
-        tmp_chg = {'criteria': {'basin': ['NA'], 'category': [1, 2, 3, 4, 5]},
+        tmp_chg = {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                    'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
         criterion.append(tmp_chg)
```
tovogt, line 262: As above, don't use unnecessary append operations, also in the test cases:
```suggestion
        criterion = [
            {'basin': 'NA', 'category': [1, 2, 3, 4, 5],
             'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply},
            ]
```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -337,10 +351,33 @@ def test_two_criterion_track(self):
...
+
         res_frequency = np.ones(4) * 0.5
         res_frequency[3] = 0.5 * 1.01875
+        res_frequency[0] = 0.5 * 0.25
+        res_frequency[1] = 0.5 * 0.325
+        res_frequency[2] = 0.5 * 1.75
         self.assertTrue(np.allclose(tc_cc.frequency, res_frequency))
 
+    def test_negative_freq_error(self):
+
```
tovogt, line 363:
```suggestion
        """Test _apply_knutson_criterion with infeasible input."""
```

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -337,10 +351,33 @@ def test_two_criterion_track(self):
...
+                      'variable': 'frequency'},
+                     ]
+
+        tc = TropCyclone()
+        tc.frequency = np.ones(2)
+        tc.basin = ['NA', 'NA']
+        tc.category = np.array([0, 1])
+        with self.assertRaises(ValueError):
+            tc._apply_knutson_criterion(criterion, 1)
+
```
tovogt, line 378: Never use more than 2 blank lines.
```suggestion
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +410,85 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+                    )
+                    tc_cc.frequency[sel_cat_chg] *= freq_scaling_cor
+                cat_larger_list += cat_chg_list
+
+        if (tc_cc.frequency < 0).any():
+            raise ValueError("The application of the given climate scenario"
+                         "resulted in at least one negative frequenciy."
+                         "This is likely due to the use of a"
+                         "non-representative event set (too small, "
+                         "incorrect reference period, ...)")
```
tovogt, line 488: Indentation
```suggestion
            raise ValueError("The application of the given climate scenario"
                             "resulted in at least one negative frequenciy."
                             "This is likely due to the use of a"
                             "non-representative event set (too small, "
                             "incorrect reference period, ...)")
```

climada/hazard/trop_cyclone.py
```diff
@@ -387,38 +410,85 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
...
+            Tropical cyclone with frequency and intensity scaled according
+            to the Knutson criterion. Returns a new instance of TropCyclone.
+        """
+
+        tc_cc = copy.deepcopy(self)
+
+        # Criterion per basin
+        for basin in np.unique(tc_cc.basin):
+
+            #basin_chg = [chg for chg in chg_int_freq if chg['basin'] == basin]
```
tovogt, line 435: Mistake?
```suggestion
```

tovogt (commented):
Mostly linting.

climada/hazard/tc_clim_change.py
chahank: > Thanks for cleaning up a bit. While you're at it, can you maybe remove all the annoying append-operations?

Excellent idea!

climada/hazard/tc_clim_change.py
chahank: Yeahhh I know, but it is so ugly ... but well, it is convention.

chahank:
Thanks @tovogt , I applied the suggested linting changes: 582072c

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -256,10 +256,9 @@ class TestClimateSce(unittest.TestCase):
 
     def test_apply_criterion_track(self):
         """Test _apply_criterion function."""
-        criterion = list()
-        tmp_chg = {'criteria': {'basin': ['NA'], 'category': [1, 2, 3, 4, 5]},
-                   'year': 2100, 'change': 1.045, 'variable': 'intensity', 'function': np.multiply}
-        criterion.append(tmp_chg)
+        criterion = [{'basin': 'NA', 'category': [1, 2, 3, 4, 5],
+                   'year': 2100, 'change': 1.045, 'variable': 'intensity'}
+                   ]
```
tovogt, line 261: Wrong indentation:
```suggestion
        criterion = [{'basin': 'NA', 'category': [1, 2, 3, 4, 5],
                      'year': 2100, 'change': 1.045, 'variable': 'intensity'}
                     ]
```

climada/hazard/tc_clim_change.py
```diff
@@ -29,106 +29,125 @@
...
+         'year': 2100, 'change': 1 - 0.284, 'variable': 'frequency'},
+        {'basin': 'SI', 'category': [3, 4, 5],
+         'year': 2100, 'change': 1, 'variable': 'frequency'},
+        {'basin': 'SI', 'category': [4, 5],
+         'year': 2100, 'change': 1, 'variable': 'frequency'},
+        {'basin': 'SI', 'category': [1, 2, 3, 4, 5],
+         'year': 2100, 'change': 1.033, 'variable': 'intensity'}
+        ]
+
+    return na + ep + wp +sp + ni + si
```
tovogt, line 129: Missing whitespace:
```suggestion
    return na + ep + wp + sp + ni + si
```

tovogt (commented):
Apart from the two little comments, it's ready to go.

chahank:
The attention to detail is impressive. Thanks.

PR #219, opened 2021-04-26
Feature/if to impf

zeliest:
I opened a new pull request, since it wasn't updating my commits.


zeliest:
Also, I updated the notebooks for the imports and variable names, but I wasn't able to run all of them because I do not have all the data or there were errors which I don't think are related to these changes. Could someone still try to run them?

chahank:
Thanks! Then I repeat my previous point here: it is not really possible to review this in its functioning besides looking at the unit tests and integration tests. @emanuel-schmid could you please run an integration test on this branch?

emanuel-schmid:
unit test results available, one test failing
https://ied-wcr-jenkins.ethz.ch/job/climada_branches/view/change-requests/job/PR-219/lastCompletedBuild/testReport/climada.entity.measures.test.test_base/TestApply/test_change_exposures_impf_pass/

chahank:
Good news! It looks like the demo files were not updated maybe?

zeliest:
Thank you Emanuel! I think it should be fixed now

chahank:
I would say that if the unit and integration tests all pass (@emanuel-schmid could you make a new one please so that we do not wait too long), we should merge as soon as possible and inform all developers about the change. The currently active pull requests might lead to some discrepancy, but hopefully only minor ones.

emanuel-schmid:
Thanks @zeliest! \
@chahank : agreed. \
There is one failure in integration tests on Windows (on Linux it is still running):
```
ERROR [1.164s]: test_calib_instance (climada.test.test_calibration.TestCalib)
Test save calib instance
  File "C:\Users\me\www\climada_python\climada\test\test_calibration.py", line 52, in test_calib_instance
    ent.exposures.gdf.impf_TC.median())
  File "C:\Users\me\anaconda3\envs\climada_v2\lib\site-packages\pandas\core\generic.py", line 5465, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'GeoDataFrame' object has no attribute 'impf_TC'
```

emanuel-schmid:
You can test it locally by calling:
```
python -m unittest climada.test.test_calibration.TestCalib.test_calib_instance 
```

ThomasRoosli:
Only a short questions: Will old saved exposures be read correctly after this pull request? If I saved an exposure to hdf5 a year ago. Will CLIMADA read that correctly after this pull request or will CLIMADA be confused, because the column "if_" will not be recognized as the old naming convention of "impf_"?

zeliest:
I fixed what was causing the error related to the impf but I still get an error in the test_calib_instance test which doesn't seem related?

zeliest:
@ThomasRoosli I haven't implemented that yet but I think it makes sense to just say if there is an if_ somehwere in the exposures dataframe when reading a file replace it by impf_

ThomasRoosli:
@zeliest that would be great 👍

chahank:
> @ThomasRoosli I haven't implemented that yet but I think it makes sense to just say if there is an if_ somehwere in the exposures dataframe when reading a file replace it by impf_


I agree, this is a very important point to keep some backward compatibility. Thus, the loading methods (excel, matlab, and else) for exposures and entity should probably be adapted to be able to read `if_` columns, issue a warning, and convert the column to `_impf`.

climada/entity/exposures/base.py
```diff
@@ -700,6 +709,11 @@ def write_raster(self, file_name, value_name='value', scheduler=None):
             raster, meta = u_coord.points_to_raster(self, [value_name], scheduler=scheduler)
             u_coord.write_raster(file_name, raster, meta)
 
+    def _replace_if_by_impf(self):
+        LOGGER.info('Replacing if_ by impf_ in Exposure geodataframe')
+        column_name = self.gdf.columns[self.gdf.columns.str.startswith('if_')][0]
+        self.gdf.rename(columns={column_name: column_name.replace("if", "impf")}, inplace=True)
```
chahank, line 715: Be careful, it is valid for an exposure to have more than one column `if_`. For example, you could have `if_TC` and `if_WF` for Tropical Cyclones and Wildfires. This will become increasingly important for multi-hazard analysis. The above code would fail in this case and only update the first instance. 

I suppose this might be a problem in other parts of the code too.

climada/entity/exposures/base.py
```diff
@@ -615,6 +622,8 @@ def read_mat(self, file_name, var_names=None):
             raise KeyError(f"Variable not in MAT file: " + str(var_err)) from var_err
 
         self.gdf = GeoDataFrame(data=exposures, crs=self.crs)
+        if any(self.gdf.loc[:, self.gdf.columns.str.startswith('if_')]):
```
chahank, line 625: Why not 
```suggestion
        if any(self.gdf.columns.str.startswith('if_')):
```

climada/entity/exposures/base.py
```diff
@@ -700,6 +709,11 @@ def write_raster(self, file_name, value_name='value', scheduler=None):
             raster, meta = u_coord.points_to_raster(self, [value_name], scheduler=scheduler)
             u_coord.write_raster(file_name, raster, meta)
 
+    def _replace_if_by_impf(self):
+        LOGGER.info('Replacing if_ by impf_ in Exposure geodataframe')
```
chahank, line 713: I suggest being more explicit here.
```suggestion
        LOGGER.warning(" 'if_'  was replaced by 'impf_' in the Exposure's GeoDataFrame. The name 'if_' is deprecated. ")
```

chahank (changes requested):
Just a few quick remarks.

climada/entity/exposures/litpop.py
```diff
@@ -1,1786 +1,1786 @@
-"""
-This file is part of CLIMADA.
```
chahank, line 2: What happened with `litpop.py` ?

emanuel-schmid: carriage return?

chahank: What is 'carriage return'?

emanuel-schmid:
> 
> 
> Only a short questions: Will old saved exposures be read correctly after this pull request? If I saved an exposure to hdf5 a year ago. Will CLIMADA read that correctly after this pull request or will CLIMADA be confused, because the column "if_" will not be recognized as the old naming convention of "impf_"?

@ThomasRoosli, @chahank & @zeliest 

Sorry for being late in the discussion. But I must mention that I'm not happy with the proposed solution. Silently juggling with if_ and impf_ names within half a dozen methods of the exposures base class makes me a bit nervous. 😱 It is quite messy and will be a annoying us for a very long time if we don't draw the line right here. Also we kind of agreed to dismiss backwards compatibility in this regard, haven't we?

Could we not provide file conversion methods for convenience instead? So when someone comes with an old hdf5 file and wants to do climada on it, the workflow will fail with a prominent warning, something like this:
"""
it looks like you have an old style climada file - perhaps you want it to adapt to the current style? just run `climada.legacy.convert("xy.hdf5", "xy-new.hdf5")`
"""

emanuel-schmid:
Other ideas for solving this?

chahank:
I was also thinking of the possibility to make a conversion function too. But is it not simpler to do it as is now, kind of implicitly? It is really only a limited set of loading methods that convert `if_` to `impf_` now, and they issue a clear warning. It also has the advantage that the several persons using excel sheets won't have problems. But I see also that it might create a long tail and is less clean than a clear cut.

The conversion methods you propose however for me look also quite messy. One would need to have for each data type a new function, that loads the data, change the `if_` to `impf_`, and save it again in the same format. This would actually require to write entirely new methods to save into format that currently have only reading methods, but no saving methods (matlab matrices, geotiffs, excel).

climada/entity/exposures/litpop.py
zeliest: I also do not know

ThomasRoosli:
> 
> 
> > Only a short questions: Will old saved exposures be read correctly after this pull request? If I saved an exposure to hdf5 a year ago. Will CLIMADA read that correctly after this pull request or will CLIMADA be confused, because the column "if_" will not be recognized as the old naming convention of "impf_"?
> 
> @ThomasRoosli, @chahank & @zeliest
> 
> Sorry for being late in the discussion. But I must mention that I'm not happy with the proposed solution. Silently juggling with if_ and impf_ names within half a dozen methods of the exposures base class makes me a bit nervous. 😱 It is quite messy and will be a annoying us for a very long time if we don't draw the line right here. Also we kind of agreed to dismiss backwards compatibility in this regard, haven't we?
> 
> Could we not provide file conversion methods for convenience instead? So when someone comes with an old hdf5 file and wants to do climada on it, the workflow will fail with a prominent warning, something like this:
> """
> it looks like you have an old style climada file - perhaps you want it to adapt to the current style? just run `climada.legacy.convert("xy.hdf5", "xy-new.hdf5")`
> """

I am happy if there is a solution that provides the convenience to be still able to use older files. I do not know which of the two suggested solutions would be better.

climada/entity/exposures/base.py
```diff
@@ -312,12 +312,12 @@ def get_impf_column(self, haz_type):
         if INDICATOR_IMPF in self.gdf.columns:
             LOGGER.info("No specific impact function column found for hazard %s."
                         " Using the anonymous 'impf_' column.", haz_type)
-            return INDICATOR_IMPF + haz_type
+            return INDICATOR_IMPF
```
chahank, line 315: That would result in wrong behaviour. The `haz_type` is required. Same below.

climada/entity/exposures/base.py
```diff
@@ -280,6 +280,46 @@ def check(self):
...
+            or the HAZ_TYPE constant of the hazard's module
+
+        Returns
+        -------
+        str
+            a column name, the first of the following that is present in the exposures' dataframe:
+            - impf_[haz_type]
+            - if_[haz_type]
+            - impf_
+            - if_
```
chahank, line 299: These two values are not correct if a `haz_type` has been defined. An empty exposures is initialialized without specific `haz_type`, thus with `if_` or `impf_`. As soon as a `haz_type` is specified, one must use `if_haz_type` (`impf_haz_type`)

emanuel-schmid:
Some `if_` occurrences seem to have got overseen. These files are probably worth to be edited:

- climada/engine/calibration_opt.py
- climada/engine/impact_data.py
- climada/engine/test/test_cost_benefit.py
- climada/engine/test/test_impact.py
- climada/entity/exposures/gdp_asset.py
- climada/entity/exposures/spam_agrar.py
- climada/entity/exposures/test/test_gdp_asset.py
- climada/entity/impact_funcs/drought.py
- climada/entity/impact_funcs/test/test_fl.py

emanuel-schmid:
And also this data file:

- climada/hazard/test/data/atl_prob_no_name.mat

(`climada.test.test_calibration.TestCalib.test_calib_instance` fails)

zeliest:
> And also this data file:
> 
>     * climada/hazard/test/data/atl_prob_no_name.mat
> 
> 
> (`climada.test.test_calibration.TestCalib.test_calib_instance` fails)

this file doesn't contain any 'if' reference. The test also fails locally for me, but because it is not looking in the right folder for the file

PR #221, opened 2021-04-27
Code Review: Methods to handle lines and polygons with CLIMADA exposures

Evelyn-M:
Dear all

This is an incomplete, work-in-progress module, which I do not intend to merge soon.
Yet, I'd like some conceptual feedback and also make sure changes aren't horrendous in a month's time or so.

The changes include the following:
Merging `util.interpolation` and `util.coordinates`, which has led to circular imports. This also removes duplicate functions. The unit tests work in the merger.

Adding the following util functions in coordinates: `interpolate_lines`, `interpolate_polygons` and the helper `dist_great_circle_allgeoms`.

Adding the new module `lines_polys_handler.py` which does the transformation from line / poly geometries into CLIMADA point exposures and the back-transformation from point impacts into line / polygon geometries, via some aggregation options. They are not yet implemented for polys.

I also added "fake" unittests for all those functions in the respective test files.


climada/util/coordinates.py
```diff
@@ -318,6 +570,56 @@ def dist_approx(lat1, lon1, lat2, lon2, log=False, normalize=True,
         raise KeyError("Unknown distance approximation method: %s" % method)
     return (dist, vtan) if log else dist
 
+def dist_great_circle_allgeoms(gdf):
```
tovogt, line 573: This function computes total lengths and perimeters of (collections of) geometrical objects. I wouldn't use `dist` in the name of this function because that would imply computing the distance *between* geometrical objects. As an alternative name, I would suggest `geodesic_lengths` or `compute_geodesic_lengths`.

climada/util/coordinates.py
```diff
@@ -76,6 +78,256 @@
...
+    # split line lengths into relative fractions acc to point_dist (e.g. 0, 0.5, 1)
+    gdf_points['distance_vector'] = gdf_points.apply(
+        lambda row: np.linspace(0, 1, num=int(np.ceil(row.length_full/
+                                                      row.length)+1)), axis=1)
+
+    # create MultiPoints along the line for every position in distance_vector
+    gdf_points['geometry'] = gdf_points.apply(
+        lambda row: MultiPoint(
+            [row.geometry.interpolate(dist, normalized=True) 
+             for dist in row.distance_vector]), axis=1)
```
tovogt, line 257: The LineString method `interpolate` does interpolation in a Cartesian coordinate system. I think there's not a lot we can do about that, but maybe we can warn the user in the docstring that the result will be inaccurate for LineStrings whose defining points are spaced very far from each other (like end points in Zürich and NYC).

climada/util/coordinates.py
```diff
@@ -76,6 +78,256 @@
...
+    trans = gdf_poly.apply(lambda row: pts_to_raster_meta(
+        row.geometry.bounds, (row.degree_dist, -row.degree_dist)), axis=1)
+    gdf_poly['trans'] = pd.DataFrame(trans.tolist(), index=trans.index).iloc[:,2]
+   
+    gdf_poly['width'] = np.floor(abs(gdf_poly.geometry.bounds.minx-
+                                     gdf_poly.geometry.bounds.maxx)/
+                                 gdf_poly['degree_dist'])
+    gdf_poly['height'] = np.floor(abs(gdf_poly.geometry.bounds.miny-
+                                      gdf_poly.geometry.bounds.maxy)/
+                                  gdf_poly['degree_dist'])
```
tovogt, line 300: Why do you manually recompute width and height even though this is already returned by `pts_to_raster_meta`?

climada/util/coordinates.py
```diff
@@ -34,28 +34,30 @@
 from fiona.crs import from_epsg
 import geopandas as gpd
 import numpy as np
+from numba import jit
 import pandas as pd
+import pyproj
```
chahank, line 39: This is a new dependency. Is it absolutely necessary? Has this been discussed with @emanuel-schmid . Please, as a general convention, before adding new dependencies in a pull request discuss it openly to see whether a) it is really needed b) it is compatible with the environment c) does not add many extra implicit dependencies.

climada/util/coordinates.py
```diff
@@ -76,6 +78,256 @@
 MAX_DEM_TILES_DOWN = 300
 """Maximum DEM tiles to dowload"""
 
+######### copied over from interpolation module #####################
```
chahank, line 81: Does this mean that nothing in this pull request is new?

climada/util/coordinates.py
```diff
@@ -76,6 +78,256 @@
...
+def _dist_sqr_approx(lats1, lons1, cos_lats1, lats2, lons2):
+    """Compute squared equirectangular approximation distance. Values need
+    to be sqrt and multiplicated by ONE_LAT_KM to obtain distance in km."""
+    d_lon = lons1 - lons2
+    d_lat = lats1 - lats2
+    return d_lon * d_lon * cos_lats1 * cos_lats1 + d_lat * d_lat
+
+def interpol_index(centroids, coordinates, method=METHOD[0],
+                   distance=DIST_DEF[1], threshold=THRESHOLD):
+    """Returns for each coordinate the centroids indexes used for
```
chahank, line 103: Could you please update the docstrings to numpy format? Otherwise this will clash with the current update of the docstrings.

chahank:
I have some questions about the use of certain methods. What would the following be used for? I do not really see concrete use cases. Also, they seem to not be used anywhere in the CLIMADA code base. This indicates to me that they should potentially be entirely removed instead of being ported.

- `interpol_index` 
- `index_nn_aprox`
- `index_nn_haversine`

tovogt:
The function `interpol_index` has originally been implemented for the `assign_centroids` functionality. Now, this functionality has been moved to the util function `assign_coordinates`. But this function still uses `interpol_index` to find the index of the geographically nearest neighbor in a given point cloud.

The functions `index_nn_approx` and  `index_nn_haversine` are helper functions for use within `interpol_index`.

climada/util/coordinates.py
tovogt: `pyproj` is not a new requirement for CLIMADA. It has been included in the requirements as an indirect dependency (through geopandas) for a long time and is used in several places in the code base. But maybe it's true that we should explicitly list it in the requirements file for the hypothetical case that geopandas might drop this dependency at some point in the future.

tovogt: Here are examples of other packages, that are explicitly imported in CLIMADA, but not listed in the requirements files: * cftime
* fiona
* mock
* netCDF4
* numpy
* pyproj
* shapefile
* shapely

chahank: Good points! It seems to me that some of these requirements should really be explicitly listed in the requirements file.

I agree that the import in general does not necessarily need to be discussed here, but this is still exactly the goal of pull requests of this type: discuss the general architecture which includes the discussion about external dependencies.

chahank:
> The function `interpol_index` has originally been implemented for the `assign_centroids` functionality. Now, this functionality has been moved to the util function `assign_coordinates`. But this function still uses `interpol_index` to find the index of the geographically nearest neighbor in a given point cloud.
> 
> The functions `index_nn_approx` and `index_nn_haversine` are helper functions for use within `interpol_index`.

Strange, when I checked the references to these functions on github it did not show them, but I see them in the branch. Fair enough. Thanks for clarifying.

climada/util/coordinates.py
Evelyn-M: Yes, indeed. I added this to the docstring

climada/util/coordinates.py
Evelyn-M: Don't know why I did this, either ;). Probably temporal code-blindness.
Thanks.

climada/util/coordinates.py
Evelyn-M: This means that i copied this part from the interpolation method (until where the next full line of ######### is). It's just a marker to facilitate the review. I will of course delete this remark later.

climada/util/coordinates.py
Evelyn-M: yep. 946d4f11

climada/util/coordinates.py
```diff
@@ -76,6 +78,273 @@
...
+    # make grid --> lon / lat for each polygon-bbox
+    gdf_points[['lon','lat']] = pd.DataFrame(gdf_points.apply(lambda row: 
+        raster_to_meshgrid(row.trans, row.width, row.height), axis=1).tolist())
+    
+    for axis in ['lon', 'lat']:
+        gdf_points[axis] = gdf_points.apply(lambda row: row[axis].flatten(), 
+                                            axis=1)
+    # filter only centroids in actual polygons
+    for i, polygon in enumerate(gdf_points.geometry):
+        in_geom = coord_on_land(lat=gdf_points['lat'].iloc[i],
```
chahank, line 333: Why implicitly assume points on land? The util function to transform from polygons to points should not assume that. There are several use cases where points on the ocean are relevant.

climada/util/coordinates.py
```diff
@@ -76,6 +78,273 @@
...
+        axis=1).tolist()
+    
+    # make grid --> lon / lat for each polygon-bbox
+    gdf_points[['lon','lat']] = pd.DataFrame(gdf_points.apply(lambda row: 
+        raster_to_meshgrid(row.trans, row.width, row.height), axis=1).tolist())
+    
+    for axis in ['lon', 'lat']:
+        gdf_points[axis] = gdf_points.apply(lambda row: row[axis].flatten(), 
+                                            axis=1)
+    # filter only centroids in actual polygons
```
chahank, line 331: If I am not mistaken, you can thus easily end up with a polygon that is mapped to no point at all (if it is between raster points). I am not sure how to solve this, but it probably is not desired behaviour, or at least it should be warned about.

One possibility is to assign a value proportional to the cell area in a given polygon. This will lead to points belonging to different polygons at the same time though.

climada/util/coordinates.py
```diff
@@ -76,6 +78,273 @@
...
+    # create MultiPoints along the line for every position in distance_vector
+    gdf_points['geometry'] = gdf_points.apply(
+        lambda row: MultiPoint(
+            [row.geometry.interpolate(dist, normalized=True) 
+             for dist in row.distance_vector]), axis=1)
+    
+    # expand gdf from MultiPoint entries to single Points per row
+    return gdf_points.explode().drop(['distance_vector', 'length_full'], axis=1)
+
+def interpolate_polygons(gdf_poly, area_point):
```
chahank, line 287: Would it be useful to provide the grid to map the polygons onto? Think for instance that one has a centroids raster and wants to map the polygons to this raster.

climada/util/coordinates.py
```diff
@@ -76,6 +78,273 @@
...
+    original LineString 
+    Important remark: LineString.interpolate() used here performs interpolation 
+    in a Cartesian coordinate system. The result will be inaccurate for 
+    LineStrings whose defining points are spaced very far from each other 
+    (like end points in Zürich and NYC)
+    
+    Parameters
+    ----------
+    gdf_lines : gpd.GeoDataframe or str
+        Geodataframe or filepath from which to read the GeoDataframe
```
chahank, line 249: I do not really like the mixing of data reading and data handling inside of a util funtion.

chahank (changes requested):
A few reflections on the `interpolate_polygons` function.

chahank:
General comment: what is implemented here is basically a version of `Exposure.set_lat_lon` for `Lines` and `Polygons` instead of points. Would it make sense to include the helper functions there? Then there could be a quite natural flow for compute impacts from geometry exposures, by simply checking in `impact.calc` if the `lat` and `lon` are set, and if not apply `set_lat_lon`. The difficulty is of course how to disaggregate by default, but maybe we can find a good compromise there. In addition, one could reaggregate the impact by default again at the end of the computation. This might again lead to aggregation default problems, but could be a way togo.

I think in past discussions this idea was rejected due to the question of how to aggregate and disaggregate. However, we might give it a tough a new here. In particular, if the util functions do not get different types of aggregations/disaggregation.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,145 @@
...
+    exp = Exposures(gdf_points)
+    if value_method == 'fixed':
+        exp.gdf['value'] = metre2_value*point_area
+    else:
+        LOGGER.warning("No other method for valuation yet implemented")
+    exp.set_lat_lon()
+    exp.check()
+    return exp
+
+def agg_point_impact_to_lines(gdf_lines, exp_points, imp_points,
```
chahank, line 78: Why are both `gdf_lines` and `exp_points` needed? The `exp_points` should contain all the information from `gdf_lines`.

climada/util/coordinates.py
Evelyn-M: That's not what it assumes. it does as said in the comment -  # filter only centroids in actual polygons
The function `coord_on_land` simply checks whether the points generated from the even grid lie within the desired polygon, and excludes the others. in short, it's just a wrapper for an sjoin func

chahank: This is confusing to me. Why use `coord_on_land` then? What happens if the coords or the centroids are in the water and the polygon also? 

Why not use sjoin directly then?

climada/util/coordinates.py
Evelyn-M: Well, if you choose your representative area per point too small, then yes. The function assumes that the user adequately judges the size of the polygons in relation to the area per point. Aka, if you have a tiny island of 1km^2, you shouldn't interpolate your polygon with a `point_area` of 10'000m2

Evelyn-M: There's also a lot of discussions online on how to best allocate a certain amount of points within an arbitrary 2D shape. There's a few fancy algorithms out there, but honestly, i think for our purpose, the commonly accepted easy-way out is good enough. And that's making an even grid of pre-determined resolution, and then cropping it to the outline of the polygon. This is how it's implemented atm

chahank: If you have say 100 polygons, and one happens to be very slim, then you get problems too, irrespective of the area per point. But I agree that this might be ok, but I would at least raise a warning if one of the polygons has no point associated.

Evelyn-M: e.g. (https://stackoverflow.com/questions/11178414/algorithm-to-generate-equally-distributed-points-in-a-polygon)

climada/util/coordinates.py
Evelyn-M: ehm. if you already have a grid at hand, you may not really need this function. with a grid, you can just use an sjoin to see which points of the grid lie inside the polygon. The idea of this func here is to help the user make a grid, where they just need to think about the representative area (resolution) per grid-point.

chahank: Ok, but then the function should only produce the grid. Another function can then be made that maps the polygons onto said grid.

climada/util/coordinates.py
chahank: Thanks, I looked again at the code for `coord_on_land`, my bad. It can indeed be misused for filtering points in polygons.

climada/util/coordinates.py
tovogt: Note that this function doesn't simply create "a grid". It creates a separate grid for each polygon. Not only the origin of each of those grids is different, but also the resolution. The reason for that is the fact that polygons are assumed to be given in lat-lon-coordinates, while the user specifies the desired resolution in meters. So, the function chooses the grid resolution according to the specific degree-meter conversion factor at the latitudinal location of each polygon.

I partly agree with @chahank: This function should probably be better separated into "subfunctions". I think the most important separation would be to have a function `_interpolate_one_polygon` that works on each polygon separately and does both generate the grid and apply it to the polygon. The function `interpolate_polygons` then just uses `apply` and `explode`. If you want to be even more modular (as suggested by @chahank), you can divide `_interpolate_one_polygon` further.

Evelyn-M: Both good points! I will implement a more modular version and also think about your use case @chahank , which was not the prime intention, and very easily implementable with the default geopandas functionalities. It might still fit somewhere here.

Evelyn-M:
> General comment: what is implemented here is basically a version of `Exposure.set_lat_lon` for `Lines` and `Polygons` instead of points. Would it make sense to include the helper functions there? Then there could be a quite natural flow for compute impacts from geometry exposures, by simply checking in `impact.calc` if the `lat` and `lon` are set, and if not apply `set_lat_lon`. The difficulty is of course how to disaggregate by default, but maybe we can find a good compromise there. In addition, one could reaggregate the impact by default again at the end of the computation. This might again lead to aggregation default problems, but could be a way togo.
> 
> I think in past discussions this idea was rejected due to the question of how to aggregate and disaggregate. However, we might give it a tough a new here. In particular, if the util functions do not get different types of aggregations/disaggregation.

I agree on the fact that it would be nice for the impact calc flow, but also difficult without making lots of assumptions on the disaggregation and re-aggregation. I'll continue for now with the current structure, and then lets see if we can make a proposal to use those functionalities with some default assumptions in the impact.calc routine

climada/util/coordinates.py
chahank: Great! I think this is really important for your design and should be taken into account right at the beginning. Even though it might not be the prime intention, a well-designed set of functions should without difficulty allow for both use cases. Whether this is possible might be a good way to test the (good) modularity of the design.

I think the most simple use-case is something like LitPop where a given value is distributed over all chosen points inside of a shape-polygon. Afterwards, there should be a function where the impacts are aggregated.

Now, the variations and difficulties come in when one chooses a way to distribute, a way to set the points inside of the polygon (a grid, centroids, grid of equal surface cells,...) and a way to aggregate the impacts.

Evelyn-M: commit a4ceae19 refactors the polygon (and the line) interpolation: a sub-function `_interpolate_one_polygon` now creates a separate grid for each polygon with the desired resolution and cuts out only those resulting points that lie inside it.

climada/util/coordinates.py
Evelyn-M: I checked the `coord_on_land` and it's basically a re-naming of the shapely-native function `vectorized.contains()` which is exactly what is needed here to see which points fall into the polygon. I use the latter directly now in commit e892d39a

climada/util/coordinates.py
Evelyn-M: A warning is now raised if a polygon has no point with the given resolution. Also, one "representative" point is then still set into the shape.

Evelyn-M:
> General comment: what is implemented here is basically a version of `Exposure.set_lat_lon` for `Lines` and `Polygons` instead of points. Would it make sense to include the helper functions there? Then there could be a quite natural flow for compute impacts from geometry exposures, by simply checking in `impact.calc` if the `lat` and `lon` are set, and if not apply `set_lat_lon`. 

I agree with this. The disagreggation (at exposure-level) is basically a set-method like the `set_lat_lon` and the `set_from_raster`, only that we set from lines or polys, respectively.
So I gave the lines and polys functions a new home in exposure.base in commit 658f9b42 .

I am against automating / doing a lot "under the hood", so the user would still be forced to run the set_from_lines / set_from_polygons incl. the resolution choice (m or m2 per interpolated point) themselves. Like they also have to run set_lat_lon themselves after exposure creation.

climada/util/coordinates.py
Evelyn-M: Changed this to allowing only gdfs

Evelyn-M:
A few general remarks, after some updates:
1) Here's the link to the [mural board](https://app.mural.co/t/flowchartcascadingcifailures5558/m/flowchartcascadingcifailures5558/1629823048943/afd90cfce01580bcf8c516b2886d2142f39a4019?sender=ud8f9b28cfdfd5d7a44c53509) where we discussed some current issues.

2) Main changes: Rather than just helper functions, we thought that the disaggregation and reaggregation could be methods of the exposure & impact instances. The first is implemented as a suggestion currently, the latter (on the impact-reaggregation), not yet.

3) Dis-aggregations and re-aggregations have to be user-defined in terms of representative area / distance per generated point on the geometry part. How values & other variables are then distributed is not so clear yet, and requires several different sub-methods. For the disaggregation, there is a `constant` option so far, that just equally distributes the value per polygon / line onto its corresponding points. @togeig has also made a suggestion to use litpop values for non-uniform redistribution of values within polygons (see the tutorial on this!) . For the re-aggregation, there is a `sum` and a `fraction` option so far.

4) The workflow is updated in the tutorial.

Suggestions, amendments, etc. are highly welcomed!!

Evelyn-M:
Update: There's a kwarg "litpop" now for the valuation method of polygons. It loads a LitPop exposure and distributes the sum of all litpop asset values in each polygon evenly over the interpolated points within each polygon.

Evelyn-M:
> Update: There's a kwarg "litpop" now for the valuation method of polygons. It loads a LitPop exposure and distributes the sum of all litpop asset values in each polygon evenly over the interpolated points within each polygon.

The LitPop valuation method for poygons, as shown in the tutorial with provinces of the Netherlands, works fine for large Polygon areas > LitPop resolution.
There is, however, a second case where LitPop could come in at hand for valuation purposes, but does not work like this:
With very small polygons (e.g. building shapes queried from OSM), this returns 0 value, since obviously, chances that a LitPop centroid falls into a small polygon, are very low.
In that case it would be useful to take the hull of all the polygons (see picture), and re-distribute the LitPop value of this hull proportionally on the small polygons.
This is not yet implemented, but I will leave it here as a comment.
image

chahank:
I propose to close this pull request since many changes have been made to the methods proposed for handling geometries and lines. New separate pull requests will be opened to handle the different proposed changes.

Evelyn-M:
Yes, we can do that

PR #223, opened 2021-04-29
Allow specifying genesis basin in read_ibtracs_netcdf

### LINKED ISSUE #222: Read IBTrACS strange basin naming
When one uses the `read_ibtracks_netcdf()` method from `climada.hazard.TCTracks` with a specified `basin`, the returned tracks are often from another basin. This might just be a labelling issue, but I think it should be corrected. 

In this example, although the selected basin is `EP`, 7/8 tracks are labelled as being from `WP`. 

```
>from climada.hazard import TCTracks
>tr = TCTracks()
>tr.read_ibtracs_netcdf(basin='EP', year_range=(1980,1981))
>tr.to_geodataframe()['basin']
0    WP
1    EP
2    WP
3    WP
4    WP
5    WP
6    WP
7    WP
8    WP
Name: basin, dtype: object
```

tovogt:
Following issue #222, this PR clarifies the docstrings of `read_ibtracs_netcdf` to inform the user about the exact meaning of the `basin` parameter.

Since some users might be interested in filtering by the genesis basin since this is the basin that ends up being attached to each track with its `"basin"` attribute, this PR adds an option `genesis_basin`.

This doesn't necessarily close #222, but is a nice convenience feature for many users.


climada/hazard/tc_tracks.py
```diff
@@ -387,36 +402,52 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
+
+        ibtracs_ds = xr.open_dataset(ibtracs_path)
+        ibtracs_date = ibtracs_ds.attrs["date_created"]
+        LOGGER.info(f"The IBTrACS data set dates from {ibtracs_date}")
+        if 180 < (np.datetime64('today') - np.datetime64(ibtracs_date)) / np.timedelta64(1, 'D'):
+            LOGGER.warning("The cached IBTrACS data set is older than 180 days. "
+                           "Very likely, a more recent version is available. "
+                           f"Consider manually removing the file {ibtracs_path} and re-running "
+                           "this function, which will download the most recent version of the "
+                           "IBTrACS data set from the official URL.")
```
tovogt, line 424: I found that users tend to use data for very recent years even though this kind of data is very often subject to changes in IBTrACS updates. But even for older data, there might be interesting updates from time to time. CLIMADA currently caches the IBTrACS data set and uses this version indefinitely. That's fine for reproducibility and stability of code, but will very often go unnoticed by the user. This update is to inform the user about this fact so that they may decide to manually switch to a more recent version of the data.

chahank:
Thanks for the quick update following #222 . However, I do not understand why one still attaches the attribute `basin` to the track object differently from the keyword argument `basin`.

climada/hazard/tc_tracks.py
chahank: This is an excellent point!

bguillod: I like, too!

climada/hazard/tc_tracks.py
```diff
@@ -289,8 +292,8 @@ def tracks_in_exp(self, exposure, buffer=1.0):
 
 
     def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=None,
-                            year_range=None, basin=None, interpolate_missing=True,
-                            estimate_missing=False, correct_pres=False,
+                            year_range=None, genesis_basin=None, basin=None,
```
chahank, line 295: Not particularly relevant, but I would write `basin=None,  genesis_basin=None` instead because I tend to view first come variables as more important and less  optional.

climada/hazard/tc_tracks.py
```diff
@@ -345,8 +348,23 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
+            because only the first *valid* eye position is considered. Possible values are 'NA'
+            (North Atlantic), 'SA' (South Atlantic), 'EP'​ (Eastern North Pacific, which includes
+            the Central Pacific region), 'WP'​ (Western North Pacific), 'SP'​ (South Pacific),
+            'SI'​ (South Indian), 'NI'​ (North Indian). If None, this filter is not applied.
+            Default: None.
         basin : str, optional
-            E.g. US, SA, NI, SI, SP, WP, EP, NA. If not provided, consider all basins.
+            If given, select storms that have at least one position in the specified basin. This
+            allows analysis of a given basin, but also means that basin-specific track sets should
+            not be combined across basins since some storms will be in more than one set. If you
```
chahank, line 364: Very clearly written!

climada/hazard/tc_tracks.py
```diff
@@ -387,36 +405,52 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
+        if not ibtracs_path.is_file():
             try:
                 download_ftp(f'{IBTRACS_URL}/{IBTRACS_FILE}', IBTRACS_FILE)
-                shutil.move(IBTRACS_FILE, fn_nc)
+                shutil.move(IBTRACS_FILE, ibtracs_path)
             except ValueError as err:
                 raise ValueError(
                     f'Error while downloading {IBTRACS_URL}. Try to download it manually and '
-                    f'put the file in {SYSTEM_DIR}') from err
+                    f'put the file in {ibtracs_path}') from err
```
chahank, line 417: I am not 100%, but I thought remembering that one should write the message as

```suggestion
                    'put the file in %s', ibtracs_path) from err
```

Can you confirm @tovogt and/or @emanuel-schmid ?

climada/hazard/tc_tracks.py
```diff
@@ -387,36 +405,52 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
+
+        ibtracs_ds = xr.open_dataset(ibtracs_path)
+        ibtracs_date = ibtracs_ds.attrs["date_created"]
+        LOGGER.info(f"The IBTrACS data set dates from {ibtracs_date}")
```
chahank, line 421: I am not sure whether the info always needs to be there. I think it is enough to have the warning if it is too old. Else, one can expect the user to know from the data is?

climada/hazard/tc_tracks.py
```diff
@@ -387,36 +405,52 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
             match &= (ibtracs_ds.basin == basin.encode()).any(dim='date_time')
             if np.count_nonzero(match) == 0:
                 LOGGER.info('No tracks in basin %s.', basin)
+        if genesis_basin is not None:
+            # Here, we only filter for the basin at *any* eye position. We will filter again later
+            # for the basin of the *first* eye position, but only after restricting to the valid
+            # time steps in the data.
+            match &= (ibtracs_ds.basin == genesis_basin.encode()).any(dim='date_time')
+            if np.count_nonzero(match) == 0:
+                LOGGER.info('No tracks in basin %s.', genesis_basin)
```
chahank, line 453: I know it is not technically 100% correct, but it is more informative to the user to say
```suggestion
                LOGGER.info('No tracks with first valid eye position in genesis basin %s.', genesis_basin)
```

climada/hazard/tc_tracks.py
```diff
@@ -521,7 +555,15 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
                 LOGGER.info("Progress: %d%%", perc)
                 last_perc = perc
             track_ds = ibtracs_ds.sel(storm=i_track, date_time=t_msk)
-            st_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
+            tr_basin_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
+            tr_genesis_basin = track_ds.basin.values[0].astype(str).item()
+
+            # Now that the valid time steps have been selected, we discard tracks that don't fit
+            # the specified basin definitions:
+            if genesis_basin is not None and tr_genesis_basin != genesis_basin:
```
chahank, line 563: Would another log message be useful here?
```suggestion
            if genesis_basin is not None and tr_genesis_basin != genesis_basin:
                LOGGER.info('No tracks with first valid eye position in genesis basin %s.', genesis_basin)
```

climada/hazard/test/test_tc_tracks.py
```diff
@@ -248,6 +248,35 @@ def test_ibtracs_correct_pass(self):
...
+        self.assertEqual(tc_track.data[0].basin[0], 'SP')
+        self.assertEqual(tc_track.data[5].basin[0], 'SI')
+
+        # genesis in NI
+        tc_track = tc.TCTracks()
+        tc_track.read_ibtracs_netcdf(
+            year_range=(1994, 1994), genesis_basin="NI", estimate_missing=True)
+        self.assertEqual(tc_track.size, 5)
+        for tr in tc_track.data:
+            np.testing.assert_array_equal(tr.basin[0], "NI")
```
chahank, line 271: Why assert array equal, arent's these two single strings?

climada/hazard/tc_tracks.py
tovogt: I think you confused this with the `logging` functions. For logging-functions it might be undesirable to format the string in case the log-level will cause the result to be discarded anyways: https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/ But this is not the case here.

climada/hazard/tc_tracks.py
```diff
@@ -387,36 +405,52 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
             except ValueError as err:
                 raise ValueError(
                     f'Error while downloading {IBTRACS_URL}. Try to download it manually and '
-                    f'put the file in {SYSTEM_DIR}') from err
+                    f'put the file in {ibtracs_path}') from err
+
+        ibtracs_ds = xr.open_dataset(ibtracs_path)
+        ibtracs_date = ibtracs_ds.attrs["date_created"]
+        LOGGER.info(f"The IBTrACS data set dates from {ibtracs_date}")
+        if 180 < (np.datetime64('today') - np.datetime64(ibtracs_date)) / np.timedelta64(1, 'D'):
```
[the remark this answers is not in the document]
tovogt: If you insist, I can use your suggested version, but I don't think it's clearer at all. Note that the two versions do not do the same thing!
```
>>> (np.datetime64('2021-05-10 10:00') - np.datetime64("2021-05-09 12:00")).item().days
0
>>> (np.datetime64('2021-05-10 10:00') - np.datetime64("2021-05-09 12:00")) / np.timedelta64(1, 'D')
0.9166666666666666
```

climada/hazard/tc_tracks.py
tovogt: Ì modified the output to mention that this refers to the given genesis basin.

climada/hazard/tc_tracks.py
tovogt: I added a log message after the loop.

climada/hazard/tc_tracks.py
```diff
@@ -521,7 +555,15 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
...
             track_ds = ibtracs_ds.sel(storm=i_track, date_time=t_msk)
-            st_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
+            tr_basin_penv = xr.apply_ufunc(basin_fun, track_ds.basin, vectorize=True)
+            tr_genesis_basin = track_ds.basin.values[0].astype(str).item()
+
+            # Now that the valid time steps have been selected, we discard tracks that don't fit
+            # the specified basin definitions:
+            if genesis_basin is not None and tr_genesis_basin != genesis_basin:
+                continue
+            if basin is not None and basin.encode() not in track_ds.basin.values:
```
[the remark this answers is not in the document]
tovogt: I added a log message after the loop.

climada/hazard/test/test_tc_tracks.py
tovogt: Well spotted, this was a copy-paste thing.

tovogt:
@chahank Thanks for your comments! I think I addressed all of them in my replies and/or in my recent commit.

climada/hazard/tc_tracks.py
chahank: Thanks for the reminder, I indeed confused both of them.

climada/hazard/tc_tracks.py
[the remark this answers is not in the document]
chahank: Sure, but in our case we care about the number of days, not about the fraction of days I would say.

chahank:
@tovogt Looks great. I didn't check the linting by myself, but from experience your linting is excellent :D.

Maybe before merging, could you consider the point I discussed in the related issue  #222 about documenting somewhere (I am not sure where it is best, but for example in the docstring of `TropCyclone.set_from_tracks`) that the `TropCyclone().basin` attribute refers to the genesis basin (or basin of the first track point)?

climada/hazard/tc_tracks.py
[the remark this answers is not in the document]
tovogt: I think it's a waste of time to argue about this, so I'm just accepting your suggested version.

climada/hazard/tc_tracks.py
```diff
@@ -625,19 +671,24 @@ def _read_file_emanuel(self, path, hemisphere='S', rmw_corr=False):
...
+
+        hem_min, hem_max = -90, 90
+        if basin == "GB":
+            if hemisphere == 'S':
+                hem_min, hem_max = -90, 0
+                basin = "S"
+            elif hemisphere == 'N':
+                hem_min, hem_max = 0, 90
+                basin = "N"
+
```
tovogt, line 691: This makes sure that the basin information contained in the data set is used if available.

climada/hazard/tc_tracks.py
```diff
@@ -625,19 +671,24 @@ def _read_file_emanuel(self, path, hemisphere='S', rmw_corr=False):
         path : str
             absolute path of file to read.
         hemisphere : str, optional
-            'S', 'N' or 'both'. Default: 'S'
+            For global data sets, restrict to northern ('N') or southern ('S') hemisphere.
+            Use 'both' to disable this restriction. Default: 'S'
```
chahank, line 675: I know I am nitpicking, but actually, the string `both` is never used in the code.

climada/hazard/trop_cyclone.py
```diff
@@ -111,6 +111,20 @@ def set_from_tracks(self, tracks, centroids=None, description='',
...
+
+        The `TropCyclone.category` attribute is set to the value of the `category`-attribute
+        of each of the given track data sets.
+
+        The `TropCyclone.basin` attribute is set to the genesis basin for each event, which
+        is the first value of the `basin`-variable in each of the given track data sets.
+
+        Optionally, the time dependent, vectorial winds can be stored using the `store_windfields`
+        function parameter (see below).
+
```
chahank, line 127: Excellent addition!

chahank:
Please look at  comment https://github.com/CLIMADA-project/climada_python/pull/223#discussion_r629454061 and address it if you wish, otherwise this is good to be merged.

PR #224, opened 2021-04-29
Reverse cumulative knutson cc

chahank:
This commit reverses the conceptual changes from merge 8084d9a894083d0ad00a0524d0da384f9d4e4374 .

The previous changes lead to negative frequency when applied to historical tracks obtained from IBTrACS.

### DIFF
--- a/climada/hazard/test/test_trop_cyclone.py
+++ b/climada/hazard/test/test_trop_cyclone.py
@@ -377,20 +377,17 @@ class TestClimateSce(unittest.TestCase):
 
     def test_negative_freq_error(self):
         """Test _apply_knutson_criterion with infeasible input."""
-        criterion = [{'basin': 'NA', 'category': [0, 1],
-                      'year': 2100, 'change': -1.1,
-                      'variable': 'frequency'},
-                     {'basin': 'NA', 'category': [1],
-                      'year': 2100, 'change': 3,
-                      'variable': 'frequency'},
+        criterion = [{'basin': 'SP', 'category': [0, 1],
+                      'year': 2100, 'change': 0.5,
+                      'variable': 'frequency'}
                      ]
 
         tc = TropCyclone()
         tc.frequency = np.ones(2)
-        tc.basin = ['NA', 'NA']
+        tc.basin = ['SP', 'SP']
         tc.category = np.array([0, 1])
         with self.assertRaises(ValueError):
-            tc._apply_knutson_criterion(criterion, 1)
+            tc._apply_knutson_criterion(criterion, 3)
 
 
 if __name__ == "__main__":
--- a/climada/hazard/test/test_trop_cyclone.py
+++ b/climada/hazard/test/test_trop_cyclone.py
@@ -37,7 +37,7 @@ TEST_TRACK = DATA_DIR.joinpath("trac_brb_test.csv")
 TEST_TRACK_SHORT = DATA_DIR.joinpath("trac_short_test.csv")
 
 CENTR_TEST_BRB = Centroids()
-#CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
+CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
 
 
 class TestReader(unittest.TestCase):
--- a/climada/hazard/test/test_trop_cyclone.py
+++ b/climada/hazard/test/test_trop_cyclone.py
@@ -37,7 +37,7 @@ TEST_TRACK = DATA_DIR.joinpath("trac_brb_test.csv")
 TEST_TRACK_SHORT = DATA_DIR.joinpath("trac_short_test.csv")
 
 CENTR_TEST_BRB = Centroids()
-CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
+#CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
 
 
 class TestReader(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


climada/hazard/test/test_trop_cyclone.py
```diff
@@ -37,7 +37,7 @@
 TEST_TRACK_SHORT = DATA_DIR.joinpath("trac_short_test.csv")
 
 CENTR_TEST_BRB = Centroids()
-CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
+#CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
```
chahank, line 40:
```suggestion
CENTR_TEST_BRB.read_mat(DATA_DIR.joinpath('centr_brb_test.mat'))
```

chahank:
@tovogt  This should not be much, I just reversed the applications of the Knutson criterion to the previous version, but I kept the cosmetic changes to the code.

climada/hazard/trop_cyclone.py
```diff
@@ -485,25 +485,17 @@ def _apply_knutson_criterion(self, chg_int_freq, scaling_rcp_year):
                 sel_cat_chg = np.isin(tc_cc.category, cat_chg_list) & bas_sel
```
tovogt, line 485: Some lines above this one (L476-477), there is an obsolete comment about cumulative frequencies that needs to be removed as well.

climada/hazard/test/test_trop_cyclone.py
```diff
@@ -370,16 +370,15 @@ def test_two_criterion_track(self):
...
+        res_frequency[1] = 0.5 * (1 + (0.7 - 1) * scale)
+        res_frequency[2] = 0.5 * (1 + (2 - 1) * scale)
+        res_frequency[3] = 0.5 * (1 + (1.025 - 1) * scale)
         self.assertTrue(np.allclose(tc_cc.frequency, res_frequency))
 
     def test_negative_freq_error(self):
         """Test _apply_knutson_criterion with infeasible input."""
         criterion = [{'basin': 'NA', 'category': [0, 1],
-                      'year': 2100, 'change': 1.1,
+                      'year': 2100, 'change': -1.1,
```
tovogt, line 381: This is a very simplistic case (and very hypothetical). A much less hypothetical would be a change factor of 0.5 (like in 'SP') together with a large scaling factor (second argument of `_apply_knutson_criterion`) like 3 (you get that for RCP 8.5 and reference year 2100). Maybe you can add this as another test case?

climada/hazard/trop_cyclone.py
```diff
@@ -485,25 +485,17 @@ def _apply_knutson_criterion(self, chg_int_freq, scaling_rcp_year):
...
-                    tc_cc.frequency[sel_cat_chg] *= freq_scaling_cor
+                    tc_cc.frequency[sel_cat_chg] *= freq_scaling
                 cat_larger_list += cat_chg_list
 
         if (tc_cc.frequency < 0).any():
             raise ValueError("The application of the given climate scenario"
                              "resulted in at least one negative frequenciy."
-                             "This is likely due to the use of a"
+                             "This is likely due to the use of a "
                              "non-representative event set (too small, "
```
tovogt, line 495: After reverting the changes, it is no longer true that negative frequencies would be due to a non-representative event set. Instead, this will happen for extremely high-emission scenarios in basins and categories with high reductions in frequency according to Knutson. For example, RCP 8.5 with reference year 2100 and 'WP' or 'SP' basin.

tovogt (commented):
Thanks! Please have a look at my comments.

climada/hazard/trop_cyclone.py
chahank: Yes, thank you!

climada/hazard/trop_cyclone.py
chahank: 1.  Correct, it was sloppy writing, but you get the point right? Instead of  making a subtraction, the scaling from climate change should be relative. Then, negative frequencies cannot happen.
2. Yes,  but if an error is discovered during this process it should be addressed right away, potentially leading to a new pull request.

tovogt: No, I'm not sure I get the point because I don't understand why we should inverse the meaning of the scaling factor. Maybe you can explain in a bit more detail?

chahank: Ok, sorry for the lack or clarity. So, the scaling factor can become negative, but this is imho an artefact of a bad implementation. 

Suppose the climate change scaling factor is 2. Then:
- Knutson factor 110%=1.1 -> change  = 1+(1.1-1)*2 = 1.2
This looks reasonable, we go from an increase by 10% to 20%.
- Knutson factor 10%=1/10=0.1 -> change = 1+(0.1-1)*2 = -0.8
This is not  reasonable. What we should expect is rather to go from a reduction of 10% to 5%=1/20=0.05.

tovogt: What makes you think that more climate change will produce less reduction in frequency?

chahank: I think the message was still unclear. Say frequency is 1, climate change factor 2. Then:

 - Knutson factor 110%=1.1 -> change = 1+(1.1-1)*2 = 1.2
This looks reasonable, we go from an increase by 10% (frequency 1.1) to 20% (frequency 1.2). 
- Knutson factor 10%=1/10=0.1 -> change = 1+(0.1-1)*2 = -0.8
This is not reasonable as frequency would be -0.8. What we should expect is rather to go from a reduction to 10% (frequency 0.1) to 5%=1/20=0.05 (frequency 0.05).

climada/hazard/trop_cyclone.py
```diff
@@ -485,25 +485,17 @@ def _apply_knutson_criterion(self, chg_int_freq, scaling_rcp_year):
                 sel_cat_chg = np.isin(tc_cc.category, cat_chg_list) & bas_sel
                 if sel_cat_chg.any():
                     freq_scaling = 1 + (chg['change'] - 1) * scaling_rcp_year
```
chahank, line 487:
```suggestion
                    if chg['change'] >= 1:
                          freq_scaling = 1 + (chg['change'] - 1) * scaling_rcp_year
                    else:
                          freq_scaling = chg['change'] / scaling_rcp_year
```

climada/hazard/trop_cyclone.py
chahank: As a possible solution (will  probably need another pull request), I propose the change in the comment below https://github.com/CLIMADA-project/climada_python/pull/224#discussion_r629339786

climada/hazard/trop_cyclone.py
tovogt: With the current implementation, a Knutson factor of 0.9 with a scaling of 3 would become 0.7. With your suggested change, the Knutson factor 0.9 would become 0.3. So, that's not very reasonable, right?

chahank: You have a good example. To be honest, I would not say that it is not reasonable. It really depends on how one does interpret the extrapolation. 

The extrapolation I suggest is "a doubling in temperature leads to a doubling of the relative frequency/intensity change if is increasing, and to a halving of the relative frequency/intensity change if it is decreasing". This seems reasonable to me, but as you are saying, it is not necessarily  the best interpolation method.

chahank: I will now merge with the old method and open an issue to discuss this point. I addressed the points which you raised in https://github.com/CLIMADA-project/climada_python/pull/223#discussion_r629454061

tovogt (approved):
Yes, please go ahead and merge if you like.

PR #226, opened 2021-05-03
Plots improvement

zeliest:
Some attempts to improve the maps:

- new colormaps suggestions, add them to constants. Separate one for impact/exposures scatterplots
- change default fontsize - still not conviced about that one
- add option to leave areas with no data white in raster plots (suggested by Thomas)
- add option to only show points of populated places without names (suggested by Sam L.)
- Antimeridian handling for function geo_im_from_array
- Add basic tests for plots in util.plot, any suggestions of other things to test are welcome


zeliest:
Screen Shot 2021-05-03 at 12 17 13

This would be the new colormaps for the scatter plots, left would be for exposures and right for impacts.

zeliest:
Screen Shot 2021-05-03 at 12 23 48

I adapted the size of the fonts on the plots, I think for most of them it looks good, but for plots with multiple subplots it doesn't work with the bigger fonts. Is there a way to adapt the fonts to the size of the figure? I could not find anything. Left is the new size, right the old

chahank:
I think for a dynamical adaptation you first have to define the figsize, and then fix the the size of the fonts depending on the figsize.

chahank:
> Screen Shot 2021-05-03 at 12 17 13
> 
> This would be the new colormaps for the scatter plots, left would be for exposures and right for impacts.

Really nice improvements! I am however not yet fully convinced by this unicolor maps. I find it quite hard to actually see differences, in particular when some regions have very low values and some very large. In addition, using whit as the lowest value, one cannot use it for missing data. It then should probably be gray for missing data?

zeliest:
@chahank which one do you consider to be unicolor? both? I also think using too many colors quickly gets chaotic. And lowest value is not completly white but rather really light colors, I will also check what it looks like with gray

zeliest:
Screen Shot 2021-05-03 at 15 54 21

I do think it makes sense to color 0s and values outside the colorbar in gray.

chahank:
Yes, gray seems to work well!

climada/entity/exposures/base.py
```diff
@@ -376,7 +376,9 @@ def plot_scatter(self, mask=None, ignore_zero=False, pop_name=True,
             mask (np.array, optional): mask to apply to eai_exp plotted.
             ignore_zero (bool, optional): flag to indicate if zero and negative
                 values are ignored in plot. Default: False
-            pop_name (bool, optional): add names of the populated places
+            pop_name : bool or str, optional
+                add names of the populated places, by default True. If set to 'markers' only the points
+                will be shown without a description.
```
chahank, line 381: Probably it is better then to change to `,None', 'names', 'markers'` or something along these lines. Mixing types  in variables is not a  very good idea most of the time.

climada/entity/exposures/base.py
```diff
@@ -415,7 +417,9 @@ def plot_hexbin(self, mask=None, ignore_zero=False, pop_name=True,
...
```
chahank, line 422: Probably it is better then to change to `,None', 'names', 'markers'` or something along these lines. Mixing types  in variables is not a  very good idea most of the time.

climada/util/plot.py
```diff
@@ -29,7 +29,7 @@
 
 import logging
 from textwrap import wrap
-
+import matplotlib
```
chahank, line 32:
```suggestion
import matplotlib as mlp
```

climada/util/plot.py
```diff
@@ -77,8 +81,9 @@ def geo_bin_from_array(array_sub, geo_coord, var_name, title,
     title : str or list(str)
         subplot title. If one provided, the same is used for all subplots.
         Otherwise provide as many as subplots in array_sub.
-    pop_name : bool, optional
-        add names of the populated places, by default True
+    pop_name : bool or str, optional
+        add names of the populated places, by default True. If set to 'markers' only the points
+        will be shown, without a description.
```
chahank, line 86: Probably it is better then to change to ,None', 'names', 'markers' or something along these lines. Mixing types in variables is not a very good idea most of the time.

climada/entity/exposures/base.py
zeliest: I think it is just a bit annoying then for all the notebooks, ... where it is written pop_name=True and it would throw an error. But maybe this option to only plot the markers is not that important and we can just forget about it

climada/util/plot.py
```diff
@@ -243,7 +249,7 @@ def geo_im_from_array(array_sub, coord, var_name, title,
         Each array (in a row or in  the list) are values at each point in corresponding
         geo_coord that are ploted in one subplot.
     coord : 2d np.array
-        (lat, lon) for each point in a row. The same grid is used for all subplots.
+        (lat, lon) for each point in a row. The same grid is used for all subploplots.
```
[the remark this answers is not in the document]
zeliest: oups

climada/util/plot.py
```diff
@@ -524,11 +540,12 @@ def add_populated_places(axis, extent, proj=ccrs.PlateCarree()):
...
-                axis.plot(point.x, point.y, color='navy', marker='o', markersize=7,
+                axis.plot(point.x, point.y, color='navy', marker='o',
                           transform=ccrs.PlateCarree(), markerfacecolor='None')
-                axis.text(point.x, point.y, place_name,
-                          horizontalalignment='right', verticalalignment='bottom',
-                          transform=ccrs.PlateCarree(), fontsize=14, color='navy')
+                if pop_name != 'markers':
+                    axis.text(point.x, point.y, place_name,
+                              horizontalalignment='right', verticalalignment='bottom',
+                              transform=ccrs.PlateCarree(), color='navy')
```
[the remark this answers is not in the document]
zeliest: I decided not to do it with the mpl.rc parameters as it didn't work when the figures where changing sizes. I added a parameter adapt_fontsize to all maps which if set to True adapts the fonts based on the size of the figure. It is also possible to set this parameter to False and to still use the mpl.rc parameters in a notebook for example. https://github.com/CLIMADA-project/climada_python/pull/226/commits/76e328217a3dfd5d85c15d74ee5d0c4197877955

PR #232, opened 2021-05-19
read_ibtracs_netcdf: raise exception for invalid storm ids

tovogt:
The `storm_id` parameter of `read_ibtracs_netcdf` refers to a unique identifier in IBTrACS. Users that specify this parameter are dealing with non-fuzzy information and will most certainly expect to have the correct ID in their hands. Still, the users are only warned in the log output, which might go unnoticed by many users.

With this PR, I propose to raise an exception whenever we find that a user-specified ID is invalid (or non-existing).

What do you think?

As an example why a user might end up specifying a wrong ID: The user might have the ID from a different source and the source might be old or in a strange format. IBTrACS IDs change from time to time, especially for events in the last 2-3 years. Furthermore, if the source contains encoding errors, the specified ID might contain invalid (maybe invisible) characters such as whitespace. In both cases, users might be able to determine the correct IDs manually - but only if they notice that there are invalid IDs in the first place. If a long list of IDs is specified, a few missing IDs will easily go unnoticed.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -348,7 +348,7 @@ class TCTracks():
         storm_id : str or list of str, optional
             IBTrACS ID of the storm, e.g. 1988234N13299, [1988234N13299, 1989260N11316].
         year_range : tuple (min_year, max_year), optional
-            Year range to filter track selection. Default: (1980, 2018)
+            Year range to filter track selection. Default: None.
         basin : str, optional
             If given, select storms that have at least one position in the specified basin. This
             allows analysis of a given basin, but also means that basin-specific track sets should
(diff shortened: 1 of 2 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -348,7 +348,7 @@ def read_ibtracs_netcdf(self, provider=None, rescale_windspeeds=True, storm_id=N
         storm_id : str or list of str, optional
             IBTrACS ID of the storm, e.g. 1988234N13299, [1988234N13299, 1989260N11316].
         year_range : tuple (min_year, max_year), optional
-            Year range to filter track selection. Default: (1980, 2018)
+            Year range to filter track selection. Default: None.
```
tovogt, line 351: If find it extremely disturbing that there is a default year range. I didn't even know that this exists even though I'm using this function regularly. So, if I specify only a basin like 'NA', there will be no tracks for the year 2019 (for example). This is totally counter-intuitive and I propose to remove the default year range.

chahank:
I fully agree! Do all the tests and the tutorials still work?

chahank:
Then I would say feel free to merge! Thanks.

PR #233, opened 2021-05-21
u_coord: function to align raster data to global reference grid

sameberenz:
util function reproject_2d_grid() reprojects (spatial up- or downsample) 2D np.ndarray to a grid using rasterio.
    The destination (target) grid is defined from target_meta and optionally
    target_res_arcsec, global_origins, and target_crs.
    This function ensures that reprojected data with the same resolution and
    global_origins are on the same global grid, i.e., no offset between 
    destination grid points for different input grids that are projected to
    the same resolution.
    
This function is required by the new litpop implementation currently under development.


tovogt:
This seems to be an extension of the already existing function `refine_raster_data`. I would prefer to have only one function `reproject_raster_data` in the end.

Furthermore, we should be transparent that this is a convenience wrapper function around `rasterio.warp.reproject`. Therefore, I suggest that we stick to the parameter names in `rasterio.warp.reproject`:

```
reproject_raster_data(source, src_crs, src_transform, dst_crs=None, dst_transform=None,
                      dst_shape=None, dst_resolution=None, dst_global_origin=None,
                      dst_buffer=None, resampling=None, conserve=None, **kwargs)
```

If omitted, `dst_crs` defaults to `src_crs`. Moreover, `dst_transform` can be determined from combinations like `dst_resolution` and `src_transform` (or `dst_global_origin`), they are determined from the other user-specified parameters. The `kwargs` are forwarded to the call of `rasterio.warp.reproject`. The output should be an array `destination` and `dst_transform`. The CRS doesn't need to be returned because it is user-specified (either from `dst_crs` or `src_crs`).

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+    meta_out : dict
+        meta data dictionary of target grid, same structure as meta_in.
+        meta_out defines the target (i.e. destination) grid.
+    res_arcsec_out : int (optional)
+        target resolution in arcsec, overrules resolution in meta_out['transform'].
+        The default is None, i.e. same resolution as target grid defined in meta_out.
+    global_origins : tuple with two numbers (lat, lon) (optional)
+        global lon and lat origins as basis for destination grid.
+        Only required if target_res_arcsec different than resolution in meta_out.
+        The default is (-180.0, 90).
```
tovogt, line 1859: I wouldn't put a default value for `global_origins`. The default prescribes a certain CRS already (lat/lon).

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+             'height': 1939,
+             'crs': CRS.from_epsg(4326),
+             'transform': Affine(0.00833333333333333, 0.0, -18.175000000000068,
+                                 0.0, -0.00833333333333333, 43.79999999999993)}
+    meta_out : dict
+        meta data dictionary of target grid, same structure as meta_in.
+        meta_out defines the target (i.e. destination) grid.
+    res_arcsec_out : int (optional)
+        target resolution in arcsec, overrules resolution in meta_out['transform'].
+        The default is None, i.e. same resolution as target grid defined in meta_out.
```
tovogt, line 1855: I would instead use a CRS-agnostic resolution parameter `dst_resolution`. Arc-seconds only make sense for certain choices of CRS. A user, that wants to specify a certain resolution in arc-seconds, can very easily convert this to degrees or meters or whatever.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+    global_origins : tuple with two numbers (lat, lon) (optional)
+        global lon and lat origins as basis for destination grid.
+        Only required if target_res_arcsec different than resolution in meta_out.
+        The default is (-180.0, 90).
+    crs_out : rasterio.crs.CRS
+        destination CRS
+        The default is crs provided in meta_out.
```
tovogt, line 1862: Rename to `dst_crs` like in original `rasterio.warp.reproject`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+    target_res_arcsec, global_origins, and target_crs.
+    This function ensures that reprojected data with the same resolution and
+    global_origins are on the same global grid, i.e., no offset between 
+    destination grid points for different input grids that are projected to
+    the same resolution.
+
+    Parameters
+    ----------
+    data : np.ndarray
+        2D-array containing the values to be reprojected
```
tovogt, line 1839: Rename to `source` like in original `rasterio.warp.reproject`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+        The default is None, i.e. same resolution as target grid defined in meta_out.
+    global_origins : tuple with two numbers (lat, lon) (optional)
+        global lon and lat origins as basis for destination grid.
+        Only required if target_res_arcsec different than resolution in meta_out.
+        The default is (-180.0, 90).
+    crs_out : rasterio.crs.CRS
+        destination CRS
+        The default is crs provided in meta_out.
+    resampling : resampling function (optional)
+        The default is rasterio.warp.Resampling.bilinear
```
tovogt, line 1864: I would use `Resampling.nearest` as a default because that's also the default in the origina `rasterio.warp.reproject`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+    meta_in : dict
+        meta data dictionary of input grid
+        Required fields in meta are 'dtype,', 'width', 'height', 'crs', 'transform'.
+        Example:
+            {'dtype': 'float32',
+             'width': 2702,
+             'height': 1939,
+             'crs': CRS.from_epsg(4326),
+             'transform': Affine(0.00833333333333333, 0.0, -18.175000000000068,
+                                 0.0, -0.00833333333333333, 43.79999999999993)}
```
tovogt, line 1849: Use parameters `src_transform` and `src_crs` like in original `rasterio.warp.reproject`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,140 @@ def apply_box(df_exp):
...
+    meta_out : dict
+        meta data dictionary of target grid, same structure as meta_in.
+        meta_out defines the target (i.e. destination) grid.
```
tovogt, line 1852: Use parameters `dst_transform` and `dst_crs` like in original `rasterio.warp.reproject`. Add `dst_shape` to specify the output array's shape (height/width).

sameberenz:
Thanks for the thorough review!
Ok, I will adapt as you suggest, even though replacing meta explodes the number of parameters – at least meta is an established form.
Also, I'd like to stick to lat, lon as a default in global_origin because otherwise it is quite annoying for users to find out what to set there as a default.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+
+    **kwargs : **kwargs
+        extra arguments for rasterio.warp.reproject.
+
+    Returns
+    -------
+    destination
+        numpy array.
+    meta_dst : dict
+        meta dict of destination.
```
tovogt, line 1888: We only need to return the `dst_transform` because the `dst_crs` and `dst_shape` are known to the user. Furthermore, `destination` and `dst_transform` are also the return values of `rasterio.warp.reproject`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+    src_transform : Affine
+        Source transform, contains info on grid, e.g.:
+            Affine(0.00833333333333333, 0.0, -18.175000000000068,
+                                 0.0, -0.00833333333333333, 43.79999999999993)
+    dst_crs : CRS, optional
+        Destination CRS. The default is src_crs.
+    dst_transform : Affine, optional
+        Destination transform. The default is src_transform.
+    dst_shape : tuple with integers, optional
+        (height, width) of destination grid. The default is source.shape.
```
tovogt, line 1854: I don't think that `source.shape` is a good default value for `dst_shape`. By default, the value should be determined from the destination CRS and resolution, together with the source raster's extent.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+        Note: Destination shape is changed if resolution changes
+    dst_resolution : number, optional
+        Destination resolution.
+        Overwrites resolution provided in dst_transform.
```
tovogt, line 1858: I suggest not to overwrite one user-specified parameter by another, but to define certain combinations of input parameters as illegal and raise an exception. The most obvious example would be `dst_transform` together with `dst_resolution` (or `dst_global_origin`).

climada/util/test/test_coordinates.py
```diff
@@ -850,6 +899,150 @@ def test_bounded_refined_raster(self):
...
+                                                           meta_list[j]['crs'],
+                                                           meta_list[j]['transform'],
+                                                           dst_crs=meta_list[i]['crs'],
+                                                           dst_transform=meta_list[i]['transform'],
+                                                           dst_shape=(meta_list[i]['height'],
+                                                                      meta_list[i]['width']),
+                                                           dst_resolution=None,
+                                                           dst_global_origin=None,
+                                                           dst_global_extent=None,
+                                                           dst_buffer=None,
```
tovogt, line 917: In the tests, you specify all the input parameters, even if they are unset (None). I strongly suggest to specify the input parameters in a way that reflects what a typical user would do.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+        Unit of resolution depends on CRS;
+        Typically but not necessarily, the unit is degree lat, lon.
+        The default is dst_transform[0].
+    dst_global_origin : tuple of numbers, optional
+        global lon and lat origins as basis for destination grid.
+        Only required if dst_resolution different than dst_transform[0].
+        The default is (-180.0, 90), change dst_global_origin for different units.
+    dst_global_extent : tuple of numbers, optional
+        Extent of global grid in unit of resoution.
+        The default is (180, 360). Change for different units than degree lat, lon.
```
tovogt, line 1868: I suggest to list x-coordinates before y-coordinate as in `dst_global_origin`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+    src_crs : CRS
+        Source CRS
+    src_transform : Affine
+        Source transform, contains info on grid, e.g.:
+            Affine(0.00833333333333333, 0.0, -18.175000000000068,
+                                 0.0, -0.00833333333333333, 43.79999999999993)
+    dst_crs : CRS, optional
+        Destination CRS. The default is src_crs.
+    dst_transform : Affine, optional
+        Destination transform. The default is src_transform.
```
tovogt, line 1852: I don't think that `src_transform` is a good default value for `dst_transform`. By default, the value should be determined from the destination CRS and resolution, together with the source raster's extent or the `dst_global_origin`.

climada/util/coordinates.py
```diff
@@ -1821,6 +1821,160 @@ def apply_box(df_exp):
...
+    dst_resolution : number, optional
+        Destination resolution.
+        Overwrites resolution provided in dst_transform.
+        Unit of resolution depends on CRS;
+        Typically but not necessarily, the unit is degree lat, lon.
+        The default is dst_transform[0].
+    dst_global_origin : tuple of numbers, optional
+        global lon and lat origins as basis for destination grid.
+        Only required if dst_resolution different than dst_transform[0].
+        The default is (-180.0, 90), change dst_global_origin for different units.
```
tovogt, line 1865: I suggest to raise an exception if the `dst_global_origin` parameter is needed, but the user didn't specify any, and the destination CRS is not in latitude/longitude. Maybe even restrict the default value to some known systems like `epsg:4326`.

tovogt (commented):
Some more conceptual comments.

climada/util/coordinates.py
sameberenz: I've put it this way to be in line with the shape from the array (source or destination), where shape = (height, width), i.e. (y, x). Either way there needs to be some thinking what is y and what is x, so I'll leave it this way

climada/util/coordinates.py
sameberenz: see https://github.com/CLIMADA-project/climada_python/pull/233#discussion_r638002248

climada/util/coordinates.py
sameberenz: I guess there is a misunderstanding here, the whole idea is to reproject to a consistent global grid based both on dst_transform and dst_resolution. However, I will avoid future confusions by not overwriting dst_transform and rather defining an new dst_transform_newres if provided, will also adjust docstring.

tovogt: Hm, I'm not sure what you mean. Why would a user provide the `dst_transform` at all, if they want the information therein to be overwritten by other parameters?

sameberenz: The idea of this function is to reproject either to a new resolution based on the source grid (src_transform and dst_resolution are required input to define the destination grid), reproject to a given grid (dst_transform defines the destinatiion grid, no dst_resolution given), or reproject to a new resolution on a grid that is globally consistent to the grid defined in dst_transform, in this case both dst_transform and dst_resolution need to be provided. This later application is exactly what I need for LitPop, i.e. project Lit-data (source) on a grid with given resolution dst_resolution that is defined consistently with the Pop-grid (defined in dst_transform). Providing both dst_transform and dst_resolution makes sure there is no offset between coordinates of Lit and Pop both reprojected to the same resolution X.

I will add an according explanation to the doc string.

sameberenz: i will also rename dst_transform to reference_transform to avoid confusion

tovogt: I still don't understand what you mean. How can a grid be "globally consistent" with a reference transformation if the resolution is overwritten by another value? What do you mean by "globally consistent"?

sameberenz:
With the latest [commit](https://github.com/CLIMADA-project/climada_python/pull/233/commits/10a7067fe01d47f3b754038ec470116ecaee3049), I have refractured the function `reproject_raster_data()` and  put the definition of grid into new function `set_global_consistent_grid()`, also updated the tests and docstrings. I hope this resolves the confusion and also meets your expectations, @emanuel-schmid ?

climada/util/coordinates.py
sameberenz: Given an equidistant 2D-grid defined by src_transform, crs and (optionally)
        src_shape, a destination grid with a new resolution (dst_resolution) is defined.
        The destination grid is globally consistent relative to a global origin.
        I.e., there is no offset in coordinates compared to another grid reprojected to
        the same dst_resolution based on a different src_transform.

tovogt:
Currently, there are three affine transformations flying around that confuse me: We have `src_transform` and `ref_transform` as specified by the user and `dst_transform` as returned by the function. Even more confusingly, the `ref_transform` is called `src_transform` in the helper function `set_global_consistent_grid`.

At first, I assumed that the function would reproject given "source" array raster data within the bounds of that "source" data set to a destination raster, possibly with a different CRS and resolution, but with the same geographic extent. The user would specify the target CRS, resolution and a global origin, and that would uniquely determine the target transformation already. From that perspective, there is no room for a `ref_transform` parameter.

Now, It seems to me like you want to optionally restrict the reprojection to some kind of window (or bounds) in the target CRS and you specify that window using a `ref_transform` and `ref_shape`. If that's what you really want, then I suggest to specify the window as bounds in the target CRS, for example, by adding a `dst_bounds` parameter (replacing `ref_transform` and `ref_shape`). This would also be the most natural input for the helper function `set_global_consistent_grid`. Specifying bounds in the form of a transform and shape is not very natural and the `ref_` prefix doesn't really make this clearer. A user can easily obtain the bounds from transform and shape using the function `rasterio.transform.array_bounds`, if necessary.

Furthermore, you seem to make a lot of manual computations that can actually be replaced by existing `rasterio` helper functions. Have a look at the docs: [rasterio.warp](https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html), [rasterio.transform](https://rasterio.readthedocs.io/en/latest/api/rasterio.transform.html), [rasterio.windows](https://rasterio.readthedocs.io/en/latest/api/rasterio.windows.html).

1. Set `global_transform = rasterio.transform.from_origin(global_origin, *dst_resolution)` (make sure that `dst_resolution` is a pair of floats for x and y resolution).
2. If `dst_bounds` is given, skip the next two steps and continue with (5).
3. Get the bounds of the source raster in the source CRS: `src_bounds = rasterio.transform.array_bounds(*source.shape, src_transform)`.
4. Translate the bounds of the source raster to the target CRS: `dst_bounds = rasterio.warp.transform_bounds(src_crs, dst_crs, *src_bounds)`.
5. If a buffer is given, manually enlarge the `dst_bounds` using the (known) resolution. However, you seem to have removed the buffer parameter in your latest commit.
6. Set `dst_window = rasterio.windows.from_bounds(*dst_bounds, global_transform)`.
7. Set `dst_window = dst_window.round_offsets(op='floor').round_shape(op='ceil')` to align the window to the global raster.
8. Set `dst_transform = rasterio.windows.transform(dst_window, global_transform)`.
9. Set `dst_shape = (int(dst_window.height), int(dst_window.width))`.

tovogt:
By the way, I suggest to change the name of the helper function `set_global_consistent_grid` because it doesn't "set" anything, but it generates a new pair of transformation and shape that describes a certain raster. ("Setting" usually means to change a given object in place.) My suggestion for a new name would be `subraster` or `get_subraster` or `compute_subraster`: Given a raster and bounds in the same CRS, the function returns a description (transform + shape) of the "subraster" within the selected bounds

sameberenz:
Thank you @tovogt , your suggestions worked well and I have rewritten and renamed the helper function to `subraster_from_bounds`. Also, I updated the tests and all seems to work out fine.
Are there any other suggestions for improvement?

tovogt (approved):
I cleaned up the docstrings and code a bit. Now it's ready to go, I think.

Issue #235, opened 2021-05-29
The rainfield result of TCRain() changes if the time_step_h of the TCTracks() is different


dhli003:
I used the TCRain().set_from_tracks(tracks, centroids) to calculate the intensity of the rainfall for tropical cyclone events. 
If I set TCTracks().equal_timestep(time_step_h=0.5) then the rainfall intensity for every single event is twice as large as that with time_step_h=1. 

In the line 180 of tc_rainfield.py (rainsum += _RCLIPER(track.max_sustained_wind.values[node], inreach, fradius_km)), the rainrate is summed up without multipling the time interval. 

I'm confused about how the unit of the result is changed from (mm/h) to (mm) in tc_rainfield.py

aleeciu:
Hi,
thanks. So the module assumed that tracks have 1h time steps. In fact, this was better documented in the matlab version https://github.com/davidnbresch/climada_module_tropical_cyclone/blob/master/docs/submodule_tc_rain.pdf. 

I now applied a time step correction and should be soon published in develop. Can you work with develop or do you need help for a work around with the current main? Essentially, as you indicate, one needs to multiply the time step of the current track node to the calcualted node's rain rate.

dhli003:
Thanks very much for the explanation and correction. The default 1h time step is actually good enough for me. I found it when I playing around with climada and thought it might be good to ask my question and study from the experts here.

emanuel-schmid:
solved with PR #236

PR #236, opened 2021-06-01
apply time step correction

aleeciu:
Applied time step correction, after #235

### DIFF
--- a/climada/hazard/tc_rainfield.py
+++ b/climada/hazard/tc_rainfield.py
@@ -178,7 +178,7 @@ def rainfield_from_track(track, centroids, dist_degree=3, intensity=0.1):
             fradius_km[pos] = np.sqrt(dd) * 111.12
 
             rainsum += _RCLIPER(track.max_sustained_wind.values[node],
-                                inreach, fradius_km)*track.time_step.values[node]
+                                inreach, fradius_km) * track.time_step.values[node]
 
     rainsum[rainsum < intensity] = 0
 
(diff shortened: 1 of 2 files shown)


climada/hazard/tc_rainfield.py
```diff
@@ -178,7 +178,7 @@ def rainfield_from_track(track, centroids, dist_degree=3, intensity=0.1):
             fradius_km[pos] = np.sqrt(dd) * 111.12
 
             rainsum += _RCLIPER(track.max_sustained_wind.values[node],
-                                inreach, fradius_km)
+                                inreach, fradius_km)*track.time_step.values[node]
```
tovogt, line 181: Linting:
```suggestion
                                inreach, fradius_km) * track.time_step.values[node]
```

emanuel-schmid:
Looks good. It would be awesome if we had it covered in a unit test. 😉

aleeciu:
> Looks good. It would be awesome if we had it covered in a unit test. 😉

Will do. So overall rainfall computed based on different timesteps can still differ by a bit. For randomly picked track for example, one could get 276259.044 mm or 276893.354 mm with 1h or 0.5h time step, respectively. What is the best way to assert such large absolute dfference but small relative difference in the unit test? assertAlmostEqual will fail I guess..

tovogt:
To test that the relative distance between the two is less than 1%, you could use something like this: 
```
np.testing.assert_allclose(276259.044, 276893.354, rtol=1e-2)
```

aleeciu:
> Looks good. It would be awesome if we had it covered in a unit test. 😉

done

aleeciu:
> To test that the relative distance between the two is less than 1%, you could use something like this:
> 
> ```
> np.testing.assert_allclose(276259.044, 276893.354, rtol=1e-2)
> ```

thanks

tovogt (approved):

PR #238, opened 2021-06-01
Feature/haz concat select

chahank:
Main points: extension of the `Hazard.concatenate` functionality to allow to concatenate hazards with different centroids. By default, the centroids are simply combined (for points just all points, for rasters a global raster containing all rasters). Optionally one can also provide a target centroids object. All Hazards are mapped onto the closest centroids.

Technically: Three* methods were added to the `Hazard.base` module. `map_matrix_to_centroids` to map hazards onto centroids, `concatenate_hazard` to concatenate a hazard list, and 'join_centroids' to find centroids common to a list of hazards.

Note1: unit tests are not yet written, waiting for general feedback first. Simple separate test cases worked well.
Note2: A bug in `Hazard.centroids.base` was corrected. `Centroids.size` now always returns an int instead of a float when it is a raster.
Note3: Docstrings in the whole file were updated to NumpyDoc format.


climada/hazard/base.py
```diff
@@ -1185,19 +1216,26 @@ def _event_plot(self, event_id, mat_var, col_name, smooth, crs_espg, axis=None,
...
-                a centroid
-            col_name (sparse matrix): Colorbar label
-            axis (matplotlib.axes._subplots.AxesSubplot, optional): axis to use
-            kwargs (optional): arguments for plot matplotlib function
-
-        Returns:
+        Parameters
+        ----------
+        centr_id: int
+        If centr_id > 0, plot mat_var
```
ChrisFairless, line 1222:
```suggestion
            If centr_id > 0, plot mat_var
```

chahank:
Made one extra function to join centroids of a list of hazards.

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+
+        Parameters
+        ----------
+        haz_src: list(climada.hazard.Hazard())
+            Hazard instances of the same hazard type
+        append : bool, optional
+            If True, append the haz_src to current object. If False,
+            replace all attributes from current objects by concatenated
+            hazards from haz_src. The default is False.
+        centtroids: climada.hazard.Centroids(), optional
```
ChrisFairless, line 1461:
```suggestion
        centroids: climada.hazard.Centroids(), optional
```

chahank:
Is this all @ChrisFairless ?

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+
+        """
+        if append:
+            haz_src = [self] + haz_src
+        self.__dict__ = copy.deepcopy(
+            concatenate_hazard(haz_src, centroids).__dict__
+            )
+
+
+def join_centroids(haz_list):
```
ChrisFairless, line 1477: It feels strange to have a method like this that takes hazards and returns centroids.
I'd suggest centroids in and centroids out (and maybe moving it to the centroids class)?
You'd then have to update line 1571 from
```
      centroids = join_centroids(haz_list)
```
to
```
      centroids = join_centroids([haz.centroids for haz in haz_list])
```

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+    """
+    Find centroids comon to all.
```
ChrisFairless, line 1479: "common to all" implies set intersection
```suggestion
    Create the union of centroids from the inputs.
```

climada/hazard/base.py
chahank: Good point.

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+    Returns
+    -------
+    centroids : climada.hazard.Centroids()
+        Centroids containing the centroids of all hazards.
+
+    """
+    centroids = copy.deepcopy(haz_list[0].centroids)
+    for haz in haz_list[1:]:
+        if not centroids.equal(haz.centroids):
+            centroids.append(haz.centroids)
```
ChrisFairless, line 1501: The duplicates removal on line 1502 might get expensive with large sets of large centroids (I didn't check), so it might also be worth running pairwise checks on the input centroid objects to spot any identical ones.

ChrisFairless:
How should `join_centroids` behave when it's given two centroids objects that have the same geometry, but differ by some other property, e.g. region_id?

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+
+    haz_concat = Hazard()
+    haz_concat.units = haz_list[0].units
+    haz_concat.centroids = centroids
+
+    #Indices for mapping matrices onto common centroids
+    hazcent_in_cent_idx_list = [
+        u_coord.assign_coordinates(haz.centroids.coord, centroids.coord)
+        for haz in haz_list
+        ]
```
ChrisFairless, line 1551: It might be useful if this object was somehow made available to the user? If I have some large exposures that I spent ages mapping to centroids, it'd be great to be able to update the with this instead of having to run `assign_coordinates` again.

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+
+    """
+    centroids = copy.deepcopy(haz_list[0].centroids)
+    for haz in haz_list[1:]:
+        if not centroids.equal(haz.centroids):
+            centroids.append(haz.centroids)
+    centroids.remove_duplicate_points()
+    return centroids
+
+def concatenate_hazard(haz_list, centroids=None):
```
ChrisFairless, line 1505: I'd really like to add the parameter `threshold`, which is then passed to `u_coord.assign_coordinates`.
- Right now it's unsettable, which means `assign_coordinates` uses its default value of 100 when matching centroids. The user should at least know this.
- If threshold is settable, then setting it to zero gives you a neat way to combine hazard objects while only retaining data defined exactly at the centroids you care about. (Though you could just subset the hazards before you combine them)

Additionally, I think the current code will fail when it can't find a centroid within the default `threshold` distance, since it'll try to map things to index -1.

ChrisFairless:
I'm thinking about the case when a user combines two Hazard objects with overlapping domains but different centroids. e.g. different resolution hazards, or hazards on differently offset grids.

Hazards like this shouldn't be combined without setting the `centroids` parameter. If they are, `combine_hazards` will mess up the footprints, since it will add a bunch of centroids with zero hazard amongst the ones with defined hazards. Can we expect users to recognise this risk, or should we add some sort of warning if we can detect it happening?

climada/hazard/base.py
chahank: I agree this could be the case, but how expensive is the pair checking between all pairs? The number of pairs growths quickly. So, while I see your point, I am not convinced this is necessary. Do you want me to do some speed tests?

I here test one edge case when all centroids basically are the same (i.e. when someone is merging hazard defined on the same centroids).

climada/hazard/base.py
chahank: Again, good points! I can easily add that possibility. 

Note however that the use case you propose would not work (precisely for the reason you said "Additionally, I think the current code will fail when it can't find a centroid within the default threshold distance, since it'll try to map things to index -1.").

Hence, a use case would be  actually the opposite: if you want to assign hazard centroids that are very fare from
the user defined centroids, you can make threshold larger. I  do not really see that this would lead to good results, but it is good for the user to be able to tune that. 

I could catch the case when centroids are beyond the threshold and inform the user that they cannot merge.

chahank: "If threshold is settable, then setting it to zero gives you a neat way to combine hazard objects while only retaining data defined exactly at the centroids you care about. (Though you could just subset the hazards before you combine them)"
This is not possible with the current implementation and would represent some extra work that for me doesn't fit into this pull request as it would require the possibility to select subsets of hazards.

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+    counter = np.array_split(np.arange(n_events), chunks + 1)
+    mapped_matrix = sparse.csr_matrix([])
+    events = []
+
+    for row in counter:
+        for n in row:
+            ar = np.zeros(int(centroids.size))
+            ar[cent_idx] = sp_matrix.getrow(n).todense()
+            events.append(sparse.csr_matrix(ar))
+        mapped_matrix = sparse.vstack(events)
```
ChrisFairless, line 1607: My sparse matrix know-how isn't the best, so I wanted to ask if there's a neater way to do this, since you're essentially doing a column permutation. Could you use a permutation matrix instead (addding any necessary all-zero columns as well)? Is there a sparse-specific way to do this?

ChrisFairless:
Do we need any new tests?

climada/hazard/base.py
chahank: Interesting idea!

Two potential objections:
- it is quite an edge case, and returning this value only for these edge cases is a strong hit to user friendliness imho.
- `u_coord.assign_coordinates` is not thaaat time consuming normally. In particular, if the centroids and the exposures lat/lon match, it is very fast.

What do you think?

climada/hazard/base.py
ChrisFairless: I think the `centroids.equal` method is very cheap to run. It checks for centroids being the same size first (almost all differing centroids will fail this), and then checks if the centroid `lat` and `lon` arrays are (near-)identical. So unless you're combining hundreds of hazard objects, it should be fast.

Having said that, maybe it doesn't matter - I assume finding the duplicate centroids isn't the slowest part here. I'm happy to leave this alone and wait to see if someone needs to speed it up later.

climada/hazard/base.py
ChrisFairless: If you're working with (near-)global hazards and exposures defined on different grids `assign_coordinates` takes forever for me. So I'd looove to avoid running it again when combining hazards.

Having said that, I can't see a really user-friendly way to provide this.

One option would be to add `hazcent_in_cent_idx_list` as an additional, optional input parameter (maybe with a better name), and then split these lines into a separate utility function. Then the user has the choice to calculate these mapping manually if they're required, and then pass them to the function. But in in most cases you just call the function and it's calculated as it is right now.

Counter-argument: it's an edge case and it's easy for me to write my a custom method when I need it.

chahank: Yes, I was also thinking of a separate util function, but came to the conclusion that it is a) quite easy to write b) not broadly useful. Hence, I kept it this way. 

"If you're working with (near-)global hazards and exposures defined on different grids assign_coordinates takes forever for me. So I'd looove to avoid running it again when combining hazards."

But in this case what would you want to concatenate which has different centroids?

climada/hazard/base.py
chahank: Then I would keep it this way. Thanks!

climada/hazard/base.py
ChrisFairless: I forgot to suggest that when the method can't find a nearby centroid within the threshold (and returns -1) I think you should drop that centroid (with a logging message). It wouldn't be too hard to snip out those columns when you're rearranging the sparse matrix.

You're right though - if you throw an error when you get a -1, then that second bullet point doesn't work. I'm also fine with you throwing an error - it's not a bad idea to make it the user's responsibility that their inputs will map sensibly.

climada/hazard/base.py
chahank: I first did not do that because there is already the method `centroids.append`. Hence, I only made this to join centroids from hazards. 

Would you still say a join  method for `centroids` is better?

climada/hazard/base.py
chahank: I think I prefer the error message. Otherwise, by simply dropping centroids the data is modified in a rather intransparent way.

chahank: The threshold has been added and a warning if beyond the threshold too. f3a872c2da5a8af9d270ba1d7b994d378d3607a9
Also for `hazard.concatenate` here 8e5422a2b8882f01874200312a7cadc0831f25e3

climada/hazard/base.py
```diff
@@ -1393,3 +1442,167 @@ def _read_att_excel(self, file_name, var_names):
...
+    return haz_concat
+
+def _map_matrix_to_centroids(sp_matrix, centroids, cent_idx):
+    """
+    Map a sparse matrix (e.g. hazard intensity or fraction) onto centroids.
+
+    Parameters
+    ----------
+    matrix: scipy.sparse.crs_matrix()
+        Input hazard. haz.centroids must be a subset of centroids
```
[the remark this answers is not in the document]
chahank: Updated docstring here bc869ef1fb92a1ad51f8b8ba27ac1629e8bbc0e5

climada/hazard/base.py
ChrisFairless: I would say so, but this version is completely fine if you don't want to put much more time into it.

climada/hazard/base.py
ChrisFairless: OK let's leave it as it is then.

(And to answer your question, I don't have a convincing use case yet, beyond maybe combining very high resolution datasets from separate basins)

climada/hazard/base.py
```diff
@@ -1393,3 +1442,204 @@ def _read_att_excel(self, file_name, var_names):
...
+    haz_concat.centroids = centroids
+
+    #Indices for mapping matrices onto common centroids
+    hazcent_in_cent_idx_list = [
+        u_coord.assign_coordinates(haz.centroids.coord, centroids.coord,
+                                   threshold=threshold)
+        for haz in haz_list
+        ]
+
+    if -1 in hazcent_in_cent_idx_list:
```
ChrisFairless, line 1582: isn't this a list of lists?
```suggestion
    if any([-1 in idx_list for idx_list in hazcent_in_cent_idx_list]):
```

climada/hazard/base.py
ChrisFairless: Nice

climada/hazard/base.py
chahank: OK, I will put it there then!

climada/hazard/base.py
chahank: Well spotted, I went too fast. Thanks!

chahank:
> I'm thinking about the case when a user combines two Hazard objects with overlapping domains but different centroids. e.g. different resolution hazards, or hazards on differently offset grids.
> 
> Hazards like this shouldn't be combined without setting the `centroids` parameter. If they are, `combine_hazards` will mess up the footprints, since it will add a bunch of centroids with zero hazard amongst the ones with defined hazards. Can we expect users to recognise this risk, or should we add some sort of warning if we can detect it happening?

You are right. When I first thought about this case, I came to the (wrong) conclusion that it doesn't matter. I will think of whether a good warning can be made.

chahank:
> How should `join_centroids` behave when it's given two centroids objects that have the same geometry, but differ by some other property, e.g. region_id?

So far simply behaves as is defined in `Centroids.append`.  Which means the centroids are just appended one to another, and then dupplicates are removed.  The `Centroids.remove_duplicates` seems to ignore anything else than geometry.

I think this is fair. Centroids with the same geometry but different attributes  do not make sense to me. Now, do we need to babysit  the use in this case?

ChrisFairless:
> You are right. When I first thought about this case, I came to the (wrong) conclusion that it doesn't matter. I will think of whether a good warning can be made.

I couldn't think of an easy way to do it, but maybe a warning that has some false positives is still worthwhile ... let me know what you think

> I think this is fair. Centroids with the same geometry but different attributes do not make sense to me. Now, do we need to babysit the use in this case?

Thanks for this answer. Let's not babysit :)

On static methods, I would trust your (and the others') python instincts here.

climada/hazard/centroids/centr.py
```diff
@@ -538,6 +538,35 @@ def append(self, centr):
                 setattr(self, var_name, np.append(var_val, centr_val).
                         astype(var_val.dtype, copy=False))
 
+    @staticmethod
+    def join_centroid(centroids_list):
```
tovogt, line 542: I suggest to apply the same logic as the built-in `set.union` method (https://docs.python.org/3/library/stdtypes.html#frozenset.union): Don't make it static, but always return a new object. That way you can call it just like a static method if needed. Moreover, why not call it `union`?
```suggestion
    def union(self, *others):
```

climada/hazard/base.py
tovogt: In fact, this is a "one-liner" with the CSR format because the number and values of non-zero entries don't change:
```
sparse.csr_matrix((sp_matrix.data, cent_idx[sp_matrix.indices], sp_matrix.indptr),
                  shape=(sp_matrix.shape[0], centroids.size))
```
Since this functionality is used only once, I wouldn't bother defining a new function for it.

tovogt:
About the `concatenate_hazard` naming: Unfortunately, we already have a method called `concatenate` that should actually be called `extend` (like Python's built-in `list.extend`) because it always modifies the current object. The name `concatenate` is reminiscent of NumPy's concatenate which always creates a new object

About overlapping centroids: For point-based centroids it's really hard to check for overlaps. You would have to take the convex hull or something similar (which might be hard to define on a sphere). On the other hand, grid-based centroids shouldn't be supported by `Centroids.append` at all, in my opinion. Currently, variables like `on_land` or `dist_coast` are just concatenated. But this is completely wrong for almost all combinations of grid-based centroids I can think of. In the long run, I would drop support for grid-based centroids anyway, because they don't actually seem to be used anywhere.

climada/hazard/centroids/centr.py
ChrisFairless: Yeah I think union is a more precise name, unless there are analogous `join_` methods elsewhere in CLIMADA?

chahank: I agree with `union` and not making it static is fine for me. Although I personally do not like the logic to have a non-static method that always returns a new object of the same type. This is for me very counter-intuitive.

chahank:
> About the `concatenate_hazard` naming: Unfortunately, we already have a method called `concatenate` that should actually be called `extend` (like Python's built-in `list.extend`) because it always modifies the current object. The name `concatenate` is reminiscent of NumPy's concatenate which always creates a new object. Maybe we should consider dropping backwards compatibility in this case...

The current `Hazard.concatenate` is actually even worse. It does by default erase the current object and replace its content with the concatenation (with the current hazard only included if it is in the list). I agree that extend would be a better name. 

But then should there be both a `hazard.concatenate` and a `hazard.extend` method? Would which returns a new object and one which extends the current hazard?

chahank:
> About overlapping centroids: For point-based centroids it's really hard to check for overlaps. You would have to take the convex hull or something similar (which might be hard to define on a sphere). On the other hand, grid-based centroids shouldn't be supported by `Centroids.append` at all, in my opinion. Currently, variables like `on_land` or `dist_coast` are just concatenated. But this is completely wrong for almost all combinations of grid-based centroids I can think of. In the long run, I would drop support for grid-based centroids anyway, because they don't actually seem to be used anywhere.

I agree. One way would be to convert grids to points in the `Centroids.append` method, but this is not that clean. I agree that the appending does not work correctly now with the raster. I see two solutions: 
a) drop support for rasters
 b) make the mapping of attributes such as `on_land` correct similarly as the mapping of the intensity matrix for the concatenate hazards.

What do you think?

climada/hazard/centroids/centr.py
chahank: Also, should the method now make the union only of the `others` or include the current `Centroids` object too?

climada/hazard/base.py
chahank: Having it as a separate function makes it quite a bit more readable, and it might be use elsewhere. But, I am also fine with having this one-liner too.

climada/hazard/centroids/centr.py
```diff
@@ -538,8 +538,7 @@ def append(self, centr):
                 setattr(self, var_name, np.append(var_val, centr_val).
                         astype(var_val.dtype, copy=False))
 
-    @staticmethod
-    def join_centroid(centroids_list):
+    def union(self, *centroids_list):
```
ChrisFairless, line 541: what about calling it `union_centroids`?
also you need to change line 1540 of hazard/base.py where it's called with its old name

ChrisFairless: no wait i'm happy with just `union`.
i really need to think twice before expressing opinions today.

chahank: :D and I have to see not to push commits that are unfinished. Thanks for the line 1540 reminder.

chahank:
Quick note: I pushed a few commits too quick and need to debug now. Sorry.

climada/hazard/centroids/centr.py
tovogt: Well, if that's counter-intuitive for you, you didn't work with pandas too much, did you? :smile: Mathematical operations are translated in python to be method calls of objects. So, `a + b` is translated to `a.__add__(b)`.  Following that scheme, the operation `A ∪ B` is translated to `A.union(B)`. And naturally, the results are new objects. So, in NumPy, when you have two arrays `x` and `y`, and you write `z = x + y` then this is the same as  `z = x.__add__(y)` and still, `z` is not the same object as `x`. Are you fine with that?

It should work just like the built-in `set.union` in my opinion. So, `self` is of course included.

tovogt:
> But then should there be both a `hazard.concatenate` and a `hazard.extend` method? Would which returns a new object and one which extends the current hazard?

Yes, I would let `Hazard.concatenate` be a static method, and `Hazard.extend` a non-static method that uses `Hazard.concatenate` and replaces `Hazard.__dict__` with the result (like you already do).

> I see two solutions:
> a) drop support for rasters
> b) make the mapping of attributes such as `on_land` correct similarly as the mapping of the intensity matrix for the concatenate hazards.

I say, drop support for rasters in the `append` method. Either by raising an error, or by automatically converting grid-based centroids to point-based centroids before appending. I don't really see the need for grid-based centroids anywhere in CLIMADA.

chahank:
> I say, drop support for rasters in the `append` method. Either by raising an error, or by automatically converting grid-based centroids to point-based centroids before appending. I don't really see the need for grid-based centroids anywhere in CLIMADA.

Yes, I agree. I just started doing this and realized that then basically `Centroids.append` and `Centroids.union` would do almost the same thing. Should we then just keep one of the two?

tovogt:
> Yes, I agree. I just started doing this and realized that then basically `Centroids.append` and `Centroids.union` would do almost the same thing. Should we then just keep one of the two?

I know that `pandas` has an `append` method that returns a new object, unlike the built-in `list.append` that returns `None` and appends the new element in-place. Intuitively, I would say that `append` should work in-place and modify the current object, while `union` is a binary (or more) operation that returns a new object. From that perspective, the way forward would be to keep both methods but have `Centroids.append` call `Centroids.union` and assign the result to `self.__dict__`.

chahank:
That sounds like a good idea to me! Thanks for the suggestion. Unless @ChrisFairless objects, I will proceed this way.

chahank:
I now made the following b54ac861183c155e1fe1be036ecff4a8a8ed2338 :

`Centroids.append(centr)` modifies the object in place and appends one single centroids centr.
`Centroids.union(centroids_list)` is a static method that makes the union of all Centroids in the centroids_list and returns a Centroids object.

Raster data is automatically converted to points and raster information is deleted.

tovogt:
I would rather have `Centroids.union` behave like the built-in `set.union` (see https://github.com/CLIMADA-project/climada_python/pull/238#discussion_r649966460): It shouldn't be a static method and it should have the signature `union(self, *others)`.

Furthermore, I would prefer to have `Centroids.union` do the concatenation of attribute values and call `remove_duplicate_points` once at the end. The method `Centroids.append` would then call `Centroids.union` and apply the result to `self`. This way, we make sure that `remove_duplicate_points` is called only once.

chahank:
@tovogt : Why should be prefer to have `centroids.union()` to asked several arguments instead of asking for a interable? Or should we account for both?

tovogt:
You can't really argue about that. I think that Centroids are similar to sets (of points), so I would like to keep the `union` API similar to the `set.union` functionality, which seems to be optimized for the union of two (or a few) explicit sets (like `a.union(b)` or `set.union(a, b, c)`). But if you feel like the main usecase of `Centroids.union` will be with many inputs that are not explicitly named, why not support both, like `min` does (like `min(1, 3)` or `min([1, 3])`)?
```
def union(self, *others):
    # ...
    if len(others) == 1 and not isinstance(others[0], Centroids):
        others = others[0]
    # ...
```

emanuel-schmid:
Nah. Don't. Having a fixed argument data type (not having to `isinstance(...)` beats the little bit of more comfort imho.

chahank:
Actually, there are 2 reasons: 1- set.union([a, b, c]) works well, hence it should remain. 2- CLIMADA should remain consistent with itself, and no other class has so far a union method of this type as far I know. 

Anyway, I am looking to satisfy your proposition, but it might come with some not necessarily wanted side effects. I will update the code soon to showcase what I mean there.

chahank:
Then I would opt for allowing only iterables rather than a number of arguments... because most developers know how to make a list, but not all know how to make a number of arguments from a list.

tovogt:
1) `set.union([a, b, c])` doesn't work, so I don't know what you want to point out.
2) No other class in CLIMADA has a union method, so there is no argument, right?

tovogt:
You just have to read the function signature to learn immediately how you make several function arguments out of a list. And that's really worth learning if you didn't know this before.

chahank:
1. set.union([1,2,3]), but set.union([{1}, {2}, {3}] does indeed not work. 
2. There are different append methods which do something closer to union than to append

For the use of *list. I must disagree. Even though it is good that programmers know that, it is very far from obvious from the method signature unless one already knows what it is about.

In the end, it does not matter that much though.

tovogt:
I still don't get your first point. Neither `set.union([1,2,3])` nor `set.union([{1}, {2}, {3}])` work in Python 3.8 or 3.9.

chahank:
For me (Python 3.8) the following code works:

```
>>> s = set([1, 2, 3])
>>> s.union([3, 4])
{1, 2, 3, 4}
```

I hope this now clarifies what is meant.

tovogt:
I see that there is some kind of typecasting where `s.union([3, 4])` is translated to `s.union({3, 4})` internally. But I don't see what you want to point out...?

chahank:
Just to make clear what I meant, even though it does not apply any more (I confused with the case `set.union ([{1}, {2}])`)

chahank:
Ok, I update the `Centroids.union` and `Centroids.append` methods. There are the following points that should be discussed (beyond the code in itself):

- In union it is checked whether the centroids are equal with `centroids.equal(cent)`. This method calls `u_coord.equal_crs(self.geometry.crs, centr.geometry.crs)`. Now, `centroids.union(cent)` is called in `TropicalCyclone.set_from_tracks()` with the situation that `self.geometry.crs` is defined, but `centr.geometry.crs` not. Then, the union fails. This was previously not the case with the `centroids.append` method that did not check for equal centroids. Solutions are: 1. adapt `TropCyclone.set_from_tracks()` 2. drop the requirement that the `Centroids.geometry.crs` are equal for the `Centroids.union()` 3. Accept union if one crs is defined and the others are empty. 

- The `Centroids.union()` method modifies the input centroids currently. First, it converts meta to lat/lon. This cannot really be avoided. One could make a copy of all the input centroids, then apply `.set_meta_to_lat_lon()`. But this creates a lot of copies.  Second set geometry points which is needed to compare the `.geometry.crs` in `.equal()`. The later could be dropped if we do not check for equal `.crs`. Otherwise it is a bit tricky since `Centroids.geometry` is not a required attribute.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
+        if self.meta and not self.lat.any():
+            self.set_meta_to_lat_lon()
+        if centr.meta and not centr.lat.any():
+            centr.set_meta_to_lat_lon()
+
+        if not u_coord.equal_crs(centr.geometry.crs, self.geometry.crs):
+            raise ValueError('Different CRS not accepted.')
+
+        self.meta = dict()
+        self.__dict__.update(self.union(centr).__dict__)
```
tovogt, line 516: I think we should implement `Centroids.union` in such a way that we can reduce this to a single line:
```suggestion
        self.__dict__ = self.union(centr).__dict__
```

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
+                    raise ValueError('Different CRS are not accepted.')
+            if not centroids.equal(cent):
+                # append all 1-dim variables
+                for (var_name, var_val), centr_val in \
+                    zip(centroids.__dict__.items(), centroids.__dict__.values()):
+                    if isinstance(var_val, np.ndarray) and var_val.ndim == 1:
+                        setattr(centroids, var_name,
+                                np.append(var_val, centr_val).astype(
+                                    var_val.dtype, copy=False
+                                    )
```
tovogt, line 563: Don't use `np.append` iteratively, but collect all the data and apply `np.cocatenate` at the end.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
+
+        centroids.set_geometry_points()
+        for cent in others:
+            if cent.meta and not cent.lat.any():
+                cent.set_meta_to_lat_lon()
+            cent.set_geometry_points()
+            if not u_coord.equal_crs(cent.geometry.crs,
+                                      centroids.geometry.crs):
+                    raise ValueError('Different CRS are not accepted.')
+            if not centroids.equal(cent):
```
tovogt, line 555: Why do we need to check for equality? Is this really a common use case where the user lists a whole Centroids object twice or even several times?

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
+            centroids = copy.deepcopy(others[0])
+            others = others[1:]
+        if centroids.meta and not centroids.lat.any():
+            centroids.set_meta_to_lat_lon()
+
+        centroids.set_geometry_points()
+        for cent in others:
+            if cent.meta and not cent.lat.any():
+                cent.set_meta_to_lat_lon()
+            cent.set_geometry_points()
```
tovogt, line 551: I think we should use `set_geometry_points` only if at least one of the centroids in the list has a `geometry` attribute set.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
+            if not u_coord.equal_crs(cent.geometry.crs,
+                                      centroids.geometry.crs):
+                    raise ValueError('Different CRS are not accepted.')
```
tovogt, line 554: Centroids needs a `crs` @property just like recently implemented by @emanuel-schmid for Exposures. The CRS will be taken from the `geometry` attribute if it's available and from `meta` otherwise and default to `DEF_CRS`.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,83 @@ def clear(self):
...
-        for (var_name, var_val), centr_val in zip(self.__dict__.items(),
-                                                  centr.__dict__.values()):
-            if isinstance(var_val, np.ndarray) and var_val.ndim == 1 and \
-            var_name not in ('lat', 'lon'):
-                setattr(self, var_name, np.append(var_val, centr_val).
-                        astype(var_val.dtype, copy=False))
+            centroids = copy.deepcopy(others[0])
+            others = others[1:]
+        if centroids.meta and not centroids.lat.any():
+            centroids.set_meta_to_lat_lon()
```
tovogt, line 545: Why do we separate out one of the `Centroids` objects in the beginning? I would start off from an empty object `centroids = Centroids()`.

tovogt (commented):
I posted some remarks directly in the code.

tovogt:
About changing the input objects with `set_meta_to_lat_lon` and `set_geometry_points`. In my opinion, this kind of change is okay because of two reasons:

1) These methods don't change existing data, but only set attributes that were previously unset. (Yes, strictly speaking, the `lat` attribute was set to an empty array before, but that's somehow equivalent to being unset.)
2) In the long run, I would prefer to deprecate grid-style centroids and only have point-style centroids anyway.

If you want, you can log a warning about these changes.

climada/hazard/centroids/centr.py
chahank: Ok, I'll see to do that. I copied that bit blindly from the older code.

climada/hazard/centroids/centr.py
chahank: For example this is how `TropCyclone.set_from_tracks()` works.

climada/hazard/centroids/centr.py
chahank: I tried that but returned empty values in the script of `TropicalCyclones.set_from_tracks()`. However, you are perfectly right. I will give it another try.

climada/hazard/centroids/centr.py
chahank: That is for the case that one of the centroids `self`  instance is empty.

climada/hazard/centroids/centr.py
chahank: Currently `Centroids().crs` returns `None` by default (without meta and geometry). Should this be changed to `DEF_CRS`?

chahank:
Ok, I think I addressed all concerns and added the needed tests if  they were missing. In addition, I had to change a few things more:

1. some centroid methods failed on empty Objects. This was changed.

2. Hazard.append now calls Hazard.extend

3. centroids.union computes optional variables if at least one centroids has them

Points to discuss:

- should centroids.union compute optional variables if at least one centroids has them and one not, or procduce and exception instead? (currently, optional variables are computed. Code is still ugly)

- Is the renaming of `Hazard.concatenate` to `Hazard.extend` a wise choice? There are now 3 methods to handle hazards: `Hazard.concatenate`, `Hazard.extend` and `Hazard.append`. I propose to drop `Hazard.extend` and only keep `Hazard.append`.

chahank:
The `.geometry` is now always computed explicitly before the concatenation of attribute is done, as the methods `.set_on_land` and `.remove_duplicate_points` compute it anyway.

climada/hazard/centroids/centr.py
```diff
@@ -548,15 +553,19 @@ def union(self, *others):
...
+            if is_area_pixel and not cent.area_pixel.any():
+                cent.set_area_pixel()
+            if is_dist_coast and not cent.dist_coast.any():
+                cent.set_dist_coast()
+            if is_on_land and not len(cent.on_land) > 0:
+                cent.set_on_land()
+            if is_region_id and not cent.region_id.any():
+                cent.set_region_id()
+            if is_elevation and not cent.elevation.any():
+                cent.set_elevation()
```
ChrisFairless, line 568: since none of these values are required to remove duplicates, is there a smart way to calculate them _after_ everything is combined and duplicates are removed? for large sets of centroids with a lot of overlapping points, it could save time.

this might be a bit of an edge case that we don't need to worry about: i don't know when you'd want to combine _N_ centroids objects where 1 < _x_ << _N_ of the objects have this calculated already.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,98 @@ def clear(self):
...
+
+        Parameters
+        ----------
+        others : any number of climada.hazard.Centroids()
+            Centroids to form the union with
+
+        Returns
+        -------
+        centroids : climada.hazard.Centroids()
+            Centroids containing the union of the centroids in centroids_list.
```
ChrisFairless, line 532: This is out of date

climada/hazard/centroids/centr.py
chahank: I tried to just put it at the end, and then realized why I put it where it is now. The `centroids.remove_duplicate_points` will select from all the numpy arrays. Hence, if one of them is not defined, it fails. 
As a compromise, the method only computes the missing values if they are missing :D

climada/hazard/centroids/centr.py
chahank: Why? I am confused...

chahank: Resolved here 
6eb8af4

climada/hazard/centroids/centr.py
ChrisFairless: I'm trying to find where that happens. Is it in `Centroids.select()`, lines 814-821 in centr.py?

This isn't very important though - in the edge case I'm describing the the user can plan around accidental large calculations, since the documentation is clear about what the code does.

climada/util/coordinates.py
```diff
@@ -890,11 +890,17 @@ def assign_coordinates(coords, coords_to_assign, method="NN", distance="haversin
         Index into `coords_to_assign`. Note that the value `-1` is used to indicate that no
         matching coordinate has been found, even though `-1` is a valid index in NumPy!
     """
+
+    if not coords.any() or not coords_to_assign.any():
+        return np.array([])
```
ChrisFairless, line 895: If `coords` is empty, I'd expect to return an arrray the size of `coords_to_assign` with all values set to -1

climada/hazard/centroids/centr.py
```diff
@@ -110,7 +110,7 @@ class Centroids():
     def __init__(self):
         """Initialize to None raster and vector"""
         self.meta = dict()
-        self.geometry = gpd.GeoSeries()
+        self.geometry = gpd.GeoSeries(crs=DEF_CRS)
```
ChrisFairless, line 113: I'm having a quick look to see if this might have unexpected consequences anywhere. Could you check out `Centroids.set_vector_file`? I think a default CRS means you now can't read in vector files without changing the empty `geometry`'s CRS to match the file, which is very unintuitive.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,98 @@ def clear(self):
         self.__init__()
 
     def append(self, centr):
-        """Append raster or points.
+        """Append centroids points.
+
+        If centr is raster it is converted to points first using
```
ChrisFairless, line 496:
```suggestion
        If self or centr are rasters they are converted to points first using
```

ChrisFairless: Also probably worth mentioning this is a wrapper around `Centroids.union` and pointing to that documentation

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+        If centroids is not None, all hazards are defined on these centroids.
+        Centoids of hazards in haz_list not in centroids are mapped onto the
+        nearest point. If a point is further than threshold from the closet
+        centroid, the concatenation fails.
+
+        Parameters
+        ----------
+        haz_list: list(climada.hazard.Hazard())
+            Hazard instances of the same hazard type
+        centtroids: climada.hazard.Centroids(), optional
```
ChrisFairless, line 1480:
```suggestion
        centroids: climada.hazard.Centroids(), optional
```

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+            elif (isinstance(attr_val_list[0], np.ndarray)
+                  and attr_val_list[0].ndim == 1):
+                setattr(haz_concat, attr_name, np.hstack(attr_val_list))
+            elif isinstance(attr_val_list[0], list):
+                setattr(haz_concat, attr_name, sum(attr_val_list, []))
+            elif isinstance(attr_val_list[0], TagHazard):
+                for tag in attr_val_list:
+                    if tag is not haz_concat.tag: haz_concat.tag.append(tag)
+
+        haz_concat.sanitize_event_ids()
```
ChrisFairless, line 1552: Is there a use case for providing a `sanitize_event_ids=True` parameter in this method, making this line optional (and instead throwing an error if the combined IDs aren't unique)? Would a user ever want event IDs to persist over the concatenation? (Or is the ID just an index variable, and they should use unique event names for this?)

climada/hazard/base.py
```diff
@@ -1101,41 +1112,6 @@ def read_hdf5(self, file_name):
 
         hf_data.close()
 
-    def concatenate(self, haz_src, append=False):
```
ChrisFairless, line 1104: I checked through the repo (and the papers repo) for anywhere that still uses this. I think the emulator tutorial still does and that's it

ChrisFairless: If you haven't already, we should mention that this is being removed in the release notes, because it could break things on branches and forks we haven't checked/can't see

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,98 @@ def clear(self):
...
-            self.lat, self.lon = np.array([]), np.array([])
+        self.__dict__.update(self.union(centr).__dict__)
+
+
+    def union(self, *others):
+        """
+        Create the union of centroids from the inputs.
+
+        The centroids are combined together point by point.
+        Rasters are converted to points and raster information is lost.
```
ChrisFairless, line 514:
```suggestion
        Rasters are converted to points and this object's self raster information is lost.
```
Is this more accurate? The `*others` don't lose `meta` information, if I understand right.

climada/hazard/centroids/centr.py
```diff
@@ -491,52 +491,98 @@ def clear(self):
         self.__init__()
 
     def append(self, centr):
-        """Append raster or points.
+        """Append centroids points.
+
+        If centr is raster it is converted to points first using
+        Centroids.set_meta_to_lat_lon. Raster information will be lost.
```
ChrisFairless, line 497: See other comment
```suggestion
        Centroids.set_meta_to_lat_lon. Raster information in `self` will be lost.
```

ChrisFairless:
I've left a few more remarks, but other than that I think we're ready to merge.

I also wanted to check: for concatenating hazards with disjoint but spatially overlapping overlapping centroids, we decided not to run checks and warnings because it's hard/unreliable?

climada/hazard/centroids/centr.py
chahank: Good question. But to: "I think a default CRS means you now can't read in vector files without changing the empty geometry's CRS to match the file, which is very unintuitive." Then we should simply change the `read_vector_files` method right?

climada/hazard/base.py
chahank: Hazard ids is just for internal indexing. The events are characterized by name and by date.

climada/hazard/base.py
chahank: That is a good point. It will be mentioned in the release notes. Should I make a deprecation notice for `self.concatenate` maybe?

climada/hazard/centroids/centr.py
chahank: Actually, only the ouput raster information is lost. So, neither the objects self, not the others.

chahank:
>I also wanted to check: for concatenating hazards with disjoint but spatially overlapping overlapping centroids, we decided not to run checks and warnings because it's hard/unreliable?

Exactly. Also, it is a quite weird edge case that should be handled by the user.

climada/hazard/centroids/centr.py
chahank: But I can remove it again.

climada/util/coordinates.py
chahank: Done here: 
59a0894

climada/hazard/centroids/centr.py
chahank: Done here 
173533f

climada/hazard/base.py
chahank: Also, what do you think about the naming:

There is now `hazard.concatenate_hazard`, `hazard.extend` and `hazard.append`. I think this is too much.

climada/hazard/centroids/centr.py
chahank: @emanuel-schmid : do you think it is a good idea to a default CRS when initiating a `Centroid`?

chahank: The main reason why I added a default CRS is for the method `Centroids.equal` which looks for equal CRS. If `Centroids.crs` is None, then `Centroids.equal(Centroids2)` will give an error 

```
  File "/climada_python/climada/hazard/centroids/centr.py", line 562, in union
    if not u_coord.equal_crs(cent.crs, self.crs):
  File "/climada_python/climada/util/coordinates.py", line 1324, in equal_crs
    return rasterio.crs.CRS.from_user_input(crs_one) == rasterio.crs.CRS.from_user_input(crs_two)
  File "anaconda3/envs/climada_210/lib/python3.8/site-packages/rasterio/crs.py", line 498, in from_user_input
    raise CRSError("CRS is invalid: {!r}".format(value))
rasterio.errors.CRSError: CRS is invalid: None

Comment (chahank) [reply climada/hazard/centroids/centr.py]: So, just to make it simple, I removed the default CRS and added it if needed in the Union part. 
a8be18d

File: climada/hazard/base.py (hunk 26 above)
Comment (chahank) [reply climada/hazard/base.py:1104]: @emanuel-schmid Any thoughts on the naming and the note about removing the method `self.concatenate` ?

**Comment by chahank** [comment]
Thanks @ChrisFairless  For the detailed review! If you could give your opinion about the naming of the `Hazard` methods, then I can close this pull request.

File: climada/hazard/base.py (hunk 26 above)
Comment (emanuel-schmid) [reply climada/hazard/base.py:1104]: about the deprecation notice - what do you mean? keep the method and log a warning like with the all but removed `IF_xy methods`? Would that even be an option in this case?

File: climada/hazard/centroids/centr.py (hunk 22 above)
Comment (emanuel-schmid) [reply climada/hazard/centroids/centr.py]: > 
> 
> @emanuel-schmid : do you think it is a good idea to a default CRS when initiating a `Centroid`?

Actually I do. While we don't have to insist that users define a CRS, it makes things easier when a geometry object always has a well defined CRS nevertheless. 
Apart from that, Exposures do have a default CRS as well, so it supports consistency within CLIMADA.

File: climada/hazard/base.py [hunk 29]
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+        Returns
+        -------
+        None.
+
+        """
+        if self.event_id.any(): #if self is not empty, concatenate with self.
+            haz_list = [self] + haz_list
+        self.__dict__ = self.concatenate_hazard(haz_list,
+                                                centroids,
+                                                threshold).__dict__
```
emanuel-schmid, line 1458: Replacing the `__dict__` object looks mighty dangerous to me. Are we _sure_ nothing unpredicted can happen here? Is there no proper way of re-initialization?

climada/hazard/centroids/centr.py
ChrisFairless: I think a default CRS is a good idea too, and I liked the implementation. My comment here was to point out that this method probably needs updating so that it can still read a file with a non-default CRS onto an empty Centroids object.

climada/hazard/base.py
ChrisFairless: @emanuel-schmid isn't the analogy with `list` fairly close? I see `append` and `extend` as operations that take either a single element or an iterable of elements and combine them into `self`.

ChrisFairless: @chahank I think this is probably too many methods, yes. With `hazard.concatenate` we've a few options:
- Remove it, as we're doing now, and hope nothing breaks in places we can't see
- Add a deprecation notice and remove it in a future release
- Just add a deprecation notice
The second of these is probably 'best practice' but our user base is not so big we couldn't do the first.

With `extend`, `append`, and `concatenate_hazard`, it is a lot. I think it's worth pointing out similar methods in the docstrings with a 'See also...', so users are aware of them. I'll add these as suggestions now.

We could also think about deprecating `hazard.append` because `hazard.extend` is almost always better. A deprecation notice would be annoying to anyone using the method, but in the long run it would encourage (slightly) better code ... I'm not sure it's worth our (and our users') time, though.

Finally, should we add the parameters `centroids` and `threshold` to `hazard.append` for consistency? It won't break backward compatibility.

climada/hazard/base.py
```diff
@@ -910,53 +948,17 @@ def calc_year_set(self):
     def append(self, hazard):
         """Append events and centroids in hazard.
```
ChrisFairless, line 949:
```suggestion
        """Append the events and centroids in hazard.
        To append multiple hazard objects, use `hazard.extend`.
        To combine a list of hazards and return a new object, use the
        static method `hazard.concatenate_hazard` 
```

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+        to points and raster data is discarded.
+
+        If centroids is not None, all hazards are defined on these centroids.
+        Centoids of hazards in haz_src not in centroids are mapped onto the
+        nearest point. If a point is further than threshold from the closet
+        centroid, the concatenation fails.
+
+        Note: centroids is modified in place and raster information is
+        destroyed.
+
```
ChrisFairless, line 1434:
```suggestion
        To combine a list of hazards and return a new object, use the
        static method `hazard.concatenate_hazard` instead. 
```

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+                                                centroids,
+                                                threshold).__dict__
+
+    @staticmethod
+    def concatenate_hazard(haz_list, centroids=None, threshold=100):
+        """
+        Concatenate events of several hazards of same type.
+
+        Centroids of all hazards must either all be rasters with the same
+        resolution, or all be points.
```
ChrisFairless, line 1466:
```suggestion
        resolution, or all be points. All centroids must have the same CRS.
```

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+
+        By default, the centroids of all hazards are combined together. If points,
+        all points are combined. If rasters, a global raster is defined.
+
+        If centroids is not None, all hazards are defined on these centroids.
+        Centoids of hazards in haz_list not in centroids are mapped onto the
+        nearest point. If a point is further than threshold from the closet
+        centroid, the concatenation fails.
+
```
ChrisFairless, line 1475:
```suggestion

        Note: To append hazards to an existing hazard object, use
        `hazard.extend` instead.
```

climada/hazard/base.py
emanuel-schmid: > @emanuel-schmid isn't the analogy with `list` fairly close? I see `append` and `extend` as operations that take either a single element or an iterable of elements and combine them into `self`.

Agreed, that's a valid perspective if we accept that Hazard has a dual nature, being a list and an object at the same time.

emanuel-schmid: > With `extend`, `append`, and `concatenate_hazard`, it is a lot. I think it's worth pointing out similar methods in the docstrings with a 'See also...', so users are aware of them. I'll add these as suggestions now.
> 
> We could also think about deprecating `hazard.append` because `hazard.extend` is almost always better. A deprecation notice would be annoying to anyone using the method, but in the long run it would encourage (slightly) better code ... I'm not sure it's worth our (and our users') time, though.
> 

Both methods are very stateful and thus - to some extent - calling for trouble. Me, personally, I wouldn't even be sad if we discarded both of them. In any case we could think about encouraging folks to use `concatenate_hazard`.

emanuel-schmid: > Finally, should we add the parameters centroids and threshold to hazard.append for consistency? It won't break backward compatibility.
> 
👍

climada/hazard/base.py
chahank: Well, I am not sure. Would looping through the dictionary and replacing the values be more sensible?

climada/hazard/centroids/centr.py
chahank: I re-added the default crs here:
9c646d4

climada/hazard/base.py
chahank: >Agreed, that's a valid perspective if we accept that Hazard has a dual nature, being a list and an object at the same time.
It kind of is. It is an object Hazard and a list of events.

chahank: >Me, personally, I wouldn't even be sad if we discarded both of them. In any case we could think about encouraging folks to use concatenate_hazard.

The reason we have an extend method is that Hazard objects can get large in memory, and thus it would be nice to have a method to append without having afterwards twice the object.

emanuel-schmid: I see your point, but the argument is somewhat mitigated by the fact that the critical maximum memory consumption is not a bit smaller, for `conatenate_hazard` is called in any case. The difference is that with `extend` the garbage collector is automatically called for action and with `concatenate_hazard` one would have to free memory explicitly...

climada/hazard/base.py
```diff
@@ -1410,3 +1413,142 @@ def _read_att_excel(self, file_name, var_names):
...
+            Default is 100.
+
+        Returns
+        -------
+        None.
+
+        """
+        if self.event_id.any(): #if self is not empty, concatenate with self.
+            haz_list = [self] + haz_list
+        self.__dict__ = self.concatenate_hazard(haz_list,
```
emanuel-schmid, line 1456:
```suggestion
        self.__dict__ = Hazard.concatenate_hazard(haz_list,
```

climada/hazard/base.py
emanuel-schmid: Not really, I'm afraid. I was rather thinking of a constructor that takes arguments and initializes data beyond their mere announcement.
Which we don't have at the moment.

However I'm still uneasy about the `extend` method. And it is partly related to the way the `__dict__` is transferred. To illustrate:
`Hazard.concatenate_hazard([Drought(), Landslide()])` is a `Hazard` object. That's fine.
`Drought().extend([Landslide()])` on the other hand is working on a `Drought` object but any `Drought` property that is not inherited from `Hazard` evaporates. In other words: `extend` bears the hazard of making `Hazard` extensions inconsistent.

chahank: Ah yes, I understand now. It should actually not be allowed to mix events from different types of hazards. I will add a check for the haz type. Also, the suggestion from below will be implemented.

climada/hazard/base.py
chahank: Yes, I see. I mean, having the garbage collector to free memory is also a plus, but not as important as the critical maximum memory.

chahank:
Thanks @ChrisFairless , @tovogt  @emanuel-schmid  for the feedback!

PR #240, opened 2021-06-02
Feature/yearsets

carmensteinmann:
Hello! 
I updated the yearsets module and would be very happy for your input!

The main changes are:
- a sampling vector (2D array) replaced the sampling dictionary
- the same event cannot occur more than once in the same year
- the user can specify lam (the number of years around which the Poisson distribution is centered)
- simplification of the function compute_imp_per_year (former compute_annual_impacts)
- update of variable names

After the discussion today on reducing the possibilities for the data type of input variables, I was also thinking of reducing the options for sampled_years to a list of years. What do you think? (This would also make the difference between sampled_years and n_sampled_years unambiguous.)

Thanks a lot for your suggestions!

Best, Carmen


climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
 
 LOGGER = logging.getLogger(__name__)
 
-def impact_yearset(event_impacts, sampled_years=None, sampling_dict=None, correction_fac=True):
+def impact_yearset(imp, sampled_years=None, sampling_vect=None, lam= None,
```
chahank, line 25: What does lam mean (why did you choose this name)?
```suggestion
def impact_yearset(imp, sampled_years=None, sampling_vect=None, lam=None,
```

chahank, line 25: Maybe the `sampled_years` should always be supplied no? This would simplify the logic further.

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-        sampling_dict : dict
-            The sampling dictionary containing two arrays:
-                selected_events (array): sampled events (len: total amount of sampled events)
-                events_per_year (array): events per sampled year
-        n_events_per_year : array
+        events_per_year : array
             Number of events per sampled year
     """
+    if not lam:
+        lam = np.sum(imp.frequency)
```
chahank, line 129:
```suggestion
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
 
     Parameters:
         n_sampled_years : int
             The target number of years the impact yearset shall contain.
-        n_annual_events : int
-            Number of events per year in given event_impacts object
-        n_input_events : int
-            Number of events contained in given event_impacts object
+        imp : climada.engine.Impact()
+            impact object containing impacts per event
```
chahank, line 120: I think this is not needed in this function. `lam = np.sum(imp.frequency)` should be used in the function that calls this one. You could then also rename the method as `sample_from_poisson` or similar.

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
 
-    if len(np.unique(event_impacts.frequency)) > 1:
-        LOGGER.warning("The frequencies of the single events in the given event_impacts "
-                       "differ among each other. Please beware that this will influence "
-                       "the resulting annual_impacts as the events are sampled uniformaly "
-                       "and different frequencies are (not yet) taken into account.")
-
+    #create sampling vector if not given as input
+    if not sampling_vect:
+        events_per_year = sample_n_events(n_sampled_years, imp, lam)
```
chahank, line 87:
```suggestion
        events_per_year = sample_n_events(n_sampled_years, lam)
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-                    "sampling_dict are used as number of sampled_years.")
-        sampled_years = list(range(1, len(sampling_dict['selected_events'])+1))
-
-    if sampling_dict and (
-            np.sum(sampling_dict['events_per_year']) != len(sampling_dict['selected_events'])):
-        raise ValueError("The sampling dictionary is faulty: the sum of selected events "
-                         "does not correspond to the number of selected events.")
+                    "in the sampling_vect differ. The number of years contained in the "
+                    "sampling_vect are used as number of sampled_years.")
+        sampled_years = list(range(1, len(sampling_vect)+1))
```
chahank, line 82:
```suggestion
if not lam:
    lam = np.sum(imp.frequency)
```

climada/util/yearsets.py
carmensteinmann: I used the same name the parameter (lambda) has in the np.random.poisson() function (https://numpy.org/doc/stable/reference/random/generated/numpy.random.poisson.html), but agree that this might not be the best name.

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-    #copy event_impacts object as basis for the annual_impacts object
-    annual_impacts = copy.deepcopy(event_impacts)
-
-    #save impact_per_year in annual_impacts
+    #save imp_per_year in yimp
     if correction_fac: #adjust for sampling error
-        correction_factor = calculate_correction_fac(impact_per_year, event_impacts)
-        annual_impacts.at_event = impact_per_year / correction_factor
+        correction_factor = calculate_correction_fac(imp_per_year, imp)
+        yimp.at_event = imp_per_year / correction_factor
```
chahank, line 99:
```suggestion
        imp_per_year = imp_per_year / calculate_correction_fac(imp_per_year, imp)
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
     else:
```
chahank, line 100:
```suggestion
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-        annual_impacts.at_event = impact_per_year
+        yimp.at_event = imp_per_year
```
chahank, line 101:
```suggestion
    yimp.at_event = imp_per_year
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-    selected_events = rng.choice(indices, size=tot_n_events,
-                                 replace=False).astype('int')
+    indices_orig = np.arange(len(freqs_orig))
 
+    freqs = freqs_orig
+    indices = indices_orig
 
-    return selected_events
+    #if check_doubling true: check that every event doesn't occur more than once per sampled year
+    check_doubling = False
```
chahank, line 168: That should be an input variable or be removed.

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-                selected_events (array) : sampled events (len: total amount of sampled events)
-                events_per_year (array) : events per sampled year
+        probab_dis = freqs/np.sum(freqs)
 
-    Returns:
-        impact_per_year: array
-            Sampled impact per year (length = n_sampled_years)
-      """
+        rng = default_rng()
+        selected_events = rng.choice(indices, size=amount_events, replace=False,
```
chahank, line 176: Whre is `amount_events` defined?

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
+                                     p=probab_dis).astype('int')
+
+        if check_doubling: #check if an event occurs more than once in a year
+            unique_events = np.unique(selected_events)
+            #resample until each event occurs max. once per year
+            while len(unique_events) != len(selected_events):
+                rng = default_rng()
+                selected_events = rng.choice(indices, size=amount_events, replace=False,
+                                             p=probab_dis).astype('int')
```
chahank, line 185: This is maybe not the most elegant way to solve the problem. In some extrem cases this might lead to a quasi-infinite loop. I would at least build in a stop condition.

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
+            check_doubling = False
+
+        idx_to_remove = [] #determine used events to remove them from sampling pool
+        for event in selected_events:
+            idx_to_remove.append(np.where(indices == event)[0][0])
```
chahank, line 190: Almost always when there is a for ... append you can replace it with a list comprehension.

```suggestion
        idx_to_remove = [np.where(indices == event)[0][0] for event in selected_events]
        #determine used events to remove them from sampling pool
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
+    Returns:
+        imp_per_year: array
+            Sampled impact per year (length = sampled_years)
+      """
 
-    return impact_per_year
+    imp_per_year = np.zeros(len(sampling_vect))
+    for year, sampled_events in enumerate(sampling_vect):
+        if sampled_events.size > 0:
+            imp_per_year[year] = np.sum(imp.at_event[sampled_events])
```
chahank, line 223: `np.sum([]) = 0.0`
```suggestion
    imp_per_year = [np.sum(imp.at_event[sampled_events]) for sampled_events in sampling_vect]
```

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-    indices = np.tile(np.arange(n_input_events), repetitions)
+    sampling_vect = []
 
-    rng = default_rng()
-    selected_events = rng.choice(indices, size=tot_n_events,
-                                 replace=False).astype('int')
+    indices_orig = np.arange(len(freqs_orig))
 
+    freqs = freqs_orig
+    indices = indices_orig
```
chahank, line 165: Is this really needed? Do you later need `freqs_orig` or `indices_orig` ?

chahank (changes requested):
Very nice improvements! It is really getting to a very lean and clear state. Congratulations. 

I have not looked at the tests in this review.

Curiosity: Do you still plan in allowing for other sampling methods than poisson distribution?

chahank:
> Hello!
> I updated the yearsets module and would be very happy for your input!
> 
> The main changes are:
> 
>     * a sampling vector (2D array) replaced the sampling dictionary
> 
>     * the same event cannot occur more than once in the same year
> 
>     * the user can specify lam (the number of years around which the Poisson distribution is centered)
> 
>     * simplification of the function compute_imp_per_year (former compute_annual_impacts)
> 
>     * update of variable names
> 
> 
> After the discussion today on reducing the possibilities for the data type of input variables, I was also thinking of reducing the options for sampled_years to a list of years. What do you think? (This would also make the difference between sampled_years and n_sampled_years unambiguous.)
> 
> Thanks a lot for your suggestions!
> 
> Best, Carmen

I also fully agree with the reduction of `sampled_years` to only a list of years!

climada/util/yearsets.py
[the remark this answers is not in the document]
carmensteinmann: If no lam is given lam = np.sum(imp.frequency) is used - is there a smoother way to implement this?

climada/util/yearsets.py
```diff
@@ -22,223 +22,227 @@
...
-      In contrast to the expected annual impact (eai) annual_impacts contains impact values that
-      differ among years (the correction factor can however be used to scale the annual_impacts
-                          to fit the eai of the events_impacts object that is used to generated it)
+      by sampling events from the impact received as input with a Poisson distribution
+      centered around n_events per year (n_events = sum(imp.frequency)).
+      In contrast to the expected annual impact (eai) yimp contains impact values that
+      differ among years (the correction factor can however be used to scale the yimp
+                          to fit the eai of the impact that is used to generated it)
 
     Parameters:
```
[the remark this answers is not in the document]
carmensteinmann: Cool, thanks for that hint! :)

climada/util/yearsets.py
carmensteinmann: It's part of the loop (for idx_year, amount_events in enumerate(events_per_year):)
It is the number of events for the particular year being sampled for.
Would n_events be clearer?

climada/util/yearsets.py
carmensteinmann: I agree. Do you have an idea for a better solution? Otherwise I'll include the stop condition for the time being.

climada/util/yearsets.py
carmensteinmann: They are used if the number of sampled events is larger than the number of input events.

climada/util/yearsets.py
carmensteinmann: The problem here is that in case of no event it's not np.sum([ ]) but np.sum(imp.at_event[ ]), which results in this error: IndexError: arrays used as indices must be of integer (or boolean) type. 
Do you have a suggestion on how to avoid this?

climada/util/yearsets.py
chahank: aaa ok, lambda is fine. Maybe call it lambda then :D ?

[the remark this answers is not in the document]
chahank: Ok, good point. Then it is the right implementation. I would just indicate in the docstring what the actual default is.

climada/util/yearsets.py
chahank: The name is fine, I just somehow missed where it is defined. Thanks!

climada/util/yearsets.py
chahank: You could for instance sample from the set with the duplicate events removed.

climada/util/yearsets.py
chahank: This should solve the problem.
```suggestion
    imp_per_year = [np.sum(imp.at_event[list(sampled_events)]) for sampled_events in sampling_vect]
```

climada/util/yearsets.py
carmensteinmann: I implemented this variable so the doubling is only checked when the same event is actually contained twice in the indices variable to reduce the use of the while loop.

climada/util/yearsets.py
carmensteinmann: that seems to be a predefined function in python..

climada/util/yearsets.py
carmensteinmann: So when my list of events to sample from becomes too short to be used for the next year I would reset it to the original list and drop the events that haven't been used so far?

climada/util/yearsets.py
carmensteinmann: Probably it doesn't make much sense to keep this as a separate function if it's literally one line, right?

carmensteinmann:
Thanks Chahan!
I'd be up to implement other sampling methods if you have a specific range of distribution of interest or an application case for a different distribution. Just let me know.

climada/util/yearsets.py
carmensteinmann: @chahank I came up with a way to get rid of the while loop without having to drop the events that hadn't been used in the preceding samplings (see my comment from yesterday). Check it out and let me know what you think :).

climada/util/yearsets.py
chahank: correct! Well, then maybe leave it as is.

climada/util/yearsets.py
chahank: Maybe could you rephrase the comment just to make it clear that this is not an option to turn off the double counting?

climada/util/yearsets.py
chahank: I do not have a strong opinion here. Please do as you prefer.

climada/util/yearsets.py
chahank: Nice! Just to be sure, you mean this commit b55a01b7ad17c99aaa54636bde392767e72ff542 right?

carmensteinmann: Yes :)

climada/util/yearsets.py
```diff
@@ -197,39 +197,30 @@ def sample_events(events_per_year, freqs_orig):
...
-
-        idx_to_remove = [np.where(indices == event)[0][0] for event in selected_events]
         #determine used events to remove them from sampling pool
+        idx_to_remove = [np.where(indices == event)[0][0] for event in selected_events]
         indices = np.delete(indices, idx_to_remove)
         freqs = np.delete(freqs, idx_to_remove)
 
         #add the original indices and frequencies to the pool if there are less events
         #in the pool than needed to fill the next sampled year
         if (idx_year < (len(events_per_year)-1)) and (
```
chahank, line 220: Looks all good to me. Just one last clarification: Why do you need the check `(idx_year < (len(events_per_year)-1))` ?

climada/util/yearsets.py
```diff
@@ -22,223 +22,254 @@
...
-        selected_events : array
-            Uniformaly sampled events (length: tot_n_events)
-      """
+    freqs = freqs_orig
+    indices = indices_orig
 
-    repetitions = np.ceil(tot_n_events/n_input_events).astype('int')
-    indices = np.tile(np.arange(n_input_events), repetitions)
+    #sample events for each sampled year
+    for _, amount_events in enumerate(events_per_year):
```
chahank, line 201:
```suggestion
    for amount_events in events_per_year:
```

chahank:
Please make a run of the tests, a quick static code analysis and then merge. Thanks for the great improvements!

PR #241, opened 2021-06-03
Feature/geopandas 0.9

emanuel-schmid:
This PR addresses the issues with updating to geopandas 0.9.
Unlike in geopandas 0.7 or before, the crs of a GeoDataFrame is not to be set by assigning the property, but with a dedicated method, `set_crs`.
Then, GeoDataFrame instantiation with a crs but without a geometry column raises a FutureWarning and ignores the submitted crs.
Due to the above changes, adaptations were necessary and inplemeneted in the python files of the climada package.

Two Exposures methods have been introduced, `set_gdf` and `set_crs`. They are preferred over direct assignment of the respective property, because they make sure the Exposures object's data is in a consistent state.

The Notebooks require also changes, which will be dealt with in a subsequent pull request.


tovogt (commented):
Very nice!

I suggest some major coding style changes in base.py that I pushed to your branch (f3d0235) so that you can have a look.

climada/entity/exposures/base.py
```diff
@@ -118,15 +118,20 @@ class Exposures():
...
+        if "geometry" in self.gdf and hasattr(self.gdf.geometry, "crs") and self.gdf.geometry.crs:
+            return self.gdf.geometry.crs
+        return self.meta.get('crs')
+
+    def __init__(self, *args,
+                 meta=None,
+                 tag=None,
+                 ref_year=DEF_REF_YEAR,
+                 value_unit=DEF_VALUE_UNIT,
+                 crs=None,
```
tovogt, line 133: IMHO, having these as explicit keyword arguments improves readability of both the docs and the code a lot.

climada/entity/exposures/base.py
```diff
@@ -197,28 +177,23 @@ def __init__(self, *args, **kwargs):
...
             crs = DEF_CRS
-            LOGGER.info('crs set to default value %s', crs)
 
         geometry = kwargs.get('geometry')
         if geometry and isinstance(geometry, str):
             raise ValueError("Exposures is not able to handle customized 'geometry' column names.")
-        if 'crs' in kwargs and not geometry and not 'geometry' in data:
-            kwargs.pop('crs')
-        if 'crs' not in kwargs and (geometry or 'geometry' in data):
-            kwargs['crs'] = crs
```
tovogt: Since `crs` is now an explicit keyword argument, we don't have to deal with it. We just pass it on to `set_gdf` where this will be dealt with.

climada/entity/exposures/base.py
```diff
@@ -445,6 +420,8 @@ def set_from_raster(self, file_name, band=1, src_crs=None, window=False,
             resampling (rasterio.warp,.Resampling optional): resampling
                 function used for reprojection to dst_crs
         """
+        if 'geometry' in self.gdf:
+            raise ValueError("there is already a geometry column defined in the GeoDataFrame")
```
tovogt, line 424: Fail early, don't read the raster data first.

climada/entity/exposures/base.py
```diff
@@ -118,12 +118,20 @@ class Exposures():
...
+            return self.gdf.geometry.crs
+        return self.meta.get('crs')
+
+    def __init__(self, *args,
+                 meta=None,
+                 tag=None,
+                 ref_year=DEF_REF_YEAR,
+                 value_unit=DEF_VALUE_UNIT,
+                 crs=None,
+                 **kwargs):
```
tovogt, line 135: It's a bit more common in CLIMADA to fill the lines with function arguments. If you prefer that, here is how that would look like:
```suggestion
    def __init__(self, *args, meta=None, tag=None, ref_year=DEF_REF_YEAR,
                 value_unit=DEF_VALUE_UNIT, crs=None, **kwargs):
```

climada/entity/exposures/base.py
emanuel-schmid: 👍 Totally!

climada/entity/exposures/base.py
emanuel-schmid: 👍

climada/entity/exposures/base.py
emanuel-schmid: 👍

climada/entity/exposures/base.py
emanuel-schmid: 👍

PR #244, opened 2021-06-08
Exposures.read_hdf5: restore CRS

### LINKED ISSUE #243: Unable to read exposure hdf5 files
With the latest changes to the 'crs' handling exposures files created with the old code are not readable anymore. 

```
from climada.entity import Exposures
exp_houses = Exposures()
exp_houses.read_hdf5("EXP_houses_VNM_v01.hdf5")
```

Error message:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-9e97ae9d3693> in <module>
      1 from climada.entity import Exposures
      2 exp_houses = Exposures()
----> 3 exp_houses.read_hdf5("EXP_houses_VNM_v01.hdf5")
      4 
      5 exp_people = Exposures()

~/Documents/Climada/climada_python/climada/entity/exposures/base.py in read_hdf5(self, file_name)
    685                     setattr(self, key, val)
    686                 if key == 'crs':
--> 687                     self.set_crs(val)
    688 
    689     def read_mat(self, file_name, var_names=None):

~/Documents/Climada/climada_python/climada/entity/exposures/base.py in set_crs(self, crs)
    278         if 'geometry' in self.gdf.columns:
    279             try:
--> 280                 self.gdf.set_crs(crs, inplace=True)
    281             except ValueError:
    282                 # restore popped crs and leave

~/opt/anaconda3/envs/climada_210/lib/python3.8/site-packages/geopandas/geodataframe.py in set_crs(self, crs, epsg, inplace, allow_override)
    776         else:
    777             df = self
--> 778         df.geometry = df.geometry.set_crs(
    779             crs=crs, epsg=epsg, allow_override=allow_override, inplace=True
    780         )

~/opt/anaconda3/envs/climada_210/lib/python3.8/site-packages/geopandas/geoseries.py in set_crs(self, crs, epsg, inplace, allow_override)
    483 
    484         if not allow_override and self.crs is not None and not self.crs == crs:
--> 485             raise ValueError(
    486                 "The GeoSeries already has a CRS which is not equal to the passed "
    487                 "CRS. Specify 'allow_override=True' to allow replacing the existing "

ValueError: The GeoSeries already has a CRS which is not equal to the passed CRS. Specify 'allow_override=True' to allow replacing the existing CRS without doing any transformation. If you actually want to transform the geometries, use 'GeoSeries.to_crs' instead.
```

tovogt:
When reading HDF5 exposure files, make sure that the CRS is restored correctly. This fixes #243 because it will also correctly handle very old files (generated with old versions of CLIMADA and/or geopandas).

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -678,13 +678,16 @@ class Exposures():
         """
         LOGGER.info('Reading %s', file_name)
         with pd.HDFStore(file_name) as store:
-            self.__init__(store['exposures'])
             metadata = store.get_storer('exposures').attrs.metadata
+            crs = None
+            if 'crs' in metadata:
+                crs = metadata['crs']
+            elif 'meta' in metadata and 'crs' in metadata['meta']:
+                crs = metadata['meta']['crs']
+            self.__init__(store['exposures'], crs=crs)
             for key, val in metadata.items():
                 if key in type(self)._metadata:
                     setattr(self, key, val)
-                if key == 'crs':
-                    self.set_crs(val)
 
     def read_mat(self, file_name, var_names=None):
         """Read MATLAB file and store variables in exposures.
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -251,6 +251,11 @@ class Exposures():
         elif not any([col.startswith(INDICATOR_CENTR) for col in self.gdf.columns]):
             LOGGER.info("%s not set.", INDICATOR_CENTR)
 
+        # check if CRS is consistent
+        if self.crs != self.meta.get('crs'):
+            raise ValueError(f"Inconsistent CRS definition, gdf ({self.crs}) attribute doesn't "
+                             f"match meta ({self.meta.get('crs')}) attribute.")
+
         # check whether geometry corresponds to lat/lon
         try:
             if (self.gdf.geometry.values[0].x != self.gdf.longitude.values[0] or
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -679,11 +679,10 @@ class Exposures():
         LOGGER.info('Reading %s', file_name)
         with pd.HDFStore(file_name) as store:
             metadata = store.get_storer('exposures').attrs.metadata
-            crs = None
-            if 'crs' in metadata:
-                crs = metadata['crs']
-            elif 'meta' in metadata and 'crs' in metadata['meta']:
-                crs = metadata['meta']['crs']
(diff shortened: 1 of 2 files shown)


climada/entity/exposures/test/test_base.py
```diff
@@ -246,17 +246,17 @@ def test_io_hdf5_pass(self):
         self.assertEqual(exp_df.crs, exp_read.crs)
```
chahank, line 246: MAybe also check the `exp_df.gdf.crs`?

climada/entity/exposures/base.py
```diff
@@ -678,13 +678,15 @@ def read_hdf5(self, file_name):
         """
         LOGGER.info('Reading %s', file_name)
         with pd.HDFStore(file_name) as store:
-            self.__init__(store['exposures'])
             metadata = store.get_storer('exposures').attrs.metadata
+            # in previous versions of CLIMADA and/or geopandas, the CRS was stored in '_crs'/'crs'
+            crs = metadata.get('crs', metadata.get('_crs'))
```
chahank, line 683: I do not really understand this line, which seems to be the crucial one.

chahank:
Thanks a lot! This solved the issue with my data file. 

Extra remark that would fit in this pull request: do you know whether the `Exposures.check()` method was adapted to check for the `.crs` in the right places?

tovogt:
Good point about adding a check. I added one.

climada/entity/exposures/base.py
tovogt: The CRS used to be stored as an explicit attribute. In geopandas previous to 0.8, this attribute was called `crs`, and starting from 0.8, it was called `_crs`. With the introduction of the `gdf` attribute in CLIMADA, CLIMADA doesn't store the GeoDataFrame's attributes in the HDF file, but uses its own `meta` attribute to store and load the CRS.

This line makes sure that an explicitly stored `crs`/`_crs` attribute (from an old file) has precedence over the CRS stored in the `meta` attribute. Only if there is no `crs`/`_crs` attribute or if the attribute's value is `None`, the CRS stored in `meta` is used.

chahank: Thanks for the explanation! So, just to be sure, no error would arise is CRS is neither of the 3 locations?

tovogt: Exactly, in that case, the local `crs` variable would get the value `None` and the would be passed on to `__init__` which then has a default way of how to deal with this case (it will default to EPSG:4326).

chahank: Great!

PR #245, opened 2021-06-08
TC surge from dynamic GeoClaw model runs

tovogt:
This adds a new module `climada.hazard.tc_surge_geoclaw` that introduces a new Hazard class called `TCSurgeGeoClaw`. This new hazard models the storm surge impact from tropical cyclone tracks. Internally, the fluid dynamics solver [GeoClaw](http://www.clawpack.org/geoclaw) is used to simulate the storm surge dynamics. Please refer to the [tutorial notebook](https://github.com/CLIMADA-project/climada_python/blob/feature/tc_surge_geoclaw/doc/tutorial/climada_hazard_TCSurgeGeoClaw.ipynb) for more information.

It's important to note that this functionality is not for the average user of CLIMADA: It requires a Fortran compiler, a non-trivial choice of input data as well as large computing power. The example of a single TC track in the tutorial notebook takes 3 hours to run even with 16 CPUs on a cluster. All this is mentioned in the code, docstrings and in the tutorial notebook.

Because of that, the usual unit tests for this module (`python -m unittest climada.hazard.test.test_tc_surge_geoclaw`) do not include non-trivial model runs. However, there is a non-trivial integration test (`python -m unittest climada.test.test_tc_surge_geoclaw`) that runs on a modern laptop computer in 1-2 hours.


climada/hazard/base.py
```diff
@@ -1066,6 +1066,8 @@ def read_hdf5(self, file_name):
         self.clear()
         hf_data = h5py.File(file_name, 'r')
         for (var_name, var_val) in self.__dict__.items():
+            if var_name != 'tag' and var_name not in hf_data.keys():
+                continue
```
chahank, line 1070: Could you please explain why this change is needed? Is this related to this module or a more general code maintenance?

climada/hazard/tc_surge_geoclaw.py
```diff
@@ -0,0 +1,1451 @@
...
+import contextlib
+import datetime as dt
+import importlib
+import logging
+import re
+import pathlib
+import site
+import subprocess
+import sys
+import warnings
```
chahank, line 34: Is this module needed? I have seen it in a previous pull request (wildfires), but it ended up not being used at all.

climada/hazard/tc_surge_geoclaw.py
```diff
@@ -0,0 +1,1451 @@
...
+CLAWPACK_SRC_DIR = CONFIG.hazard.tc_surge_geoclaw.clawpack_src_dir.dir()
+"""Directory for Clawpack source checkouts (if it doesn't exist)"""
+
+GEOCLAW_WORK_DIR = CONFIG.hazard.tc_surge_geoclaw.geoclaw_work_dir.dir()
+"""Base directory for GeoClaw run data."""
+
+KN_TO_MS = (1.0 * ureg.knot).to(ureg.meter / ureg.second).magnitude
+NM_TO_KM = (1.0 * ureg.nautical_mile).to(ureg.kilometer).magnitude
+MBAR_TO_PA = (1.0 * ureg.mbar).to(ureg.pascal).magnitude
+KM_TO_DEG = 1.0 / (60 * NM_TO_KM)
```
chahank, line 75: I was confused about this also in the other review of wildfire. Why do a conversion between degrees and km with a constant? The ratio depends on the coordinate system right? Also, the longitude conversion depends on the latitude value...

climada/hazard/tc_surge_geoclaw.py
```diff
@@ -0,0 +1,1451 @@
...
+"""Unit conversion factors."""
+
+
+class TCSurgeGeoClaw(Hazard):
+    """TC storm surge heights in m, modeled using GeoClaw.
```
chahank, line 80:
```suggestion
    """TC storm surge heights in meters (m), modeled using GeoClaw.
```

climada/hazard/tc_surge_geoclaw.py
```diff
@@ -0,0 +1,1451 @@
...
+
+    Attributes
+    ----------
+    gauge_data : list of lists of dicts
+        For each storm and each gauge, a dict containing the `location` of the gauge, and
+        (for each landfall event) `base_sea_level`, `topo_height`, `time` and `height_above_geoid`
+        information.
+        Due to this format, this data will NOT be stored when using `write_hdf5`. However, you
+        can manually pickle it in a separate file if needed.
```
chahank, line 89: Could you maybe write a helper method `save_gauge_data` and `load_gauge_data` if this makes sense?

climada/hazard/tc_surge_geoclaw.py
```diff
@@ -0,0 +1,1451 @@
...
+        return new_haz
+
+    def write_hdf5(self, *args, **kwargs):
+        gauge_data = self.gauge_data
+        delattr(self, "gauge_data")
+        Hazard.write_hdf5(self, *args, **kwargs)
+        self.gauge_data = gauge_data
+
+
+def get_centroids_from_tracks(tracks, res_deg, buffer_deg):
```
chahank, line 225: Should this method maybe be part of the `TCTracks` module instead? I can see it being useful for genereal applications with Tropical cyclones tracks.

chahank:
Very nice addition. I only looked quickly at the top and made a few comments. This is not a real review yet.

One question came to me: why does the module use the tracks instead of the wind footprints? Is it for the temporal component? If it could work with the footprints, it could also compute the storm surge for the winter storms europe (if this makes sense).

climada/hazard/base.py
tovogt: If the Hazard class has attributes that are not stored in the HDF file, then the reading of the HDF file will fail. I think that's a general problem. For example if a new attribute is added to a Hazard class at some point, then old HDF files won't be read any more. In case of the TCSurgeGeoClaw class, there is a class attribute (`gauge_data`) that is not written to the HDF file because it has a format that is not compatible with the HDF format. When the HDF file is read later, the reading of this attribute shouldn't be enforced (it will fail because it's not present in the HDF file).

climada/hazard/tc_surge_geoclaw.py
tovogt: It's needed to suppress irrelevant warnings when reading the output of GeoClaw:
https://github.com/CLIMADA-project/climada_python/blob/964d976f096bf991bb14d94df156ba0d811518e9/climada/hazard/tc_surge_geoclaw.py#L561-L564

climada/hazard/tc_surge_geoclaw.py
tovogt: That constant is perfectly fine: It stores the degree distance at the equator of one kilometer. That's a constant, so it makes sense to store it. When actually doing a conversion from km to degrees, you would do the following:
```
lat = y * KM_TO_DEG
lon = x * KM_TO_DEG * np.cos(lat)
```
However, if you don't care about a factor 2 and you are in a TC context, then you can even decide to ignore the cosine factor. For example, in case of this module, the factor is used to give a rough estimate of the size of the coastal area that is actually threatened to be flooded.

tovogt:
The track input is the kind of input data that is accepted by GeoClaw. The reason for this is that the solver requires wind fields at arbitrary points in time and space. The solver is an adaptive mesh solver that will refine temporal and spatial resolution on the fly as needed. So the temporal and spatial resolution of wind fields needs to be refined on the fly as well.

climada/hazard/tc_surge_geoclaw.py
chahank: Could we maybe change the name of the variable to `KM_TO_DEG_EQUA` or something similar?

chahank:
I see, that is what I suspected. Do you know how the computed windfields differ from the ones obtained by the Holland model?

tovogt:
GeoClaw supports three wind field models: Holland 1980, Holland 2010 and Chavas et al. 2016. When I started working on this module, only Holland 1980 was working properly, so that's what I'm using here.

climada/hazard/tc_surge_geoclaw.py
tovogt: Done.

climada/hazard/tc_surge_geoclaw.py
tovogt: Done.

climada/hazard/tc_surge_geoclaw.py
tovogt: Done.

chahank:
The TropCyclone module uses Holland 2008. I suppose this is the same as Holland 2010?

tovogt:
No, Holland 2008 is a small modification of Holland 1980:
https://github.com/CLIMADA-project/climada_python/blob/70919a9a69a5f8940674cbc821b118e0d39208f2/climada/hazard/trop_cyclone.py#L666-L681
The main point of Holland 2008 is that it doesn't trust the measured maximum wind speed, but uses a refined wind-pressure relationship instead.

Holland 2010 splits the wind field decay curve at the radius of maximum wind so that the old 1980 model is used close to the eye, and a new decay is applied when it comes to the outskirts of the storm. From a first glance at the implementation in GeoClaw, I wouldn't use it now because I don't understand its choice of some parameters (xn, vn, rn) without justification.

I quickly checked the CLE model and found that it's still non-functional. It's marked as NotImplemented by GeoClaw (even though there are some code fragments) and a code cleanup is discussed in several places (e.g. https://github.com/clawpack/geoclaw/pull/467).

climada/hazard/tc_surge_geoclaw.py
tovogt: Done.

ChrisFairless:
This is super impressive!!
I've only skimmed the code, so again, not a proper review. I wanted to ask a few broad questions though,
- How did you decide on the model setup? Are there recommended settings, or is it a matter of trial and error?
- Have you run any validations? The ability to match to gauge data means you can check how well the hazard performs
- How does GeoCLAW think about tides? I know a 6-hour change in landfall can turn a wet day into a catastrophe, and for future climate change/counterfactual/tail-event studies, it might be interesting to perturb this.
- From the tutorial it looks like GeoCLAW uses a double-nested model grid. Is it one-way or two-way nested? If it's one-way, are there ways to speed up the simulations by checking the low-resolution output before committing to high resolution?

tovogt:
Thanks for your questions, @ChrisFairless !

- I started from examples in the GeoClaw repository and then there was a lot of trial and error. There are so many parameters to choose, it's really quite overwhelming. The point of this module is to provide a setup that takes that burden from you, while still producing decent results for many use cases.
- A validation study is currently work in progress. We use tide gauge data and inundated areas as detected from satellite imagery.
- No tides at all, currently. When the modeling starts, the water is at rest. The monthly mean sea level in the region of landfall is used for initialization. Then the TC's wind friction at the surface is the only forcing for the fluid dynamics.
- GeoClaw uses adaptive meshes, there is no grid in the strict sense, it's a finite-element method. I'm not entirely sure what you mean by "checking the low-resolution output", but that seems to be the whole point of an adaptive-mesh strategy.
- I don't understand the question about Matthew. What's the problem about this storm? A fluid dynamics modeler like GeoClaw doesn't care about a concept like "landfall". It just gets a wind forcing at the water surface, that's it.

ChrisFairless:
Thanks! I'll be excited to see how it performs in validation.

I'm not a surge modeller so I don't know how important it is to model storm tide rather than storm surge. Maybe we can chat more in-depth elsewhere.

I'd misunderstood the mesh system - I'd assumed the plots in the tutorial example showed simulation domains. This is also why I asked about Matthew, assuming that it would struggle to place these domains without a landfall location, but that's not the case!

tovogt:
Closing this in favor of https://github.com/CLIMADA-project/climada_petals/pull/2 and #269

PR #246, opened 2021-06-11
Feature/crop prod input

sameberenz:
adding an alternative method (set_from_area_and_yield_nc4) to class CropProduction(Exposure) that allows expert users to initiate exposure of annual crop production [t/y] from two netcdf files containing yield [t/ha/y] and cultivated area [ha] information.


climada/entity/exposures/crop_production.py
```diff
@@ -372,9 +372,126 @@ def set_from_isimip_netcdf(self, input_dir=None, filename=None, hist_mean=None,
...
+        when using this data for publication.
+
+        Parameters
+        ----------
+        crop_type : str
+            Crop type, e.g. 'mai' for maize, or 'ric', 'whe', 'soy', etc.
+        i_crop_yield : int
+            crop layer in yield input data set. Index typically starts with 1.
+            In the default yield input file, the crop indices are as follows:
+            1: maize, 2: wheat, 3: rice(!), 4: soybean(!).
```
carmensteinmann, line 404:
```suggestion
            1: maize, 2: wheat, 3: rice(!), 4: soybean(!).
            (!): layer number differs between default input files.
```

carmensteinmann (changes requested):
Hi Sam!
Overall the code is easy to read, I don't have many comments. 
Regarding the structure: I would suggest splitting this function into two. 
1. an expert function that REQUIRES all your input variables
2. a wrapper calling on the first function for the dataset you're using. 

This would get rid of a lot of the if conditions. Moreover, I think the default values don't make much sense if they are used on an individual basis and we could expect an expert user to provide them. What do you think? 

Also, the second function could be simplified by having a variable called irr instead of yield_var and area_var, setting them both either to irr, noirr or combined.

sameberenz:
Thanks for your suggestions @carmensteinmann. i have split the methods as you suggested. what do you think?

carmensteinmann:
@sameberenz I think it looks much clearer now, thanks a lot for adapting the code. 
I have one more note regarding variable names. I would suggest consistently using x_yield and x_area for all variables. Also, I think the names i_crop_yield and i_crop_area are quite unclear. 
My suggestions: 
- set_from_area_and_yield_nc4(self, crop_type, layer_yield, layer_area, filename_yield, filename_area, var_yield,                                    var_area, bbox=BBOX, input_dir=INPUT_DIR) and using data_area / data_yield within the function.
-layer_yield / layer_area instead of crop_idx_yield and crop_idx_area in set_from_spam_ray_mirca()

sameberenz:
Thanks for the suggestions @carmensteinmann , i have remaned the variables accordingly. What do you think?

climada/entity/exposures/crop_production.py
```diff
@@ -372,9 +371,158 @@ def set_from_isimip_netcdf(self, input_dir=None, filename=None, hist_mean=None,
...
+                                    var_area, bbox=BBOX, input_dir=INPUT_DIR):
+        
+        """
+        Set crop_production exposure from cultivated area [ha] and
+        yield [t/ha/year] provided in two netcdf files with the same grid.
+
+        Both input files need to be netcdf format and come with dimensions
+        'lon', 'lat' and 'crop'. The information which crop type is saved in which
+        crop layer in each input files needs to be provided manually via
+        the parameters 'i_crop_*'.
```
carmensteinmann, line 387:
```suggestion
        the parameters 'layer_*'.
```

carmensteinmann (approved):
For me it's ready to be merged.

PR #247, opened 2021-06-11
Hotfix/tc read legacy hdf5 files

emanuel-schmid:
This PR addresses the issue #248. But only superficially, for one particular tc_tracks file in hdf5 format.

### DIFF
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -443,8 +443,8 @@ class TropCyclone(Hazard):
         ])
         new_haz.orig = np.array([track.orig_event_flag])
         new_haz.category = np.array([track.category])
-        new_haz.basin = track.basin if isinstance(track.basin, str) \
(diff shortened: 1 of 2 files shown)


climada/hazard/trop_cyclone.py
```diff
@@ -443,7 +443,8 @@ def _tc_from_track(self, track, centroids, coastal_idx, model='H08',
         ])
         new_haz.orig = np.array([track.orig_event_flag])
         new_haz.category = np.array([track.category])
-        new_haz.basin = [str(track.basin.values[0])]
+        new_haz.basin = track.basin if isinstance(track.basin, str) \
+                        else [str(track.basin.values[0])]
```
tovogt, line 447:
```suggestion
        new_haz.basin = [track.basin if isinstance(track.basin, str)
                         else str(track.basin.values[0])]
```

tovogt (approved):
Well spotted, thanks for the fix!

PR #249, opened 2021-06-13
Feature/ec codes tc forecast

manniepmkam:
The bufr files from ECMWF that contains the TC tracks forecast recently has a file format change (see [here](https://confluence.ecmwf.int/display/FCST/Update+to+Tropical+Cyclone+tracks)). The old version of the `TCForecast` cannot support the new file format change. Therefore, I propose to use package `ecCodes` maintained by ECMWF themselves to extract the TC track forecast data. `ecCodes` has already been installed in CLIMADA since it is a dependent package of `cfgrib`, no extra installation is needed here.

I rewrote the `tc_tracks_forecast.py` using `ecCodes` and follow [this example](https://confluence.ecmwf.int/display/ECC/bufr_read_tropical_cyclone). I have not modified the original unit test but currently it is failing. The possible reason is that in the new version I also include the forecast in timestep 0, whereas the original code only start at timestep 1. Will double check and possibly amend the unit test too.

The rewrite of this module is not completed but please feel free to have a look and make comments.


climada/hazard/tc_tracks_forecast.py
```diff
@@ -83,7 +87,8 @@ class TCForecast(TCTracks):
         data (list(xarray.Dataset)): Same as in parent class, adding the
             following attributes
                 - ensemble_member (int)
-                - is_ensemble (bool)
+                - is_ensemble (bool; if True, the simulation is a high resolution
+                               deterministic run)
```
ChrisFairless, line 91: (if I've understood this correctly)
```suggestion
                - is_ensemble (bool; if False, the simulation is a high resolution
                               deterministic run)
```

ChrisFairless:
From talking with @manniepmkam while she was writing this, it's clear that BUFR can be a nightmare to work with. It's so hard to read a file in a way that's intuitive to someone who hasn't also worked with BUFR. This means that whenever things break (as with this ECMWF update to their BUFR message formats) it's a huge amount of work to fix.

Hopefully using ecCodes will help a little here, as it makes the reading-in part of the code simpler.

I'd also really like to see some documentation (or links to documentation) of
- how BUFR files are navigated and variables are extracted
- the structure of ECMWF TC track messages, the relevant variables for us, and which bits are analysis/deterministic runs/ensemble runs and how we use them

basically everything the next person will need to update things the next time this breaks. (It doesn't need to stop this pull request from being merged though.) I don't know if it should be done as detailed comments within the code or as a separate tutorial/guide?

climada/hazard/tc_tracks_forecast.py
```diff
@@ -74,6 +75,9 @@
 
 LOGGER = logging.getLogger(__name__)
 
+MISSING_DOUBLE = ec.CODES_MISSING_DOUBLE
+MISSING_LONG = ec.CODES_MISSING_LONG
+"""Missing double and integers in ecCodes """
```
chahank, line 80: Is this a module constant or something that should be adjusted as configurables?

climada/hazard/tc_tracks_forecast.py
```diff
@@ -83,7 +87,8 @@ class TCForecast(TCTracks):
         data (list(xarray.Dataset)): Same as in parent class, adding the
             following attributes
                 - ensemble_member (int)
-                - is_ensemble (bool)
+                - is_ensemble (bool; if True, the simulation is a high resolution
```
chahank, line 90: That name sounds quite obscure. Why would the word 'ensemble' = 'high resolution deterministic run'? I actually do not see why this would be a base class variable. Not all forecast models have the concept of 'high resolution deterministic run', which in itself is unclearly defined as it is unclear what 'high resolution' and 'deterministic' mean. What is a 'deterministic' run for a tropical cyclone track forecast model?

climada/hazard/tc_tracks_forecast.py
```diff
@@ -179,76 +184,156 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
+        timestamp_origin = np.datetime64(timestamp_origin)
+        
+        # number of timestep
+        n_timestep = 0
+        while True:
+            n_timestep = n_timestep + 1
+            try:
+                ec.codes_get_array(bufr, "#%d#timePeriod" % n_timestep)
+            except ec.CodesInternalError:
+                LOGGER.info('Key values not found: end of the time period')
```
chahank, line 215: Please expand this logging info. This is not clear enough.

climada/hazard/tc_tracks_forecast.py
```diff
@@ -179,76 +184,156 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
+            ec.codes_get(bufr, 'year'), ec.codes_get(bufr, 'month'),
+            ec.codes_get(bufr, 'day'), ec.codes_get(bufr, 'hour'),
+            ec.codes_get(bufr, 'minute'),
+        )
+        timestamp_origin = np.datetime64(timestamp_origin)
+        
+        # number of timestep
+        n_timestep = 0
+        while True:
+            n_timestep = n_timestep + 1
```
chahank, line 211: Could you read the number of times steps from the file first? That would make a cleaner loop.

climada/hazard/tc_tracks_forecast.py
```diff
@@ -179,76 +184,156 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
+        lon_init = ec.codes_get_array(bufr, '#2#longitude')
+        pre_init = ec.codes_get_array(bufr, '#1#pressureReducedToMeanSeaLevel')
+        max_wind_init = ec.codes_get_array(bufr, '#1#windSpeedAt10M')
+        
+        if len(lat_init) == len(ens_no) and len(max_wind_init) == len(ens_no):
+            for ind_ens in range(len(ens_no)):
+                latitude[ind_ens] = np.array(lat_init[ind_ens])
+                longitude[ind_ens] = np.array(lon_init[ind_ens])
+                pressure[ind_ens] = np.array(pre_init[ind_ens])
+                max_wind[ind_ens] = np.array(max_wind_init[ind_ens])
```
chahank, line 238: Please use list (dict) comprehension here

climada/hazard/tc_tracks_forecast.py
```diff
@@ -179,76 +184,156 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
+                break
+            
+        # ensemble members number
+        ens_no = ec.codes_get_array(bufr, "ensembleMemberNumber")
+        
+        # initiate a data dictionary to hold values
+        latitude = collections.defaultdict(dict)
+        longitude = collections.defaultdict(dict)
+        pressure = collections.defaultdict(dict)
+        max_wind = collections.defaultdict(dict)
```
chahank, line 225: Why use a default dict and then later not use the default?

climada/hazard/tc_tracks_forecast.py
```diff
@@ -179,76 +184,156 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
-            )
-
-        # query the bufr message
+            print('pressure', pre_init)
+            print('max_wind', max_wind_init)
+            for ind_ens in range(len(ens_no)):
+                latitude[ind_ens] = np.array(lat_init[0])
+                longitude[ind_ens] = np.array(lon_init[0])
+                pressure[ind_ens] = np.array(pre_init[0])
+                max_wind[ind_ens] = np.array(max_wind_init[0])
```
chahank, line 246: PLease use list comprehensions

chahank:
Thanks for taking care of these technical changes!

climada/hazard/tc_tracks_forecast.py
ChrisFairless: interesting question. we'd only want to change these missing values if ECMWF change their missing values in the future (very unlikely imo). in that situation the current code would allow the user to update ecCodes manually without having to modify any code themselves.

on the other hand, with the CLIMADA release cycle it might be three months before it's fixed in CLIMADA, and the user may want to manually specify these values instead. but this situation is really unlikely, so i wouldn't prepare for it.

climada/hazard/tc_tracks_forecast.py
ChrisFairless: `is_ensemble` should be `False` when the track is from the high-res deterministic model (see above suggestion). this is how it's implemented in the code.

this setup is true for a lot of meteorological products, where an agency provides _n_ ensemble model simulations (here 51, i think) plus their 'deterministic forecast', a higher resolution better model run as their single best guess. (the term 'deterministic' here is chosen to contrast with the 'probabilistic' ensemble members.) for something like the TCForecast it makes sense to think of this as a 52nd ensemble member. it should be distinguishable somehow in case e.g. you want to exclude it from an analysis, or you want to plot a representative track for the storm

climada/hazard/tc_tracks_forecast.py
ChrisFairless: alternately, could you run the loop starting at line 250 without knowing n_timesteps first?
i.e. replace the for loop at line 250 with this while loop?

climada/hazard/tc_tracks_forecast.py
chahank: Ok, I see. But how does this work with backward compatibility? Will this code just stop working after every update of CLIMADA or ECMWF?

climada/hazard/tc_tracks_forecast.py
chahank: OK, thanks for the explanation. The name is however still quite cryptic ihmo. Maybe choose another name (unless it is some general standard). And could you maybe expand the docstring accordingly?

climada/hazard/tc_tracks_forecast.py
chahank: I am a bit worried about the while loop which does not look like it has a very clear stopping point.

climada/hazard/tc_tracks_forecast.py
ChrisFairless: it'll stop working after every big update to ECMWF BUFR formats, but there's no way to prepare for that (e.g. the latest update replaced ~12 variables with a single integer that gives the same variables via a lookup).

the hope is that, by using ECMWF's ecCodes to read files, some breaking updates to the forecast messages can be handled by updating ecCodes instead of rewriting CLIMADA.

backward compatibility will always be a nightmare. we have two options, i guess:
- people have to check out old versions of CLIMADA to read old-format BUFR messages (this is how it works now)
- we add a 'version' parameter to `read_one_bufr_tc`, and point to different methods based on that parameter

i'd suggest sticking with what we've got until someone needs to work with archived messages, since that's not yet a use case

manniepmkam: So this the the missing doubles and missing integers defined in the ECMWF's bufr files. Theses missing constants are defined in the ecCodes constant values. I assume that if they change the bufr formats, they will also update the constants value from their library.

chahank: What do you mean with "i'd suggest sticking with what we've got until someone needs to work with archived messages, since that's not yet a use case"? 

I would say that having access to past/archived forecasts is absolutely crucial for research. Or I am missing something here?

manniepmkam: Using ecCodes can access to past forecast tracks. Their recent changes of the BUFR files involve the change of the 'descriptors' (see [here](https://confluence.ecmwf.int/display/FCST/Update+to+Tropical+Cyclone+tracks)), that is a 6 digits number that is used to access the information. The old codes were written evolving the old descriptors and there's is no easy fix to read the new forecast tracks.

The missing constants defined are the same for past and present BUFR files. In a very unlikely case that they may change the values of these missing constants, but then it should be their responsibility to update `eccodes.CODES_MISSING_DOUBLE` and `eccodes.CODES_MISSING_LONG`.

climada/hazard/tc_tracks_forecast.py
manniepmkam: This is a while loop to obtain the number of time steps. I changed to `LOGGER.info('No. of timestep: %d', n_timestep)`

chahank: This is not a very good way to handle this. You are catching some arbitrary exception and then you do not handle it. I strongly suggest you rewrite this loop such that you do not use an exception as a stoping criterion.

climada/hazard/tc_tracks_forecast.py
manniepmkam: This has been updated.

climada/hazard/tc_tracks_forecast.py
manniepmkam: This part has been changed by list (dict) comprehension.

climada/hazard/tc_tracks_forecast.py
manniepmkam: Same as the above.

climada/hazard/tc_tracks_forecast.py
manniepmkam: Replaced by list (dict) comprehension.

manniepmkam:
To do list of this pull requests:
- Update the test files. Currently the unit test fails, one of the reason is that this version of code includes time step 0 forecast whereas the old version starts with time step 1. That is, if the forecast is retrieved at 00 UTC, the storm tracks include the 00 UTC forecast (time step 0). However in the old version, the storm tracks start at 6 UTC (time step 1). That may contribute to the failure.
- Include a test that test the downloading and processing of the latest forecast.
- code cleaning.

climada/hazard/tc_tracks_forecast.py
ChrisFairless: we also haven't seen a public archive of forecasts, so if they exist i assume they're behind a paywall

emanuel-schmid:
This Pull Request affects only files that are moved to climada_petals.
Hence it should be resubmitted there.

Issue #254, opened 2021-06-16
TCTrack basin ambiguous data types


emanuel-schmid:
The `xarray.Dataset` basin attribute can be a list of strings or a string. And in the first case it is not necessarily one for each track position but possibly one for all.

In any case a mere string should be prohibited, but also the "one for all" approach looks risky.

I suggest to make sure all `TCTracks` object have the same data structure in this regard, at creation time.
Each `xarray.Dataset` in the `data` list should have a list of strings as `basin` attribute that has the same size as `time`.

tovogt:
> Each xarray.Dataset in the data list should have a list of strings as basin attribute that has the same size as time.

That's already the case! I don't understand this issue.

emanuel-schmid:
Yes, my bad. I cannot find any evidence for the "one for all" approach either. 
However there is still the case where tracks were read from files and have a single string basin. As fixed here
 https://github.com/CLIMADA-project/climada_python/pull/247/commits/aed8968029e271a7fea99355149cdcffd50b05ac

The issue is reduced then to: the patch above should have happened earlier, when the track was created from file, not when it was used by `_tc_from_track`.

tovogt:
There is no way we can prevent users from changing our object's attribute to values that we don't support. I still don't get your point.

emanuel-schmid:
I want to check all the methods `tc_tracks.TCTracks.read_xy` and make sure they don't come up with at track that has not an array of basins.
That is the reason why the hotfix above was made in the first place. Only that it does not help much because the underlying problem, that the track is corrupt, comes up again a bit later.

tovogt:
Ah, okay, I see that there is a `read_netcdf` function that I haven't checked. I think that not many users use that function to load pre-computed tracks. Most users will use `pickle` because that produces a single file instead of a directory with thousands of files.

Please look here for a fix of `read_netcdf`: #255

emanuel-schmid:
Fixed by #255

PR #255, opened 2021-06-17
TCTracks.read_netcdf: convert legacy basin attribute on the fly

tovogt:
Following #254, this PR makes sure that track data that is read in using `TCTracks.read_netcdf` has a basin variable instead of the legacy basin string attribute.


emanuel-schmid:
Excellent. 😃

PR #257, opened 2021-06-22
Feature/fix rc wrapper

sameberenz:
put mask file name of wrapper method in CONFIG for better maintenance and ignore `mask*` files in hazard creation


carmensteinmann (approved):
Moving the filename into config seems reasonable. Thanks for adapting this!

PR #259, opened 2021-06-29
Feature/data api client

emanuel-schmid:
This pull introduces a novel util module, `climada.util.api_client` that exposes a `Client` which is a python wrapper around the data api at [climada.ethz.ch](https://climada.ethz.ch/rest/docs).
At present state, the wrapper provides convenience methods for querying the rest interface and download data.

In case a file has already been downloaded before, the `Client` notices this and just points to the already present file.
The download is supposed to work also in multi processing mode. If a download process is running that has been started from another `Client` , this will be noticed and any further attempt will fail with an according message for as long as the download is in progress.
To achieve this, a sqlite database is used as a locking device.

The `Client` currently depends on an open internet connection to access the rest api at climada.ethz.ch. Offline mode is not yet possible. This will (possibly) be handled in a future PR.

It is planned to extend the Client, to not just download files but immediately turn them into CLIMADA objects such as Centroids or Exposures. This will be handled in a future PR.

A user guide jupyter notebook will be added in the climada/doc/guide directory with a future PR.


climada/util/api_client.py
```diff
@@ -0,0 +1,472 @@
...
+        params = {
+            'data_type': data_type,
+            'name': name,
+            'version': version,
+            'status': '' if status is None else status,
+        }
+        params.update(properties if properties else dict())
+        return [DatasetInfo.from_json(ds) for ds in Client._request_200(url, params=params)]
+
+    def get_dataset(self, data_type=None, name=None, version=None, properties=None,
```
zeliest, line 224: is this function really necessary? Can't get_datasets just do both?

climada/util/api_client.py
```diff
@@ -0,0 +1,472 @@
...
+        try:
+            self._download(url=remote_url, path=local_path, replace=True)
+            dlf.enddownload = datetime.utcnow()
+            dlf.save()
+        except Exception:
+            dlf.delete_instance()
+            raise
+        return Download.get(Download.path==path_as_str)
+
+    def download_file(self, local_path, fileinfo, check=checksize, retries=3):
```
zeliest, line 377: this one could also have a automatically configured path as in download_dataset based on the datatype,... that it belongs to?

climada/util/api_client.py
```diff
@@ -0,0 +1,472 @@
...
+
+        Returns
+        -------
+        list of DataTypeInfo
+        """
+        url = f'{self.host}/rest/data_types'
+        params = {'data_type_group': data_type_group} \
+            if data_type_group else {}
+        return [DataTypeInfo(**jobj) for jobj in Client._request_200(url, params=params)]
+
```
zeliest, line 299: maybe also add a function which allows to get information on all the data_type belonging to a data_type_group, for example get all hazards, or all exposure

climada/conf/climada.conf
```diff
@@ -79,5 +79,10 @@
         }
     },
     "log_level": "WARNING",
-    "max_matrix_size": 1000000000
+    "max_matrix_size": 1000000000,
+    "data_api": {
+        "host": "https://climada.ethz.ch",
+        "chunk_size": 8192,
```
zeliest, line 85: does this mean that the chunk_size is fixed, would it make sense to adapt it to the size of the files?

zeliest (approved):

zeliest:
Thank you Emanuel, everything seems to work well. I just commented on some minor points.

climada/conf/climada.conf
emanuel-schmid: No, but it could make sense to adapt it to the download capacity of the internet connection. However I really think it hardly matters. I haven't tested it but if I had to guess I'd say anything between a 1K and 1M would do and the performance loss/gain would be in the low percentages.

climada/util/api_client.py
emanuel-schmid: The default usage of the client is meant to go along these lines:
```
client = Client()
dataset_info = client.get_dataset(data_type='rainfall', country='CHE')
downloaded_files = client.download_dataset(dataset)
do_something_with_(downloaded_files )
```
Like this, `get_dataset` would raise an Exception if there were more then one datasets matching the search parameters.
So if no exception is raised, one can be sure that the search parameters are sufficient to return the one and only desired dataset.
If `get_datasets` is used instead, one has to check for ambiguity right afterwards, to prevent arbitrary behaviour.

climada/util/api_client.py
emanuel-schmid: Yes, but not so easy: all information on how to automatically build the path is in dataset_info, which is not available.
To include such an option would load a lot of unessential complexity upon `download_file` which I want to avoid.

climada/util/api_client.py
emanuel-schmid: 🤔 ? 
I was thinking that's exactly what `get_data_types` does.

zeliest: Ah yes sorry

PR #262, opened 2021-07-01
Fix/integr tests lp

sameberenz:
- fix some more integration test fails on jenkins that were caused by call to GPW v4.11 2020 file not saved on jenkins, by setting reference year close to 2015 (2015 file is available on jenkins)
- restored two accidentially deleted `nightlight` functions that are required for `blackmarble` for legacy reasons: `load_nightlight_nasa` and `read_bm_file`.


emanuel-schmid:
🙌 Thanks!

PR #268, opened 2021-07-26
Fix Centroids.union and Hazard.concat

tovogt:
This PR fixes several problems with `Centroids.union` and `Hazard.concat`, the most important are the following:

- The function `Centroids.union` now avoids to call `set_geometry_points` if not necessary. This is important because `set_geometry_points` is really slow and should only be used if really needed.
- The function `Hazard.concat` will now make sure that the concatenated object is of the same subclass as the input objects. So, if you input several objects of type `TropCyclone`, the concatenated object will also be of type `TropCyclone`. Before this PR, the concatenated object would always be of the base type `Hazard`. I also added a regression test for this.

I comment on other changes directly in the code.


climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,48 @@ def union(self, *others):
...
-            and not np.all([cent.area_pixel.any() for cent in cent_list])
-        is_dist_coast = np.any([cent.dist_coast.any() for cent in cent_list]) \
-            and not np.all([cent.dist_coast.any() for cent in cent_list])
-        is_on_land = np.any([len(cent.on_land) > 0 for cent in cent_list]) \
-            and not np.all([len(cent.on_land) > 0 for cent in cent_list])
-        is_region_id = np.any([cent.region_id.any() for cent in cent_list]) \
-            and not np.all([cent.region_id.any() for cent in cent_list])
-        is_elevation = np.any([cent.elevation.any() for cent in cent_list]) \
-            and not np.all([cent.elevation.any() for cent in cent_list])
+        # restrict to non-empty centroids
```
tovogt, line 558: Since this function's operation is like the union of sets, empty centroid objects should not change the result at all. We remove them in the very first step to avoid manual checks for emptiness of objects later in the code.

climada/hazard/centroids/centr.py
```diff
@@ -793,9 +781,12 @@ def remove_duplicate_points(self, scheduler=None):
         """
         if not self.lat.any() and not self.meta:
             return self
-        self.set_geometry_points(scheduler)
-        geom_wkb = self.geometry.apply(lambda geom: geom.wkb)
-        sel_cen = geom_wkb.drop_duplicates().index
+        if self.lat.size > 0:
+            coords_view = self.coord.astype(np.float64).view(dtype='float64,float64')
+            sel_cen = np.sort(np.unique(coords_view, return_index=True)[1])
```
tovogt, line 786: If the points are stored in `self.lat` and `self.lon`, it's much faster to use these two lines to identify duplicate points than applying `set_geometry_points` and using `drop_duplicates` afterwards.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -135,7 +135,7 @@ def test_union(self):
         np.testing.assert_array_equal(cent.lon, [0, -1, -2, 3])
         np.testing.assert_array_equal(cent.on_land, [True, True, False, False])
 
-        cent = Centroids().union(cent1)
+        cent = Centroids.union(cent1)
```
tovogt, line 138: No need to create a new object here.

climada/hazard/centroids/test/test_vec_ras.py
```diff
@@ -139,19 +139,18 @@
...
 ])
 
+def data_vector():
+    vec_data = gpd.GeoDataFrame(crs={'init': 'epsg:32632'})
+    vec_data['geometry'] = list(zip(VEC_LON, VEC_LAT))
+    vec_data['geometry'] = vec_data['geometry'].apply(Point)
+    vec_data['lon'] = VEC_LON
+    vec_data['lat'] = VEC_LAT
+    vec_data['value'] = np.arange(VEC_LAT.size) + 100
+    return vec_data.lat.values, vec_data.lon.values, vec_data.geometry
```
tovogt, line 149: Since this function is used in several classes in this module, it doesn't make sense to have it as a static method of one of the classes.

climada/hazard/trop_cyclone.py
```diff
@@ -214,12 +215,14 @@ def set_from_tracks(self, tracks, centroids=None, description='',
...
                     last_perc = perc
-                self.append(
+                tc_haz.append(
                     self._tc_from_track(track, centroids, coastal_idx,
                                         model=model, store_windfields=store_windfields,
                                         metric=metric))
             if last_perc < 100:
                 LOGGER.info("Progress: 100%")
+        LOGGER.debug('Append events.')
+        self.__dict__ = TropCyclone.concat([self] + tc_haz).__dict__
```
tovogt, line 225: Without this line, the function is broken for people that use `pool`.

climada/util/coordinates.py
```diff
@@ -634,6 +634,8 @@ def coord_on_land(lat, lon, land_geom=None):
     if lat.size != lon.size:
         raise ValueError('Wrong size input coordinates: %s != %s.'
                          % (lat.size, lon.size))
+    if lat.size == 0:
+        return np.empty((0,), dtype=bool)
```
tovogt, line 638: For compatibility, this function should be able to handle size-0 arrays.

chahank:
Just as a quick feedback: many of the awkward changes that you are correcting might break other CLIMADA modules such as `TropCyclone`. I struggled quite a bit to make these changes compatible with what is already there. If you can improve them keeping compatibility, this is great. So, please run the tests for all modules to check compatibility.

climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,48 @@ def union(self, *others):
...
-            and not np.all([len(cent.on_land) > 0 for cent in cent_list])
-        is_region_id = np.any([cent.region_id.any() for cent in cent_list]) \
-            and not np.all([cent.region_id.any() for cent in cent_list])
-        is_elevation = np.any([cent.elevation.any() for cent in cent_list]) \
-            and not np.all([cent.elevation.any() for cent in cent_list])
+        # restrict to non-empty centroids
+        cent_list = [c for c in (self,) + others if c.size > 0 or c.meta]
+        if len(cent_list) == 0 or len(others) == 0:
+            return copy.deepcopy(self)
+        self = cent_list[0]
```
chahank, line 562: That looks like a very strange line. Replace self with and `=` ?

climada/hazard/trop_cyclone.py
chahank: But replacing the `__dict__` of the object is also quite sub-optimal. Can one do without it?

climada/hazard/centroids/centr.py
tovogt: I expected that you wouldn't like this, so I won't argue, but just changed it in my latest commit :smile:

tovogt:
All the tests should be working, I even fixed some tests that were failing unrelated to the changes introduced here (the xarray problems).

climada/hazard/trop_cyclone.py
tovogt: How do you mean it's sub-optimal? What's wrong with it? There are already several places in CLIMADA with this kind of assignment to `self.__dict__`.

chahank: Having it in older code doesn't make it optimal :D . More seriously, first of all it is not recommend to do so: https://docs.python.org/3/library/stdtypes.html 
```
A special attribute of every module is __dict__. This is the dictionary containing the module’s symbol table. Modifying this dictionary will actually change the module’s symbol table, but direct assignment to the __dict__ attribute is not possible (you can write m.__dict__['a'] = 1, which defines m.a to be 1, but you can’t write m.__dict__ = {}). Modifying __dict__ directly is not recommended.
```
I think what would be better is  to have a proper way to initialize the object and then use this. Here I would like to ask @emanuel-schmid to give his input, as he can do that much better than I could.

tovogt: No, that quotation clearly refers to the `__dict__` attribute of Python modules. Here, we have a class instance's `__dict__` attribute.

chahank:
>     * The function `Hazard.concat` will now make sure that the concatenated object is of the same subclass as the input objects. So, if you input several objects of type `TropCyclone`, the concatenated object will also be of type `TropCyclone`. Before this PR, the concatenated object would always be of the base type `Hazard`. I also added a regression test for this.

Probably this is a good idea. Could you please add to the docstring that the type of the first object in the list is used?

climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,47 @@ def union(self, *others):
...
-        is_region_id = np.any([cent.region_id.any() for cent in cent_list]) \
-            and not np.all([cent.region_id.any() for cent in cent_list])
-        is_elevation = np.any([cent.elevation.any() for cent in cent_list]) \
-            and not np.all([cent.elevation.any() for cent in cent_list])
+        # restrict to non-empty centroids
+        cent_list = [c for c in (self,) + others if c.size > 0 or c.meta]
+        if len(cent_list) == 0 or len(others) == 0:
+            return copy.deepcopy(self)
 
+        # check if all centroids agree
```
chahank, line 563:
```suggestion
        # check if all centroids are identical
```

climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,47 @@ def union(self, *others):
...
+        if all([cent_list[0].equal(cent) for cent in cent_list[1:]]):
```
chahank, line 564: I think this will fail if a centroids as no defined `crs` (which was one of the reasons to `set_geometry_points`). This was addressed by initializing centroids with a default `crs`. However, older data might still fail. Probably not a big deal.

climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,47 @@ def union(self, *others):
...
         for cent in cent_list:
             if cent.meta and not cent.lat.any():
                 cent.set_meta_to_lat_lon()
+
+        # make sure that all Centroids have the same CRS
+        for cent in cent_list:
             if cent.crs is None:
                 cent.geometry = cent.geometry.set_crs(DEF_CRS)
-            cent.set_geometry_points()
             if not u_coord.equal_crs(cent.crs, cent_list[0].crs):
```
chahank, line 576: This can be removed as it is checked in line 564

climada/hazard/centroids/centr.py
```diff
@@ -555,60 +555,47 @@ def union(self, *others):
...
+                for cent in cent_list:
+                    if not getattr(cent, attr).size > 0:
+                        fun_name = f"set_{attr}{'_points' if attr == 'geometry' else ''}"
+                        getattr(Centroids, fun_name)(cent)
+
+        # create new Centroids object and set concatenated attributes
+        centroids = Centroids()
+        centroids.meta = {}
+        for attr_name, attr_val in vars(cent_list[0]).items():
             attr_val_list = [getattr(cent, attr_name) for cent in cent_list]
```
chahank, line 592: Technically, this could be moved in the if clauses such that it is not computed when the attribute is neither a 1D numpy array nor a geoseries.

climada/hazard/centroids/centr.py
chahank: If both `self.lat` and `self.geometry` are defined, which one is faster?

climada/hazard/centroids/test/test_centr.py
chahank: Agreed.

climada/hazard/storm_europe.py
```diff
@@ -241,7 +241,7 @@ def read_cosmoe_file(self, fp_file, run_datetime, event_date=None,
 
         # read intensity from file
         ncdf = xr.open_dataset(fp_file)
-        ncdf = ncdf.assign_coords(date=('time',ncdf["time"].dt.floor("D")))
+        ncdf = ncdf.assign_coords(date=('time',ncdf["time"].dt.floor("D").values))
```
chahank, line 244: I am not sure to understand the need of this change. Coulb you please comment?

chahank:
Maybe it would also be useful to add a check to see whether the hazard are all of the same sub-class?

chahank:
Great job @tovogt ! I think that we should only settle the discussion regarding `__dict__` replacement (here input from @emanuel-schmid is needed).

chahank:
This is slightly out of this pull request, but I noted that the added method `hazard.change_centroids` has an issue: the method will have troubles if two or more centroids are mapped onto the same closest centroids. If you have a quick fix, it would be good to add it in this pull request. Else, it is for another time.

climada/hazard/storm_europe.py
tovogt: Starting with xarray 0.19.0, this will raise an Error. Before, it was only a deprecation warning. https://github.com/pydata/xarray/pull/5630/files#diff-43c76e9be8425b5b6897dcecab4b240c32580447455c0c8c0b9b7fd84ce8a15dL119-R121

This already fails on Jenkins, so I thought we can fix it quickly in this PR.

tovogt:
> Maybe it would also be useful to add a check to see whether the hazard are all of the same sub-class?

This is already part of this PR: https://github.com/CLIMADA-project/climada_python/pull/268/files#diff-3d9f1d416e26f7c178b30a55b7c8e4cfcfdc8c7099416572cf3f04dd1e7fb155R1479-R1482

climada/hazard/centroids/centr.py
tovogt: This shouldn't fail because there is now a fallback CRS in `Centroids.crs`: https://github.com/CLIMADA-project/climada_python/pull/268/files#diff-0ec37b2aaa4e9ad606f7859d74412d0af63acc0712fbc49baed2a7935487e4edR1013-R1015

chahank:
> > Maybe it would also be useful to add a check to see whether the hazard are all of the same sub-class?
> 
> This is already part of this PR: https://github.com/CLIMADA-project/climada_python/pull/268/files#diff-3d9f1d416e26f7c178b30a55b7c8e4cfcfdc8c7099416572cf3f04dd1e7fb155R1479-R1482

Oh sorry, I missed that!

tovogt:
> This is slightly out of this pull request, but I noted that the added method `hazard.change_centroids` has an issue: the method will have troubles if two or more centroids are mapped onto the same closest centroids. If you have a quick fix, it would be good to add it in this pull request. Else, it is for another time.

Done in 71a5ba9c

climada/hazard/centroids/centr.py
tovogt: It's definitely much faster to use pure NumPy functionality wherever possible. In this case, the operation `self.geometry.apply(lambda geom: geom.wkb)` alone is already factor 10 slower than the two lines above. The `geom_wkb.drop_duplicates().index` operation then doesn't make much of a difference.

tovogt:
The discussion about `__dict__` can continue in #271 and #272

PR #269, opened 2021-07-27
TCTracks.generate_centroids

tovogt:
Originally, this change was part of #245 (now moved to https://github.com/CLIMADA-project/climada_petals/pull/2).

This adds a method `generate_centroids` to the `TCTracks` class that returns a Centroids object covering the region affected by the TC tracks with the specified resolution.

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1070,6 +1070,8 @@ class Hazard():
         self.clear()
         hf_data = h5py.File(file_name, 'r')
         for (var_name, var_val) in self.__dict__.items():
+            if var_name != 'tag' and var_name not in hf_data.keys():
+                continue
             if var_name == 'centroids':
                 self.centroids.read_hdf5(hf_data.get(var_name))
             elif var_name == 'tag':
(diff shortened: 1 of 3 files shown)


climada/hazard/base.py
```diff
@@ -1070,6 +1070,8 @@ def read_hdf5(self, file_name):
         self.clear()
         hf_data = h5py.File(file_name, 'r')
         for (var_name, var_val) in self.__dict__.items():
+            if var_name != 'tag' and var_name not in hf_data.keys():
+                continue
```
tovogt, line 1074: If the Hazard class has attributes that are not stored in the HDF file, then the reading of the HDF file will fail. That's a general problem. For example, if a new attribute is added to a Hazard class at some point, then old HDF files won't be read any more. This is motivated from the TCSurgeGeoClaw class (https://github.com/CLIMADA-project/climada_petals/pull/2), which has a class attribute (`gauge_data`) that is not written to the HDF file because it has a format that is not compatible with the HDF format. When the HDF file is read later, the reading of this attribute shouldn't be enforced (it will fail because it's not present in the HDF file).

PR #271, opened 2021-07-28
Implement Hazard.append/concat without __dict__

tovogt:
As noted in a discussion about #268, the use of `__dict__` in `Hazard.append` is avoidable. This PR shows how we can avoid it while being 100% backwards compatible.

I did some performance improvements, bug fixes and cleanups along the way. 

To avoid the use of `__dict__` in `TropCyclone.set_from_tracks`, see #272.

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1399,16 +1399,27 @@ class Hazard():
     def append(self, *others):
         """Append the events and centroids to this hazard object.
 
-        All of the given hazards must be of the same type as self. The centroids of all hazards
-        must have the same CRS.
+        All of the given hazards must be of the same type and use the same units as self. The
+        centroids of all hazards must have the same CRS.
 
-        The centroids of all hazards are combined together. All raster centroids are converted to
-        points and raster data is discarded.
+        The following kinds of object attributes are processed:
+
+        - All centroids are combined together using `Centroids.union`.
+
+        - Lists, 1-dimensional arrays (NumPy) and sparse CSR matrices (SciPy) are concatenated.
+        Sparse matrices are concatenated along the first (vertical) axis.
+
+        - All `tag` attributes are appended to `self.tag`.
+
+        For any other type of attribute: A ValueError is raised if an attribute of that name is
+        not defined in all of the non-empty hazards at least. However, there is no check that the
+        attribute value is identical among the given hazard objects. The initial attribute value of
+        `self` will not be modified.
 
         Note: Each of the hazard's `centroids` attributes might be modified in place in the sense
         that missing properties are added, but existing ones are not overwritten. In case of raster
         centroids, conversion to point centroids is applied so that raster information (meta) is
-        lost.
+        lost. For more information, see `Centroids.union`.
 
         Parameters
         ----------
@@ -1417,11 +1428,12 @@ class Hazard():
 
         Raises
         ------
-        TypeError
+        TypeError, ValueError
 
         See Also
         --------
-        Hazard.concat: concatenate 2 or more hazards
+        Hazard.concat : concatenate 2 or more hazards
+        Centroids.union : combine centroids
         """
         if len(others) == 0:
             return
@@ -1495,30 +1507,24 @@ class Hazard():
         """
         Concatenate events of several hazards of same type.
 
-        Centroids of all hazards must either all be rasters with the same
-        resolution, or all be points.
-
-        The centroids of all hazards are combined together.
-        All raster centroids are converted to points and raster data
-        is discarded.
+        This function creates a new hazard of the same class as the first hazard in the given list
+        and then applies the `append` method. Please refer to the docs of `Hazard.append` for
+        caveats and limitations of the concatenation procedure.
 
         Parameters
         ----------
-        haz_list: list of climada.hazard.Hazard objects
+        haz_list : list of climada.hazard.Hazard objects
             Hazard instances of the same hazard type (subclass).
 
         Returns
         -------
-        haz_concat: instance of climada.hazard.Hazard
+        haz_concat : instance of climada.hazard.Hazard
             This will be of the same type (subclass) as all the hazards in `haz_list`.
-
-        Raises
-        ------
-        ValueError
 
         See Also
         --------
-        hazard.centroids.Centroids.union: combine centroids
+        Hazard.append : append hazards to a hazard in place
+        Centroids.union : combine centroids
         """
         haz_concat = haz_list[0].__class__()
         haz_concat.tag.haz_type = haz_list[0].tag.haz_type
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1396,22 +1396,23 @@ class Hazard():
         self.intensity = sparse.csr_matrix(dfr.values[:, 1:num_events + 1].transpose())
         self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape, dtype=float))
 
-    def append(self, *haz_list):
+    def append(self, *others):
         """Append the events and centroids to this hazard object.
 
-        All of the given hazards must be of the same type as self.
-        The centroids of all hazards must have the same CRS.
+        All of the given hazards must be of the same type as self. The centroids of all hazards
+        must have the same CRS.
 
-        The centroids of all hazards are combined together.
-        All raster centroids are converted to points and raster data
-        is discarded.
+        The centroids of all hazards are combined together. All raster centroids are converted to
+        points and raster data is discarded.
 
-        Note: `self.centroids` is modified in place and centroid raster information (meta)
-        is destroyed.
+        Note: Each of the hazard's `centroids` attributes might be modified in place in the sense
+        that missing properties are added, but existing ones are not overwritten. In case of raster
+        centroids, conversion to point centroids is applied so that raster information (meta) is
+        lost.
 
         Parameters
         ----------
-        haz_list : one or more climada.hazard.Hazard objects
+        others : one or more climada.hazard.Hazard objects
             Hazard instances to append to self
 
         Raises
@@ -1422,9 +1423,9 @@ class Hazard():
         --------
         Hazard.concat: concatenate 2 or more hazards
         """
-        if len(haz_list) == 0:
+        if len(others) == 0:
             return
-        haz_list = (self,) + haz_list
+        haz_list = [self] + list(others)
         haz_list_nonempty = [haz for haz in haz_list if haz.size > 0]
 
         for haz in haz_list:
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1457,7 +1457,7 @@ class Hazard():
 
         units = {haz.units for haz in haz_list if haz.units != ''}
         if len(units) > 1:
-            raise ValueError(f"The hazards use different units: {units}. "
+            raise ValueError(f"The given hazards use different units: {units}. "
                              "The hazards are incompatible and cannot be concatenated.")
         elif len(units) == 0:
             units = {''}
@@ -1502,8 +1502,8 @@ class Hazard():
         self.sanitize_event_ids()
 
 
-    @staticmethod
-    def concat(haz_list):
+    @classmethod
+    def concat(cls, haz_list):
         """
         Concatenate events of several hazards of same type.
 
@@ -1513,6 +1513,12 @@ class Hazard():
 
         For centroids, tags, lists, arrays and sparse matrices, the remarks in `Hazard.append`
         apply. All other attributes are copied from the first object in `haz_list`.
+
+        Note that `Hazard.concat` can be used to concatenate hazards of a subclass. The result's
+        type will be the subclass. However, calling `concat([])` (with an empty list) is equivalent
+        to instantiation without init parameters. So, `Hazard.concat([])` is equivalent to
+        `Hazard()`. If `HazardB` is a subclass of `Hazard`, then `HazardB.concat([])` is equivalent
+        to `HazardB()` (unless `HazardB` overrides the `concat` method).
 
         Parameters
         ----------
@@ -1529,6 +1535,8 @@ class Hazard():
         Hazard.append : append hazards to a hazard in place
         Centroids.union : combine centroids
         """
+        if len(haz_list) == 0:
+            return cls()
         haz_concat = haz_list[0].__class__()
         haz_concat.tag.haz_type = haz_list[0].tag.haz_type
         for attr_name, attr_val in vars(haz_list[0]).items():
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1540,7 +1540,8 @@ class Hazard():
         haz_concat = haz_list[0].__class__()
         haz_concat.tag.haz_type = haz_list[0].tag.haz_type
         for attr_name, attr_val in vars(haz_list[0]).items():
-            # only copy simple attributes like "units" to save memory
+            # to save memory, only copy simple attributes like
+            # "units" that are not explicitly handled by Hazard.append 
             if not (isinstance(attr_val, (list, np.ndarray, sparse.csr.csr_matrix))
                     or attr_name in ["tag", "centroids"]):
                 setattr(haz_concat, attr_name, copy.deepcopy(attr_val))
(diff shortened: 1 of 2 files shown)


climada/hazard/base.py
```diff
@@ -1326,7 +1325,7 @@ def _read_att_mat(self, data, file_name, var_names):
         self.frequency = np.squeeze(data[var_names['var_name']['freq']])
         self.orig = np.squeeze(data[var_names['var_name']['orig']]).astype(bool)
         self.event_id = np.squeeze(
-            data[var_names['var_name']['even_id']].astype(np.int, copy=False))
+            data[var_names['var_name']['even_id']].astype(int, copy=False))
```
tovogt, line 1328: See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

climada/hazard/base.py
```diff
@@ -1345,8 +1344,7 @@ def _read_att_mat(self, data, file_name, var_names):
         except ValueError as err:
             raise ValueError('Size missmatch in fraction matrix.') from err
         except KeyError:
-            self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape,
-                                                      dtype=np.float))
+            self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape, dtype=float))
```
tovogt, line 1347: See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
                              f'{dfr.shape[0]} != {self.centroids.size}')
 
         self.intensity = sparse.csr_matrix(dfr.values[:, 1:num_events + 1].transpose())
-        self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape,
-                                                  dtype=np.float))
+        self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape, dtype=float))
```
tovogt, line 1397: See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
...
+        Hazard.concat: concatenate 2 or more hazards
         """
-        hazard._check_events()
-        if type(self) != type(hazard):
-            raise TypeError(f"hazard is of class {type(self)} which is "
-                            f" different from {type(hazard)}")
+        if len(haz_list) == 0:
+            return
+        haz_list = (self,) + haz_list
+        haz_list_nonempty = [haz for haz in haz_list if haz.size > 0]
```
tovogt, line 1428: For the actual concatenation of array/list type attributes later, only non-empty Hazard objects are relevant, so we keep track of those separately.

climada/hazard/base.py
```diff
@@ -1469,63 +1519,10 @@ def concat(haz_list):
...
-                setattr(haz_concat, attr_name, sum(attr_val_list, []))
-            elif isinstance(attr_val_list[0], TagHazard):
-                for tag in attr_val_list:
-                    if tag is not haz_concat.tag: haz_concat.tag.append(tag)
-
-        haz_concat.sanitize_event_ids()
-
+        haz_concat.tag.haz_type = haz_list[0].tag.haz_type
+        haz_concat.units = haz_list[0].units
+        haz_concat.append(*haz_list)
```
tovogt, line 1525: By delegating the static concat-operation to the non-static append-operation, we avoid the use of `__dict__`.

climada/hazard/test/test_base.py
```diff
@@ -925,15 +925,15 @@ def test_hazard_pass(self):
...
         self.assertEqual(hazard.frequency.shape, (n_events,))
         self.assertEqual(hazard.frequency[0], 0.01)
         self.assertEqual(hazard.frequency[n_events - 2], 0.001)
 
-        self.assertEqual(hazard.intensity.dtype, np.float)
+        self.assertEqual(hazard.intensity.dtype, float)
         self.assertEqual(hazard.intensity.shape, (n_events, n_centroids))
 
-        self.assertEqual(hazard.fraction.dtype, np.float)
+        self.assertEqual(hazard.fraction.dtype, float)
```
tovogt, line 936: See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

climada/hazard/test/test_base.py
```diff
@@ -966,15 +966,15 @@ def test_hazard_pass(self):
...
         self.assertEqual(hazard.frequency.shape, (n_events,))
 
-        self.assertEqual(hazard.intensity.dtype, np.float)
+        self.assertEqual(hazard.intensity.dtype, float)
         self.assertEqual(hazard.intensity.shape, (n_events, n_centroids))
         self.assertEqual(hazard.intensity[12, 46], 12.071393519949979)
         self.assertEqual(hazard.intensity[13676, 49], 17.228323602220616)
 
-        self.assertEqual(hazard.fraction.dtype, np.float)
+        self.assertEqual(hazard.fraction.dtype, float)
```
tovogt, line 977: See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
                              f'{dfr.shape[0]} != {self.centroids.size}')
 
         self.intensity = sparse.csr_matrix(dfr.values[:, 1:num_events + 1].transpose())
-        self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape,
-                                                  dtype=np.float))
+        self.fraction = sparse.csr_matrix(np.ones(self.intensity.shape, dtype=float))
 
-    def append(self, hazard):
-        """Append the events and centroids in hazard.
+    def append(self, *haz_list):
```
chahank, line 1399: This is a bit too much maybe, but `haz_list` is here not a list but a tuple.

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
...
+
+        units = {haz.units for haz in haz_list if haz.units != ''}
+        if len(units) > 1:
+            raise ValueError(f"The hazards use different units: {units}. "
+                             "The hazards are incompatible and cannot be concatenated.")
+        elif len(units) == 0:
+            units = {''}
+        self.units = units.pop()
+
+        attributes = sorted(set.union(*[set(vars(haz).keys()) for haz in haz_list]))
```
chahank, line 1453: What is the sorted needed for here?

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
...
+        for attr_name in attributes:
+            if not all(hasattr(haz, attr_name) for haz in haz_list_nonempty):
+                raise ValueError(f"Attribute {attr_name} is not shared by all hazards. "
+                                 "The hazards are incompatible and cannot be concatenated.")
+
+        # append all tags (to keep track of input files and descriptions)
+        for haz in haz_list:
```
chahank, line 1460: Should it be `haz_list` or `haz_list_nonempty`? Not sure what is conceptually more sensible.

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
...
+            if haz.tag is not self.tag:
+                self.tag.append(haz.tag)
+
+        # map individual centroids objects to union
+        centroids = Centroids.union(*[haz.centroids for haz in haz_list])
```
chahank, line 1465: Should it be `haz_list` or `haz_list_nonempty`? Not sure what is conceptually more sensible.

climada/hazard/base.py
```diff
@@ -1396,47 +1394,99 @@ def _read_att_excel(self, file_name, var_names):
...
+                    sparse.csr_matrix(
+                        (matrix.data, cent_idx[matrix.indices], matrix.indptr),
+                        shape=(matrix.shape[0], centroids.size)
+                        )
+                    for matrix, cent_idx in zip(attr_val_list, hazcent_in_cent_idx_list)
+                    ], format='csr'))
+            elif isinstance(attr_val_list[0], np.ndarray) and attr_val_list[0].ndim == 1:
+                setattr(self, attr_name, np.hstack(attr_val_list))
+            elif isinstance(attr_val_list[0], list):
+                setattr(self, attr_name, sum(attr_val_list, []))
```
chahank, line 1486: There we must think a bit about what to do with non-standard parameters. For instance `WildFire` objects have attributes of the custom dataclasses `FirmsParams()` and `ProbaParams()`.

climada/hazard/base.py
chahank: But why not just make a copy of the first element in list, and then append the rest?

```
haz_concat = copy.deepcopy(haz_list[0])
haz_concat.append(*haz_list[1:])
```

tovogt: In the extreme case where the first element is an extremely large Hazard, this would copy the large Hazard's data in memory here and then copy it again when concatenating the attributes so that there is a moment in the execution of the code where you have three times the memory requirements of that Hazard, instead of just twice. This sounds quite hypothetical, but I run into out-of-memory problems all the time because of all kinds of reasons, so I wouldn't be surprised to see this scenario at some point.

climada/hazard/base.py
tovogt: This makes sure that we iterate over the attributes in a reproducible and consistent way in all of the following for-loops. From my experience, you should never iterate over a Python set. I think that shouldn't even be supported because it leads to inconsistent behavior when running the very same code twice on the same setup.

climada/hazard/base.py
tovogt: I only use `haz_list_nonempty` for cases where it obviously doesn't make a difference whether to include them or not.

In this case, it makes a difference: Empty hazards can still contain tag information about files and descriptions. For example, they could have been loaded from a file that doesn't contain any events because of the way it has been generated (maybe as part of batch output, year sets, regional sets etc.).

climada/hazard/base.py
tovogt: As I said above, I only use haz_list_nonempty for cases where it obviously doesn't make a difference whether to include them or not.

In this case, it makes a difference: When the hazards have been generated in a batch-kind of fashion, some of the hazards might contain no events, but they might still cover a region that is not covered by the other hazards.

climada/hazard/base.py
tovogt: In my opinion, somebody who introduces the non-standard parameters must deal with this. If there is absolutely no way to adapt the non-standard parameters to fit to the current implementation of the concatenation, then that developer can create a new PR and suggest changes. But we can't a priori have support for all kinds of non-standard parameters. By supporting lists we are extremely flexible already. Other objects might have Pandas Series or all kinds of list-like objects in the future. But I suggest to support this, only as soon as somebody asks for it.

climada/hazard/base.py
tovogt: Solved in cd8720ef

climada/hazard/base.py
tovogt: In the meantime, I had a look at the `WildFire` class. Since `FirmsParams` and `ProbaParams` aren't iterables and don't depend on the event size, your comment doesn't seem to be related to this part of the code, but is of a more general concern.

When subclasses add attributes that are not event-specific, but ensemble-specific. It's still up to the creators of those subclasses to make sure that they work as expected with `append`. Currently, since the append-operation is in-place, the attributes of the `self` object remain untouched unless they are event-specific (list/array) or tag attributes.

A more complex approach would be for subclasses to define "compatibility attributes": a class-specific set of attributes that need to be identical in order for objects to be appendable. For the base class, those attributes would be `haz_type` and `units`. Subclasses would extend that set similar to the current use of the `vars_opt` class attribute.

tovogt:
Thanks for the feedback, @chahank !

Can we please focus on the main topic of this PR, namely avoiding `__dict__` without losing backwards compatibility? I don't have time to discuss (or implement) more fundamental changes to the concatenation procedure currently, and would appreciate if we could settle this (and #272) efficiently. If you feel like fundamental changes to the concatenation procedure are required (e.g. to support exotic attribute types), then please open a new PR or issue for that.

climada/hazard/base.py
chahank: I think here the order of iteration is irrelevant, so the sorting would not be strictly needed. But I do not mind it.

climada/hazard/base.py
chahank: I agree. My question was whether this is the desired functionality (e.g. that the centroids of the empty hazard are added). I am fine with it.

climada/hazard/base.py
tovogt: The order of iteration is irrelevant if and only if everything works as expected. If you have a bug in your code, it can be extremely hard to debug due to inconsistent behavior. You obviously never had to deal with this kind of bug, otherwise you would mind. :D

climada/hazard/base.py
chahank: >When subclasses add attributes that are not event-specific, but ensemble-specific, it's still up to the creators of those subclasses to make sure that they work as expected with append

I do not think this is a feasible requirement. How should a developer in the future that when developing a Hazard they should specifically go look into the code of the `hazard.append` method to ensure compatibility?

chahank:
While in principle I agree with you that focusing one thing at a time in a pull request is sensible, the developer structure of CLIMADA simply does not allow for that. We do not have a set of developers dedicated to collect issues and address them. Hence, whenever an important issue arises, it should be dealt with immediately, otherwise, it gets lost. Opening a new issue is in principle possible, but comes with an unnecessary overhead imho, which just results in the issue never being addressed.

In the point in case, the quick minimal solution is to hint to how this works in the docstring. So that users and developers are aware that non list/np.array/crs_matrix/tags elements are not appended/concatenated.

climada/hazard/base.py
tovogt: Yes, that's how inheritance works. This is nothing that we "require", but it's just a fact. When you inherit from a class that somebody else defines, then you yourself have to make sure that you don't break the base class's functionality. It's always like that. There is no alternative.

chahank: I agree with the principle. But if the issue is hidden in the code, that becomes increasingly difficult.

tovogt:
The procedure is now explained in detail in the docstring:
https://github.com/CLIMADA-project/climada_python/blob/3d80eaa59a58118d84a270f8eb49209aa847dcfc/climada/hazard/base.py#L1405-L1416

climada/hazard/base.py
chahank: However, this will erase all the non-list attributes, which would be a different behaviour from `append`. This is not necessarily surprising, still noteworthy.

tovogt: Okay, then I propose to copy all non-list attributes:
https://github.com/CLIMADA-project/climada_python/blob/5bc9eb4155cf662988e7ce54a06ff9de4faedbcc/climada/hazard/base.py#L1542-L1546

climada/hazard/base.py
```diff
@@ -1396,136 +1394,157 @@ def _read_att_excel(self, file_name, var_names):
...
+        --------
+        Hazard.append : append hazards to a hazard in place
+        Centroids.union : combine centroids
+        """
+        if len(haz_list) == 0:
+            return cls()
+        haz_concat = haz_list[0].__class__()
+        haz_concat.tag.haz_type = haz_list[0].tag.haz_type
+        for attr_name, attr_val in vars(haz_list[0]).items():
+            # only copy simple attributes like "units" to save memory
```
chahank, line 1543:
```suggestion
            # to save memory, only copy simple attributes like
            # "units" that are not explicitly handled by Hazard.append 
```

climada/hazard/base.py
chahank: I just made a suggestion to make it very clear that the 'not simple' attributes are explicitly handled by `Hazard.append`.

chahank:
Thanks a lot for the improvements. From my side it is ready to be merged (just look at the one small suggestion for in-code comment). Maybe @emanuel-schmid would like to comment, so I leave it un-merged for the moment.

emanuel-schmid:
Excellent. Thanks a lot!

PR #272, opened 2021-07-28
Implement TropCyclone.from_track without __dict__

tovogt:
As noted in a discussion about #268, the use of `__dict__` in `TropCyclone.set_from_tracks` is avoidable. This PR shows how we can avoid it by introducing a static constructor method `TropCyclone.from_tracks`. While `set_from_tracks` is not removed for backwards compatibility, it is deprecated and replaced by `from_tracks` in all of the tutorials.

This PR is based on #271, so make sure to merge that before looking into this one.

### DIFF
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -109,16 +109,17 @@ class TropCyclone(Hazard):
     """Name of the variables that aren't need to compute the impact."""
 
     def __init__(self, pool=None):
-        """Empty constructor."""
-        Hazard.__init__(self, HAZ_TYPE)
+        """Initialize values.
+
+        Parameters
+        ----------
+        pool : pathos.pool, optional
+            Pool that will be used for parallel computation when applicable. Default: None
+        """
+        Hazard.__init__(self, haz_type=HAZ_TYPE, pool=pool)
         self.category = np.array([], int)
         self.basin = []
         self.windfields = []
-        if pool:
-            self.pool = pool
-            LOGGER.info('Using %s CPUs.', self.pool.ncpus)
-        else:
-            self.pool = None
 
     def set_from_tracks(self, *args, **kwargs):
         """This function is deprecated, use TropCyclone.from_tracks instead."""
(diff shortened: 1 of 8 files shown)


climada/hazard/trop_cyclone.py
```diff
@@ -216,16 +232,20 @@ def set_from_tracks(self, tracks, centroids=None, description='',
                     LOGGER.info("Progress: %d%%", perc)
                     last_perc = perc
                 tc_haz.append(
```
chahank, line 234: I cannot make a suggestion directly for all here, but I suggest to rename `tc_haz` to `tc_haz_list`.

```suggestion
                tc_haz_list.append(
```

climada/hazard/trop_cyclone.py
```diff
@@ -216,16 +232,20 @@ def set_from_tracks(self, tracks, centroids=None, description='',
...
+                                                  model=model, store_windfields=store_windfields,
+                                                  metric=metric, intensity_thres=intensity_thres))
             if last_perc < 100:
                 LOGGER.info("Progress: 100%")
-        LOGGER.debug('Append events.')
-        self.__dict__ = TropCyclone.concat([self] + tc_haz).__dict__
+
+        LOGGER.debug('Concatenate events.')
+        haz = TropCyclone.concat(tc_haz)
+        haz.pool = pool
```
chahank, line 243: Is this still needed? What is the use of having the pool attached to self with the static method approach?

chahank:
Thanks for the proposal @tovogt ! I think it is a good improvement which keeps backwards compatibility for some time.

climada/hazard/trop_cyclone.py
tovogt: Fixed in 97b4e31b

climada/hazard/trop_cyclone.py
tovogt: Currently, the pool attribute is a standard attribute of the `Hazard` base class. We should retain it for consistency with the base class and for backwards compatibility. Users might have code where the pool that has been used to generate a Hazard is reused later and accessed through the `pool` attribute.

chahank:
From my point of view, this is ready for merge. Maybe @emanuel-schmid wants to add something, hence I leave it unmerged for the moment.

emanuel-schmid:
very well, very well. 🙌

PR #276, opened 2021-08-03
Clean up the logger setup

tovogt:
Now that the logger setup for climada_petals is fixed (see https://github.com/CLIMADA-project/climada_petals/pull/3), there are also some small simplifications we can apply to the core's logging setup.

I explained the changes by adding inline review comments here on GitHub (see below).


climada/util/config.py
```diff
@@ -321,3 +312,4 @@ def _fetch_conf(directories, config_name):
     Path(Path.home(), '.config'),  # ~/.config directory
     Path.cwd(),  # current working directory
 ], CONFIG_NAME))
+LOGGER.setLevel(getattr(logging, CONFIG.log_level.str()))
```
tovogt, line 315: `config.py` is the file where both `LOGGER` and `CONFIG` are initialized. It doesn't make sense to me to make the final step of setting to log-level of `LOGGER` according to `CONFIG` outside of this file. It should happen right here, once the `CONFIG` is loaded.

climada/__init__.py
```diff
@@ -93,10 +91,4 @@ def setup_climada_data(reload=False):
...
                 copyfile(src, path)
 
-
-def init():
-    setup_climada_data()
-    setup_logging(CONFIG.log_level.str())
-
-
-init()
+setup_climada_data()
```
tovogt, line 94: The `init` function is not needed after https://github.com/CLIMADA-project/climada_petals/pull/3 is settled.

climada/util/config.py
```diff
@@ -48,14 +47,6 @@ def remove_handlers(logger):
 LOGGER.addHandler(CONSOLE)
 
 
-def setup_logging(log_level='DEBUG'):
-    """Setup logging configuration"""
-    remove_handlers(LOGGER)
-    LOGGER.propagate = False
-    LOGGER.setLevel(getattr(logging, log_level))
-    LOGGER.addHandler(CONSOLE)
```
tovogt, line 56: Apart from updating the log-level (which I now moved to the end of the file), this only repeats configuration steps that are already applied in the lines above.

PR #277, opened 2021-08-11
Bugfix/savefig forecast

ThomasRoosli:
Fix Error that occurs in new environment v2.1.1 upwards, that was not catched in the tests. The bug is fixed and the tests updated to catch the error in the future.

### DIFF
--- a/climada/engine/test/test_forecast.py
+++ b/climada/engine/test/test_forecast.py
@@ -26,13 +26,14 @@ import geopandas as gpd
 import matplotlib.pyplot as plt
 import fiona
 from cartopy.io import shapereader
+from pathlib import Path
 
 from climada import CONFIG
 from climada.hazard.storm_europe import StormEurope
 from climada.entity.exposures.base import Exposures, INDICATOR_IMPF
 from climada.entity import ImpactFuncSet
 from climada.entity.impact_funcs.storm_europe import ImpfStormEurope
-from climada.engine.forecast import Forecast
+from climada.engine.forecast import Forecast, FORECAST_PLOT_DIR
 from climada.util.constants import WS_DEMO_NC
 
 HAZ_DIR = CONFIG.hazard.test_data.dir()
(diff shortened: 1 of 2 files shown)


climada/engine/test/test_forecast.py
```diff
@@ -162,6 +155,21 @@ def test_Forecast_plot(self):
                     for f in source:
                         if f['properties']['adm0_a3'] == 'CHE':
                             sink.write(f)
+        #test plotting functions
+        forecast.plot_imp_map(run_datetime=dt.datetime(2017,12,31),
+                              polygon_file=str(cantons_file),
+                              save_fig=True,close_fig=True)
```
chahank, line 161:
```suggestion
                              save_fig=True, close_fig=True)
```

climada/engine/test/test_forecast.py
```diff
@@ -162,6 +155,21 @@ def test_Forecast_plot(self):
...
+        map_file_name = (forecast.summary_str(dt.datetime(2017,12,31)) +
+                         '_impact_map' +
+                         '.jpeg')
+        map_file_name_full = Path(FORECAST_PLOT_DIR) / map_file_name
+        map_file_name_full.absolute().unlink(missing_ok=True)
```
chahank, line 166: Should it not fail if the file was not created?
```suggestion
        map_file_name_full.absolute().unlink(missing_ok=False)
```

climada/engine/test/test_forecast.py
```diff
@@ -162,6 +155,21 @@ def test_Forecast_plot(self):
...
+        forecast.plot_hist(run_datetime=dt.datetime(2017,12,31),
+                           save_fig=False,close_fig=True)
```
chahank, line 168:
```suggestion
                           save_fig=False, close_fig=True)
```

chahank:
Small comments, otherwise good for me.

PR #278, opened 2021-08-12
Bugfix/forecst number abbrevations

ThomasRoosli:
I fixed a bug in the number abbrevations used in the labeling of the axis (before when the value was 10 it resulted in 0 trillion, now it correctly results in 10, whilst the rest stays the same, e.g. 1000 results in 1 thousand etc...).

In the same pull request I also removed an unnecessary conversion of a Config Path to a string and back to a Path.


chahank:
All good, thanks!

PR #289, opened 2021-09-23
Feature/api wrapper functions

zeliest:
Add wrapper functions to client to be able to directly open Climada objects, see Climada_api_client notebook to see how these can be used. Still work in progress :)


climada/util/api_client.py
```diff
@@ -486,3 +493,145 @@ def purge_cache(self, local_path):
         """
         dlf = Download.get(Download.path==str(local_path.absolute()))
         dlf.delete_instance()
+
+    def get_hazard(self, hazard_type=None, data_dir=SYSTEM_DIR):
```
zeliest, line 497: the functions gets_hazard get_exposures and get_litpop_default do very similar things, I think I could simplify them

climada/util/api_client.py
```diff
@@ -486,3 +493,145 @@ def purge_cache(self, local_path):
...
+            if os.path.isfile(os.path.join(data_dir, dataset.files[0].file_name)):
+                LOGGER.info('The file already exists and it was not downloaded again.')
+            exposures = Exposures()
+            exposures.read_hdf5(os.path.join(data_dir, dataset.files[0].file_name))
+            exposures_list.append(exposures)
+        exposures_concat = Exposures()
+        exposures_concat = exposures_concat.concat(exposures_list)
+        return exposures_concat
+
+    def get_litpot_default(self, country=None, data_dir=SYSTEM_DIR):
```
zeliest, line 562: I think that this function is useful on top of get_exposures() as mostly people will use the default and I think it may be easier without choosing options

climada/util/api_client.py
```diff
@@ -486,3 +493,145 @@ def purge_cache(self, local_path):
...
+
+    def _get_data(self, type):
+        datasets = self.get_datasets(data_type=type)
+        properties_keys = np.unique([dataset.properties.keys() for dataset in datasets])
+        # find common properties between "groups" of datasets:
+        properties_keys = set(properties_keys[0]).intersection(*properties_keys)
+        user_properties_input = {}
+        # get user input to differentiate between these groups
+        user_properties_input = self._select_properties(datasets, properties_keys, user_properties_input)
+        datasets = self.get_datasets(data_type=type, properties=user_properties_input)
```
zeliest, line 598: any idea how to not query the datasets a second time here with the extra properties?

zeliest:
The exposures part doesn't work for now, I need to update the lit pop files which have some problems.  I will then also add the exposures to the tutorial.

climada/util/api_client.py
```diff
@@ -486,3 +493,145 @@ def purge_cache(self, local_path):
...
+            for country in user_properties_input['country_name']:
+                properties = user_properties_input.copy()
+                properties['country_name'] = country
+                datasets_properties.append(properties)
+        except:
+            datasets_properties = [user_properties_input]
+        datasets = [self.get_dataset(data_type=type, properties=properties) for properties in datasets_properties]
+        return datasets
+
+    def _select_properties(self, datasets, properties_keys, user_properties_input):
```
zeliest, line 614: I still have to make sure that only possible combinations of properties are suggested here

emanuel-schmid:
The approach with interactive methods is interesting and the functionality is definitely useful. _But_. The whole climada_python package is by nature a library and interactive methods are kind of odd there. Frankly, I don't think we should go in this direction.
I'd rather suggest to move the functionality, that guides a user to the right dataset selection, to the CLIMADA webservice where Chris is working on.
For the climada.util.api_client module, I'm about to write a suggestion that removes the interactive part. Sorry. I hope you're not too disappointed. Meanwhile, as you clearly demonstrate, there is a need to extend the get_datasets method in a way that one can select e.g. multiple countries at once. I'll also suggest something for that.

climada/util/api_client.py
```diff
@@ -486,3 +533,166 @@ def purge_cache(self, local_path):
         """
         dlf = Download.get(Download.path==str(local_path.absolute()))
         dlf.delete_instance()
+
+    def get_hazard(self, hazard_type, dump_dir=SYSTEM_DIR, **kwargs):
```
zeliest, line 537: should we maybe set a limit to the number of datasets that are downloaded in these functions by default? if you simply do get_hazard() or get_exposures() you get thousands of datasets downloaded it may be a bit dangerous for a new user that doesn't know anything about these functions?

climada/util/api_client.py
```diff
@@ -486,3 +533,166 @@ def purge_cache(self, local_path):
...
+            for dsf in dataset.files:
+                if dsf.file_format == 'hdf5':
+                    exposures_file = self.download_file(target_dir, dsf)
+                    exposures = Exposures()
+                    exposures.read_hdf5(exposures_file)
+                    exposures_list.append(exposures)
+        if not exposures_list:
+            raise ValueError("no exposures files found in datasets")
+
+        exposures_concat = Exposures()
```
zeliest, line 665: Here I only think it makes sense to concat country datasets but not different litpop datasets with different parameters

climada/util/api_client.py
```diff
@@ -486,3 +533,166 @@ def purge_cache(self, local_path):
...
+                if dsf.file_format == 'hdf5':
+                    hazard_file = self.download_file(target_dir, dsf)
+                    hazard = Hazard()
+                    hazard.read_hdf5(hazard_file)
+                    hazard_list.append(hazard)
+        if not hazard_list:
+            raise ValueError("no hazard files found in datasets")
+
+        hazard_concat = Hazard()
+        hazard_concat = hazard_concat.concat(hazard_list)
```
zeliest, line 599: it shouldn't be possible to concat different RCP datasets, I think only countries make sense

ChrisFairless:
> The approach with interactive methods is interesting and the functionality is definitely useful. _But_. The whole climada_python package is by nature a library and interactive methods are kind of odd there. Frankly, I don't think we should go in this direction. I'd rather suggest to move the functionality, that guides a user to the right dataset selection, to the CLIMADA webservice where Chris is working on. For the climada.util.api_client module, I'm about to write a suggestion that removes the interactive part. Sorry. I hope you're not too disappointed. Meanwhile, as you clearly demonstrate, there is a need to extend the get_datasets method in a way that one can select e.g. multiple countries at once. I'll also suggest something for that.

@zeliest let's talk about this - i've been working on something similar, but only for the web tool i'm making. you could maybe help me generalise.

PR #291, opened 2021-10-04
Example: Deprecate set_ methods in favor of static constructors

tovogt:
This PR replaces the `LitPop.set_countries` method by a static contructor method `LitPop.from_countries`.

**Please don't merge!** This is a reference example for the upcoming work session on the deprecation of `set_` methods in favor of static constructors in CLIMADA. I provide some commentary about the following aspects directly in the code:

1. Rename the `set_`-method, e.g. `set_countries` becomes `from_countries`.
2. Mark as `@staticmethod` or `@classmethod`.
3. Remove references to `self`, initialize a new object instead.
4. Make sure the new object is returned at the end.
5. Create a generic `set_`-method for backwards compatibility, including a deprecation warning.
6. Replace all occurrences of the `set_`-method in the rest of the CLIMADA code base, including tests and tutorials. In this case, the bash command `grep -r set_countries ./`, executed from the CLIMADA's git directory was very helpful.


climada/engine/uncertainty_quantification/input_var.py
```diff
@@ -66,8 +66,7 @@ class InputVar():
     Categorical variable function: LitPop exposures with m,n exponents in [0,5]
         import scipy as sp
         def litpop_cat(m, n):
-            exp = Litpop()
-            exp.set_country('CHE', exponent=[m, n])
+            exp = Litpop.from_countries('CHE', exponent=[m, n])
```
tovogt, line 69: This is a minimal example of the anti-pattern that we want to avoid by introducing the static constructors.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -69,16 +68,21 @@ class LitPop(Exposures):
...
 
-    def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
-                      fin_mode='pc', total_values=None, admin1_calc=False,
-                      reference_year=DEF_REF_YEAR, gpw_version=GPW_VERSION,
-                      data_dir=SYSTEM_DIR):
+    def set_countries(self, *args, **kwargs):
+        """This function is deprecated, use LitPop.from_countries instead."""
+        LOGGER.warning("The use of LitPop.set_countries is deprecated."
+                       "Use LitPop.from_countries instead.")
+        self.__dict__ = LitPop.from_countries(*args, **kwargs).__dict__
```
tovogt, line 75: This is the function structure to include for backwards compatibility, including a deprecation warning.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -69,16 +68,21 @@ class LitPop(Exposures):
...
+
+    @classmethod
```
tovogt, line 77: In this case, I decided to use a `@classmethod` so that we have a reference `cls` to the class that the developer is using to call this constructor. In the future, if somebody implements a derived class of `LitPop` (say `AdvancedLitPop`), this method can still be used as `AdvancedLitPop.from_countries(...)` and will return an instance of `AdvancedLitPop` instead of `LitPop`.

In some cases, the usage of `@staticmethod` might make sense, but in general, a `@classmethod` is safer.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -69,16 +68,21 @@ class LitPop(Exposures):
...
+    def from_countries(cls, countries, res_arcsec=30, exponents=(1,1),
```
tovogt, line 78: In this case, I decided to rename `set_countries` to `from_countries`. The prefix `from_` generally indicates that this is a constructor method (cf. `DataFrame.from_dict` in the pandas package).

However, you don't need to adhere to the original `set_`-method name so strictly. For example, `from_country_list` would have been an alternative in this case.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -176,8 +180,7 @@ def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
                           f'year: {reference_year}, financial mode: {fin_mode}, ' \
                           f'exp: {exponents}, admin1_calc: {admin1_calc}'
 
-        Exposures.__init__(
-            self,
+        exp = cls(
```
tovogt, line 183: Within the constructor, identify the location where the object initialization takes place and make sure to create a new instance instead of overwriting `self`.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -190,25 +193,26 @@ def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
...
         try:
             rows, cols, ras_trans = u_coord.pts_to_raster_meta(
-                (self.gdf.longitude.min(), self.gdf.latitude.min(),
-                 self.gdf.longitude.max(), self.gdf.latitude.max()),
-                u_coord.get_resolution(self.gdf.longitude, self.gdf.latitude))
-            self.meta = {
+                (exp.gdf.longitude.min(), exp.gdf.latitude.min(),
+                 exp.gdf.longitude.max(), exp.gdf.latitude.max()),
+                u_coord.get_resolution(exp.gdf.longitude, exp.gdf.latitude))
+            exp.meta = {
```
tovogt, line 199: Replace `self` by the new object everywhere in the constructor function.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -190,25 +193,26 @@ def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
...
                 'transform': ras_trans,
             }
         except ValueError:
             LOGGER.warning('Could not write attribute meta, because exposure'
                            ' has only 1 data point')
-            self.meta = {'crs': self.crs}
-        self.check()
+            exp.meta = {'crs': exp.crs}
+        exp.check()
+        return exp
```
tovogt, line 210: Don't forget to return the new object instance at the end of the constructor function!

climada/entity/exposures/litpop/litpop.py
```diff
@@ -190,25 +193,26 @@ def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
...
 
     def set_nightlight_intensity(self, countries=None, shape=None, res_arcsec=15,
                         reference_year=DEF_REF_YEAR, data_dir=SYSTEM_DIR):
         """
-        Wrapper around `set_countries` / `set_custom_shape`.
+        Wrapper around `from_countries` / `set_custom_shape`.
```
tovogt, line 215: Replace all occurrences of the deprecated `set_`-method by the new constructor, including comments, docstrings, examples, tests and tutorials.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -233,6 +237,7 @@ def set_nightlight_intensity(self, countries=None, shape=None, res_arcsec=15,
         if countries is not None and shape is not None:
             raise ValueError("Not allowed to set both `countries` and `shape`. Aborting.")
         if countries is not None:
+            # TODO: replace with LitPop.from_countries
             self.set_countries(countries, res_arcsec=res_arcsec,
```
tovogt, line 241: In some cases, it might make sense to postpone the replacement of the `set_`-method with the new constructor to a later point. Make sure to add a `TODO` comment in those cases.

In this example, the `set_nightlight_intensity` is a `set_`-method that needs to be replaced by a constructor, only in a later work step. We decide to leave the reference to `self.set_countries` within this method for now, but will replace it as soon as we deal with the `set_nightlight_intensity` method in one of the next work steps.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -508,11 +515,11 @@ def set_custom_shape(self, shape, total_value, res_arcsec=30, exponents=(1,1),
             self.meta = {'crs': self.crs}
 
     @staticmethod
-    def _set_one_country(country, res_arcsec=30, exponents=(1,1), fin_mode=None,
+    def _from_country(country, res_arcsec=30, exponents=(1,1), fin_mode=None,
```
tovogt, line 518: In this case, an internal method (prefixed with underscore) is a `set_`-method that our new constructor `LitPop.from_countries` relies on. Fortunately, it's only *called* `set_`, but is actually a proper constructor already. We only need to rename the function. In this case, we don't care about backwards compatibility because this is an internal function and not part of the official API.

doc/tutorial/1_main_climada.ipynb
```diff
@@ -668,8 +668,7 @@
    "source": [
     "from climada.entity.exposures import LitPop\n",
     "\n",
-    "exp_litpop = LitPop()\n",
-    "exp_litpop.set_country('Puerto Rico', res_arcsec = 120)  # We'll go lower resolution than default to keep it simple\n",
+    "exp_litpop = LitPop.from_countries('Puerto Rico', res_arcsec = 120)  # We'll go lower resolution than default to keep it simple\n",
```
tovogt, line 671: Don't forget to use the new constructor method in all tutorials, including Jupyter notebooks. You can search through `ipynb` files using the `grep` shell command.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -508,11 +515,11 @@ def set_custom_shape(self, shape, total_value, res_arcsec=30, exponents=(1,1),
...
                          total_value=None, reference_year=DEF_REF_YEAR,
                          gpw_version=GPW_VERSION, data_dir=SYSTEM_DIR):
         """init LitPop exposure object for one single country
-        See docstring of set_countries() for detailled description of parameters.
+        See docstring of from_countries() for detailled description of parameters.
```
tovogt, line 522: Replace all occurrences of the deprecated `set_`-method, not only in the code, but also in the docstrings and comments.

tovogt (commented):
I provide some commentary about the following aspects directly in the code:

1. Rename the `set_`-method, e.g. `set_countries` becomes `from_countries`.
2. Mark as `@staticmethod` or `@classmethod`.
3. Remove references to `self`, initialize a new object instead.
4. Make sure the new object is returned at the end.
5. Create a generic `set_`-method for backwards compatibility, including a deprecation warning.
6. Replace all occurrences of the `set_`-method in the rest of the CLIMADA code base, including tests and tutorials. In this case, the bash command `grep -r set_countries ./`, executed from the CLIMADA's git directory was very helpful.
7. Further issues that might arise in the process.

chahank:
Great example. I think we should decide on a preference between `static` and `class` methods. Imho, for these use cases `classmethod` might be more appropriate (to keep the classes and subclasses).

tovogt:
I agree about the use of `classmethod`!

I changed my posts to only mention the `staticmethod` in the comment, and not in the overview check list.

climada/entity/exposures/litpop/litpop.py
sameberenz: What would be reasons / cases to use `@staticmethod`?

tovogt: That would be a hypothetical case, where the `set_`-method is not really a pure constructor intrinsically, but does something more involved. I haven't looked at all the `set_` methods in detail, but some of them might be different from this example. They might create a new object from a given object or from a list of objects, like `Hazard.concat`. In that case the type of the returned object is determined using input parameters instead of the `cls` argument.

climada/entity/exposures/litpop/litpop.py
```diff
@@ -125,6 +129,10 @@ def set_countries(self, countries, res_arcsec=30, exponents=(1,1),
         Raises
         ------
         ValueError
+
+        Returns
+        -------
+        LitPop
```
tovogt, line 135: Make sure to mention in the docstring that the constructor method returns a (new) object. Go through the existing docstrings and amend any references to the deprecated approach, e.g. "clearing" or "filling" of the object.

tovogt:
I have rerun all the tutorials except from `climada_util_yearsets.ipynb` (non-functional), `climada_util_earth_engine.ipynb` (non-functional), and `climada_engine_uncertainty.ipynb` (too long, didn't want to wait). The three excluded ones shouldn't be significantly affected by this PR anyways.

PR #307, opened 2021-10-18
deprecate Hazard.set_vector, replace with Hazard.from_vector

ChrisFairless:
Part of the big antipattern hunt.

@tovogt this will cause a couple of merge conflicts with the Centroids methods it calls

I didn't find any use of this method in CLIMADA core or petals (scripts or notebooks).


ChrisFairless:
thanks to @sameberenz for reviewing

PR #309, opened 2021-10-18
Feature/deprecate set in entity and measures

chahank:
This branch needs the update in the hazard read methods before merging.

### DIFF
--- a/climada/entity/entity_def.py
+++ b/climada/entity/entity_def.py
@@ -132,10 +132,10 @@ class Entity(object):
         meas_set = MeasureSet.from_excel(file_name, description)
 
         return cls(
-            exposures = exp,
-            disc_rates = dr,
-            impact_func_set = impf_set,
-            measure_set = meas_set
+            exposures=exp,
+            disc_rates=dr,
+            impact_func_set=impf_set,
+            measure_set=meas_set
             )
 
     def read_excel(self, *args, **kwargs):


climada/entity/entity_def.py
```diff
@@ -50,14 +50,34 @@ class Entity(object):
...
+        impact_func_set : climada.entity.ImpactFuncSet, optional
+            The impact function set. The default is None (empty ImpactFuncSet()).
+        measure_set : climada.entity.Measures, optional
+            The measures. The default is None (empty MeasuresSet().
+
+        Returns
+        -------
+        None.
+
+        """
```
tovogt, line 73: Leave out the return section if there is no return value.

climada/entity/entity_def.py
```diff
@@ -69,22 +89,26 @@ def read_mat(self, file_name, description=''):
...
-
-        self.measures = MeasureSet()
-        self.measures.read_mat(file_name, description)
-
-    def read_excel(self, file_name, description=''):
+        return cls(
+            exposures = Exposures.from_mat(file_name),
+            disc_rates = DiscRates.from_mat(file_name, description),
+            impact_func_set = ImpactFuncSet.from_mat(file_name, description),
+            measure_set = MeasureSet.from_mat(file_name, description)
```
tovogt, line 101: Don't use spaces around the = sign when used to indicate a keyword argument ([PEP 8](https://www.python.org/dev/peps/pep-0008/))

climada/entity/entity_def.py
```diff
@@ -96,23 +120,33 @@ def read_excel(self, file_name, description=''):
...
+
+        dr = DiscRates.from_excel(file_name, description)
+        impf_set = ImpactFuncSet.from_excel(file_name, description)
+        meas_set = MeasureSet.from_excel(file_name, description)
+
+        return cls(
+            exposures = exp,
+            disc_rates = dr,
+            impact_func_set = impf_set,
+            measure_set = meas_set
```
tovogt, line 142: As above, don't use spaces around the = sign.

PR #313, opened 2021-10-20
update NONISO_REGIONS with new territories

### LINKED ISSUE #296: potential bug when calling util.coordinates.get_country_code()
There is something weird going on with `util.coordinates.get_country_code()`. In particular, if one tries to pass Brazil's lats and lons to get, trivially, Brazil's ISO3 codes:

```
from climada_petals.entity.exposures import BlackMarble
import climada.util.coordinates as u_coord

bm = BlackMarble()
bm.set_countries(['Brazil'])

reg_id = u_coord.get_country_code(bm.gdf.latitude, bm.gdf.longitude)
```

she gets:

`raise LookupError(f'Unknown country identifier: {country}') from None LookupError: Unknown country identifier: Brazilian I.`

which i think is due to some _country.NAME_ taken from the _nat_earth_shape_ in `util.coordinates.get_country_geometries()` not being found by the _pycountry_ package called afterwards by `util.coordinates.country_to_code()`.

I discovered this issue by noticing that `test_crop_production.py` in `climada_petals.entitity.exposures.test` crashes on my machine, but not on @emanuel-schmid's nor Jenkins apparently.

- Can someone check if the above code runs?
- If yes, why can't I run it ? I am using a freshly installed climada-env
- If not, this should be fixed.

aleeciu:
this should resolve #296

### DIFF
--- a/climada/util/constants.py
+++ b/climada/util/constants.py
@@ -164,6 +164,8 @@ NONISO_REGIONS = [
     dict(name="Akrotiri", alpha_2="XA", alpha_3="XXA", numeric="901"),
     dict(name="Baikonur", alpha_2="XB", alpha_3="XXB", numeric="902"),
     dict(name="Bajo Nuevo Bank", alpha_2="XJ", alpha_3="XXJ", numeric="903"),
+    dict(name="Bir Tawil", alpha_2="XQ", alpha_3="XXQ", numeric="919"),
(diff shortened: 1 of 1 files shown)


tovogt:
Looks fine to me. We are slowly running out of letters for the mock-up alpha_2-codes though :smile:

PR #315, opened 2021-10-25
config: introduce installation directory location

emanuel-schmid:
This PR is for making it possible to point to the installation directory of envipath in the config files.\
The notation to do this are three preceeding dots: `...`.

Example:
```
{
  "test_directory": ".../climada/test"
}
```
points to the directory where `climada.test` sources can be found.


climada/util/config.py
```diff
@@ -213,11 +220,18 @@ def dir(self, index=None, create=True):
...
         if create:
             path.mkdir(parents=True, exist_ok=True)
         return path.absolute()
 
+    @staticmethod
+    def _expand_source_dir(path):
+        parts = path.parts
+        if parts[0] == '...':
+            return Path(Config.SOURCE_DIR, *parts[1:])
+        return Path(*parts)
```
tovogt, line 233: Since this function accesses the class, wouldn't it be more natural to have a class method instead of a static method?
```suggestion
    @classmethod
    def _expand_source_dir(cls, path):
        parts = path.parts
        if parts[0] == '...':
            return Path(cls.SOURCE_DIR, *parts[1:])
        return Path(*parts)
```

emanuel-schmid: Agreed. Cleaner that way. (Even though `Config` is not meant to be extended)

PR #322, opened 2021-11-02
Fix deprecated +init CRS specs

### LINKED ISSUE #321: upgrade rasterio to 1.2.7
rastrio 1.2.7/8 are more distinctive with regard to CRS comparison, some CRS that used to be equal (`util.coordinates.equal_crs`) are different now.

run `climada.util.test.test_coordinates.TestRasterMeta.test_points_to_raster_pass` and `climada.util.test.test_plot.TestFuncs.test_get_transform_3035_pass` to reproduce,

tovogt:
This fixes #321 by removing all deprecated `+init` CRS specs from the code base.


climada/util/coordinates.py
```diff
@@ -1296,7 +1295,15 @@ def to_crs_user_input(crs_obj):
     ValueError
         if type(crs_obj) has the wrong type
     """
-    if type(crs_obj) in [dict, int]:
+    def _is_deprecated_init_crs(crs_dict):
+        return (isinstance(crs_dict, dict)
+                and "init" in crs_dict
+                and all(k in ["init", "no_defs"] for k in crs_dict.keys())
+                and ("no_defs" not in crs_dict or crs_dict['no_defs'] == True))
```
emanuel-schmid, line 1302:
```suggestion
                and crs_dict.get("no_defs", True) == True)
```

tovogt: Thanks, good idea!

PR #325, opened 2021-11-17
Feature/admin1 geometries

sameberenz:
- Add function `get_admin1_geometries` returning GDF with geometries of admin 1 regions within given countries.
- Adds flexibility to existing function `get_admin1_info` with regard to country identifier provided.

(I have not added the functionality to extract the geometry of a single admin 1 region in here - but this could be added with a wrapper function.)

### DIFF
--- a/climada/util/coordinates.py
+++ b/climada/util/coordinates.py
@@ -1139,12 +1139,12 @@ def get_country_code(lat, lon, gridded=False):
         region_id[region_id == -1] = 0
     return region_id
 
-def get_admin1_info(countries):
+def get_admin1_info(country_names):
     """Provide Natural Earth registry info and shape files for admin1 regions
 
     Parameters
     ----------
-    countries : list or str
+    country_names : list or str
         string or list with strings, either ISO3 code or names of countries, e.g.:
         ['ZWE', 'GBR', 'VNM', 'UZB', 'Kenya', '051']
         For example, for Armenia, the following inputs work:
@@ -1157,18 +1157,18 @@ def get_admin1_info(countries):
     admin1_shapes : dict
         Shape according to Natural Earth.
     """
-
-    if isinstance(countries, (str, int, float)):
-        countries = [countries]
-    if not isinstance(countries, list):
-        raise TypeError("Invalid type for input parameter 'countries'")
+    if isinstance(country_names, (str, int, float)):
+        country_names = [country_names]
+    if not isinstance(country_names, list):
+        LOGGER.error("country_names needs to be of type list, str, int or float")
+        raise TypeError("Invalid type for input parameter 'country_names'")
     admin1_file = shapereader.natural_earth(resolution='10m',
                                             category='cultural',
                                             name='admin_1_states_provinces')
     admin1_recs = shapefile.Reader(admin1_file)
     admin1_info = dict()
     admin1_shapes = dict()
-    for country in countries:
+    for country in country_names:
         if isinstance(country, (int, float)):
             country = f'{int(country):03d}' # transform numerric code to str
         country = pycountry.countries.lookup(country).alpha_3 # get iso3a code
--- a/climada/util/coordinates.py
+++ b/climada/util/coordinates.py
@@ -1139,13 +1139,14 @@ def get_country_code(lat, lon, gridded=False):
         region_id[region_id == -1] = 0
     return region_id
 
-def get_admin1_info(country_names):
+def get_admin1_info(countries):
     """Provide Natural Earth registry info and shape files for admin1 regions
 
     Parameters
     ----------
-    country_names : list
-        list with ISO3 names of countries, e.g. ['ZWE', 'GBR', 'VNM', 'UZB']
+    countries : list
+        list with strings, either ISO3 alpha code or names of countries, e.g.:
+        ['ZWE', 'GBR', 'VNM', 'UZB', 'Kenya']
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


climada/util/coordinates.py
```diff
@@ -1139,13 +1139,16 @@ def get_country_code(lat, lon, gridded=False):
         region_id[region_id == -1] = 0
     return region_id
 
-def get_admin1_info(country_names):
+def get_admin1_info(countries):
```
ChrisFairless, line 1142: Changing the parameter name here ... this will break any code that names the parameter, e.g. `x = get_admin1_info(country_names='ARM')`

I've looked through core and petals and I can't find any code that does this, so it's prooobably OK? It depends on how important it is to rename the parameter, and how long this method has been part of CLIMADA.

climada/util/coordinates.py
```diff
@@ -1155,23 +1158,81 @@ def get_admin1_info(country_names):
...
 
-    if isinstance(country_names, str):
-        country_names = [country_names]
+    if isinstance(countries, (str, int, float)):
+        countries = [countries]
+    if not isinstance(countries, list):
+        raise TypeError("Invalid type for input parameter 'countries'")
     admin1_file = shapereader.natural_earth(resolution='10m',
                                             category='cultural',
                                             name='admin_1_states_provinces')
```
ChrisFairless, line 1167: All of the analogous methods in this module have `resolution` as a parameter. Probably worth adding here as well.

Separately, they also have `extent` or `bounds` parameter, but I think that might be an issue for another time!

climada/util/coordinates.py
```diff
@@ -1155,23 +1158,81 @@ def get_admin1_info(country_names):
...
         for rec, rec_shp in zip(admin1_recs.records(), admin1_recs.shapes()):
-            if rec['adm0_a3'] == iso3:
-                admin1_info[iso3].append(rec)
-                admin1_shapes[iso3].append(rec_shp)
+            if rec['adm0_a3'] == country:
+                admin1_info[country].append(rec)
+                admin1_shapes[country].append(rec_shp)
     return admin1_info, admin1_shapes
 
+def get_admin1_geometries(countries):
```
ChrisFairless, line 1183: Repeating myself: it makes sense to add a `resolution` parameter (and a `bounds`/`extent` parameter at a later time)

climada/util/coordinates.py
```diff
@@ -1155,23 +1158,81 @@ def get_admin1_info(country_names):
...
+        country = pycountry.countries.lookup(country).alpha_3 # get iso3a code
+        admin1_info[country] = list()
+        admin1_shapes[country] = list()
         for rec, rec_shp in zip(admin1_recs.records(), admin1_recs.shapes()):
-            if rec['adm0_a3'] == iso3:
-                admin1_info[iso3].append(rec)
-                admin1_shapes[iso3].append(rec_shp)
+            if rec['adm0_a3'] == country:
+                admin1_info[country].append(rec)
+                admin1_shapes[country].append(rec_shp)
```
ChrisFairless, line 1180: There's a chance of a confusing error here if the ISO codes in pycountry are different from those in Natural Earth (which could happen with different versions of the packages). We should check that every country that pycountry identifies is successfully extracted from Natural Earth.
e.g. add a check that every list in `admin1_info` has length > 0

climada/util/coordinates.py
```diff
@@ -1155,23 +1158,81 @@ def get_admin1_info(country_names):
...
+    for country in admin1_info.keys():
+        # fill admin 1 region names and codes to GDF for single country:
+        gdf_tmp = gpd.GeoDataFrame(columns = gdf.columns)
+        gdf_tmp.admin1_name = [record['name'] for record in admin1_info[country]]
+        gdf_tmp.iso_3166_2 = [record['iso_3166_2'] for record in admin1_info[country]]
+        geoseries = gpd.GeoSeries()
+        for shape in admin1_shapes[country]:
+            # fill shape into GeoSeries,
+            # this automatically transforms shape to Polygon object:
+            geoseries = geoseries.append(gpd.GeoSeries(shape))
```
ChrisFairless, line 1227: could this be a list comprehension?
```suggestion
       geoseries = gpd.GeoSeries([gpd.Geoseries(shape) for shape in admin1_shapes[country]])
```

climada/util/coordinates.py
ChrisFairless: There's also the edge case to worry about, where a user provides duplicate countries in their list. Do we care about this?

My understanding right now is that `get_admin1_info(['CHE', 'CHE'])` returns the same as `get_admin1_info('CHE')` (but does the calculations twice). This seems like safe behaviour to be honest, but I wanted to check.

ChrisFairless:
Looks good!! A few comments above.

Your suggestion of a wrapper function to extract a single admin1, rather than the list for a country: I don't think we need this. The returned geodataframe is easy enough to subset, and it wouldn't reduce the time of calculation (which  I assume is mostly taken up by reading the Natural Earth shapefile). If a use case shows up, maybe.

climada/util/coordinates.py
sameberenz: good point - no urgent need to change that parameter's name, so I willl leave it

climada/util/coordinates.py
sameberenz: for 'admin_1_states_provinces', only one resolution is available (10m), so no choice here unfortunately

climada/util/coordinates.py
sameberenz: for 'admin_1_states_provinces', only one resolution is available (10m), so no choice here unfortunately

PR #327, opened 2021-11-19
hazard.Hazard.from_raster: call the constructor without haz_type

emanuel-schmid:
This PR fixes issue https://github.com/CLIMADA-project/climada_petals/issues/15.

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -277,9 +277,16 @@ class Hazard():
         if files_fraction is not None and len(files_intensity) != len(files_fraction):
             raise ValueError('Number of intensity files differs from fraction files: %s != %s'
                              % (len(files_intensity), len(files_fraction)))
-        if haz_type is None:
-            haz_type = cls().tag.haz_type
-        haz = cls(haz_type, pool)
+
+        if haz_type and cls is Hazard:
+            haz = cls(haz_type=haz_type, pool=pool)
+        else:
+            haz = cls(pool=pool)
+        
+        if haz_type and cls is not Hazard:
+            LOGGER.warn("haz_type argument ('%s') is ignored, haz_type is determined by class",
+                        haz_type)
+
         haz.tag.file_name = str(files_intensity) + ' ; ' + str(files_fraction)
 
         haz.centroids = Centroids.from_raster_file(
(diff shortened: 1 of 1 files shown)


tovogt:
Looks good to me :+1:

climada/hazard/base.py
```diff
@@ -277,9 +277,16 @@ def from_raster(cls, files_intensity, files_fraction=None, attrs=None,
...
             raise ValueError('Number of intensity files differs from fraction files: %s != %s'
                              % (len(files_intensity), len(files_fraction)))
-        if haz_type is None:
-            haz_type = cls().tag.haz_type
-        haz = cls(haz_type, pool)
+
+        if haz_type and cls is Hazard:
+            haz = cls(haz_type=haz_type, pool=pool)
+        else:
+            haz = cls(pool=pool)
```
chahank, line 284: I am not sure this is optimal. This will break as soon as someone makes a hazard with `haz_type` as argument right? Maybe this is very unlikely though...

tovogt: As I said in CLIMADA-project/climada_petals#15, there are infinitely many ways how any of this can fail. This PR is supposed to be a fix for the current state of the code *as is*. In a future version, we need to require new child classes to implement unit tests also for the inherited (and not only for the added) methods. In that case, this kind of problem can be discussed once it is actually introduced.

climada/hazard/base.py
```diff
@@ -277,9 +277,16 @@ def from_raster(cls, files_intensity, files_fraction=None, attrs=None,
...
+        
+        if haz_type and cls is not Hazard:
+            LOGGER.warn("haz_type argument ('%s') is ignored, haz_type is determined by class",
+                        haz_type)
```
emanuel-schmid, line 288:
```suggestion
        if haz_type is not None:
            try:
                haz = cls(haz_type=haz_type, pool=pool)
            except TypeError:
                haz = cls(pool=pool)
                LOGGER.warn("haz_type argument ('%s') is ignored, haz_type is determined by class",
                            haz_type)
        else:
            haz = cls(pool=pool)
```

climada/hazard/base.py
emanuel-schmid: I agree with @tovogt and I have some doubts about how well that will go if someone introduces haz_type as truly variable in a Hazard class. However - alternatively one can just _try_ to use the haz_type as an argument in the constructor. (See suggestion.)
Any other thinkable deviation of the constructor pattern for subclasses of  Hazard, `def __init__(self, pool)`, will still break.

Issue #328, opened 2021-11-22
Conversion of ISO numeric-3 codes to FAO country code does not work for integers


CyrilWillimann:
In climada_python/coordinates.py, the method _country_iso2faocode_ is supposed to accept both integers and arrays but currently only works for arrays. See screenshot below. Workaround with _fao_code = u_coord.country_iso2faocode(np.array([iso_code])_) works fine but maybe this could be improved.
![Screenshot 2021-11-22 at 13 44 09](https://user-images.githubusercontent.com/93337700/142864188-3d8bff0a-6407-4006-9d96-eefc2d6796d0.png)

carmensteinmann:
@chahank I would suggest adapting the docstring and only allowing for arrays. Should I push that change to climada_python/develop?

chahank:
I also think this is a good idea. Could you please do a branch and a pull request as usual?

carmensteinmann:
Sure

carmensteinmann:
Suggested changes: #329

emanuel-schmid:
@CyrilWillimann : thank you for pointing this out. The pydoc string has been modified in the develop branch, so that it is not contradicting the method's signature anymore.
This also means that the workaround suggested by you is now the "official" way to handle single codes.

PR #329, opened 2021-11-22
Correct docstring of country_iso2faocode method in util/coordinates

carmensteinmann:
PR fixing #328

### DIFF
--- a/climada/util/coordinates.py
+++ b/climada/util/coordinates.py
@@ -2173,7 +2173,7 @@ def country_iso2faocode(input_iso):
 
     Parameters
     ----------
-    input_iso : array
+    input_iso : iterable of int
         ISO numeric-3 code(s) of country/countries
 
     Returns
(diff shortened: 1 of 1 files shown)


climada/util/coordinates.py
```diff
@@ -2173,13 +2173,13 @@ def country_iso2faocode(input_iso):
 
     Parameters
     ----------
-    input_iso : int or array
-        ISO numeric-3 codes of countries (or single code)
+    input_iso : array
```
emanuel-schmid, line 2176:
```suggestion
    input_iso : iterable of int
```

climada/util/coordinates.py
```diff
@@ -2173,13 +2173,13 @@ def country_iso2faocode(input_iso):
...
+        ISO numeric-3 code(s) of country/countries
 
     Returns
     -------
-    output_faocode : int or array
-        FAO country codes of countries (or single code)
+    output_faocode : array
```
emanuel-schmid, line 2181:
```suggestion
    output_faocode : numpy.array
```

PR #331, opened 2021-12-03
Guide_Installation: add other workaround option to install spyder

sameberenz:
As this workaroudn is prefered by many (experienced) users, it should be added to the guide.


sameberenz:
c.f. https://github.com/CLIMADA-project/climada_python/blob/doc/guide_installation/doc/guide/Guide_Installation.ipynb for the adjusted workaround description @emanuel-schmid

emanuel-schmid:
Thanks a lot! 🙌

PR #332, opened 2021-12-07
Add kdtree nearest neighbour search

chahank:
This pull request adds the possibility to use a kTree for nearest neighbors matching using the Euclidian distance as metrics. This is much more efficient than a Ball tree with Haversine metrics for gridded data. 

Since CLIMADA data is often provided on a grid, even though the `meta` attribute might not be set, one could even make this the default method for the `Exposure.assign_centroids` method.


chahank:
Test methods pending.

climada/util/interpolation.py
```diff
@@ -186,3 +191,45 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
 
     # Copy result to all exposures and return value
     return np.squeeze(assigned[inv])
+
+
+def index_nn_euclidian(centroids, coordinates, threshold=THRESHOLD):
+    """Compute the neareast centroid for each coordinate using a Ball
+    tree with haversine distance.
```
tovogt, line 198:
```suggestion
    """Compute the neareast centroid for each coordinate using a k-d tree.
```

tovogt (commented):
Looks okay to me, though I didn't test it since unit tests are pending anyways.

chahank:
What do you think about the idea of making the kd-tree the default ?

tovogt:
S

tovogt:
Sorry for closing accidentally.

Making k-d trees the default should be safe for most use cases since a threshold of 100 km (less than 1 degree) is the default already. With this threshold in place, the most problematic thing to consider is data close to the antimeridian, which includes all global datasets.

climada/util/interpolation.py
```diff
@@ -186,3 +191,44 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
...
+    tree = sp.spatial.KDTree(np.radians(centroids))
+    # Select unique exposures coordinates
+    _, idx, inv = np.unique(coordinates, axis=0, return_index=True,
+                            return_inverse=True)
+
+    # query the k closest points of the n_points using dual tree
+    dist, assigned = tree.query(np.radians(coordinates[idx]), k=1, p=2, workers=-1)
+
+    # Raise a warning if the minimum distance is greater than the
+    # threshold and set an unvalid index -1
```
Evelyn-M, line 226: Ok, I see we've been doing this also for the other two methods; but is setting an index of -1 always save? I.e. in no circumstance it will just silently take the last index in the array as a result, if I use this afterwards?

Evelyn-M:
Very general comment: I find the kD-Tree implementation very useful, been using one similarly. But, would it be worth having the number of nearest neighbours returned as a kwarg option? I think all of the three sub-methods for the `interpol_index` have this as an easily settable parameter.
I for example need different numbers of NNs within a threshold (or even _all_ ), and I could imagine that one would not want to interpolate to a random centroid if a coordinate is located exactly in the middle of, say 4 grid corners.

climada/util/interpolation.py
tovogt: It's definitely not save, but there is no alternative.

chahank: Well, a Nan could also have been used, but -1 is probably still better.

chahank:
> I could imagine that one would not want to interpolate to a random centroid if a coordinate is located exactly in the middle of, say 4 grid corners.

Why not? What would you do in such a case? You have to assign a single centroid... Also, this is a very contrived example imho.

climada/util/interpolation.py
tovogt: Not if the data type is `int`. For `NaN` we need to switch the data type to `float`.

chahank:
> Making k-d trees the default should be safe for most use cases since a threshold of 100 km (less than 1 degree) is the default already. With this threshold in place, the most problematic thing to consider is data close to the antimeridian, which includes all global datasets. If making k-d trees the default we should consider normalizing the longitudinal coordinates before doing any further processing. But normalizing will not fix the problem of global data sets. For global datasets, you might run into problems in Fiji or Russia, depending on the location of the 0-meridian as selected by the longitude-normalization algorithm.

I agree. But, this problem might be no different from the current default settings. Whenever a grid (meta parameter) is set, the interpolation is also just based on Euclidean metrics... or does it wrap around?

climada/util/interpolation.py
chahank: Sure, but that could be worth it in principle. But probably not in the current state.

tovogt:
There is no issue with wrapping around when a ball tree with haversine distances is used.

chahank:
> There is no issue with wrapping around of coordinates when a ball tree with Haversine distances is used.

That is not the default method. This is the first fall-back method, when the meta is not defined.

tovogt:
Agreed: In case of an existing meta-attribute, wrapping around is an issue. This is unavoidable with the grid-based nearest-neighbor assignment algorithm implemented in `u_coord.assign_grid_points`.

chahank:
I made some changes to the code to:

1- add tests for the euclidian kd-tree
2- optionally (default is True) revert to Haversine in a strip of length Threshold around the anti-meridian if present

chahank:
> Very general comment: I find the kD-Tree implementation very useful, been using one similarly. But, would it be worth having the number of nearest neighbours returned as a kwarg option? I think all of the three sub-methods for the `interpol_index` have this as an easily settable parameter.

This could be useful indeed. But, so far only the NearestNeighbor methods are implemented. Having arbirtrary numbers or neighbours would require quite some rewriting.

tovogt:
Thanks for adding tests! However, I find it a bit unsatisfactory that we only have tests for the cases where the functions agree while, in fact, they are not doing the same in general. I think, there should be at least some test cases, where the function outputs differ from each other.

chahank:
Good points. I can add tests were they do not agree. Would you then test explicitly for disagreement between methods?

tovogt:
> I can add tests were they do not agree. Would you then test explicitly for disagreement between methods?

Take centroids like `(lon=-160, lat=60), (lon=-140, lat=60), ..., (lon=180, lat=-60)` (20 degree steps in lat and lon) and set the threshold to something very large (e.g. 10,000 km). Then check that the nearest neighbor of the point `(lon=9, lat=49.9)` is `(lon=0, lat=40)` in euclidean metric, and `(lon=0, lat=60)` in both the great circle (haversine) as well as the `approx` distance. On the other hand, for the point `(lon=9, lat=49.5)`, the point `(lon=0, lat=40)` is closest both in euclidean and haversine distance, but still `(lon=0, lat=60)` is closest in the `approx` distance. Then another test for the point `(lon=-175, lat=0)` which should give interesting results if `check_antimeridian` is set to `False`.

climada/util/interpolation.py
```diff
@@ -186,3 +202,95 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
...
+        coordinates that have assigned so far
+
+    Returns
+    -------
+    np.array
+        with so many rows as coordinates containing the centroids indexes
+
+    """
+
+    coord_strip_bool =  coordinates[:, 1] + np.pi < 1.5*threshold/EARTH_RADIUS_KM
```
tovogt, line 282: This assumes that all centroids and coordinates have longitudes in `[-180, 180]`. The recomputation will only fix assignments close to lon=180. If, however, the coordinates are in `[0, 360]` (or something completely different), the recomputation will not make it worse, but it will also fail to fix the assignments close to lon=0.

One solution would be for `index_antimeridian_strip` to automatically guess the `mid_lon` from the input coordinates, by computing:
```
lon_min = min(centroids[:, 1].min(), coordinates[:, 1].min())
lon_max = max(centroids[:, 1].max(), coordinates[:, 1].max())
if lon_max - lon_min > 180:
    raise ValueError("...")
mid_lon = 0.5 * (lon_max - lon_min)
antimeridian = mid_lon + 180
```

climada/util/interpolation.py
```diff
@@ -186,3 +202,95 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
...
+    coord_strip_bool += coordinates[:, 1] - np.pi >  -1.5*threshold/EARTH_RADIUS_KM
```
tovogt, line 283:
```suggestion
    coord_strip_bool |= coordinates[:, 1] - np.pi >  -1.5 * threshold / EARTH_RADIUS_KM
```

climada/util/interpolation.py
```diff
@@ -186,3 +202,95 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
...
+    if np.any(coord_strip_bool):
+        coord_strip = coordinates[coord_strip_bool]
+        cent_strip_bool = centroids[:, 1] + np.pi < 2.5*threshold/EARTH_RADIUS_KM
+        cent_strip_bool += centroids[:, 1] - np.pi >  -2.5*threshold/EARTH_RADIUS_KM
```
tovogt, line 287:
```suggestion
        cent_strip_bool |= centroids[:, 1] - np.pi >  -2.5 * threshold / EARTH_RADIUS_KM
```

climada/util/interpolation.py
```diff
@@ -186,3 +202,95 @@ def index_nn_haversine(centroids, coordinates, threshold=THRESHOLD):
...
+    # threshold and set an unvalid index -1
+    num_warn = np.sum(dist * EARTH_RADIUS_KM > threshold)
+    if num_warn:
+        LOGGER.warning('Distance to closest centroid is greater than %s'
+                       'km for %s coordinates.', threshold, num_warn)
+        assigned[dist * EARTH_RADIUS_KM > threshold] = -1
+
+    if check_antimeridian:
+        assigned = index_antimeridian_strip(np.radians(centroids),
+                                            np.radians(coordinates[idx]),
```
tovogt, line 250: Why do you convert everything to radians? If you really need the conversion, you can still do it within the function `index_antimeridian_strip`. I suggest to stick to degrees for compatibility with the other functions in this module.

tovogt (commented):
Some comments about the new `check_antimeridian` feature.

climada/util/interpolation.py
chahank: Good point. It is in CLIMADA sometimes implicitly assumed that the range is `[-180, 180]`, but it mustn't be the case.

climada/util/interpolation.py
chahank: Thanks! I was looking for the "or/equal", but somehow missed it.

climada/util/interpolation.py
chahank: The spaces were remove in a ill attempt to not have lines of more than 80 characters :D

tovogt: In CLIMADA, the convention is to have at most 100 characters per line:
https://github.com/CLIMADA-project/climada_python/blob/30cd7de5363807363bddfe2275cb2eaa248bd1a1/.pylintrc#L218

But yes, you should probably shorten the line by assigning `2.5 * threshold / EARTH_RADIUS_KM` to a variable (like `pad`).

climada/util/interpolation.py
chahank: Yes, it is better indeed.

climada/util/interpolation.py
chahank: I think it should be `mid_lon = 0.5 * (lon_max + lon_min)` though.

tovogt: Yes, sure. Well spotted :+1:

chahank:
> check_antimeridian

Thanks, these were great examples. I implemented these 
4c84a1b

climada/util/interpolation.py
```diff
@@ -278,17 +285,25 @@ def index_antimeridian_strip(centroids, coordinates, threshold, assigned):
...
 
     """
-
-    coord_strip_bool =  coordinates[:, 1] + np.pi < 1.5*threshold/EARTH_RADIUS_KM
-    coord_strip_bool += coordinates[:, 1] - np.pi >  -1.5*threshold/EARTH_RADIUS_KM
+    lon_min = min(centroids[:, 1].min(), coordinates[:, 1].min())
+    lon_max = max(centroids[:, 1].max(), coordinates[:, 1].max())
+    if lon_max - lon_min > 360:
+        raise ValueError("Longitudinal coordinates need to be normalized"
+                         "to 360degrees")
```
tovogt, line 292:
```suggestion
                         "to a common 360 degree range")
```

climada/util/interpolation.py
```diff
@@ -278,17 +285,25 @@ def index_antimeridian_strip(centroids, coordinates, threshold, assigned):
...
+    mid_lon = 0.5 * (lon_max + lon_min)
+    antimeridian = mid_lon + 180
+
+    thres_deg = np.degrees(threshold / EARTH_RADIUS_KM)
+    coord_strip_bool =  coordinates[:, 1] + antimeridian < 1.5 * thres_deg
```
tovogt, line 297:
```suggestion
    coord_strip_bool = coordinates[:, 1] + antimeridian < 1.5 * thres_deg
```

climada/util/interpolation.py
```diff
@@ -79,22 +80,24 @@ def interpol_index(centroids, coordinates, method=METHOD[0],
     threshold : float
         distance threshold in km over which no neighbor will
         be found. Those are assigned with a -1 index
+    kwargs: keyword arguments, optional
```
tovogt, line 83:
```suggestion
    kwargs: dict, optional
```

climada/util/interpolation.py
```diff
@@ -79,22 +80,24 @@ def interpol_index(centroids, coordinates, method=METHOD[0],
...
+        keyword arguments to be passed on the interpolation method
```
tovogt, line 84:
```suggestion
        keyword arguments to be passed on to the interpolation method
```

climada/util/interpolation.py
```diff
@@ -119,7 +122,9 @@ def index_nn_aprox(centroids, coordinates, threshold=THRESHOLD,
         be found. Those are assigned with a -1 index
     check_antimedirian: bool
         if True, recomputes the nn in a strip with lon size equal to threshold
-        around the antimeridian using the Haversine distance.
+        around the antimeridian using the Haversine distance. The antimeridian
+        is guessed from both coorindates and centroids, and is assumed
```
tovogt, line 126:
```suggestion
        is guessed from both coordinates and centroids, and is assumed
```

climada/util/interpolation.py
```diff
@@ -221,7 +226,9 @@ def index_nn_euclidian(centroids, coordinates, threshold=THRESHOLD,
         are assigned with a -1 index
     check_antimedirian: bool
         if True, recomputes the nn in a strip with lon size equal to threshold
-        around the antimeridian using the Haversine distance.
+        around the antimeridian using the Haversine distance. The antimeridian
+        is guessed from both coorindates and centroids, and is assumed
```
tovogt, line 230:
```suggestion
        is guessed from both coordinates and centroids, and is assumed
```

climada/util/test/test_interpolation.py
```diff
@@ -316,6 +315,45 @@ def test_euc_antimer_warning(self):
...
 
+    def test_diff_outcomes(self):
+        """Different NN interpolation outcomes"""
+        threshold = 100000
+
+        # Define centroids
+        lons = np.arange(-160, 180+1, 20)
+        lats = np.arange(-60, 60+1, 20)
+        lons = np.repeat(lons, len(lats))
+        lats = np.tile(lats, int(len(lons) / len(lats)))
```
tovogt, line 326:
```suggestion
        lats, lons = [arr.ravel() for arr in np.meshgrid(lats, lons)]
```

climada/util/test/test_interpolation.py
```diff
@@ -316,6 +315,45 @@ def test_euc_antimer_warning(self):
...
+        kwargs_list = [
+            {'check_antimeridian':False},
+            {},
+            {'check_antimeridian':False}
+            ]
+
+        for dist, ref, kwargs in zip(dist_list, ref_neighbors, kwargs_list):
+            neighbors= u_interp.interpol_index(centroids, exposures, 'NN',
+                                            dist, threshold=threshold,
+                                            **kwargs)
```
tovogt, line 353:
```suggestion
            neighbors= u_interp.interpol_index(centroids, exposures, 'NN',
                                               dist, threshold=threshold,
                                               **kwargs)
```

climada/util/test/test_interpolation.py
```diff
@@ -316,6 +315,45 @@ def test_euc_antimer_warning(self):
...
+            self.assertTrue(np.array_equal(neighbors, ref))
```
tovogt, line 354: More informative output in case of failure:
```suggestion
            np.testing.assert_array_equal(neighbors, ref)
```

tovogt (commented):
Thanks! Looks great apart from some details, see comments below.

chahank:
Thanks again for the outstanding review as usual!

tovogt:
Everything looks great. Just one little thing that strikes me now: The spelling "Euclidean" seems to be more common than "Euclidian", see e.g. (https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.euclidean.html). Would you mind changing the spelling everywhere to "Euclidean"?

chahank:
:D I was hesitating on that one already! I will change it.

chahank:
Euclidean is done 
1f254d5 !

tovogt:
Perfect, ready to be merged if @Evelyn-M doesn't have any further comments. By the way, concerning the question of returning the first `k` nearest neighbors instead of only the single nearest neighbor: I suggest to have that discussion in a separate place, maybe open a new PR with your suggested implementation, or a new issue.

chahank:
Agreed.

I think @Evelyn-M is away for some weeks now, so let us wait one or two days, and if there is no response I would merge. We can always reopen the case if needed.

climada/util/interpolation.py
```diff
@@ -59,7 +60,8 @@ def dist_sqr_approx(lats1, lons1, cos_lats1, lats2, lons2):
     return d_lon * d_lon * cos_lats1 * cos_lats1 + d_lat * d_lat
 
 def interpol_index(centroids, coordinates, method=METHOD[0],
-                   distance=DIST_DEF[1], threshold=THRESHOLD):
+                   distance=DIST_DEF[1], threshold=THRESHOLD,
```
chahank, line 63:
```suggestion
                   distance=DIST_DEF[2], threshold=THRESHOLD,
```

chahank: Shall we make the 'euclidian' distance the default as discussed before?

tovogt: Yes, let's do it. But please mention in the docstring below that `'euclidean'` is the default and also mention that there is a slower but more accurate alternative `'haversine'`.

chahank: Good point, will do! Thanks.

chahank:
I just realized, wouldn't it be more sensible if the method `interpol_index` actually does not have any default set? This is not really a visible method. Also, I checked that no at least no unittest fails (except in the `interpolation` module test specifically) if the defaults are removed.

climada/util/interpolation.py
chahank: I now added docstrings to explain the default use of 'euclidian' and that 'haversine' can be used when higher accuracy is needed. I hope it is clear 
3fb7778

chahank:
> I just realized, wouldn't it be more sensible if the method `interpol_index` actually does not have any default set? This is not really a visible method. Also, I checked that no at least no unittest fails (except in the `interpolation` module test specifically) if the defaults are removed.

In the end I did not yet implement this change and would wait for @tovogt opinion.

tovogt:
> I just realized, wouldn't it be more sensible if the method `interpol_index` actually does not have any default set?

Yes, just do whatever you want. I think the whole `climada.util.interpolation` module has a misleading API. It doesn't really have a lot to do with interpolation. At least, it's never used for interpolation purposes in the CLIMADA context. The `interpol_index` function is quite legacy and should be removed. The function `u_coord.assign_coordinates` can easily take it's place and directly call the nearest neighbor helper functions, here is what this would look like: #344

chahank:
Thanks for the great help and work @tovogt !!

PR #333, opened 2021-12-07
Feature/hazard select

chahank:
Allow to select the centroids of hazards by 'extent'. In addition, a method to reduce the hazard to the smallest rectangle of lat/lon containing all the non-zero intensity values.

Tests pending.


climada/hazard/base.py
```diff
@@ -760,6 +775,34 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
...
+        --------
+        self.select: Method to select centroids by lat/lon extent
+
+        """
+
+        cent_nz = self.intensity.nonzero()[1]
+        lon_nz = self.centroids.lon[cent_nz]
+        lat_nz = self.centroids.lat[cent_nz]
+        ext = u_coord.latlon_bounds(lat=lat_nz, lon=lon_nz, buffer=buffer)
+        return self.select(extent = [ext[0], ext[2], ext[1], ext[3]])
```
tovogt, line 804:
```suggestion
        return self.select(extent=(ext[0], ext[2], ext[1], ext[3]))
```

climada/hazard/base.py
```diff
@@ -760,6 +775,34 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
...
+            Copy of the Hazard with centroids reduced to minimal box. All other
+            hazard properties are carried over without changes.
+
+        See also
+        --------
+        self.select: Method to select centroids by lat/lon extent
+
+        """
+
+        cent_nz = self.intensity.nonzero()[1]
```
tovogt, line 800: This will create a lot of duplicate indices which creates unnecessary overhead in the call to `latlon_bounds` below. I suggest to reduce the intensity matrix to a single row before extracting non-zero columns:
```suggestion
        cent_nz = (self.intensity != 0).sum(axis=0).nonzero()[1]
```

climada/hazard/base.py
```diff
@@ -719,6 +723,14 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
             if not np.any(sel_cen):
                 LOGGER.info('No hazard centroids with region %s.', str(reg_id))
                 return None
+        if extent is not None:
+            cent_ext = self.centroids.select(extent=extent)
+            cent_ext_view = cent_ext.coord.view(dtype='float64,float64').reshape(-1)
+            cent_view = self.centroids.coord.view(dtype='float64,float64').reshape(-1)
+            sel_cen &= np.isin(cent_view, cent_ext_view)
```
tovogt, line 730: Instead of trying to reproduce the selection of `Centroids.select`, I suggest to add a new method `Centroids.select_mask` that implements the lines https://github.com/CLIMADA-project/climada_python/blob/1d8da77b06e72dc095f3d00b9e22c88ad8e6635a/climada/hazard/centroids/centr.py#L963-L974. It returns the `sel_cen` mask from given `reg_id` and/or `extent` parameters. This function should be used in `Centroids.select` and it should be used here (in `Hazard.select`) to compute the `sel_cen` mask to be used below.

climada/hazard/base.py
```diff
@@ -743,9 +755,12 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
...
             elif var_name == 'centroids':
                 if reg_id is not None:
-                    setattr(haz, var_name, var_val.select(reg_id))
+                    new_cent = var_val.select(reg_id=reg_id, extent=extent)
+                elif extent is not None:
+                    new_cent = cent_ext
                 else:
-                    setattr(haz, var_name, var_val)
+                    new_cent = var_val
+                setattr(haz, var_name, new_cent)
```
tovogt, line 763: This should be:
```
if reg_id is None and extent is None:
    new_cent = var_val
else:
    new_cent = var_val.select(sel_cen=sel_cen)
setattr(haz, var_name, new_cent)
```

tovogt (changes requested):
A useful feature! The implementation could be improved a bit, see my comments in the code.

climada/hazard/base.py
chahank: Good point. Not sure which one is faster, but I think your suggestion is.

climada/hazard/base.py
chahank: That is a good idea.

climada/hazard/base.py
```diff
@@ -679,6 +680,9 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
             Select only historical (True) or only synthetic (False) events.
         reg_id : int, optional
             Region identifier of the centroids' region_id attibute.
+        extent: tuple(float, float, float, float), optional
+            Extent of centroids as (min_lon, max_lon, min_lat, max_lat).
```
Evelyn-M, line 684: Just a general comment, since I see that it's the same order in the centroids module, but having min_lon, max_lon, min_lat, max_lat order for a bounding box, instead of the classic x_min, y_min, x_max, y_max is a bit confusing, I find.

climada/hazard/base.py
```diff
@@ -760,6 +775,34 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
         haz.sanitize_event_ids()
         return haz
 
+    def select_tight_cent(self, buffer=0.0):
+        """
+        Reduce hazard to centroids to minimal box containing all non-zero
+        intensity points.
```
Evelyn-M, line 781: This is a bit cryptic. Maybe 
```suggestion
        Reduce hazard to those centroids spanning a minimal box which contains all non-zero
        intensity points.
```

Evelyn-M: Very general question: could the hazard be reduced to only those centroids that are non-zero? I.e. shape-agnostic. Since a "minimal bounding box" would still contain a ton of zero-intensity centroids, depending on the hazard path.

climada/hazard/base.py
```diff
@@ -760,6 +775,34 @@ def select(self, event_names=None, date=None, orig=None, reg_id=None, reset_freq
...
+
+        Parameters
+        ----------
+        buffer : float, optional
+            Buffer of box in degrees. The default is 0.0.
```
Evelyn-M, line 786: The buffer is in those units in which the coordinates are, right?

climada/hazard/base.py
chahank: I agree... but this is the code legacy we have.

climada/hazard/base.py
chahank: One could, but this is not necessarily a good idea. Because then an exposure point might be assigned a centroid with non-zero intensity, although it is closer to a (now removed) centroid with zero intensity.

chahank: Now that I think of it, this means the default buffer should be equal to the default 'assign_centroids' threshold.

climada/hazard/base.py
chahank: Great catch.

climada/hazard/base.py
chahank: Also, since github removed the comments on the suggestion commit:

- Reducing a hazard to only those centroids that are non-zero is dangerous because then exposures might be assigned centroids with non-zero intensity, even though a centroid (now removed) with zero intensity would have been closer.

- The default buffer has been made the same as the default max distance for `assign_centroids`

chahank:
I updated the code implementing the suggestions. I also added tests.

climada/hazard/base.py
chahank: done here 011f307

climada/hazard/base.py
chahank: Done.

climada/hazard/base.py
chahank: Any strong opinions about it?

tovogt: This always reminds me of the `extent` keyword in matplotlib's `imshow` function which also uses this unusual ordering. I never understood why they decided to do it like this, and why they stick to it.

The cleanest solution would be to require a data class (like `matplotlib.transforms.BboxBase` or `matplotlib.transforms.Bbox`) or at least a named tuple (like `rasterio.coords.BoundingBox`). For backwards compatibility, we could still support a tuple, but raise a deprecation warning, then convert the tuple while assuming the legacy order.

A user doesn't need to import that data class type to use this function, but we could alternatively support a `dict` that is automatically converted to the data class type, as in this example:
```
@dataclass
class BoundingBox:
    left: float
    bottom: float
    right: float
    top: float
```
Usage:
```
>>> extent = dict(left=10, right=20, bottom=0, top=100)
>>> bounds = BoundingBox(**extent)
>>> bounds
BoundingBox(left=10, bottom=0, right=20, top=100)
```
However, we don't need to come to a decision now. Just keep this in mind for future discussions. For now, I would suggest to stick to the confusing legacy tuple format (for backwards compatibility)...

climada/hazard/test/test_base.py
```diff
@@ -452,6 +452,73 @@ def test_select_reg_id_pass(self):
...
+        self.assertTrue(np.array_equal(sel_haz.intensity.toarray(),
+                                       haz.intensity[:,:-1].toarray()))
+        self.assertEqual(sel_haz.event_name, haz.event_name)
+        self.assertIsInstance(sel_haz, Hazard)
+        self.assertIsInstance(sel_haz.intensity, sparse.csr_matrix)
+        self.assertIsInstance(sel_haz.fraction, sparse.csr_matrix)
+
+        haz = dummy_hazard()
+
+        sel_haz = haz.select_tight(buffer=10)
```
tovogt, line 502:
```suggestion
        haz.intensity[:, -1] = 0.0
        
        # small buffer: zero field is discarded
        sel_haz = haz.select_tight(buffer=0.1)
        
        self.assertTrue(np.array_equal(sel_haz.centroids.coord.squeeze(),
                                       haz.centroids.coord[:-1, :]))

        # large buffer: zero field is retained
        sel_haz = haz.select_tight(buffer=10)
```

climada/hazard/base.py
chahank: Excellent points.

climada/hazard/test/test_base.py
chahank: Why not? done here 
2fec89c Thanks.

chahank:
Thanks @tovogt  and @Evelyn-M  for the review! In the last commit, I just updated this PR with the changes from PR #332 which changed the location of the coordinate Threshold.

tovogt (approved):

PR #334, opened 2021-12-16
Clearer install instructions for setting paths

tovogt:
As brought up by @aleeciu on Slack, the installation instructions currently have one FAQ part that points to a non-existing section of the document. This is because the part about how to set the path to `climada_python` using the command `conda develop` (which is buggy) has been replaced by the current version of the instructions that use (the less buggy) `pip install -e` instead.

Here I improve the wording of some of the installation instructions and fix the above-mentioned FAQ part.


doc/guide/Guide_Installation.ipynb
```diff
@@ -29,35 +29,39 @@
     "\n",
     "### Conda\n",
     "\n",
-    "It's highly recommanded to use [Conda](https://docs.conda.io/projects/conda/en/latest/) for the installation. Without Conda the installation of the dependencies can be cumbersome. If it is not already installed, download the latest version of [Anaconda](https://www.anaconda.com/) or [Miniconda](https://conda.io/miniconda.html) and execute it.\n",
+    "It's highly recommanded to install CLIMADA into a [Conda](https://docs.conda.io/projects/conda/en/latest/) environment. Without Conda, the installation of the dependencies can be cumbersome. If it is not already installed, download the latest version of [Anaconda](https://www.anaconda.com/) or [Miniconda](https://conda.io/miniconda.html) and execute it.\n",
```
tovogt, line 32: Here, I changed the wording because it seems like some people think this means "It's highly recommended to use Conda [instead of pip] for the installation". Following that, people seem to wonder about the use of `pip install -e` in the subsection below.

doc/guide/Guide_Installation.ipynb
```diff
@@ -29,35 +29,39 @@
...
     "```bash\n",
     "cd $[installation directory]\n",
     "conda env create -n climada_env -f climada_petals/requirements/env_climada.yml\n",
     "conda activate climada_env\n",
     "```\n",
     "\n",
-    "### Install climada core and petals\n",
+    "### Install CLIMADA's paths for core and petals\n",
+    "\n",
+    "Include the CLIMADA paths into the Conda environment as follows:\n",
```
tovogt, line 46: Since people often think that the word "install" means that some files are copied to some other location, I made the wording more clear. Otherwise, people might think that there are still paths to be set after installation.

doc/guide/Guide_Installation.ipynb
```diff
@@ -73,28 +77,33 @@
...
-    "\n",
+    "Start Anaconda, create a new python_env environment (just click create and enter the name, then press Create) and install latest Spyder in there (currently 4.2.5). Start this version of Spyder. After Spyder has started, navigate to *Preferences > Python Interpreter > Use the following interpreter* and paste the output of the following terminal command into the text box:\n",
     "```bash\n",
     "conda activate climada_env \n",
-    "conda install spyder-kernels \n",
     "python -c \"import sys; print(sys.executable)\" \n",
-    "--> returns a path, like /Users/XXX/opt/anaconda3/envs/climada_env/bin/python\n",
+    "# this returns a path, like /Users/XXX/opt/anaconda3/envs/climada_env/bin/python\n",
     "```\n",
+    "Now, obtain the version of the package `spyder-kernels` (currently 1.10.2) that has automatically been installed along with Spyder in the Anaconda environment and install the same version of `spyder-kernels` in the `climada_env` environment:\n",
```
tovogt, line 86: In a different thread on Slack we found that it's crucial to install a compatible version of `spyder-kernels` in your `climada_env`. If you don't specify a version, the latest develop version will be installed which is for Spyder 5 and not for the (stable) Spyder 4.

doc/guide/Guide_Installation.ipynb
```diff
@@ -73,28 +77,33 @@
...
+    "conda activate climada_env \n",
+    "conda install spyder-kernels=1.10.2\n",
+    "```\n",
+    "\n",
+    "Now, start a new IPython console in Spyder and run ``tests_install.py``.\n",
     "\n",
     "## FAQs\n",
     "\n",
-    "* ModuleNotFoundError; CLIMADA libraries are not found. Try to include *climada_python-x.y.z* path in the environment *climada_env* path as suggested in Section 2 of [Install environment with Anaconda]. If it does not work you can always include the path manually before executing your code:\n",
+    "* `ModuleNotFoundError`; the CLIMADA packages are not found. Try to include the paths to `climada_python` and `climada_petals` into the environment `climada_env` using `pip install -e` as suggested in [Install environment with Anaconda or Miniconda]. If it does not work, you can always include the path manually before executing your code:\n",
```
tovogt, line 97: Before this PR, this section used to refer to a now non-existent section of the installation instructions.

aleeciu:
looks good to me @tovogt, thanks. Actually some of the points raised stemmed from a misreading from myside of what in the previous version, but it now certainly reads better.

PR #337, opened 2022-01-04
Feature/bugfix exposures write raster

ChrisFairless:
`Exposures.write_raster` - wasn't updated when gdf was added to Exposures


chahank:
Looks good to me. 

The unitest seems not to have seen this error. So, bonus point would be to add one.

ChrisFairless:
I've got another PR coming up for this method, so I'll implement it there

PR #341, opened 2022-01-05
Improve `Exposures.write_raster` flexibility

ChrisFairless:
`Exposures.write_raster` calls `util.coordinates.points_to_raster` to write to a raster file. This method is able to write multiple bands to a file. This PR extends/fixes `Exposures.write_raster` for the same functionality without messing with backward compatibility.

This also updates the documentation and adds unit tests.


climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
             **metadata
         )
 
-    def write_raster(self, file_name, value_name='value', scheduler=None):
+    def write_raster(self, file_name, value_name=['value'], scheduler=None):
```
chahank, line 909:
```suggestion
    def write_raster(self, file_name, value_name='value', scheduler=None):
```

As agreed in the Developers' Guidelines, we should not use mutables (here lists) as defaults.

chahank, line 909:
```suggestion
    def write_raster(self, file_name, value_name='value', scheduler=None):
```

As agreed in the Developers' Guidelines, we should not use mutables (here a list) as defaults.

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
         """Write value data into raster file with GeoTiff format
 
         Parameters
         ----------
         file_name : str
-            name output file in tif format
+            output tif file to write to
```
chahank, line 915:
```suggestion
            name (and path) of the output file.
```

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
+        value_name : str or list of str
+            column(s) in the geodataframe containing data. If multiple values are provided, each column will be
+            written to a separate band in the output raster
```
chahank, line 918:
```suggestion
            written to a separate band in the output raster
            The default is 'value'.
```

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
+        scheduler : str, optional
+            used for dask map_partitions.
+            “threads”, “synchronous” or “processes”
```
chahank, line 921:
```suggestion
            “threads”, “synchronous” or “processes”
            The default is None.
```

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
-        if self.meta and self.meta['height'] * self.meta['width'] == len(self.gdf):
+        if not isinstance(value_name, list):
+            value_name = [value_name]
+
+        # Simple case: one raster band, every grid value defined in the Exposures.
+        # Then we can skip a slow points_to_raster call:
+        if self.meta and \
+                'height' in self.meta.keys() and \
+                len(value_name) == 1 and \
+                self.meta['height'] * self.meta['width'] == len(self.gdf):
```
chahank, line 931: This has a weird and hard-to-read indention.

chahank, line 931: Will this not cause an error if `meta['height']` or `meta['width']` are not defined? 

Also, the clause `'height' in self.meta.keys()` is probably redundant. OR, it should read:

```suggestion
        if self.meta and 'height' in self.meta and 'width' in self.meta:
                if len(value_name) == 1 and self.meta['height'] * self.meta['width'] == len(self.gdf):
```

Note that checking if something is in a dict by default looks at the keys.

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
-    def write_raster(self, file_name, value_name='value', scheduler=None):
+    def write_raster(self, file_name, value_name=['value'], scheduler=None):
         """Write value data into raster file with GeoTiff format
 
         Parameters
         ----------
         file_name : str
-            name output file in tif format
+            output tif file to write to
+        value_name : str or list of str
```
chahank, line 916: Should `value_name` potentially be replaced by a plural? E.g. `values`

climada/entity/exposures/base.py
```diff
@@ -906,15 +906,29 @@ def copy(self, deep=True):
...
+            column(s) in the geodataframe containing data. If multiple values are provided, each column will be
+            written to a separate band in the output raster
+        scheduler : str, optional
+            used for dask map_partitions.
+            “threads”, “synchronous” or “processes”
         """
-        if self.meta and self.meta['height'] * self.meta['width'] == len(self.gdf):
+        if not isinstance(value_name, list):
+            value_name = [value_name]
+
```
chahank, line 925: I would add the correct file ending by default:

`file_name = Path(file_name).with_suffix('.tiff')`

Although, this might be something to be added to the `write_raster` method rather than here.

climada/entity/exposures/test/test_base.py
```diff
@@ -264,6 +265,35 @@ def test_io_hdf5_pass(self):
...
+        test_file = Path(DATA_DIR, "test_write_raster.tif")
+        exp = Exposures.from_raster(HAZ_DEMO_FL, window=Window(10, 20, 50, 60))
+        exp.write_raster(test_file, 'value')
+        read_meta, read_data = u_coord.read_raster(test_file)
+        self.assertEqual(read_meta['transform'], exp.meta['transform'])
+        self.assertEqual(read_meta['width'], exp.meta['width'])
+        self.assertEqual(read_meta['height'], exp.meta['height'])
+        self.assertTrue(u_coord.equal_crs(read_meta['crs'], exp.meta['crs']))
+        self.assertEqual(read_data.shape, (1, exp.gdf.shape[0]))
+        np.testing.assert_array_equal(read_data[0], exp.gdf['value'])
```
chahank, line 279:
```suggestion
        self.assertEqual(read_data.shape, (1, exp.gdf.shape[0]))
        np.testing.assert_array_equal(read_data[0], exp.gdf['value'])
        test_file.unlink()
```
Delete the file after the test.

climada/entity/exposures/test/test_base.py
```diff
@@ -264,6 +265,35 @@ def test_io_hdf5_pass(self):
...
+
+    def test_write_raster_multiband_pass(self):
+        """Test write_raster function with two (identical) bands."""
```
chahank, line 282: Why two identical ones? Why not two different ones?

climada/entity/exposures/test/test_base.py
```diff
@@ -264,6 +265,35 @@ def test_io_hdf5_pass(self):
...
+        exp.gdf['value2'] = exp.gdf['value'] + 1
+        exp.write_raster(test_file, ['value', 'value2'])
+        read_meta, read_data = u_coord.read_raster(test_file, band=[1, 2])
+        self.assertEqual(read_meta['transform'], exp.meta['transform'])
+        self.assertEqual(read_meta['width'], exp.meta['width'])
+        self.assertEqual(read_meta['height'], exp.meta['height'])
+        self.assertTrue(u_coord.equal_crs(read_meta['crs'], exp.meta['crs']))
+        self.assertEqual(read_data.shape, (2, exp.gdf.shape[0]))
+        np.testing.assert_array_equal(read_data[0], exp.gdf['value'])
+        np.testing.assert_array_equal(read_data[1], exp.gdf['value2'])
```
chahank, line 294:
```suggestion
        np.testing.assert_array_equal(read_data[0], exp.gdf['value'])
        np.testing.assert_array_equal(read_data[1], exp.gdf['value2'])
        test_file.unlink()
```

chahank (changes requested):
A few comments. Overall its a nice streamlining!

climada/entity/exposures/base.py
ChrisFairless: That would break backward compatibility, not that anyone seems to use this. I'm happy to leave it as it is.

climada/entity/exposures/base.py
ChrisFairless: Yeah it would be better in `write_raster`.
But would anyone ever provide a file name without an extension? I can't imagine that.

Unless this is to enforce `.tiff` extensions over `.tif`, in which case, why?

climada/entity/exposures/base.py
ChrisFairless: Would you go for the nested `if`s over the awkward indentation?
Thanks for the tip about dictionary keys.

climada/entity/exposures/test/test_base.py
ChrisFairless: Oh lol, they're not actually identical in the implemented version.

ChrisFairless, line 282:
```suggestion
        """Test write_raster function with two bands."""
```

ChrisFairless:
Thanks @chahank, I've accepted most of these :)

climada/entity/exposures/base.py
chahank: I do. Also, it is a simple thing to do :).

climada/entity/exposures/base.py
chahank: Yes, it makes the code much clearer and safer.

chahank: Can you please still fix this indention?

Also, in line with the other pull request #343:

There is now a big repetition of the lines 
```        if self.meta and \
                'height' in self.meta.keys() and \
                len(value_name) == 1 and \
                self.meta['height'] * self.meta['width'] == len(self.gdf)
```
                
 This is ugly, and can lead to issues in the future. It would be better to
 a) fix the issue with inconsistent meta data at its roots
b) simply add a function to the exposures that makes this check. 

Thus, this can be removed from here, and be maintenained easily, and be read easily without the need for cryptic and quickly obsolete in-line comments.

chahank (changes requested):

climada/entity/exposures/base.py
ChrisFairless: yeah i am in favour of `Exposures.is_raster` and `Hazard.is_raster` methods

ChrisFairless: I think that's easier than reviewing/rewriting everything `meta`, for now.

If you're ok with that I _won't_ fix this indentation and will replace it with an `is_raster` check in another pull request.

chahank: I think this is good path forward. Could you also open an issue about the general handling of  `meta` ?

climada/entity/exposures/base.py
ChrisFairless: Implemented in 2553a0e8

climada/entity/exposures/base.py
ChrisFairless: Can do

ChrisFairless:
@chahank is it OK to merge this? I'd like to implement `is_raster` on a fresh branch

climada/entity/exposures/base.py
chahank: This should eventually be replaced by the `is_meta` method.

chahank (approved):
This pull request get's closed so that another one can be opened which will implement a `is_raster` method.

climada/entity/exposures/base.py
```diff
@@ -912,9 +912,25 @@ def write_raster(self, file_name, value_name='value', scheduler=None):
...
+        # Simple case: one raster band, every grid value defined in the Exposures.
+        # Then we can skip a slow points_to_raster call:
+        if self.meta and \
+                'height' in self.meta.keys() and \
+                len(value_name) == 1 and \
+                self.meta['height'] * self.meta['width'] == len(self.gdf):
```
ChrisFairless, line 933:
```suggestion
        if u_coord.grid_is_regular exp.gdf[['latitude', 'longitude']]:
```
@chahank is this alternate fix too dirty? It's not ugly

ChrisFairless: dammit, this doesn't fix the need for meta data. ignore this.

chahank: I am a bit confused. The point here is to fix the problem when the `latitude` and `longitude` do not correspond to the `meta` right? This here test only the `latitude` and `longitude`, not the `meta`.

emanuel-schmid:
I'm closing this, since the tests don't run through and there will be another PR for the `is_raster` method, supposed to include a fix for the problem addressed herein.

PR #343, opened 2022-01-06
Make stricter check when using  raster approach to assign centroids

ChrisFairless:
In `Exposures.assign_centroids`, the first thing the method tries involves rasters, since this is fast. It decides this is possible if the Hazard's centroids have a `meta` property.

However a `Centroids` object having a `meta` isn't equivalent to it being a complete raster. For example, `Centroids.union` creates a `Centroids` object with a its `meta` as `{}`, and calling `Hazard.lat_lon_to_meta` will create a `meta` for its centroids, regardless of whether the Centroids are defined at every grid point of the implied raster (side thought: is this correct behaviour?)

I've improved the check so that it makes sure that the number of centroids matches the number of grid cells defined by the meta.

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -390,7 +390,9 @@ class Exposures():
                     str(self.gdf.shape[0]), str(hazard.centroids.size))
         if not u_coord.equal_crs(self.crs, hazard.centroids.crs):
             raise ValueError('Set hazard and exposure to same CRS first!')
-        if hazard.centroids.meta:
+        if hazard.centroids.meta and \
+                hazard.centroids.meta['height'] and \
+                hazard.centroids.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
             assigned = u_coord.assign_grid_points(
                 self.gdf.longitude.values, self.gdf.latitude.values,
                 hazard.centroids.meta['width'], hazard.centroids.meta['height'],
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -391,7 +391,7 @@ class Exposures():
         if not u_coord.equal_crs(self.crs, hazard.centroids.crs):
             raise ValueError('Set hazard and exposure to same CRS first!')
         if hazard.centroids.meta and \
-                hazard.centroids.meta['height'] and \
+                'height' in hazard.centroids.meta and \
                 hazard.centroids.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
             assigned = u_coord.assign_grid_points(
                 self.gdf.longitude.values, self.gdf.latitude.values,
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -392,7 +392,7 @@ class Exposures():
             raise ValueError('Set hazard and exposure to same CRS first!')
         if hazard.centroids.meta and \
                 'height' in hazard.centroids.meta and \
-                hazard.centroids.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
+                hazard.centroids.lat.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
             assigned = u_coord.assign_grid_points(
                 self.gdf.longitude.values, self.gdf.latitude.values,
                 hazard.centroids.meta['width'], hazard.centroids.meta['height'],


climada/entity/exposures/base.py
```diff
@@ -390,7 +390,9 @@ def assign_centroids(self, hazard, method='NN', distance='haversine',
                     str(self.gdf.shape[0]), str(hazard.centroids.size))
         if not u_coord.equal_crs(self.crs, hazard.centroids.crs):
             raise ValueError('Set hazard and exposure to same CRS first!')
-        if hazard.centroids.meta:
+        if hazard.centroids.meta and \
+                hazard.centroids.meta['height'] and \
```
ChrisFairless, line 394:
```suggestion
                'height' in hazard.centroids.meta and \
```

climada/entity/exposures/base.py
```diff
@@ -390,7 +390,9 @@ def assign_centroids(self, hazard, method='NN', distance='haversine',
...
+                hazard.centroids.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
```
ChrisFairless, line 395:
```suggestion
                hazard.centroids.lat.size == hazard.centroids.meta['height'] * hazard.centroids.meta['width']:
```

ChrisFairless: omg apologies for having to edit a two-line pull request _twice_

chahank:
I think this requires a more in-depth check in the CLIMADA code-base to make the use of rasters cleaner. It is correct that post-manipulations of the raster data might lead to inconsistencies (such as the one you examplified above). I think we should not start patching down-stream methods such a `assign_centroids`. Instead, the methods that lead to inconsistent raster data must be addressed.

Also, your patch will lead to errors whenever the `meta = {}` as the call `hazard.centroids.meta['height']` will produce a `Keyerror`. Basically, this will break the code whenever `meta = {}`.

ChrisFairless:
I agree that we probably need a bit more standardisation of all things raster.

For now I wanted to add this check because the bug showed up live in my code (I wasn't just looking for another easy pull request).

The tests all pass.

ChrisFairless:
>Also, your patch will lead to errors whenever the meta = {} as the call hazard.centroids.meta['height'] will produce a Keyerror.

Is this true? I thought lazy evaluation would mean that it would never try to read `meta['height']` if the earlier check for its existence had failed.

chahank:
> Is this true? I thought lazy evaluation would mean that it would never try to read `meta['height']` if the earlier check for its existence had failed.

This is in principle true, yet it is a dangerous pattern imho.

chahank:
> I agree that we probably need a bit more standardisation of all things raster.
> 
> For now I wanted to add this check because the bug showed up live in my code (I wasn't just looking for another easy pull request).
> 
> The tests all pass.

But I do not think that this fixes your problem. The problem in your case was that you made an inconsistent use of the CLIMADA methods which lead to inconsistent data. Granted, the CLIMADA code did not prevent you from doing this. 

For the moment, I argue against this quick and dirty pull request that does not resolve the problem at hand which is created by a not attentive user. I think the problem should be solved indeed, but not like this. 

What do you think @tovogt @emanuel-schmid ?

ChrisFairless:
In that case, the correct fix for my use case would be to change `Centroids.union` so that it sets `centroids.meta = None` rather than `centroids.meta = {}`.

Do you know if that would break anything?

chahank:
I would avoid that quick fix again. This is not fixing the general problem that inconsistent raster and non-raster data can be contained in a CLIMADA object.

I agree that as an intermediate fix, the methods `is_raster` would be better.

ChrisFairless:
OK then, I'll close this without merging and put together something with `is_raster` methods.

Quick question: naming. What's better `is_raster` or `has_raster_metadata`?

chahank:
Tough one. 

`is_raster`: a bit unclear, since it could be a raster even if there is no meta attribute. But, the method could actually check for that too.

`has_raster_metadata`: a bit confusing, as this method would be checking wether the raster is compatible with the point data, and not checking the existence of raster metadata per se. 

So, I would opt for the former.

emanuel-schmid:
>  Quick question: naming. What's better is_raster or has_raster_metadata?

I also opt for the former. The latter looks like a question about the implementation rather then the data.

tovogt:
I said this in earlier discussions: In my opinion, the centroids object should always be a point centroids, there shouldn't be a meta attribute at all. I think, CLIMADA should completely get rid of the meta attribute everywhere. But CLIMADA should come with easy-to-use convenience functions that automatically detect that point centroids "look like" a raster. And for that, the name "is_raster" would be perfect, in my opinion.

chahank:
Thanks @tovogt for this summary of passed conversation! Very important.

ChrisFairless:
After talking with @chahank I'm inclined to agree. The current `meta` property is so quick to calculate you can just do it on the fly when you need it, which means it won't disagree with its parent object. And I think the change would make it clearer that CLIMADA's raster functionality is (primarily) for read/write purposes.

But worth talking more with the community about this. I know @mmyrte loves his CLIMADA rasters.

For now I'll just implement an `is_raster` method.

mmyrte:
Well, sure, I like the idea of using vectorised operations where possible, plus it fits _most_ of the work done with CLIMADA - but that doesn't mean that the simplicity you win by dropping any pretense of raster support is not worth it.

Or in other words: I trust you guys to do the right thing. Raster processing can be done outside CLIMADA if need be.

ChrisFairless:
I'm closing this PR without a merge to implement an `is_raster` method instead.

Thanks for the great discussion, all.

PR #345, opened 2022-01-10
from_ibtracs_netcdf: store selected agency for each track variable

tovogt:
This PR makes sure that the data source (provider/reporting agency) of each track and variable generated in `TCTracks.from_ibtracs_netcdf` is stored so that users can better understand where a particular time series came from. The information is stored in the `data_provider` attribute of each track.

Example: `ibtracs_mixed:lat(official_3h),lon(official_3h),wind(official_3h),pres(official_3h),rmw(usa),poci(usa),roci(usa)` means that latitude, longitude, max. wind and min. pressure have been taken from the officially reporting agency while the radius of maximum wind, the environmental pressure and the storm radius have been taken from the `usa` agency.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -506,7 +506,8 @@ class TCTracks():
         elif isinstance(provider, str):
             provider = [provider]
 
-        for tc_var in ['lat', 'lon', 'wind', 'pres', 'rmw', 'poci', 'roci']:
+        phys_vars = ['lat', 'lon', 'wind', 'pres', 'rmw', 'poci', 'roci']
+        for tc_var in phys_vars:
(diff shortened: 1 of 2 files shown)


simonameiler (approved):
Having the data source (provider/reporting agency) of each track and variable generated in TCTracks.from_ibtracs_netcdf stored in the data_provider attribute of each track is a very useful addition. It allows to distinguish between the two versions of the IBTrACS data: https://www.ncdc.noaa.gov/ibtracs/index.php?name=climatology, the IBTrACS - WMO and the IBTrACS - US (JTWC + NHC), which come with notable differences in basins outside the North Atlantic and Eastern Pacific.

Thank you very much, this is good to be merged.

PR #349, opened 2022-01-18
HDF5 file IO for TCTracks

tovogt:
For Hazard and Exposure objects, CLIMADA comes with handy methods to read and write to/from HDF5 files. For `TCTracks` objects, users have to either use pickle (whick has its own issues) or the `write_netcdf` functionality.

The problem with the NetCDF IO is that all tracks are stored in separate files. On many setups, especially with network file systems, the access times for many small files can be extremely bad.

That's why this PR adds IO methods `write_hdf` and `from_hdf` for `TCTracks`, in line with the corresponding functionality for Hazard and Exposure objects. All tracks are stored in a single hdf5 file.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -22,6 +22,7 @@ Define TCTracks: IBTracs reader and tracks manager.
 __all__ = ['CAT_NAMES', 'SAFFIR_SIM_CAT', 'TCTracks', 'set_category']
 
 # standard libraries
+import contextlib
 import datetime as dt
 import itertools
 import logging
@@ -1474,16 +1475,13 @@ def _xr_to_netcdf_multi(path, ds_dict, encoding=None):
         keyword parameter in `xr.Dataset.to_netcdf`. Default: None
     """
     path = str(pathlib.Path(path).expanduser().absolute())
-    store = NetCDF4DataStore.open(path, "w", "NETCDF4", None)
-    try:
+    with contextlib.closing(NetCDF4DataStore.open(path, "w", "NETCDF4", None)) as store:
         writer = ArrayWriter()
         for group, dataset in ds_dict.items():
             store._group = group
             unlimited_dims = dataset.encoding.get("unlimited_dims", None)
             encoding = None if encoding is None or group not in encoding else encoding[group]
             dump_to_store(dataset, store, writer, encoding=encoding, unlimited_dims=unlimited_dims)
-    finally:
-        store.close()
 
 def _xr_open_dataset_multi(path, prefix=""):
     """Read multiple xarray Datasets from groups contained in a single NetCDF4 file
@@ -1507,9 +1505,8 @@ def _xr_open_dataset_multi(path, prefix=""):
         Note that an empty string ("") is a valid group name and refers to the root group.
     """
     path = str(pathlib.Path(path).expanduser().absolute())
-    store = NetCDF4DataStore.open(path, "r", "NETCDF4", None)
     ds_dict = {}
-    try:
+    with contextlib.closing(NetCDF4DataStore.open(path, "r", "NETCDF4", None)) as store:
         groups = [g for g in _xr_nc4_groups_from_store(store) if g.startswith(prefix)]
         store_entrypoint = StoreBackendEntrypoint()
         LOGGER.info('Reading %d datasets from %s', len(groups), path)
@@ -1518,8 +1515,6 @@ def _xr_open_dataset_multi(path, prefix=""):
             ds = store_entrypoint.open_dataset(store)
             ds.load()
             ds_dict[group] = ds
-    finally:
-        store.close()
     return ds_dict
 
 def _xr_nc4_groups_from_store(store):
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -1264,7 +1264,7 @@ class TCTracks():
         tr.data = data
         return tr
 
-    def write_hdf5(self, file_name, complevel=4):
+    def write_hdf5(self, file_name, complevel=5):
         """Write TC tracks in NetCDF4-compliant HDF5 format.
 
         Parameters
@@ -1272,8 +1272,8 @@ class TCTracks():
         file_name: str or Path
             Path to a new HDF5 file. If it exists already, the file is overwritten.
         complevel : int
-            Specifies a compression level (0-9) for the gzip compression of the data.
-            A value of 0 or None disables compression. Default: 4
+            Specifies a compression level (0-9) for the zlib compression of the data.
+            A value of 0 or None disables compression. Default: 5
         """
         # initialise the data set with the size information
         xr.Dataset(attrs={'size': self.size}).to_netcdf(file_name, mode="w", format="NetCDF4")
@@ -1288,7 +1288,8 @@ class TCTracks():
 
             # change dtype from bool to int to be NetCDF4-compliant
             track.attrs['orig_event_flag'] = int(track.attrs['orig_event_flag'])
-            track.to_netcdf(file_name, mode="a", group=f"track{i}")
+            encoding = {var: dict(zlib=True, complevel=complevel) for var in track.data_vars}
+            track.to_netcdf(file_name, mode="a", group=f"track{i}", encoding=encoding)
             track.attrs['orig_event_flag'] = bool(track.attrs['orig_event_flag'])
 
         if last_perc != 100:
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -1263,6 +1263,51 @@ class TCTracks():
         tr.data = data
         return tr
 
+    def write_hdf(self, file_name, complevel=5):
+        """Write TC tracks in hdf5 format.
+
+        Parameters
+        ----------
+        file_name: str
+            File name to write, with h5 format
+        complevel : int
+            Specifies a compression level (0-9) for the data.
+            A value of 0 or None disables compression. Default: None
+        """
+        with pd.HDFStore(file_name, mode="w", complevel=complevel) as store:
+            attr_keys = sorted(set.union(*[set(d.attrs.keys()) for d in self.data]))
+            attrs = {key: [d.attrs[key] for d in self.data] for key in attr_keys}
+            store["attrs"] = pd.DataFrame(attrs)
+            for i, tr in enumerate(self.data):
+                store[f"data/track{i}"] = tr.to_dataframe()
+
+    @classmethod
+    def from_hdf(cls, file_name):
+        """Create new TCTracks object from hdf5 file
+
+        Parameters
+        ----------
+        file_name : str
+            File name of a file that has been generated with `TCTracks.write_hdf`.
+
+        Returns
+        -------
+        tracks : TCTracks
+            TCTracks with data from the given hdf5 file.
+        """
+        data = []
+        with pd.HDFStore(file_name, mode="r") as store:
+            df_attrs = store["attrs"]
+            for i, (_, attrs) in enumerate(df_attrs.iterrows()):
+                df = store[f"data/track{i}"]
+                tr = df.to_xarray().set_coords(["lat", "lon"])
+                tr['basin'] = tr['basin'].astype('<U2')
+                for key in attrs.index:
+                    tr.attrs[key] = attrs[key]
+                data.append(tr)
+        tracks = cls()
+        tracks.data = data
+        return tracks
 
     def to_geodataframe(self, as_points=False, split_lines_antimeridian=True):
         """Transform this TCTracks instance into a GeoDataFrame.
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -34,6 +34,7 @@ from pathlib import Path
 import cartopy.crs as ccrs
 import cftime
 import geopandas as gpd
+import h5py
 import matplotlib.cm as cm_mp
 from matplotlib.collections import LineCollection
 from matplotlib.colors import BoundaryNorm, ListedColormap
@@ -1263,7 +1264,7 @@ class TCTracks():
         tr.data = data
         return tr
 
-    def write_hdf(self, file_name, complevel=5):
+    def write_hdf(self, file_name, complevel=4):
         """Write TC tracks in HDF5 format.
 
         Parameters
@@ -1271,15 +1272,16 @@ class TCTracks():
         file_name: str or Path
             Path to a new HDF5 file. If it exists already, the file is overwritten.
         complevel : int
-            Specifies a compression level (0-9) for the data.
-            A value of 0 or None disables compression. Default: 5
+            Specifies a compression level (0-9) for the gzip compression of the data.
+            A value of 0 or None disables compression. Default: 4
         """
-        with pd.HDFStore(file_name, mode="w", complevel=complevel) as store:
-            attr_keys = sorted(set.union(*[set(d.attrs.keys()) for d in self.data]))
-            attrs = {key: [d.attrs[key] for d in self.data] for key in attr_keys}
-            store["attrs"] = pd.DataFrame(attrs)
+        with h5py.File(file_name, 'w') as store:
+            store.create_dataset('size', (), dtype=int, data=self.size)
             for i, tr in enumerate(self.data):
-                store[f"data/track{i}"] = tr.to_dataframe()
+                tr_bytes = tr.to_netcdf()
+                store.create_dataset(
+                    f'track{i}', (1,), dtype=f'|S{len(tr_bytes)}', data=[tr_bytes],
+                    compression='gzip', compression_opts=complevel)
 
     @classmethod
     def from_hdf(cls, file_name):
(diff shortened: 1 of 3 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -956,10 +958,16 @@ def from_simulations_storm(cls, path, years=None):
         # a bug in the data causes some storm tracks to be double-listed:
         tracks_df = tracks_df.drop_duplicates(subset=["year", "tc_num", "time_delta"])
 
-        # conversion of units and time
+        # conversion of units
         tracks_df['rmw'] *= (1 * ureg.kilometer).to(ureg.nautical_mile).magnitude
         tracks_df['wind'] *= (1 * ureg.meter / ureg.second).to(ureg.knot).magnitude
+
+        # convert from 10-minute to 1-minute sustained winds
+        tracks_df['wind'] /= 0.88
```
tovogt, line 966: This was a long-standing issue: All read functions in CLIMADA make sure that track wind speeds are 1-minute sustained winds, except this one. So, I fixed it.

chahank: Great! Could you just comment year why the number `0.88`, I do not get it right away.

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
                 basin = track.basin
                 del track.attrs['basin']
-                track['basin'] = ("time", np.full(track.time.size, basin))
+                track['basin'] = ("time", np.full(track.time.size, basin, dtype="<U2"))
             data.append(track)
         tr = cls()
         tr.data = data
         return tr
 
+    def write_hdf(self, file_name, complevel=4):
```
chahank, line 1267:
```suggestion
    def write_hdf5(self, file_name, complevel=4):
```
I think this is so far the most common naming of these methods in the CLIMADA code-base.

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+        """Write TC tracks in HDF5 format.
+
+        Parameters
+        ----------
+        file_name: str or Path
+            Path to a new HDF5 file. If it exists already, the file is overwritten.
```
chahank, line 1273: In the rest of CLIMADA it is usual to automatically overwrite. Do you think this is a good idea? This is not necessarily to make a change here right now, unless it is decided that overwriting should not be the default behavior.

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+        complevel : int
+            Specifies a compression level (0-9) for the gzip compression of the data.
+            A value of 0 or None disables compression. Default: 4
+        """
+        with h5py.File(file_name, 'w') as store:
+            store.create_dataset('size', (), dtype=int, data=self.size)
+
+            LOGGER.info('Writing %d tracks to %s', self.size, file_name)
```
chahank, line 1281:
```suggestion
```
Is this really necessary info?

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+            last_perc = 0
+            for i, tr in enumerate(self.data):
+                perc = 100 * i / self.size
+                if perc - last_perc >= 10:
+                    LOGGER.info("Progress: %d%%", perc)
+                    last_perc = perc
```
chahank, line 1287: Is the writing so long that this is needed?

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+
+                tr['basin'] = tr['basin'].astype('<U2')
```
chahank, line 1289: Do tracks only have basin as a non-data attribute? Would it not be safer to make a loop in case extra attributes are set (would of course only work for certain attribute types though)...

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+
+        Returns
+        -------
+        tracks : TCTracks
+            TCTracks with data from the given HDF5 file.
+        """
+        data = []
+        with h5py.File(file_name, 'r') as store:
+            size = store.get('size')[()]
+            LOGGER.info('Reading %d tracks from %s', size, file_name)
```
chahank, line 1315: Really needed?

climada/hazard/tc_tracks.py
```diff
@@ -1249,12 +1258,70 @@ def from_netcdf(cls, folder_name):
...
+            for i in range(size):
+                tr_ds = store.get(f'track{i}')
+                tr_bytes = tr_ds[0].ljust(int(str(tr_ds.dtype)[2:]), b'\x00')
```
chahank, line 1318: I do not understand what is going on in this line exactly, but I trust that it is doing the right operation.

climada/hazard/tc_tracks.py
```diff
@@ -1473,6 +1540,7 @@ def _read_one_gettelman(nc_data, i_track):
     tr_ds = xr.Dataset.from_dataframe(tr_df.set_index('time'))
     tr_ds.coords['lat'] = ('time', tr_ds.lat.values)
     tr_ds.coords['lon'] = ('time', tr_ds.lon.values)
+    tr_ds['basin'] = tr_ds['basin'].astype('<U2')
```
chahank, line 1543: Great addition.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -308,6 +308,26 @@ def test_read_legacy_netcdf(self):
...
+            self.assertEqual(set(tr.attrs.keys()), set(tr_read.attrs.keys()))
+            self.assertEqual(set(tr.variables), set(tr_read.variables))
+            self.assertEqual(set(tr.coords), set(tr_read.coords))
+            for key in tr.attrs.keys():
+                self.assertEqual(tr.attrs[key], tr_read.attrs[key])
+            for v in tr.variables:
+                self.assertEqual(tr[v].dtype, tr_read[v].dtype)
+                np.testing.assert_array_equal(tr[v].values, tr_read[v].values)
+            self.assertEqual(tr.sid, tr_read.sid)
+
```
chahank, line 330:
```suggestion
path.unlink()
```
Maybe remove the file after the test?

climada/hazard/test/test_tc_tracks.py
```diff
@@ -308,6 +308,26 @@ def test_read_legacy_netcdf(self):
             self.assertEqual(tr.basin.shape, tr.time.shape)
             np.testing.assert_array_equal(tr.basin, "SP")
 
+    def test_hdf_io(self):
+        """Test writting and reading hdf5 TCTracks instances"""
+        path = DATA_DIR.joinpath("tc_tracks.h5")
+        tc_track = tc.TCTracks.from_ibtracs_netcdf(
+            provider='usa', year_range=(1993, 1994), basin='EP', estimate_missing=True)
+        tc_track.write_hdf(path)
```
chahank, line 316: This is only testing for the ibtracs. But, seeing that the methods for gettelman and emanuel were also changed, can one do a test for these too? Or is there an issue with the available datasets?

chahank (changes requested):
Excellent addition!

Note for future: such a method would also be very useful for other climada objects, in particular the Impact and CostBenefit objects.

climada/hazard/tc_tracks.py
tovogt: As this is part of the `read_simulations_storm` function, I just took the value from the STORM paper Bloemendaal et al. 2020 that is cited in the function docstring. With my latest commit, this is also mentioned in the inline comment. It's also a very common factor for this kind of conversion, see the docstring to `IBTRACS_AGENCY_1MIN_WIND_FACTOR`.

climada/hazard/tc_tracks.py
tovogt: Well spotted, I somehow missed this! :+1:

climada/hazard/tc_tracks.py
tovogt: Yes, this is absolutely fine. It's the same for functions like `pandas.DataFrame.to_csv` or `xarray.Dataset.to_netcdf` or `pickle.dump`.

climada/hazard/tc_tracks.py
tovogt: By default, this log level is not printed to the console. For power users like me who manually decide to enable this log level, it's interesting to know when certain IO operations start and end and how long they take.

climada/hazard/tc_tracks.py
tovogt: See my comment above.

climada/hazard/tc_tracks.py
tovogt: No, this is a misunderstanding. This is just changing the dtype to make sure the data is compatible with NetCDF3. In my latest commit, I mention this in an inline comment.

climada/hazard/tc_tracks.py
tovogt: See my comment above.

climada/hazard/tc_tracks.py
tovogt: By default, numpy strips trailing \x00 in byte strings, so we need to undo that. In my latest commit, I mention this in an inline comment.

climada/hazard/test/test_tc_tracks.py
tovogt: Good idea!

climada/hazard/test/test_tc_tracks.py
tovogt: The change only affects the dtype of the `basin` variable. But yes, let's add unit tests for that.

tovogt:
Thanks for the quick feedback! I think I addressed all of your comments in my latest commit.

tovogt:
I have to mark this as WIP because I found that there are alternative implementations (data formats) and I'm not sure which one to choose.

My first proposal would encode each track separately as a NetCDF3 file and then store its (compressed) byte representation in the hdf5 file. This is not very elegant since it is not really using the hdf5 structure.

An alternative file format is in bf878c8 which produces NetCDF4-compliant HDF5-files that can also be inspected with NetCDF-tools like `ncdump`. However, the resulting hdf5 file is much (factor 10!) larger, even with zlib compression enabled. That's because it has huge amounts of metadata and metadata is not compressed. In the first format, the NetCDF headers (metadata) are compressed as well.

When gzipping the files in the second format, the size problem is more than solved. However, it's not possible to read and write directly to/from gzipped hdf5 files when using `xarray`. We would first need to write the uncompressed data to disk and then gzip it. Similarly, when reading the file, we would need to uncompress the data on disk (which requires a lot of space) and then read it from there.

tovogt:
@chahank I'm sorry, but I have to ask again for your review. I didn't change the API or the tests, but I changed the file format and the implementation. The produced HDF5-file is now completely NetCDF4-compliant and can be read with external NetCDF-tools like `ncdump`. This is much more elegant for most purposes, but requires factor 4-5 more space on disk then the data format you reviewed at first. For example, storing the appx. 3000 tracks from the global IBTrACS data set in 1980-2019 requires 87 MB on disk. This doesn't sound like a lot, but when generating 100 synthetic random walk tracks for each event, you will end up with almost 9 GB of data.

However, users that have problems with disk space can still manually gzip their files.

climada/hazard/tc_tracks.py
chahank: Thanks! Would it then make sense to pack it into a constant?

climada/hazard/tc_tracks.py
chahank: Sure, some functions do it this, some differently :). I personally do not like automatic overwrite, but if you prefer it this way I have no objections.

climada/hazard/tc_tracks.py
chahank: But then you should use 'debug'? Info gets printed by default in CLIMADA, at least when using it e.g. in a notebook. I think this is too verbose.

tovogt: No, the default log level is `WARNING`, see here:
https://github.com/CLIMADA-project/climada_python/blob/2111c496a77cd2bad2f603f8e28adcf9c7914124/climada/conf/climada.conf#L56

I think that `debug` should be reserved for developing purposes. So you only activate this level when you are working on the package's code. The `info` level is for users of the package who are interested to know what happens inside of the package.

climada/hazard/tc_tracks.py
tovogt: Yes, why not. Added in 9347bd30

climada/hazard/tc_tracks.py
```diff
@@ -1275,25 +1279,22 @@ def write_hdf5(self, file_name, complevel=5):
...
-
             # change dtype from bool to int to be NetCDF4-compliant
             track.attrs['orig_event_flag'] = int(track.attrs['orig_event_flag'])
-            encoding = {var: dict(zlib=True, complevel=complevel) for var in track.data_vars}
-            track.to_netcdf(file_name, mode="a", group=f"track{i}", encoding=encoding)
-            track.attrs['orig_event_flag'] = bool(track.attrs['orig_event_flag'])
-
-        if last_perc != 100:
-            LOGGER.info("Progress: 100%")
+        try:
```
chahank, line 1293: When using a try block, you probably should do something if there is a particular exception right? At least give a message that it failed or something.

climada/hazard/tc_tracks.py
```diff
@@ -1275,25 +1279,22 @@ def write_hdf5(self, file_name, complevel=5):
...
+                for i, track in enumerate(self.data)
+            }
+            ds_dict = {f'track{i}': track for i, track in enumerate(self.data)}
+            ds_dict[""] = xr.Dataset(attrs={'size': self.size})
+
+            LOGGER.info('Writing %d tracks to %s', self.size, file_name)
+            _xr_to_netcdf_multi(file_name, ds_dict, encoding=encoding)
+        finally:
+            for i, track in enumerate(self.data):
+                track.attrs['orig_event_flag'] = bool(track.attrs['orig_event_flag'])
```
chahank, line 1303: Why do you change the track object after saving to file?

climada/hazard/tc_tracks.py
```diff
@@ -1450,6 +1451,35 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+    ds_dict : dict whose keys are group names and values are xr.Dataset
+        Each xr.Dataset in the dict is stored in the group identified by its key in the dict.
+        Note that an empty string ("") is a valid group name and refers to the root group.
+    encoding : dict whose keys are group names and values are dict, optional
+        For each dataset/group, one dict that is compliant with the format of the `encoding`
+        keyword parameter in `xr.Dataset.to_netcdf`. Default: None
+    """
+    path = str(pathlib.Path(path).expanduser().absolute())
+    store = NetCDF4DataStore.open(path, "w", "NETCDF4", None)
+    try:
```
chahank, line 1478: You should handle the exception in some way at least.

climada/hazard/tc_tracks.py
```diff
@@ -1480,6 +1478,63 @@ def _xr_to_netcdf_multi(path, ds_dict, encoding=None):
...
+    Returns
+    -------
+    ds_dict : dict whose keys are group names and values are xr.Dataset
+        Each xr.Dataset in the dict is taken from the group identified by its key in the dict.
+        Note that an empty string ("") is a valid group name and refers to the root group.
+    """
+    path = str(pathlib.Path(path).expanduser().absolute())
+    store = NetCDF4DataStore.open(path, "r", "NETCDF4", None)
+    ds_dict = {}
+    try:
```
chahank, line 1512: Again, I think exceptions should be handled in some way.

chahank:
Thanks @tovogt . I went through the code, and made a few remarks. 

I think having a method that makes use of the files easier and compatible at the cost of memory is the right choice. And, as you said, for user where it is critical they may use extra compressing methods.

climada/hazard/tc_tracks.py
```diff
@@ -1310,12 +1309,11 @@ def from_hdf5(cls, file_name):
         tracks : TCTracks
             TCTracks with data from the given HDF5 file.
         """
-        size = xr.open_dataset(file_name).attrs['size']
-
-        LOGGER.info('Reading %d tracks from %s', size, file_name)
+        ds_dict = _xr_open_dataset_multi(file_name, prefix="track")
+        track_no = sorted(int(key[5:]) for key in ds_dict.keys())
```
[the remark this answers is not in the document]
tovogt: Because the order of tracks in the `TCTracks` object should be restored to the same order it had when saving.

climada/hazard/tc_tracks.py
```diff
@@ -1480,6 +1478,63 @@ def _xr_to_netcdf_multi(path, ds_dict, encoding=None):
     finally:
         store.close()
 
+def _xr_open_dataset_multi(path, prefix=""):
```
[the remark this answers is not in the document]
tovogt: I created a feature request for this here: pydata/xarray#6174 I don't think we currently need such a functionality as a util function in CLIMADA. But once we stumble upon another place in CLIMADA that might need this, feel free to move the function to the util submodule.

climada/hazard/tc_tracks.py
tovogt: No, when using io operations with a handler, it's best practice to have a `try ... finally` clause even if you don't know which kind of exception might happen. This applies to all programming languages that have this clause. In Python, this is usually obfuscated by the `with ...:` clause. There is even a built-in method for the special case of calling the `close` method of an object: https://docs.python.org/3/library/contextlib.html#contextlib.closing

I changed the code to use the contextlib functionality in 2facdaab

climada/hazard/tc_tracks.py
tovogt: This reverses the change that has been applied before saving to the file, see the comments in fb1655b8

climada/hazard/tc_tracks.py
tovogt: No, see my comment above.

climada/hazard/tc_tracks.py
tovogt: No, see my comment above.

climada/hazard/tc_tracks.py
chahank: Makes sense. I did not understand that the goal was to have a `close` method. I like the change to the context manager.

chahank:
Looks good to go to me, thanks for the updates!

Issue #348, opened 2022-01-17
Reviewing how `meta` works in CLIMADA


ChrisFairless:
We've been talking a lot about `meta` properties in CLIMADA Hazard and Centroids objects, how they have changed over time, and how it's now causing bugs. This issue brings it all together.

Previous conversations are in #340, #341 and #343.

### The issue

Historically the `meta` property of CLIMADA classes was used to store raster metadata. So if you wanted to check that an object had a defined raster transformation you could just check for the existence of its `meta` property.

This check shows all over the place in CLIMADA. But as the use of `meta` has expanded and changed it's no longer a reliable check:
- The `meta` property is sometimes set with no raster information. e.g. `Exposures.meta` sometimes stores CRS information and nothing else, and after appending or union-ing Centroids and Hazards the meta is set to an empty dictionary. There are `if` statements in CLIMADA that assume the existence of a `meta` is equivalent to full raster metadata, and the code then falls over when that metadata isn't available.
- Earlier CLIMADA functionality allowed you to describe your geographic data with _either_ `lat` and `lon` properties _or_ the `meta` property, with the idea that they were equivalent and interchangeable for raster operations. So there are methods that assume that if a `meta` is defined, it contains all the relevant information about an object. For example, `Centroids.equal` assumes two Centroids objects are identical if they have the same `meta` properties. Some methods, such as `Centroids.from_raster_file` read a raster file and _only_ set the `meta` property, not `lat` and `lon`. This isn't really how we use CLIMADA any more.
- A consequence of allowing two ways of describing geographic properties is that `meta`s can get out of sync with lat-lon data, so checks on a meta don't necessarily give accurate information about lat and lon coordinates, and vice versa.

In summary, **the `meta` can no longer be used to infer anything about a CLIMADA class**.

### Quick fixes failed

I spent some of last week attempting a quick fix, which was an `is_raster` method for the Centroids and Exposures classes. The idea was `if x.is_raster()` would replace existing `if x.meta` checks.

I ran into issues about what the 'source of truth' is for some of these checks, however, and decided that I needed to discuss things with you all.

For example,
- `is_raster` can be calculated without looking at an object's `meta`. So should the method care if the `meta` isn't defined? Or if it's defined but empty? Or if it's defined and disagrees with the lat-lon data?
- On the other hand, if a `meta` is defined but there's no lat-lon data, what should happen (and should this functionality be allowed any more?)
[...]

Issue #342, opened 2022-01-06
Can we have a one-line initialisation of `ImpactFuncSet`?


ChrisFairless:
When I create an impact function set from some impact function `impf` I always do it with
```
impfset = ImpactFuncSet()
impfset.append(impf)
```

In the spirit of not initialising empty classes, I'd like to either
- write an `ImpactFuncSet.from_impf` method which initialises from a list of impact functions
- modify `ImpactFuncSet.__init__` to take a list of impact functions on initialisation

Is this a good idea? Any preferences?

chahank:
Great point!

I would argue for making a good `__init__`. Ideally, all elements of the impact function set should be set at initialization.

ChrisFairless:
Yeah a good `__init__` would be better. But would that mess with CLIMADA's, uh, 'style', where our `__init__` methods never take parameters, and instead initialise with `from_*` methods?

chahank:
I agree, but we can try to improve in the future. Some classes already have it. 

I would propose to do the `__init__` as it does not hurt.

@emanuel-schmid : do you have any feedback here?

peanutfun:
One-line initialization of `ImpactFuncSet` was implemented in #568

PR #350, opened 2022-01-19
WIP: Feature/set imp matrix

chahank:
This is a first step in an attempt to attack the problem of methods that modify CLIMADA objects in place which can lead to inconsistencies (such as e.g. discussed for the `meta` parameter in #348 ). This would also be a step towards better `__init__` methods as discussed e.g. in #342.

Here we add methods that compute the `impact` attributes `at_event`, `eai_exp` and `aai_agg` directly from the `imp_mat` and the `frequency` (no testing).

These methods are so far not embedded in the rest of the method from the class `impact`, but can in a future iteration if this pull request is deemed useful. Furthermore, these methods come in handy for instance in the handling of line and polygons exposures c.f. https://github.com/CLIMADA-project/climada_python/tree/feature/lines_polygons_exp or multi impact yearsets c.f. https://github.com/CLIMADA-project/climada_python/tree/feature/sets .

Feedback on the proposed methods would be appreciated.


Evelyn-M:
Small detail, the references issue above, @chahank, does not match the  `__init__` methods discussion you refer to. Can you find the right hashtag?

climada/engine/impact.py
```diff
@@ -1111,6 +1111,76 @@ def _cen_return_imp(imp, freq, imp_th, return_periods):
...
+        Parameters
+        ----------
+        imp_mat : sparse.csr_matrix
+            matrix num_events x num_exp with impacts.
+        Returns
+        -------
+        at_event : np.array
+            impact for each hazard event
+        """
+        return np.squeeze(np.asarray(np.sum(imp_mat, axis=1)))
```
Evelyn-M, line 1166: So currently the `at_event` is calculated from an `impact` variable (line 1033) which considers potential deductibles, covers etc. The imp_mat does not previously account for those, right? Aka the at_event would have to be modified afterwards, if such conditions exist?

Evelyn-M: (similary for eai_exp,  etc.)

Evelyn-M:
I find this way of defning the attributes very neat and clear, in general.

ChrisFairless:
I missed some of these conversations: what risk are we mitigating here? Are there parts of an Impact object that get out of sync? That is, is there a risk that `Impact.imp_mat` is overwritten after `eai_exp`, `at_event` or `aai_agg` are set?

If that's the case we should consider other options:
- Don't let that happen: any method that would modify `imp_mat` returns a fresh Impact object
- Make properties `eai_exp`, `at_event` or `aai_agg` methods as well, meaning they're recalculated whenever they're accessed

Skimming the code for the class, I think the only method that breaks this rule is `Impact._exp_impact`, which is hidden and only used in `Impact.calc` anyway, so the risk is low. Has it become an issue elsewhere?

climada/engine/impact.py
chahank: Yes, this is then to be added in this method, good point! My issue with the current implementation is that the `at_event` and `eai_exp` are computed in a very obscure manner inside of the `.calc` method.

chahank:
>I missed some of these conversations: what risk are we mitigating here? Are there parts of an Impact object that get out of sync? That is, is there a risk that Impact.imp_mat is overwritten after eai_exp, at_event or aai_agg are set?

Currently, the `Impact` objects are initialized with an empty constructors, which is then filled in a intransparent manner by the method `.calc`. This is sub-optimal, an prevents the use of `Impact` objects in other parts of the code, such as the `lines_polygons_handler` or the `yearsets`.

chahank:
>If that's the case we should consider other options:

    Don't let that happen: any method that would modify imp_mat returns a fresh Impact object
    Make properties eai_exp, at_event or aai_agg methods as well, meaning they're recalculated whenever they're accessed

Yes and no. The problem is not with the modification of the `imp_mat` only, but with the whole class. Instead of trying to apply a restriction on future methods handling `Impact`, it would be cleaner to make the `__init__` or other initialization methods more consistent.

ChrisFairless:
It's nice these are all static methods as well - it means if you're doing something really janky with impacts you can do it outside of the class

chahank:
> These make sense then. Is the idea that Impact.__init__ would basically be Impact.calc?
Not sure yet. Probably not. I would like a generic `__init__` which can then be used in `Impact.calc` which would become a class method probably.

climada/engine/impact.py
chahank: Best would probably be to have a class method `calc_impact` and `calc_residual_impact` (if with insurance stuff).

chahank: If I am not mistaken, the deductible all act on the impact matrix as implemented right now? Thus, the `imp_mat` is actually not the risk matrix, but the residual risk matrix after insurance deductibles have been applied. 

My suggestion would thus be to disentangle that as it is unclear at the moment.

chahank:
Thanks for the discussion! The main points were included in a larger refactoring of the `Impact` class. The work is currently on the branch [feature/refactor_impact_calc](https://github.com/CLIMADA-project/climada_python/tree/feature/refactor_impact_calc). Please feel free to come to help with the refactoring efforts.

chahank:
This pull request was addressed in PR #436

PR #351, opened 2022-01-20
WIP: Feature/lines polygons exp

chahank:
This pull request is to discuss the updates of the util functions to handle lines and polygons. This is a follow-up of #221.

It is now possible to compute directly the impact for exposures with polygons and/or lines geometries. There are methods to disaggregate polygons/lines to points, with wrappers for geodataframes and exposures. There are methods to aggregate the impact per point to geometries. 

The final outcome is an `Impact` object, which has one new attribute containing the geometries of the original exposure. This allows plotting the impact per geometry with a new method.


climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+    """
+
+    #discretize exposure
+    exp_pnt = exp_geom_to_pnt(
+        exp=exp, res=res,
+        to_meters=to_meters, disagg=disagg
+        )
+    exp_pnt.assign_centroids(haz)
+
+    return calc_impact_pnt_agg(exp_pnt=exp_pnt, impf_set=impf_set, haz=haz, agg=agg)
```
Evelyn-M, line 95: I'd consider refactoring a bit differently. ATM, there are three functions that have as output the aggregated impact per geometry [calc_geom_impact --> calc_impact_pnt_agg --> impact_pnt_agg]. Especially the first two funcs basically only differ in signature in that the first contains the original exposure and the second the disaggregated exposure, so I'd suggest merging the second into the first

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+        The set of impact functions.
+    haz : Hazard
+        The hazard instance.
+    res : float
+        Resolution of the disaggregation grid.
+    to_meters : bool, optional
+       If True, the geometries are projected to an equal area projection before
+       the disaggregation. res is then in meters. The exposures are
+       then reprojected into the original projections before the impact
+       calculation. The default is False.
```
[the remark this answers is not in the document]
chahank: Yes, exactly. PLease feel free to modify and make clearer!

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+        mask = np.concatenate([np.ones(l) / l for l in geom_sizes])
+    elif agg == 'sum':
+        mask = np.ones(len(col_geom))
+    else:
+        raise ValueError("Please choose a valid aggregation method. agg=%s is not valid", agg)
+    csr_mask = sp.sparse.csr_matrix(
+        (mask, (row_pnt, col_geom)),
+         shape=(len(row_pnt), len(np.unique(col_geom)))
+        )
+    return imp_pnt.imp_mat.dot(csr_mask)
```
Evelyn-M, line 211: Did you implement this for efficiency reasons like that? 
One could also do a one-liner using a multi-indexed pandas-df for the imp_mat and doing a df.groupby(outer level).sum / df.groupby(outer level).avg, instead. The advantage of that being readability. But perhaps it gets slow for _very_ large matrices.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+
+def calc_geom_impact(
+        exp, impf_set, haz, res,
+        to_meters=False, disagg=None, agg='sum'
+        ):
+    """
+    Compute impact for exposure with (multi-)polygons and/or (multi-)lines.
+    Lat/Lon values in exp.gdf are ignored, only exp.gdf.geometry is considered.
+
+    The geometries are first disaggrated to a grid with resolution res*res.
```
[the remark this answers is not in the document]
chahank: Yes. For polygons, it is a raster, for lines a line. So one is with resolution res**2, the other with res.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+    disagg : string, optional
+        if 'avg' average value over points
+        if 'surf' value per points is equal to area per point (res*res)
+            for polygons, and distance per point (res) for lines.
+        if 'None' value is unchanged (or 1 if no value is defined)
+        Disaggregation method. The default is None.
+    agg : string, optional
+        If 'agg', the impact is averaged over all points in each geometry.
+        If 'sum', the impact is summed over all points in each geometry.
+        The default is 'sum'.
```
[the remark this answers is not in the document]
chahank: Actually, I just saw that it should be `'avg'` and not `'agg'`. Makes sense no?

climada/util/lines_polys_handler.py
chahank: Could be an idea. I decided to do more methods to account for the different signatures. Thus, a user can choose to implement its own disaggregation method, and still use the calc impact. That was the idea to have first and second. What do you think?

climada/util/lines_polys_handler.py
chahank: Yes. Whenever one is dealing with sparse matrices, one should never cast them into dense ones unless absolutely necessary. Also, linear algebra operations are very efficient. The code here is rather optimized and at least my first implementation along the lines you suggest was 3-4 orders of magnitude (1000 - 10000x) slower.

climada/util/lines_polys_handler.py
chahank, line 47:
```suggestion
    The geometries are first disaggregated to points. Polygons: grid with resolution res*res. Lines: points with length res.
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,886 @@
...
+    ----------
+    exp : Exposures
+        The exposure instance with exp.gdf.geometry containing (multi-)polygons
+        and/or (multi-)lines
+    impf_set : ImpactFuncSet
+        The set of impact functions.
+    haz : Hazard
+        The hazard instance.
+    res : float
+        Resolution of the disaggregation grid.
```
chahank, line 60:
```suggestion
        Resolution of the disaggregation grid (polygon) or line (lines).
```

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
chahank: Is the line `res is then in meters` not enough?

climada/util/lines_polys_handler.py
chahank: Basically, if you can use a mask and a matrix multiplication instead of indexing/loops, always do so.

chahank:
Commit fbe841f fixes a bug when the index of the original geometries is not integer. Now it works, but it should be more thoroughly tested.

chahank:
Commit 2f172a9  updates the tutorial with the latest API. It is not complete but explains the basics. An example of an exposure with both polygons and lines could be interesting to add for example.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
Evelyn-M: I think I skipped it. I'll make a proposition to have it already mentioned in `res`.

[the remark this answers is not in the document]
Evelyn-M: Right. didnt see it

climada/util/lines_polys_handler.py
Evelyn-M: Right, that's a valid point. On the other hand, one question would be how realistic it is that someone does a non-regular disaggregation of the exposure that is not offered within  the `exp_geom_to_pnt` function, while still providing a `exp_pnt` in the next step that fulfills the structural requirements of that variable (e.g. a multiindex, a geometry column, an original geometry column, ..).
The advantage of having one wrapper function that does it all, is that it shows the three crucial steps immediately and comprehensively, i.e. exposure-disaggregation, classic impact calc, impact re-aggregation.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+    sparse.csr_matrix
+        matrix num_events x num_polygons with impacts.
+
+    """
+    # aggregate impact
+    mi = gdf_pnt.index
+    col_geom = mi.get_level_values(level=0)
+    def convert_str_idx_to_int(idx):
+        """Converts string multi-index level 0 to integer index"""
+        return np.sort(np.unique(idx, return_inverse=True)[1])
```
Evelyn-M, line 201: Why is it necessary to define this in here? Or just to "hide it away" because it's a util func of a util func?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+        mask = np.ones(len(col_geom))
+    else:
+        raise ValueError("Please choose a valid aggregation method. agg=%s is not valid", agg)
+    csr_mask = sp.sparse.csr_matrix(
+        (mask, (row_pnt, col_geom)),
+         shape=(len(row_pnt), len(np.unique(col_geom)))
+        )
+    return imp_pnt.imp_mat.dot(csr_mask)
+
+def plot_eai_exp_geom(imp_geom, centered=False, figsize=(9, 13), **kwargs):
```
Evelyn-M, line 218: can we extend this to a general `plot_impact_geom` function where we have as a kwarg the different usual impact variable options `eai`, `at_event`, `aai`, etc.? 
Especially for a single event calculation, I don't like the frequent "misuse" of eai (being the expected _annual_ impact) to plot the event impact.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+
+def _gdf_line_to_pnt(gdf, res, to_meters, disagg):
+    """
+    Disaggregate exposures with (multi-)lines
+    geometries to points.
+
+    Parameters
+    ----------
+    exp : Exposures
```
Evelyn-M, line 323:
```suggestion
    gdf : gpd.GeoDataFrame
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+        The exposure instance with exp.gdf.geometry containing (multi-)polygons
```
Evelyn-M, line 324:
```suggestion
        Geodataframe of an exposure instance with gdf.geometry containing (multi-)polygons
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+def _gdf_poly_to_pnt(gdf, res, to_meters, disagg):
+    """
+    Disaggregate exposures with (multi-)polygons
+    geometries to points.
+
+    Parameters
+    ----------
+    exp : Exposures
+        The exposure instance with exp.gdf.geometry containing (multi-)polygons
```
Evelyn-M, line 371: gdf

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+        gdf_pnt = disagg_gdf_avg(gdf_pnt)
+    elif disagg == 'surf':
+        gdf_pnt = disagg_gdf_val(gdf_pnt, res * res)
+    elif disagg is None and 'value' not in gdf_pnt.columns:
+        gdf_pnt['value'] = 1
+
+    return gdf_pnt
+
+
+def disagg_gdf_avg(gdf_pnts):
```
Evelyn-M, line 411: I'd suggest a slightly different naming, since the gdf was already disaggregated in its geometry ("rasterized" as you call it) at that point.
E.g. `disagg_vals_avg`

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+    group = gdf_pnts.groupby(axis=0, level=0)
+    gdf = group.value.mean() / group.value.count()
+
+    gdf = gdf.reindex(gdf_pnts.index, level=0)
+    gdf_agg['value'] = gdf
+
+    return gdf_agg
+
+
+def disagg_gdf_val(gdf_pnts, value_per_pnt):
```
Evelyn-M, line 440: dito.
E.g. `disagg_vals_cnst`

climada/util/lines_polys_handler.py
chahank: I found this a tad bit more clear to read... but yes, one could also do with a comment.

climada/util/lines_polys_handler.py
chahank: I guess so yes, good idea!

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+    gdf_points = _swap_geom_cols(
+        gdf_points, geom_to='geometry_orig', new_geom='geometry_pnt'
+        )
+    gdf_points = gdf_points.explode()
+    gdf_points.index = gdf_points.index.set_levels(idx, level=0)
+    return gdf_points
+
+def _interp_one_poly(poly, res):
+    """
+    Disaggragate a single polygon to points
```
Evelyn-M, line 535:
```suggestion
    Disaggregate a single polygon to points
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+    height, width, trafo = u_coord.pts_to_raster_meta(poly.bounds, (res, res))
+    x_grid, y_grid = u_coord.raster_to_meshgrid(trafo, width, height)
+    in_geom = sh.vectorized.contains(poly, x_grid, y_grid)
+
+    if sum(in_geom.flatten()) > 1:
+        return shgeom.MultiPoint([(x, y) for x, y in
+                            zip(x_grid[in_geom], y_grid[in_geom])])
+    else:
+        return shgeom.MultiPoint([poly.representative_point()])
```
Evelyn-M, line 562: should this return also a warning/info? because it assigns a point even though it's not representative of the usual expectable area per point. Esp. for value assignment / disaggregation later with a fixed value per point (assuming a certain underlying area), this could lead to distortions if this happens very often

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+    """
+
+    if poly.is_empty:
+        return shgeom.MultiPoint([])
+
+    repr_pnt = poly.representative_point()
+    lon_0, lat_0 = repr_pnt.x, repr_pnt.y
+
+    project = pyproj.Transformer.from_proj(
```
Evelyn-M, line 590: short comment to this trafo here

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+    if sum(in_geom.flatten()) > 1:
+        return shgeom.MultiPoint([(x, y) for x, y in
+                            zip(x_grid[in_geom], y_grid[in_geom])])
+    else:
+        return shgeom.MultiPoint([poly.representative_point()])
+
+def _interp_one_poly_m(poly, res, orig_crs):
+    """
+    Disaggragate a single polygon to points. Resolution in meters.
```
Evelyn-M, line 566: Explanation to how the crs needs to be adjusted to a geographically suitable one in metres.

climada/util/lines_polys_handler.py
chahank: But what is the downside of having an extra method? There is a method that does all in one already...

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+    return impact_agg
+
+def aggregate_impact_mat(imp_pnt, gdf_pnt, agg):
+    """
+    Aggregate impact matrix given geodataframe or disaggregated polygons.
+
+    Parameters
+    ----------
+    impact_pnt : Impact
```
chahank, line 181:
```suggestion
    imp_pnt : Impact
```

climada/util/lines_polys_handler.py
chahank:
```suggestion
    gdf : GeoDataFrame
```

climada/util/lines_polys_handler.py
chahank, line 371:
```suggestion
    gdf : GeoDataFrame
        The geodataframe instance with gdf.geometry containing (multi-)polygons
```

climada/util/lines_polys_handler.py
chahank: Or 'disagg_gdf_values_avg`?

climada/util/lines_polys_handler.py
chahank: I do not like `cnst` so much as it is not clear what `constant` means imho.

climada/util/lines_polys_handler.py
chahank: Maybe... it would be hard though to interpret for the user right? The open method has a warning in the docstring about that.

climada/util/lines_polys_handler.py
chahank: This is a private method, so I kept the comments to a minimum. But it could help future developers. What do you mean though? Here any crs can be used... The method does the projections itself.

climada/util/lines_polys_handler.py
chahank: Ohhh actually, I suppose the code should be using `poly_to_equalarea_pro` here instead. Could you adjust this?

climada/util/lines_polys_handler.py
Evelyn-M: Readability. At the moment, the overall function `calc_geom_impact` returns `calc_impact_pnt_agg` returns `impact_pnt_agg`. So one needs to scroll through more functions to get the main line of what the overall wrapper does.

climada/util/lines_polys_handler.py
Evelyn-M: also good.

climada/util/lines_polys_handler.py
Evelyn-M: `disagg_gdf_values_fixed`?

climada/util/lines_polys_handler.py
Evelyn-M: mmmh, how / where? didn't find it.

climada/util/lines_polys_handler.py
Evelyn-M: Yes yes. I mean the fact that the crs is projected in the code. Would be useful to other developers that are not geographers.

climada/util/lines_polys_handler.py
Evelyn-M: That pre-answers my next comment :D

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+    See also
+    --------
+    * util.coordinates.compute_geodesic_lengths()
+    """
+
+    # Needed because gdf.explode() requires numeric index
+    idx = gdf_lines.index.to_list() #To restore the naming of the index
+
+    gdf_points = gdf_lines.copy().reset_index(drop=True)
+    line_lengths = gdf_lines.length
```
Evelyn-M, line 679: Does a gdf have a defined `.length` attribute by default?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,891 @@
...
+
+def line_to_pnts_m(gdf_lines, res):
+
+    """ Convert a GeoDataframe with LineString geometries to
+    Point geometries, where Points are placed at a specified distance
+    (in meters) along the original LineString. Each line is reduced to
+    at least two points.
+
+
+
```
Evelyn-M, line 709:
```suggestion
```suggestion



```
```

Comment (Evelyn-M) [inline climada/util/lines_polys_handler.py:709]:
```suggestion
```suggestion



```
```

File: climada/util/lines_polys_handler.py [hunk 21]
```diff
@@ -0,0 +1,891 @@
...
+        new_geom is renamed to 'geometry'
+
+    Returns
+    -------
+    gdf_swap : GeoDataFrame
+        Copy of gdf with the new geometry column
+
+    """
+    gdf_swap = gdf.rename(columns = {'geometry': geom_to})
+    gdf_swap.rename(columns = {new_geom: 'geometry'}, inplace=True)
```
Evelyn-M, line 802:
```suggestion
    gdf_swap = gdf.rename(columns = {'geometry': geom_to,
    new_geom: 'geometry'}, inplace=True})
```

chahank: Not sure, have you tried the tests? Should work I think.

climada/util/lines_polys_handler.py
chahank: Yes. 

> In the case of a (Multi)Polygon it measures the length of its exterior (i.e. perimeter).

For Points it is 0.

climada/util/lines_polys_handler.py
chahank: Readability for developers? This might be true. But I would say the use of the methods is there. So, maybe choosing clearer names would be a solution?

climada/util/lines_polys_handler.py
chahank: It looks like it got lost over the iterations. You are right, one should add it in the docstring of the open method.

climada/util/lines_polys_handler.py
chahank: The cleanest would probably be to make a sub-class `Impact_geom` or similar with the extra methods and attributes.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,832 @@
...
+        shgeom.MultiPoint([
+            line.interpolate(dist, normalized=True)
+            for dist in fractions
+            ])
+        for line, fractions in zip(gdf_points.geometry, line_fractions)
+        ]
+
+    gdf_points = _swap_geom_cols(
+        gdf_points, geom_to='geometry_orig', new_geom='geometry_pnt'
+        )
```
Evelyn-M, line 637: same question goes for the `poly_to_points`: is it necessary to store the polygon / line geometry  in the interpolated gdfs? I think this blows up the dataframe size quite a bit, but we don't really need it?

chahank: In my opinion, I think the pros out way the cons. 1) Can you make some tests to see whether the size really increases? Because in general, the original geometries will be in a much smaller numbers than the points. Thus, I suspect that the extra memory needed would be only marginal (multi-index dataframes store their data intelligently as far as I know). 2) Not having both geometries in the same object means that external book-keeping must be made. This is in general very cumbersome and prone to errors. The only other solution I see would be to introduce another column with unique id numbers in order to be sure to know what point corresponds to what geometry. But I think this does not significantly reduce the memory requirements, and is vulnerable to errors. 3) In any case a user must keep both geometries somewhere. Thus, keeping it in the same, multi-index and efficient dataframe doing the book-keeping seems a good idea.

What do you think?

Evelyn-M: Removed this method and changed the syntax a bit in the main impact_calc function, so the original geometries don't need to be carried over.

chahank: What do you mean? It is always necessary to keep the original geometries...

chahank: The point is: I think the dataframe should have both the original geometries and the point geometries, otherwise the risk for inconsistent data is very high as external book-keeping is required.

Evelyn-M: ah, didn't see your earlier comment. Will check the memory part; the bookkeeping is indeed cumbersome unless using the top-level all-in-one function.

Evelyn-M: Re-added it, since memory seems to not make a huuuuge difference.

chahank:
Most of the tests still need to be written. But, if @Evelyn-M agrees, it is time for an in-depth review. The final tests will be supplied as soon as all the code remarks are worked-in.

emanuel-schmid:
In the tutorial, second cell, the function `gdf_poly` sets the province names manually, like this:
```
    # extract the NL polygons
    prov_names = {'Netherlands': ['Groningen', 'Drenthe',
                                  'Overijssel', 'Gelderland',
                                                    'Limburg', 'Zeeland',
                                                    'Noord-Brabant', 'Zuid-Holland',
                                                    'Noord-Holland', 'Friesland',
                                                    'Flevoland', 'Utrecht']
```

Can't they be read from natural earth?

If not - they must be set into a different order, since they are wrong, in the sense of: in the returned data frame the index doesn't match the polygons.

emanuel-schmid:
> If not - they must be set into a different order, since they are wrong, in the sense of: in the returned data frame the index doesn't match the polygons.

The wrong order was fixed with commit 6ca9f46.

manniepmkam:
I have read the tutorial, very well-written and easy to follow the workflow. One minor thing that could be change is the colour scheme of the polygon exposure that can be changed (cell 4). Currently it shares the same colour scheme as the impact plots (cell 7 and 9), which gives me a slight confusion when I first read the tutorial.

image
image

Evelyn-M:
The one thing that still gives me a bit of a headache is the `disagg` kwarg in the methods and its mix between actually disaggregating and newly assigning values to the exposure.
Also, the `disagg`=`area` is very hard to understand for a user.
Also, in some instances i find that i'd like to assign an `area`*m(2)-value as value, and not just the length / area; this is currently not possible.
So thinking how this could be solved

chahank:
That sounds like a good idea!

How would you further do the "area" assignment in this logic? Maybe allow for the combination:

`dissag_met = 'fix'` and `dissag_val = 'area'` ?

Or another strategy completely?

Evelyn-M:
> That sounds like a good idea!
> 
> How would you further do the "area" assignment in this logic? Maybe allow for the combination:
> 
> `dissag_met = 'fix'` and `dissag_val = 'area'` ?
> 
> Or another strategy completely?

Either that, or dropping the area option altogether. I think this wouldn't be soo tragic either, since the user manually has to decide for a `res` value, anyways, and hence can just enter the numeric res or res^2 value as `dissag_val` once more, if they want to assign that.

chahank:
The only edge case when this would not work is when the exposures has both lines and polygons. Because then lines should have the value `res` and the polygons `res**2`.

chahank:
I made the methods with two arguments. The only issue is with the example explained above: if an exposures has both lines and polygons, it is now not possible to directly get lengths and areas. But this is fine imho.

Evelyn-M:
> I made the methods with two arguments. The only issue is with the example explained above: if an exposures has both lines and polygons, it is now not possible to directly get lengths and areas. But this is fine imho.

Yep, looks good! I added a few details and updated the tutorial accordingly.
Think we can then write the unittests as a last step?

chahank:
Yes, it is time I think to write unit tests! Maybe we should also build in a few checks in the code (e.g. to warn the user about the issue with lines/polygons length/area, or to warn about incorrect combinations of disaggregation methods/values).

emanuel-schmid (commented):
Looks quite good to me in general. All methods are nicely nicely disentangled. 🎉

Just one comment about the tutorial (I haven't gone through the whole of it yet):\
is the cell starting with "in this tutorial..." and ending with "... brief overview" possibly misplaced? There is neither an obvious context to _load data_ nor a brief overview following.

climada/util/test/test_coordinates.py
```diff
@@ -27,18 +27,149 @@
...
+                     46, 46, 46, 46, 46, 46, 46, 46, 46, 46, 46,
+                     46, 46, 46, 45, 45, 45, 45, 45, 45, 45, 45,
+                     45, 45, 45, 45, 45, 45])
+
+def def_ref_50():
+    """Default output reference for maximum distance threshold 50km"""
+    return np.array([46, 46, 36, -1, 36, 46, 46, 46, 46, 46, 46, 36, 46, 46,
+                     36, 46, 46, 46, 46, 46, 46, 46, 46, 46, 46, 46, 46, 46,
+                     46, 46, 46, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 45,
+                     45, 45, 45, 45, 45, 45, 45, 45])
```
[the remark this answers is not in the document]
chahank: Yes, the tests and the tutorial need writing and re-writing.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,829 @@
...
+    if np.any(poly_mask):
+        gdf_pnt = gpd.GeoDataFrame(
+            pd.concat(
+                [gdf_pnt, _gdf_poly_to_pnt(exp.gdf[poly_mask], res, to_meters, disagg)]
+            ))
+    if np.any(line_mask):
+        gdf_pnt = gpd.GeoDataFrame(
+            pd.concat(
+            [gdf_pnt, _gdf_line_to_pnt(exp.gdf[line_mask], res, to_meters, disagg)]
+            ))
```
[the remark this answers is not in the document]
chahank: Maybe it actually does... I will try it!

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+        Disaggregation method of the shapes's original value onto its inter-
+        polated points. 'avg': Divide the value evenly over all the new points;
+        'fix': Replicate the value onto all the new points. Default is 'avg'.
+        Works in combination with the kwarg 'disagg_val'.
+    disagg_val: float, optional
+        Specifies from where and what number should be taken as the value, which
+        is to be disaggregated according to the method provided in disagg_met.
+        None: default. The shape's value is taken from the exp.gdf.value column.
+        Will throw an error if no value column is present.
+        float: This given number will be disaggregated according to the method.
```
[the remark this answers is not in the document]
chahank: This is probably a good idea indeed. This makes the method slightly less flexible, but clearer.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+    agg : string, optional
+        Aggregation method of the point impacts into impact for respective
+        parent-geometry.
+        If 'avg', the impact is averaged over all points in each geometry.
+        If 'sum', the impact is summed over all points in each geometry.
+        The default is 'sum'.
```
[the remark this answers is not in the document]
chahank: Should we do the same with the `disagg_met`?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+        The set of impact functions.
+    haz : Hazard
+        The hazard instance.
+    res : float
+        Resolution of the disaggregation grid (polygon) or line (lines).
+    to_meters : bool, optional
+       If True, res is interpreted as meters, and geometries are projected to
+       an equal area projection for disaggregation.  The exposures are
+       then projected back to the original projections before  impact
+       calculation. The default is False.
```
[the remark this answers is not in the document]
chahank: I think I prefer the `to_meters` version because one can use any unit in principle. When `to_meters = False`, the disaggregation is simply done on the coordinates provided, and thus `res` is in the same unit as the provided coordinates. In this way, the user can use any projection for their coordinate. This needs neither be in `meters` nor in `degrees`.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+    containing the geometries.
+
+    Parameters
+    ----------
+    impact_pnt : Impact
+        Impact object with impact per exposure point (lines of exp_pnt)
+    exp_pnt : Exposures
+        Exposures with a gdf featuring a multi-index, as obtained from
+        disaggregation method exp_geom_to_pnt(). First level indicating
+        membership of original geometries, second level the disaggregated points
```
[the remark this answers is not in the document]
chahank: Yes, well spotted. The docstrings need some more care and are not all complete.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+
+    line_mask, poly_mask = _line_poly_mask(exp.gdf)
+    gdf_geom = exp.gdf.copy()
+
+    if disagg_val is not None:
+        gdf_geom.value = disagg_val
+
+    if ((disagg_val is None) and ('value' not in gdf_geom.columns)):
+        raise ValueError('There is no value column in the exposure gdf to'+
+                         ' disaggregate from. Please set disagg_val explicitly.')
```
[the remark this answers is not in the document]
chahank: Probably this check should be put in the method that is really using the parameter instead of here in the exposed method maybe? Also, I agree that it can be removed here probably.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+
+    # Needed because gdf.explode() requires numeric index
+    idx = gdf.index.to_list() #To restore the naming of the index
+
+    gdf_points = gdf.copy().reset_index(drop=True)
+
+    if isinstance(res, tuple):
+        x_grid, y_grid = res
+        gdf_points['geometry_pnt'] = gdf_points.apply(
+            lambda row: _interp_one_poly_grid(row.geometry, x_grid, y_grid), axis=1)
```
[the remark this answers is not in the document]
chahank: This is for the moment a dirty fix to allow users to use their own grid. Should we rather make a method with another signature?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+    frequency : np.array
+        annual frequency of event
+
+    Returns
+    -------
+    float
+        average annual impact aggregated
+
+    """
+    return sum(at_event * freq)
```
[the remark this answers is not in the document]
chahank: Will be removed, is taken care of in PR #436

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
emanuel-schmid: actually yes, imho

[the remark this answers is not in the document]
chahank: Ok... This will mean about 3 to 4 more methods as all the parent (exposed methods) also need adjustment then.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
Evelyn-M: ehm, no. 
While I see the "one thing at a time argument", this is really impractical in practice, and i strongly advocate against it, since I use the latter syntax quite often for my stuff.
I need to experiment with different resolutions relatively often, and then flexibly want to set the value  of the interpolated points to the resolution. The track-keeping is much more cumbersome if I have to  re-assign an (at that place still wrong!) value of `res` to the exposure-gdf first, and then do the impact calc.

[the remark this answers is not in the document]
chahank: Good point too.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
chahank: It does not work as for an empty geodataframe the method `_gdf_poly_to_pnt` fails currently.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,811 @@
...
+        gdf_pnt = gpd.GeoDataFrame(
+            pd.concat(
+                [gdf_pnt, gdf_poly_to_pnt(gdf_geom[poly_mask], res, to_meters, disagg_met)]
+            ))
+    if np.any(line_mask):
+        gdf_pnt = gpd.GeoDataFrame(
+            pd.concat(
+            [gdf_pnt, gdf_line_to_pnt(gdf_geom[line_mask], res, to_meters, disagg_met)]
+            ))
+
```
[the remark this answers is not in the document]
chahank: I works now.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
chahank: Solved.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
chahank: Another way would be to have the signature `(... res, to_meters=False, grid=None...)` with
 - if `grid = (x_grid, y_grid)` ignore `res, to_meters` and disaggregate on the grid
 - if `grid=None` and `to_meters=False` disaggregate with resolution `res` on the same `CRS` as the gdf
 - if `grid=None` and `to_meter=True` disaggregate with resolution `res` in meters on a locally (for each polygon) defined equidistant projection.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,812 @@
...
+from climada.engine import Impact
+from climada.util import coordinates as u_coord
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def calc_geom_impact(
+        exp, impf_set, haz, res,
+        to_meters=False, disagg_met='avg', disagg_val=None, agg='sum'):
```
emanuel-schmid:
```suggestion
        to_meters=False, disagg_met='avg', disagg_val=None, agg=AggMethod.SUM):
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,812 @@
...
+        None: default. The shape's value is taken from the exp.gdf.value column.
+        Will throw an error if no value column is present.
+        float: This given number will be disaggregated according to the method.
+        The default is None.
+    agg : string, optional
+        Aggregation method of the point impacts into impact for respective
+        parent-geometry.
+        If 'avg', the impact is averaged over all points in each geometry.
+        If 'sum', the impact is summed over all points in each geometry.
+        The default is 'sum'.
```
emanuel-schmid, line 83:
```suggestion
    agg : AggMethod, optional
        The default is AggMethod.SUM
```

emanuel-schmid (commented):
Showing an example of how to use an Enum in place of a str.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
chahank: Actually, this was on purpose. When a dataframe is empty but has an index (e.g., after being masked with an empty mask), then `pd.concat` will always give an `Index` instead of a `MultiIndex`. This is unwanted behaviour here.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
Evelyn-M: This is the "last" function using the `disagg_val` parameter. It's therefore in the correct place. Analogously with the function `exp_geom_to_grid`

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,812 @@
...
+    impact_agg.coord_exp = np.array([repr_pnts.y, repr_pnts.x]).transpose()
+
+    # Add original geometries for plotting
+    impact_agg.geom_exp = exp_pnt_gdf.xs(0, level=1)\
+        .set_geometry('geometry_orig')\
+            .geometry.rename('geometry')
+
+    return impact_agg
+
+
```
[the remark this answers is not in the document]
Evelyn-M: @emanuel-schmid I tried implementing this logic. Could you check that this is how it's done now in the full code?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1065 @@
...
+from climada.util import coordinates as u_coord
+
+LOGGER = logging.getLogger(__name__)
+
+
+class DisaggMethod(Enum):
+    """
+    Disaggregation Method for the ... function
+
+	AVG : the geometry's value is equally averaged over all its interpolated points
```
emanuel-schmid, line 42:
```suggestion
	DIV: the geometry's distributed to equal parts over all its interpolated points
```

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
emanuel-schmid: @Evelyn-M Thanks! 😃 I've amended one small change. When checking for the value of an enum the preferable way is to

```python
if AggMethod(agg_met) is AggMethod.SUM:
```

instead of

```python
if agg_met.value == 'sum':
```

With the former, there is a lot of checking done in the background, and the error message - if any - points directly to the problem: `ValueError: 'fix' is not a valid AggMethod`

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
emanuel-schmid: Still - when `disagg_val` is not None, the whole calculation is kind of done on a different `Exposures` object.
Why isn't the `Exposures` of interest not given in the first place?
Instead of
```
exp_grid = exp_geom_to_grid(exp, disagg_val=disagg_val)
```
one could always
```
exp.gdf.value = disagg_val
exp_grid = exp_geom_to_grid(exp)
```
And if the original exposures must be kept (for what reason?), the user can also take care of that.
Bottom line: the function seems to be doing too many things at a time.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1059 @@
...
+        Exposures with a double index geodataframe, first level indicating
+        membership of the original geometries of exp,
+        second for the point disaggregation within each geometries.
+
+    """
+
+    gdf_geom = exp.gdf.copy()
+
+    if disagg_val is not None:
+        gdf_geom.value = disagg_val
```
emanuel-schmid, line 445: Is the copy still necessary if `disagg_val` _is_ `None`?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1059 @@
...
+                         ' disaggregate from. Please set disagg_val explicitly.')
+
+    gdf_pnt = gdf_to_grid(gdf_geom, grid)
+
+    # disaggregate value column
+    if DisaggMethod(disagg_met) is DisaggMethod.DIV:
+        gdf_pnt = _disagg_values_div(gdf_pnt)
+
+    # set lat lon and centroids
+    exp_pnt = exp.copy()
```
emanuel-schmid, line 458:
```suggestion
    exp_pnt = exp.copy(deep=False)
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1059 @@
...
+    # a dataframe with a multi-index breaks the multi-index
+    gdf_pnt = gpd.GeoDataFrame([])
+    if pnt_mask.any():
+        gdf_pnt = gpd.GeoDataFrame(pd.concat([
+            gdf_pnt,
+            gdf[pnt_mask]
+        ]))
+    if line_mask.any():
+        raise AttributeError("The dataframe contains lines. Lines cannot be disaggregated onto a "
+                             "fixed grid.")
```
emanuel-schmid, line 574: Why is it that points can be disaggregated onto a fixed grid but lines cannot?

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1059 @@
...
+        to represent values per polygon / line (first index).
+
+    Returns
+    -------
+    gdf_disagg : gpd.GeoDataFrame
+        The value per geometry are evenly distributed over the points per geometry.
+
+    """
+
+    gdf_disagg = gdf_pnts.copy()
```
emanuel-schmid, line 603: Is it really necessary to make a deep copy here? I doubt `groupy` makes any changes on the original dataframe and `reindex` is making a copy of the index by itself (unless called with `copy=False`)

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1059 @@
...
+    -------
+    poly : shapely Polygon
+        Polygon in desired projection
+    """
+
+    project = _get_pyproj_trafo(orig_crs, dest_crs)
+    return sh.ops.transform(project.transform, poly)
+
+
+def _line_to_pnts(gdf_lines, res, to_meters=False):
```
emanuel-schmid, line 875:
```suggestion
def _line_to_pnts(gdf_lines, res, to_meters):
```

climada/util/lines_polys_handler.py
Evelyn-M: strictly speaking not, but how would you handle the next line then?
https://github.com/CLIMADA-project/climada_python/blob/3420912b8544897b328faee83a1df840e5604038/climada/util/lines_polys_handler.py#L451 

we would have to run the `gdf_to_grid` then either with `gdf_geom` or with `exp.gdf` depending on the case. That's also adding an "unnecessary" line?

climada/util/lines_polys_handler.py
Evelyn-M: allowing to keep points in a mixed point / poly dataframe results in having a disaggregated dataframe in the end that has single indices (in the point section) and multi-indices (in the poly section). This then cause issues in the re-aggregation, which expects strictly a multi-index.

Hence I suggest to either:
- allow disaggregation on _any_ mixed line / point / poly dataframe. That would need a method to temporarily set a multi-index on original point geometries and a method to disaggregate a line onto a grid, too.
- allow disaggregation to grid only for polygon gdf

I advocate for now for the latter, and keep the first as an extension option in the issues section.

climada/util/lines_polys_handler.py
Evelyn-M: sounds reasonable.

climada/util/lines_polys_handler.py
Evelyn-M: Keep this as an extension idea in the issues section.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
Evelyn-M: I cannot find this anymore. @chahank, you adopted Emanuels suggestion and added a separate method, right? Then this can be resolved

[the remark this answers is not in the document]
chahank: Yes, it is now a separate method.

climada/util/lines_polys_handler.py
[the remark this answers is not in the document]
Evelyn-M: I wouldnt want to assign a value to the non-disaggregated (original) exp.gdf, as very often I have to experiment with different resolutions for disaggregation and hence different values that need to be set.
In order to avoid assigning errors (say, once i tried to disaggregate to 300m, and hence set the value of each point to 300, next time i try to disaggregate to 500m but forget to set the value to 500 ) and to avoid meaningless assignments (neither 300 nor 500 mean anything as a value on the original gdf, where my objects have arbitrary lengths / areas), i advocate for leaving it this way.

climada/util/lines_polys_handler.py
emanuel-schmid: One could
```python 
    if disagg_val is not None:
        exp = exp.copy()
        exp.gdf.value = disagg_val
```
and then
```
    gdf_pnt = gdf_to_grid(exp.gdf, grid)
```
This could save quite some memory and time in case disagg_val is None and cost only a little more in case it isn't. (I don't care all too much about whether the line is necessary, it's the calculation it triggers that worries me 😁  )

Evelyn-M: Fair enough, thanks!

Evelyn-M:
I have troubles with the right syntax for using the Enum classes.
Please see the example below.
It seems to be a difference whether I directly load the classes into memory (within the script where they are defined) vs. when I load them into another script via the definition script. 
When I assign the disagg_met or agg_met then from "outside" (e.g. as `u_lp.DisaggMethod.DIV`), this is not recognized inside the script as the same `DisaggMethod.DIV`
What is the right method then to assign a disagg_met or agg_met as a kwarg in one of the functions, then?

image

emanuel-schmid:
what! 😱 
As I can't easily reproduce this outrageous behavior ...

image

 - could you perhaps provide an MRE?
Many thanks!

Evelyn-M:
> 

```
from enum import Enum
import climada.util.lines_polys_handler as u_lp

class DisaggMethod(Enum):
    """
    Disaggregation Method for the ... function

	DIV : the geometry's distributed to equal parts over all its interpolated points
	FIX : the geometry's value is replicated over all its interpolated points
    """
    DIV = 'div'
    FIX = 'fix'

```
```

class AggMethod(Enum):
    """
    Aggregation Method for the aggregate_impact_mat function

	SUM : the impact is summed over all points in the polygon/line
    """
    SUM = 'sum'


disagg_met = u_lp.DisaggMethod.DIV
disagg_met is u_lp.DisaggMethod.DIV
disagg_met is DisaggMethod.DIV
```

Evelyn-M:
Strange.. I re-tested it with this code, and it keeps reproducing exactly what I showed earlier.
I agree that you're doing the exact same thing though..

emanuel-schmid:
Alright! 😌 
This is a redefinition of `DisaggMethod` not an import. Now it all makes perfect sense. Like this you created _another_ Enum, with the same name and the same values but still - not the same thing. The `is` and even the '==` comparison will return `False`.
The way to deal with this: just don't recreate the Enum, import it.

Evelyn-M:
ah, ok. so that means to also never use this class "from the inside" when working directly on the module. good to know

Evelyn-M:
The unit tests for `test_calc_geom_impact_polys` are failing when run from the command line, since different numbers are returned from calculating it locally vs. from there.
I have no idea why this is the case.

When I run the same calculation locally, (
````
imp1 = u_lp.calc_geom_impact(
            exp_poly, impf_set, haz, res=0.1, to_meters=False, 
            disagg_met=u_lp.DisaggMethod.DIV,
            disagg_val=None, agg_met=u_lp.AggMethod.SUM)````, ) I get `imp1.aai_agg=2182703.085366719`. From the cmd line, this yields `24599.05777`. Any chance that it "sees" a different gdf?
The same tests in the `test_calc_geom_impact_lines` pass.


```
FAIL: test_calc_geom_impact_polys (test_lines_polys_handler.TestGeomImpactCalcs)
test calc_geom_impact() with polygons
  File "/Users/evelynm/climada_python/climada/util/test/test_lines_polys_handler.py", line 178, in test_calc_geom_impact_polys
    self.assertAlmostEqual(imp1.aai_agg, 2182703.085366719, 3)
AssertionError: 24599.057772318487 != 2182703.085366719 within 3 places (2158104.0275944006 difference)
````

**Comment by Evelyn-M** [comment]
The current methods fail with a point-only gdf and with mixed-geometry gdfs.

I gathered 2 examples to start with in the test file, `test_calc_geom_impact_points` and `test_calc_geom_impact_mixed`.

One approach could be to work only with single-geometry type gdfs (lines OR polygons), and tackle this in a future release.

**Comment by Evelyn-M** [comment]
To-Dos are collected in issue #472

**Comment by emanuel-schmid** [comment]
> It seems that the data API Client doesn't return me the exp_poly properly from which the gdf_poly is derived (it stopped working altogether now). So all the values in this test need to be adjusted accordingly by someone who can load the current gdf / exp.

What!? 😱 How did it stop working? How is it not returning it properly.
I can probably load the gdf/exp's, if you hand them to me and tell me how you want to get them back from the API.

**Comment by Evelyn-M** [comment]
> What!? 😱 How did it stop working? How is it not returning it properly. I can probably load the gdf/exp's, if you hand them to me and tell me how you want to get them back from the API.

So I'm loading the exp_poly as indicated in the beginning of the test script. However, suddenly it tells me that `base` is not an allowed exp type, only `litpop` and `crop-sth`. Hence I cannot load it anymore at all. Not sure why; re-starting a few times might help ^^

**Comment by chahank** [comment]
> The current methods fail with a point-only gdf and with mixed-geometry gdfs.
> 
> I gathered 2 examples to start with in the test file, `test_calc_geom_impact_points` and `test_calc_geom_impact_mixed`.
> 
> One approach could be to work only with single-geometry type gdfs (lines OR polygons), and tackle this in a future release.

This was solved in acf8a06a5665fa9d37b3c2769b4862ff2228616b . The methods failed because the points geodataframe had no `geometry_orig` column and no multi-index.

**Comment by chahank** [comment]
> The current methods fail with a point-only gdf and with mixed-geometry gdfs.
> 
> I gathered 2 examples to start with in the test file, `test_calc_geom_impact_points` and `test_calc_geom_impact_mixed`.
> 
> One approach could be to work only with single-geometry type gdfs (lines OR polygons), and tackle this in a future release.

This was solved in https://github.com/CLIMADA-project/climada_python/commit/acf8a06a5665fa9d37b3c2769b4862ff2228616b

**Comment by chahank** [comment]
> It seems that the data API Client doesn't return me the exp_poly properly from which the gdf_poly is derived (it stopped working altogether now). So all the values in this test need to be adjusted accordingly by someone who can load the current gdf / exp.

For me locally it works all good for now.

File: climada/util/lines_polys_handler.py [hunk 42]
```diff
@@ -0,0 +1,1073 @@
...
+
+    # Add original geometries for plotting
+    impact_agg.geom_exp = exp_pnt_gdf.xs(0, level=1)\
+        .set_geometry('geometry_orig')\
+            .geometry.rename('geometry')
+
+    return impact_agg
+
+
+def _aggregate_impact_mat(imp_pnt, gdf_pnt, agg_met):
```
emanuel-schmid, line 189:
```suggestion
def _aggregate_impact_mat(imp_pnt, gdf_pnt):
```

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1073 @@
...
+        Impact object with impact per point (rows of gdf_pnt)
+    gdf_pnt : gpd.GeoDataFrame
+        Exposures geodataframe with a multi-index, as obtained from disaggregation
+        method exp_geom_to_pnt(). First level indicating
+        membership of original geometries, second level the disaggregated points
+    agg_met : AggMethod
+        Aggregation method of the point impacts into impact for respective
+        parent-geometry.
+        If 'SUM', the impact is summed over all points in each geometry.
+        The default is 'SUM'.
```
emanuel-schmid, line 206:
```suggestion
```

climada/util/lines_polys_handler.py
chahank: I do not understand, why should we remove the aggregation method from the aggregation function? commit 88785411f63103d13e515574aded5bcf18e95cb7 proposes a solution.

climada/util/lines_polys_handler.py
```diff
@@ -0,0 +1,1077 @@
...
+    col_geom = gdf_pnt.index.get_level_values(level=0)
+    # Converts string multi-index level 0 to integer index
+    col_geom = np.sort(np.unique(col_geom, return_inverse=True)[1])
+    row_pnt = np.arange(len(col_geom))
+
+    if agg_met is AggMethod.SUM:
+        mask = np.ones(len(col_geom))
+    else:
+        raise NotImplementedError(
+            'The available aggregation methods are %s', AggMethod._member_names_)
```
emanuel-schmid, line 224:
```suggestion
            f'The available aggregation methods are {AggMethod._member_names_}')
```

PR #352, opened 2022-01-24
Fix infinite frequency

chahank:
The method `StormEurope.from_footprints` resulted in a hazard with infinite frequency when reading a single footprint or a set of footprints from the same year.

This fixes the bug by taking as divisor in the frequency as `np.max([(last_year(haz.date) - first_year(haz.date)), 1])`.

In addition, the docstring for `from_cosmoe_file` and `from_icon_grid` were update to make it clear that the frequency is NOT in 1/years, but 1/(number of runs/ensemble members).

### DIFF
--- a/climada/hazard/storm_europe.py
+++ b/climada/hazard/storm_europe.py
@@ -171,7 +171,7 @@ class StormEurope(Hazard):
         haz.event_id = np.arange(1, len(haz.event_id) + 1)
         haz.frequency = np.divide(
             np.ones_like(haz.date),
-            (last_year(haz.date) - first_year(haz.date))
+            np.max([(last_year(haz.date) - first_year(haz.date)), 1])
         )
 
         haz.tag = TagHazard(
@@ -262,6 +262,9 @@ class StormEurope(Hazard):
         Works for MeteoSwiss model output of
         COSMO-1E (11 members, resolution 1.1 km, forecast period 33-45 hours)
         COSMO-2E (21 members, resolution 2.2 km, forecast period 5 days)
+
+        The frequency of all storms is equal to 1/11 and 1/21 for COSMO-1E
+        and COSMO-2E, respectively.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


climada/hazard/storm_europe.py
```diff
@@ -263,6 +263,9 @@ def from_cosmoe_file(cls, fp_file, run_datetime, event_date=None,
         COSMO-1E (11 members, resolution 1.1 km, forecast period 33-45 hours)
         COSMO-2E (21 members, resolution 2.2 km, forecast period 5 days)
 
+        The frequency of all storms is equal to 1/11 and 1/21 for COSMO-1E
+        and COSMO-2E, respectively.
```
chahank, line 267:
```suggestion
        The frequency of each event is informed by their probability in the ensemble forecast and is equal to 1/11 or
        1/21 for COSMO-1E or COSMO-2E, respectively.
```

climada/hazard/storm_europe.py
```diff
@@ -384,6 +387,8 @@ def from_icon_grib(cls, run_datetime, event_date=None, model_name='icon-eu-eps',
         because of the given file structure with 1-hour, 3-hour and
         6-hour maxima provided.
 
+        The frequency for one event is 1/(number of runs).
```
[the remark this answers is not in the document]
chahank: Do you know the 'suggestion' feature for Github comments? You may use these also in the future if you want to :D

chahank, line 390:
```suggestion
        The frequency for one event is 1/(number of ensemble members)
```

chahank:
Thanks @ThomasRoosli I update the docstrings as you suggested. Ready to merge?

chahank:
> Very good changes. Only one request is important: "frequency is equal to 1/(number of members)". the current phrasing "1/(number of runs)" could be missunderstood as several runs with different initial times. What do you think?

Did I address the correct docstring here?

chahank:
I will merge now. If there are issues we can re-open this request.

PR #355, opened 2022-01-24
Fix/cartopy0.19 compatibility

ThomasRoosli:
Pragmatic fix to ensure cartopy0.19 compatibility.

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -735,7 +735,13 @@ def get_transformation(crs_in):
     try:
         units = crs_epsg.proj4_params['units']
     except KeyError:
-        units = '°'
+        if crs_epsg.proj4_params == {}: # if crs_epsg is _EPSGProjection under cartopy 0.20
+            try:
+                units = crs_epsg.to_dict()['units']
+            except KeyError:
+                units = '°'
+        else: # crs_epsg.proj4_params exists under cartopy <= 0.19, but does not contain ['units']
+            units = '°'
     return crs_epsg, units
 
 
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -733,7 +733,7 @@ def get_transformation(crs_in):
         crs_epsg = ccrs.PlateCarree()
 
     try:
-        units = crs_epsg.to_dict()['units']
+        units = crs_epsg.proj4_params['units']
     except KeyError:
         units = '°'
     return crs_epsg, units
--- a/requirements/env_climada.yml
+++ b/requirements/env_climada.yml
@@ -4,7 +4,7 @@ channels:
   - defaults
 dependencies:
   - bottleneck>=1.3
-  - cartopy>=0.20.0
+  - cartopy>=0.18
   - cfgrib>=0.9.7
   - contextily>=1.0,<1.2  # 1.2 has removed the deprecated contextily.sources
   - dask>=2.25


emanuel-schmid:
Because cartopy 0.20 requires proj 8.0 and this cannot be installed easily on ubuntu as it seems (https://github.com/SciTools/iris/issues/4468#issuecomment-997997068).

The use case is an installation without conda.

Jenkins will have whatever suits conda when specified >=0.19 (or even 0.18)

Are there reasons not to?

tovogt:
Maintaining compatibility with several versions of a package is always a lot of work, and you would at least need to run all unit tests against all versions that you would like to support.

For example, for me to review this PR, I would need to set up a new conda environment with cartopy 0.19.0, and one with 0.18.0 and see whether all tests succeed. I should probably also read the cartopy changelogs to understand what else might have changed between 0.19.0 and 0.20.0 that might have been overlooked in this PR. 

As for the problem with proj 8.0: This only applies to pure pip-users, right? In a conda environment, this is no problem at all.

emanuel-schmid:
The problem with proj 8.0 applies to pure pip-users, right.

For the version compatibility testing - I agree, it's not really feasible. Allthough in theory we could run these tests in Jenkins with a few configuration changes, in practice it wouldn't scale. The current integration tests last about an hour in total. When we start running them on different versions and, worse, combinations of different versions, "we're gonna need a bigger boat".

What about keeping the requirements/env_climada.yml file as it is, and commit only the changes in climada/util/plot.py?

The pure pip installation doesn't see the yml file and the changes in plot.py seem unsuspicious.

Thus we would have something like an inofficial backwards-compatibility for this release. With some luck we can keep it up until proj 8.0 is available on ubuntu out of the box.

tovogt:
I made an effort to setup two legacy environments, one with `cartopy==0.18.0` and one with `cartopy==0.19.0.post1`, and in both envs, the unittests `climada.util.test.test_plot` fail at the same positions:
```
======================================================================
ERROR: test_get_transform_3035_pass (climada.util.test.test_plot.TestFuncs)
Check that assigned attribute is correctly set.
  File "/home/tovogt/code/climada_python/climada/util/test/test_plot.py", line 47, in test_get_transform_3035_pass
    self.assertIsInstance(res, cartopy._epsg._EPSGProjection)
AttributeError: module 'cartopy' has no attribute '_epsg'

======================================================================
FAIL: test_get_transform_3395_pass (climada.util.test.test_plot.TestFuncs)
Check that assigned attribute is correctly set.
  File "/home/tovogt/code/climada_python/climada/util/test/test_plot.py", line 41, in test_get_transform_3395_pass
    self.assertIsInstance(res, cartopy.crs.Mercator)
AssertionError: <cartopy.crs.PlateCarree object at 0x7fd1773dc7c0> is not an instance of <class 'cartopy.crs.Mercator'>
```
So, I'm sorry to say this, but this is the kind of review I prefer not to spend my time with: maintaining support for legacy versions of third-party packages. I understand that this might be of interest for a certain group of CLIMADA users. While I, personally, recommend that you refrain from providing this kind of support in general, it's of course completely up to you to do it anyway. But I would prefer not to be involved in this kind of work. No offence! :peace_symbol: :slightly_smiling_face:

emanuel-schmid:
Oh-uh. Awfully sorry for that wast of time! 😳 \
I simply assumed that the tests would pass for cartopy 0.18 without checking myself. \
We'll never do that again. Promised. ☮️

PR #356, opened 2022-01-24
Add change hazard centroids from meta

chahank:
This small fix allows the hazard change centroids to use the meta from the centroids.

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1734,16 +1734,21 @@ class Hazard():
                 self.centroids.lon, self.centroids.lat,
                 centroids.meta['width'], centroids.meta['height'],
                 centroids.meta['transform'])
+            if -1 in new_cent_idx:
+                raise ValueError("At least one hazard centroid is out of"
+                                 "the raster defined by centroids.meta."
+                                 " Please choose a larger raster.")
         else:
             new_cent_idx = u_coord.assign_coordinates(
                 self.centroids.coord, centroids.coord, threshold=threshold
             )
+            if -1 in new_cent_idx:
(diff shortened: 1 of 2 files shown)


emanuel-schmid:
Naive remarks:
- This introduces another occurrence of the `meta` attribute. Is that not problematic?
- There is no unit test - or is this about improving performance?

chahank:
> This introduces another occurrence of the meta attribute. Is that not problematic?
Indeed. But I would say this removes an inconsistency rather than adding one. The way to deal with meta needs an overhaul.
> There is no unit test, right - or is this about improving performance?
This is about being more consistent and much faster indeed. Basically, if you want to project say a Hazard computed for one country onto a grid of the whole world, the current implementation requires to make a dense matrix of all coordinates. This can quickly grow to 100GB or more of required memory. With this fix, one can use the meta an then one only needs the coordinates of the country points which is manageable memory-wise.

chahank:
Unit test added.

Evelyn-M:
Indeed, looks sensible!

chahank:
@emanuel-schmid : any objections?

emanuel-schmid:
Actually yes. Or at least I wished for some convincing about the drop_duplicates part which - at first sight - seems contradictory to the "do one thing at a time" dogma. 🧐

emanuel-schmid:
I've moved the drop_duplicates part to a separate PR, #359

emanuel-schmid:
Judging from the unit-test, the arbitrary removal of concurring centroids seems to be a hazardous feature with the potential of distorting the data.

Shouldn't the fraction and intensity values be aggregated in a more sensible way if we were to allow such a of pooling of centroids?

Issue #353, opened 2022-01-24
shapefile.Reader can't cope with cartopy's changed shapefile format


emanuel-schmid:
Cartopy's `shapereader.natural_earth` method read/writes files into the `~/.local/share/cartopy/natural_earth` directory, using a cache system that obviously checks for presence of the files and skips unnecessary writing.

When this file cache is emptied, and the method is called it writes new files, which have a different format than what they used to have back in 2020. The updated version of cartopy, 0.20, cannot be blamed, the same files are created with cartopy 0.18 and 0.19.

When the files in the new format are read with `shapefile.Reader` the result is heavily distorted, all strings have an arbitrarily sized series of '\x000' at their end, obviously because the `trim` method doesn't remove them properly. Also, the encoding seems to be different.

Suggestions on how to deal with this are welcome.

emanuel-schmid:
My personal favorite would be to replace `shapefile.Reader` (or even the whole package) with cartopy's `shapereder` somehow.\
Is it possible? At what costs?

tovogt:
Apparently, there are two changes at the same time:

1. There has been a new release of Natural Earth (v5.0.0) on 2021-12-08 (https://github.com/nvkelso/natural-earth-vector/releases/tag/v5.0.0). In the old versions, string fields used to be padded with spaces `\x20`. In the new version, string fields are padded with `\x00`. Apparently, this came as a surprise for `shapefile.Reader`: https://github.com/GeospatialPython/pyshp/commit/60b7aed05a0cd01e6c02bec0c5377d5988913571 But the new version also changes some of the contents, at least the ordering. For example, this check will fail now since the ordering of US states has changed: https://github.com/CLIMADA-project/climada_python/blob/2111c496a77cd2bad2f603f8e28adcf9c7914124/climada/util/test/test_coordinates.py#L589
2. Since cartopy version 0.20.2 (https://github.com/SciTools/cartopy/commit/6d787b01e122eea68b67a9b2966e45877755a52d), the `*.prj` and `*.cpg` files that were part of the Natural Earth files for many years, are not ignored (discarded) any more.  The `*.cpg` file contains the information about the encoding which has been `UTF-8` already for a long time (at least 5 years). Since the `*.cpg` was missing, the reader assumed the encoding to be `latin-1`. That's why we had a workaround for that:
[...]

PR #357, opened 2022-01-25
Fixes for Natural Earth 5.0.0 compatibility

tovogt:
As discussed in #353, some data has changed in Natural Earth 5.0.0 and will cause several problems. Furthermore, this gets rid of `shapefile.Reader` in favor of `shapereader.Reader`.

### DIFF
--- a/climada/util/test/test_coordinates.py
+++ b/climada/util/test/test_coordinates.py
@@ -864,7 +864,8 @@ class TestGetGeodata(unittest.TestCase):
         self.assertEqual(len(admin1_info['CHE']), 26)
         self.assertEqual(len(admin1_shapes['IDN']), 33)
         self.assertEqual(len(admin1_info['USA']), 51)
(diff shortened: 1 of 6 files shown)


climada/util/test/test_coordinates.py
```diff
@@ -864,7 +864,7 @@ def test_get_admin1_info_pass(self):
         self.assertEqual(len(admin1_info['CHE']), 26)
         self.assertEqual(len(admin1_shapes['IDN']), 33)
         self.assertEqual(len(admin1_info['USA']), 51)
-        self.assertEqual(admin1_info['USA'][1][4], 'US-WA')
+        self.assertEqual(admin1_info['USA'][1]['iso_3166_2'], 'US-ID')
```
tovogt, line 867: Note that this won't work when running the unit tests with the old Natural Earth versions because the ordering of US states has changed.

PR #358, opened 2022-01-25
Feature/data api 1.0

emanuel-schmid:
This PR contains several updates on the climada API client

- adaptations to the [CLIMADA Data API 1.0](https://climada.ethz.ch/data-api/v1/docs#/). Most important changes of the API:
  - data type structure has changed: properties are returned as lists of dictionaries
  - datasets have a default limit

- the unfortunately named methods `get_dataset`, `get_datasets`, `get_data_type` and `get_data_types` have been renamed to less confusing `get_dataset_info`, `list_dataset_infos`, `get_data_type-info` and `list_data_type_infos`.

- in order to figure out what is available from the data api and how to select the appropriate dataset, a new method `get_property_values` was introduced.

- concatenation of CLIMADA objects within the methods `get_hazard` and `get_exposures` was entirely skipped.

- the tutorial was generously overhauled


chahank:
This is a very good improvement, in particular in user-friendliness, of the CLIMADA API. Some of the methods and their name are still rather confusing as discussed in https://github.com/CLIMADA-project/climada_python/wiki/data-api-client-design . 

It is a good idea to push this update for the upcoming release as the usability of the API has much improved, even though the API public methods have probably not yet reached their final form.

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -43,10 +43,10 @@
 
 
 apiclient = Client()
-ds = apiclient.get_dataset(name=TEST_UNC_OUTPUT_IMPACT)
+ds = apiclient.get_dataset_info(name=TEST_UNC_OUTPUT_IMPACT, status='test_dataset')
```
chahank, line 46: It is a good idea to add a label for test data.

climada/test/test_api_client.py
```diff
@@ -35,38 +35,40 @@ class TestClient(unittest.TestCase):
...
     def test_data_types(self):
         """"""
-        exdts = Client().get_data_types("exposures")
+        exdts = Client().list_data_type_infos("exposures")
         self.assertTrue('litpop' in [exdt.data_type for exdt in exdts])
 
     def test_datasets(self):
         """"""
-        datasets = Client().get_datasets(status=None, name='FAOSTAT_data_producer_prices')
+        datasets = Client().list_dataset_infos(status=None, name='FAOSTAT_data_producer_prices')
```
chahank, line 51: Should not all datasets have a defined `status` ?

climada/test/test_api_client.py
```diff
@@ -35,38 +35,40 @@ class TestClient(unittest.TestCase):
...
         """"""
         client = Client()
 
-        dataset = client.get_dataset(name='FAOSTAT_data_producer_prices')
+        dataset = client.get_dataset_info(name='FAOSTAT_data_producer_prices', status='test_dataset')
         self.assertEqual(dataset.version, 'v1')
         self.assertEqual(len(dataset.files), 1)
         self.assertEqual(dataset.files[0].file_size, 26481)
-        self.assertEqual(dataset.data_type.data_type, 'crop_production')
+        self.assertEqual(dataset.data_type, DataTypeShortInfo('crop_production', 'exposures'))
```
chahank, line 62: What is `DataTypeShortInfo` ?

chahank:
>concatenation of CLIMADA objects within the methods get_hazard and get_exposures was entirely skipped.
What happens now when two or more hazards or exposures are downloaded?

climada/test/test_api_client.py
```diff
@@ -144,7 +146,6 @@ def test_get_hazard(self):
         self.assertEqual(np.max(hazard.centroids.region_id), 756)
```
chahank, line 146: Should not more than one hazard instance be returned now since concatenation is not part of `get_hazard` anymore?

climada/util/api_client.py
```diff
@@ -80,15 +79,23 @@ class DataTypeInfo():
     """data_type data from CLIMADA data API."""
     data_type:str
     data_type_group:str
+    status: str
     description:str
-    properties:list = None
+    properties:list
```
chahank, line 84:
```suggestion
    properties:list(dict)
```

chahank, line 84: Can one indicate that it should be a list of dict?

climada/util/api_client.py
```diff
@@ -80,15 +79,23 @@ class DataTypeInfo():
...
+
+
+@dataclass
+class DataTypeShortInfo():
+    """data_type essentials from CLIMADA data API."""
```
chahank, line 89: This docstring is not clear. What is essentials?

climada/util/api_client.py
```diff
@@ -218,7 +227,7 @@ def _filter_datasets(datasets, multi_props):
             pdf = pdf[pdf[prop].isin(selection)]
         return [datasets[i] for i in pdf.index]
 
-    def get_datasets(self, data_type=None, name=None, version=None, properties=None,
+    def list_dataset_infos(self, data_type=None, name=None, version=None, properties=None,
```
chahank, line 230: What does `name` refer too here? Is there a dataset property with is called `name` ?

climada/util/api_client.py
```diff
@@ -234,19 +243,20 @@ def get_datasets(self, data_type=None, name=None, version=None, properties=None,
             search parameters for dataset properties, by default None
             any property has a string for key and can be a string or a list of strings for value
         status : str, optional
-            valid values are 'preliminary', 'active', 'expired', and 'test_dataset',
+            valid values are 'preliminary', 'active', 'expired', 'test_dataset' and None
```
chahank, line 246: Why have the status `None` ?

climada/util/api_client.py
```diff
@@ -302,8 +312,8 @@ def get_dataset(self, data_type=None, name=None, version=None, properties=None,
             raise Client.NoResult("there is no dataset meeting the requirements")
         return jarr[0]
 
-    def get_dataset_by_uuid(self, uuid):
-        """Returns the data from 'https://climada/rest/dataset/{uuid}' as DatasetInfo object.
+    def get_dataset_info_by_uuid(self, uuid):
+        """Returns the data from 'https://climada.ethz.ch/data-api/v1/dataset/{uuid}' as DatasetInfo object.
```
chahank, line 316: Will this have to be updated with each release of the api? That looks like a road to have inconsistent docstrings...

climada/util/api_client.py
```diff
@@ -319,10 +329,10 @@ def get_dataset_by_uuid(self, uuid):
         NoResult
             if the uuid is not valid
         """
-        url = f'{self.host}/rest/dataset/{uuid}'
+        url = f'{self.url}/dataset/{uuid}'
         return DatasetInfo.from_json(Client._request_200(url))
 
-    def get_data_types(self, data_type_group=None):
+    def list_data_type_infos(self, data_type_group=None):
         """Returns all data types from the climada data API
```
chahank, line 336: This returns more than the data types right? It returns the full info on all the data types no?

climada/util/api_client.py
```diff
@@ -335,12 +345,12 @@ def get_data_types(self, data_type_group=None):
...
         """
-        url = f'{self.host}/rest/data_types'
+        url = f'{self.url}/data_type'
         params = {'data_type_group': data_type_group} \
             if data_type_group else {}
         return [DataTypeInfo(**jobj) for jobj in Client._request_200(url, params=params)]
 
-    def get_data_type(self, data_type):
+    def get_data_type_info(self, data_type):
         """Returns the data type from the climada data API
```
chahank, line 354: That seems contractictory. The `data_type` is specified as an input parameter, so the method cannot return the data type.

climada/util/api_client.py
```diff
@@ -578,12 +561,8 @@ def get_hazard(self, hazard_type, dump_dir=SYSTEM_DIR, max_datasets=1, **kwargs)
             Directory where the files should be downoladed. Default: SYSTEM_DIR
             If the directory is the SYSTEM_DIR, the eventual target directory is organized into
             dump_dir > hazard_type > dataset name > version
-        max_datasets : int, optional
-            Download limit for datasets. If a query matches is matched by more datasets than this
-            number, a ValueError is raised. Setting it to 0 or a negative number inactivates the
-            limit. Default is 10.
-        **kwargs :
-            additional parameters passed on to get_datasets
+       **kwargs :
```
chahank, line 564: I am not convinced by the used of `kwargs` here. This is one of the most exposed methods from the API. Thus, the docstring should not be obscure. It should be very clear to the user what the input parameters are.

climada/util/api_client.py
```diff
@@ -578,12 +561,8 @@ def get_hazard(self, hazard_type, dump_dir=SYSTEM_DIR, max_datasets=1, **kwargs)
             Directory where the files should be downoladed. Default: SYSTEM_DIR
```
chahank, line 561: Where is `SYSTEM_DIR` defined? Can it be changed by the user? Is this set in the config? This is a very exposed method and should give information that can be immediately digested by the user if possible.

climada/util/api_client.py
```diff
@@ -595,20 +574,17 @@ def get_hazard(self, hazard_type, dump_dir=SYSTEM_DIR, max_datasets=1, **kwargs)
...
-        datasets = self.get_datasets(data_type=hazard_type, **kwargs)
-
-        self._check_datasets_for_concatenation(datasets, max_datasets)
+        dataset = self.get_dataset_info(data_type=hazard_type, **kwargs)
+        return self.to_hazard(dataset, dump_dir)
 
-        return self.to_hazard(datasets, dump_dir)
-
-    def to_hazard(self, datasets, dump_dir=SYSTEM_DIR):
+    def to_hazard(self, dataset, dump_dir=SYSTEM_DIR):
```
chahank, line 580: Should this maybe be a private method? Or what is the goal of this method when compare to `get_hazard`?

climada/util/api_client.py
```diff
@@ -654,32 +630,29 @@ def get_exposures(self, exposures_type, dump_dir=SYSTEM_DIR, max_datasets=10, **
...
-        datasets = self.get_datasets(data_type=exposures_type, **kwargs)
-
-        self._check_datasets_for_concatenation(datasets, max_datasets)
-
-        return self.to_exposures(datasets, dump_dir)
+        dataset = self.get_dataset_info(data_type=exposures_type, **kwargs)
+        return self.to_exposures(dataset, dump_dir)
 
-    def to_exposures(self, datasets, dump_dir=SYSTEM_DIR):
+    def to_exposures(self, dataset, dump_dir=SYSTEM_DIR):
```
chahank, line 648: Same question as for `to_hazard`. Should this be private?

climada/util/api_client.py
```diff
@@ -738,14 +710,49 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
             raise ValueError("country must be string or list of strings")
         return self.get_exposures(exposures_type='litpop', dump_dir=dump_dir, properties=properties)
 
+    def get_property_values(self, list_dataset_infos, known_property_values=None,
```
chahank, line 713: Maybe `get_valid_property_values` ?

climada/util/api_client.py
```diff
@@ -738,14 +710,49 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+                            ignore_properties=['date_creation', 'climada_version', 'country_iso3num']):
```
chahank, line 714: Why exclude the `country_iso3num` ?

chahank:
Above are a few comments / question on the update API. I hope they are useful. The tutorial was not looked at for this review.

climada/test/test_api_client.py
emanuel-schmid: Same as the property keys, the status is technically not restricted (yet) to a limited vocabulary. This provides a way to collect all datasets regardless of their status.

climada/test/test_api_client.py
emanuel-schmid: A @dataclass containing data_type and data_type_group, without description, status and property list.

climada/util/api_client.py
emanuel-schmid: 👍  replaced by 'name and group"

climada/util/api_client.py
emanuel-schmid: True, it may lead to inconsistencies but it's still useful to see a valid address in the docs, rather than a reference to the config file or pattern like '...api/vXY/dataset...', up for guessing.

In any case I don't expect many updates in the api url, and only if the client itself has major updates in the DatasetInfo class a change of that url is going to be required. (note that the version v1 is there to stay, new versions will not replace just amend)

climada/util/api_client.py
emanuel-schmid: right

climada/util/api_client.py
emanuel-schmid: You got a point there. What would you suggest? data_type_name? If we change it we should change it all over the place, including the `DataTypeInfo` class.

climada/util/api_client.py
chahank: But does it not include more also?

climada/util/api_client.py
chahank: Is it the name of the file? or the uuid? or how is this name chosen/set?

emanuel-schmid: Arbitrary by the person who creates the dataset.

emanuel-schmid: I.g. the filename is most probably the dataset name plus file extension.

climada/util/api_client.py
emanuel-schmid: obviously not

climada/util/api_client.py
emanuel-schmid: Ok. I've added hints in the doc strings.

climada/util/api_client.py
emanuel-schmid: I wondered too. The use case is: you have already all dataset info objects collected, the files are downloaded and you're ready to do calcuate the impact, but then someone turns off the WLAN. 😏\
After all them method is a short cut that skips a GET request.

climada/util/api_client.py
emanuel-schmid: s.a.

climada/test/test_api_client.py
emanuel-schmid: 👍 right. the test hasn't been fixed - until now

climada/util/api_client.py
emanuel-schmid: Can't tell. It's included again.

emanuel-schmid:
Although the PR is merged (in sight of release 3.1), review comments are still welcome, @manniepmkam 😀

manniepmkam:
The tutorial is well-written and easy to follow. But the litpop data retrieval confuses me a bit. I will post the relevant lines in the next comment.

PR #359, opened 2022-01-31
Feature/hazard change cent drop duplicates

emanuel-schmid:
This PR introduces the `drop_duplicates` option of the `Hazard.change_centroids` method as a new feature.
The feature was originally proposed in PR #356 but held back for release 3.1.

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1734,16 +1734,21 @@ class Hazard():
                 self.centroids.lon, self.centroids.lat,
                 centroids.meta['width'], centroids.meta['height'],
                 centroids.meta['transform'])
+            if -1 in new_cent_idx:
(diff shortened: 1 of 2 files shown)


emanuel-schmid:
@chahank  Judging from the unit-test, the arbitrary removal of concurring centroids seems to be a hazardous feature with the potential of distorting the data.

Shouldn't the fraction and intensity values be aggregated in a more sensible way if we were to allow such a of pooling of centroids?

chahank:
I fully agree @emanuel-schmid . The feature with drop_duplicates was too hastily implemented and as you say better aggregation methods should be used. This is a task for the next release.

chahank:
I would close this pull request. This feature requires more care and should be addressed in the general issue related to raster and vector data  #348 and (re)-projections.

PR #361, opened 2022-02-03
Feature/cartopy 0.20 alleviation

emanuel-schmid:
As discussed on Feb. 2, this PR makes the `get_transformation` method more tolerant with regard to `to_dict` being present and looks first into the proj4_params dictionary for units.

Included is a softened test for its unit test on epsg:3035, which does not check the precise type but rather its behavior (duck type)

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -737,12 +737,14 @@ def get_transformation(crs_in):
     try:
         units = (crs_epsg.proj4_params.get('units')
                 # As of cartopy 0.20 the proj4_params attribute is {} for CRS from an EPSG number
-                # (see issue raised https://github.com/SciTools/cartopy/issues/1974 
+                # (see issue raised https://github.com/SciTools/cartopy/issues/1974
                 # and longterm discussion on https://github.com/SciTools/cartopy/issues/813).
                 # In these cases the units can be fetched through the method `to_dict`.
                 or crs_epsg.to_dict().get('units', '°'))
-    except AttributeError:  # Earlier than cartopy 0.20, the method `to_dict` is not defined, but
-                # this block nevertheless handles the case when there is no 'units' in proj4_params.
+    except AttributeError:
+                # This happens in setups with cartopy<0.20, where `to_dict` is not defined.
+                # Officially, we require cartopy>=0.20, but there are still users around that
+                # can't upgrade due to https://github.com/SciTools/iris/issues/4468
         units = '°'
     return crs_epsg, units
 


emanuel-schmid:
I'm not very confident about the inline comments, don't hesitate to improve.

climada/util/test/test_plot.py
```diff
@@ -44,7 +44,8 @@ def test_get_transform_3395_pass(self):
     def test_get_transform_3035_pass(self):
         """Check that assigned attribute is correctly set."""
         res, unit = u_plot.get_transformation('epsg:3035')
-        self.assertIsInstance(res, cartopy._epsg._EPSGProjection)
+        self.assertIsInstance(res, cartopy.crs.Projection)
+        self.assertEqual(res.epsg_code, 3035)
```
tovogt, line 48: Much better, thanks!

climada/util/plot.py
```diff
@@ -735,8 +735,14 @@ def get_transformation(crs_in):
...
-        units = crs_epsg.to_dict()['units']
-    except KeyError:
+        units = (crs_epsg.proj4_params.get('units')
+                # As of cartopy 0.20 the proj4_params attribute is {} for CRS from an EPSG number
+                # (see issue raised https://github.com/SciTools/cartopy/issues/1974 
+                # and longterm discussion on https://github.com/SciTools/cartopy/issues/813).
+                # In these cases the units can be fetched through the method `to_dict`.
+                or crs_epsg.to_dict().get('units', '°'))
+    except AttributeError:  # Earlier than cartopy 0.20, the method `to_dict` is not defined, but
+                # this block nevertheless handles the case when there is no 'units' in proj4_params.
```
tovogt, line 745: Thanks for inserting the comments! This is really helpful for CLIMADA devs that consider removing or changing those lines in the future. I would also mention why some users might be interested in cartopy<0.20 compatibility even though we officially require 0.20:
```suggestion
        units = (crs_epsg.proj4_params.get('units')
                # As of cartopy 0.20 the proj4_params attribute is {} for CRS from an EPSG number
                # (see issue raised https://github.com/SciTools/cartopy/issues/1974
                # and longterm discussion on https://github.com/SciTools/cartopy/issues/813).
                # In these cases the units can be fetched through the method `to_dict`.
                or crs_epsg.to_dict().get('units', '°'))
    except AttributeError:
                # This happens in setups with cartopy<0.20, where `to_dict` is not defined.
                # Officially, we require cartopy>=0.20, but there are still users around that
                # can't upgrade due to https://github.com/SciTools/iris/issues/4468
```

tovogt (commented):
Thanks! :+1:

climada/util/test/test_plot.py
emanuel-schmid: dito, thanks go to @ThomasRoosli

emanuel-schmid:
@ThomasRoosli : if this works for you, we'll merge

ThomasRoosli (approved):
Thanks, I tested it on my machine, and it works with our setup. Very nice implemention using "or".  I would be very happy if this can be merged.

PR #367, opened 2022-03-01
TCTracks.equal_timestep: fix for non-normalized longitudes

tovogt:
I just noticed that there is an entry in IBTrACS where longitude values are in a very strange format. For the storm with IBTrACS ID "2016055S15202", the agency NADI reported longitudinal coordinates jumping from 204.30 to -155.73 and back to 204.20 (see http://ibtracs.unca.edu/index.php?name=v04r00-2016055S15202). When restricting to the official 6-hourly reporting intervals, the problem is not present, but many people make use of the unofficial 3-hourly data.

I never expected that something like this would exist, but now that we know that it exists, we have to make sure that CLIMADA normalizes all longitudinal values before applying interpolation. This PR fixes it. And it might also fix problems with other data sets that we don't know about currently.

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -1428,7 +1428,7 @@ class TCTracks():
             method = ['linear', 'quadratic', 'cubic'][min(2, track.time.size - 2)]
 
             # handle change of sign in longitude
-            lon = track.lon.copy()
+            lon = u_coord.lon_normalize(track.lon.copy(), center=0)
             if (lon < -170).any() and (lon > 170).any():
                 # crosses 180 degrees east/west -> use positive degrees east
                 lon[lon < 0] += 360
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -541,9 +541,12 @@ class TCTracks():
             ibtracs_ds[f'{tc_var}_agency'] = ('storm', selected_ags[preferred_idx.values])
 
             if tc_var == 'lon':
-                # By IBTrACS default, no longitude should be <= -180, but this is not true for some
-                # agencies, so we have to manually enforce this policy:
-                ibtracs_ds[tc_var].values[(ibtracs_ds[tc_var] <= -180).values] += 360
+                # Most IBTrACS longitudes are either normalized to [-180, 180] or to [0, 360], but
+                # some aren't normalized at all, so we have to make sure that the values are okay:
+                lons = ibtracs_ds[tc_var].values.copy()
+                lon_valid_mask = np.isfinite(lons)
+                lons[lon_valid_mask] = u_coord.lon_normalize(lons[lon_valid_mask], center=0.0)
+                ibtracs_ds[tc_var].values[:] = lons
 
                 # Make sure that the longitude is always chosen positive if a track crosses the
                 # antimeridian:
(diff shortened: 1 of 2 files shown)


climada/hazard/tc_tracks.py
```diff
@@ -1428,7 +1428,7 @@ def _one_interp_data(track, time_step_h, land_geom=None):
             method = ['linear', 'quadratic', 'cubic'][min(2, track.time.size - 2)]
 
             # handle change of sign in longitude
-            lon = track.lon.copy()
+            lon = u_coord.lon_normalize(track.lon.copy(), center=0)
```
tovogt, line 1431: The normalization step introduces a little computational overhead, but I'm quite sure that it's neglectable compared to the other operations in this function.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -622,6 +622,13 @@ def test_interp_track_pass(self):
             with self.assertRaises(ValueError, msg=msg) as _cm:
                 tc_track.equal_timestep(time_step_h=time_step_h)
 
+        # test for tracks with non-normalized longitude
+        tc_track = tc.TCTracks.from_ibtracs_netcdf(provider="nadi", storm_id="2016055S15202")
+        self.assertAlmostEqual(tc_track.data[0].lon.values[0], 203.7, places=1)
+        self.assertAlmostEqual(tc_track.data[0].lon.values[3], -155.7, places=1)
```
tovogt, line 628: We check that this storm's IBTrACS data really has this issue before doing the real test below because IBTrACS data might change in the future. It also demonstrates the problem to other people who might look into the tests in the future.

climada/hazard/tc_tracks.py
ChrisFairless: worth it

ChrisFairless: would it be worth normalising the longitudes when we read in the tracks instead of when we interpolate?

we already make one adjustment to longitude when reading in (tc_tracks.py lines 527-534), and this way we can be more confident that CLIMADA will 'play nice' with all the tracks.

do you know what the overhead is to normalise longitudes for the full event set? it looks like it would still be quite cheap?

climada/hazard/test/test_tc_tracks.py
tovogt: It's not a bug. In the IBTrACS specifications, it's not said that the longitude values of the reporting agencies adhere to any specific format. So, it's well possible that this will stay like this forever since it's literally what the agency reported. However, there's also no guarantee that it won't change in the future.

tovogt: With [eaed9cf](https://github.com/CLIMADA-project/climada_python/pull/367/commits/eaed9cf0c6ec965da59958dc8dc340fcead83f7d), the test data is chosen in a more consistent way.

climada/hazard/tc_tracks.py
ChrisFairless: thanks, and agreed

climada/hazard/test/test_tc_tracks.py
ChrisFairless: looks good

ChrisFairless:
i'm happy for this to merge!

tovogt:
Thanks @ChrisFairless for the fast and helpful review! Now that Jenkins confirms that all tests are fine, we can safely merge this.

PR #370, opened 2022-03-03
Save impact matrices in CostBenefit calculations?

ChrisFairless:
### The issue

The `CostBenefit.calc` method takes a parameter `save_imp`. When True the method saves the Impact object for every impact calculation it makes (which is several). These Impact objects don't have impact matrices saved. I'd like to change things so that they do.

**Reasons not to save the impact matrix**
- Large cost-benefit calculations with multiple adaptation measures create several Impact objects. Storing them all in memory will crash analyses on machines with smaller RAM. With a large enough hazard and exposure this is most of us.

**Reasons to save the impact matrix**
- Many of my favourite things to do with the Impact objects require an impact matrix. In particular, if you want to plot maps of impacts with different adaptation measures applied, you need the impact matrix.

**An alternate solution**
- Make the method more flexible: have two parameters `save_imp` and `save_mat` which let you choose whether to save the Impact and, if so, whether to save its impact matrix.

### This pull request

CostBenefit calculates its impacts using the `Measure.calc_impact` method. If we want it to save the impact matrix, we need to update this method to take a `save_mat` parameter and pass it to `Impact.calc`. I think it makes sense to implement this regardless of everything else in the pull request, and is done in commit 51d3d9fdfa1fe6bc10c4254e5d0266d21ea2a434 for easy cherry picking.

The next commit 7ac95cd582a1279cc6a74894486b899e7808f5b7 updates `CostBenefit` to save impact matrices when the `save_imp` parameter is True.

Finally, one outstanding issue is when risk transfer is involved. When risk transfer parameters are defined the impact calculation `Measure.calc_impact` also calls `Impact.calc_risk_transfer` which always returns an impact object with a 0 x 0 impact matrix. I don't know the risk transfer codebase so well, so I haven't changed anything here.

### DIFF
--- a/climada/engine/cost_benefit.py
+++ b/climada/engine/cost_benefit.py
@@ -752,14 +752,14 @@ class CostBenefit():
             Default: average annual impact (aggregated).
         save_imp : bool, optional
             activate if Impact of each measure is
-            saved. Default: False.
+            saved. Also saves impact matrix. Default: False.
         """
         impact_meas = dict()
 
         # compute impact without measures
         LOGGER.debug('%s impact with no measure.', when)
         imp_tmp = Impact()
-        imp_tmp.calc(exposures, imp_fun_set, hazard)
+        imp_tmp.calc(exposures, imp_fun_set, hazard, save_imp)
         impact_meas[NO_MEASURE] = dict()
         impact_meas[NO_MEASURE]['cost'] = (0, 0)
         impact_meas[NO_MEASURE]['risk'] = risk_func(imp_tmp)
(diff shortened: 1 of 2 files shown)


chahank:
I agree with the general premise, but not sure yet 100% convinced with the proposed solution. I will prepare a more detailed analysis.

chahank:
> Make the method more flexible: have two parameters save_imp and save_mat which let you choose whether to save the Impact and, if so, whether to save its impact matrix.

If this is to be implemented, we should definitely not by default set `save_mat=True`.

chahank:
> > Why would you need the impact matrix for this? What you would need is the eai_exp vector no? Could you provide more precise use cases for the impact matrix?
> 
> Plotting impacts by return period with `Impact.plot_rp_imp` (calling `Impact.local_exceedance_imp`) needs the impact matrix

But why would you do that on a CostBenefit object?

ChrisFairless:
>But why would you do that on a CostBenefit object?
It saves me running `Measures.calc_impact` for every measure and for present and future. Those calculations have to be done anyway ifor the CostBenefit, and the results (without the impact matrix) are already saved.

chahank:
For this example, I disagree. If you really want the individual data, you probably should compute them separately. Having objects that can quickly blow up the memory is not a very sustainable design choice.

ChrisFairless:
OK. Then we just need to merge the first commit https://github.com/CLIMADA-project/climada_python/commit/51d3d9fdfa1fe6bc10c4254e5d0266d21ea2a434 which adds the `save_mat` parameter to `Measures.calc_impact` to let us do that. I can put it in a new PR.

chahank:
All right! Long-time we can refactor CostBenefit taking your point into account. Now, allowing the `Measures` to `sava_mat` is a good idea.

chahank:
@ChrisFairless : will you implement the suggested changes?

emanuel-schmid:
there is already a follow up PR

PR #374, opened 2022-03-07
Feature/tutorial1 update

frqqyy:
- Corrected some spelling mistakes
- Added a piece of guiding information into the part "exposure", because a data set must be downloaded to run the next block. Without this data set, there will be an error. But in the latest version, there is no information saying which data set should be downloaded, or from where, or where the data set should be put into. Only the error information gives a bit of a hint.
- Changed one link to a relative path (This link didn't work before). But I didn't make it to build the readthedocs file, so I don't really know if it works...


chahank:
@frqqyy : did you not also plan to reduce the number of tracks used in the tropical cyclone example to make the code faster and more readable? To do so, you simply choose tracks in a certain time bracket, e.g. `year_range=(2000,2005)`. But mind that you should choose the year range such that the storm `2017260N12310` from later in the tutorial is also included (or choose another storm).

chahank:
This `inside of the “climada” folder you installed.` is slightly misleading. When installing CLIMADA, a folder for internal data is by default created in `~/climada/`. This is where the folder `data` is located.

frqqyy:
> @frqqyy : did you not also plan to reduce the number of tracks used in the tropical cyclone example to make the code faster and more readable? To do so, you simply choose tracks in a certain time bracket, e.g. `year_range=(2000,2005)`. But mind that you should choose the year range such that the storm `2017260N12310` from later in the tutorial is also included (or choose another storm).

I tried, but it didn't bring much to the time this block will take... I still can't see the output after waiting for a very long time...

frqqyy:
> This `inside of the “climada” folder you installed.` is slightly misleading. When installing CLIMADA, a folder for internal data is by default created in `~/climada/`. This is where the folder `data` is located.

Thanks a lot! I changed this sentence

chahank:
> > @frqqyy : did you not also plan to reduce the number of tracks used in the tropical cyclone example to make the code faster and more readable? To do so, you simply choose tracks in a certain time bracket, e.g. `year_range=(2000,2005)`. But mind that you should choose the year range such that the storm `2017260N12310` from later in the tutorial is also included (or choose another storm).
> 
> I tried, but it didn't bring much to the time this block will take... I still can't see the output after waiting for a very long time...

Please try to restart the kernel, and use only a few tracks.

emanuel-schmid:
The frequency curve plot in cell 21 has changed quite a bit. Before it was
image

now it is
image

Do we know why?

emanuel-schmid:
I'd suggest to replace _all_ the links to tutorials within the same repository with relative path to the notebooks instead of the html pages of the documentation, as they work in both ways, to jump from notebook to notebook _and_ from html doc file to html doc file. These three are remaining:

```bash
 egrep "climada-python.*\.html" doc/tutorial/1_main_climada.ipynb
    "- [**Hazard**](https://climada-python.readthedocs.io/en/latest/tutorial/climada_hazard_Hazard.html):\n",
    "    - [Exposures](https://climada-python.readthedocs.io/en/latest/tutorial/climada_entity_Exposures.html):\n",
    "- [**Engine**](https://climada-python.readthedocs.io/en/latest/tutorial/climada_engine_Impact.html):\n",
```

frqqyy:
> # plotting tracks can be very time consuming, depending on the number of tracks. So we choose only a few here, by limiting the time range to one year
> tracks_2017 = TCTracks.from_ibtracs_netcdf(provider='usa', basin='NA', year_range = (2017, 2017))
> tracks_2017 .plot();

Thank you for the suggestion! I already changed it

chahank:
> > The frequency curve plot in cell 21 has changed quite a bit. Before it was image
> > now it is image
> > Do we know why?
> 
> Hmmm sorry I don't know, I didn't change anything in this part

The graphics appear to be the same data, except that in the second one the axis extends to 175. Maybe some high-risk events were added?

frqqyy:
> > > The frequency curve plot in cell 21 has changed quite a bit. Before it was image
> > > now it is image
> > > Do we know why?
> > 
> > 
> > Hmmm sorry I don't know, I didn't change anything in this part
> 
> The graphics appear to be the same data, except that in the second one the axis extends to 175. Maybe some high-risk events were added?

Maybe the dataset was updated since the last time the codes were executed?

emanuel-schmid:
Yes, maybe. 🤷

PR #380, opened 2022-03-15
Fix bug in get_gridcellarea function in util/coordinates

carmensteinmann:
Hello!

I came across an issue with the unit / order of magnitude of this function. 
Assuming that the max grid cell area should be 2500km^2 (50x50km), I am not sure why we had added *100 in the first place - or am I getting mixed up?

Thanks for having a look!

Best, Carmen

### DIFF
--- a/climada/util/coordinates.py
+++ b/climada/util/coordinates.py
@@ -323,7 +323,7 @@ def dist_approx(lat1, lon1, lat2, lon2, log=False, normalize=True,
         raise KeyError("Unknown distance approximation method: %s" % method)
     return (dist, vtan) if log else dist
 
-def get_gridcellarea(lat, resolution=0.5, unit='km2'):
+def get_gridcellarea(lat, resolution=0.5, unit='ha'):
     """The area covered by a grid cell is calculated depending on the latitude
         1 degree = ONE_LAT_KM (111.12km at the equator)
         longitudal distance in km = ONE_LAT_KM*resolution*cos(lat)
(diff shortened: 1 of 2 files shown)


climada/util/coordinates.py
```diff
@@ -342,9 +342,9 @@ def get_gridcellarea(lat, resolution=0.5, unit='km2'):
     """
 
     if unit == 'm2':
-        area = (ONE_LAT_KM * resolution)**2 * np.cos(np.deg2rad(lat)) * 100 * 1000000
+        area = (ONE_LAT_KM * resolution)**2 * np.cos(np.deg2rad(lat)) * 1000000
     else:
-        area = (ONE_LAT_KM * resolution)**2 * np.cos(np.deg2rad(lat)) * 100
+        area = (ONE_LAT_KM * resolution)**2 * np.cos(np.deg2rad(lat))
```
sameberenz, line 347: This correction makes sense to me. I have also double checked with some example coordinates using this external tool: https://www.engr.scu.edu/~emaurer/tools/calc_cell_area_cgi.pl and it looks alright.

sameberenz (approved):

bguillod:
The factor 100 could be for a conversion from km2 to ha. Could it be that somewhere else in Climada this unit is (wrongly) expected? Might be good to check everywhere the function is used before merging, in case you haven’t yet done so

carmensteinmann:
@bguillod thanks a lot for that comment. Thinking about it, probably the crop production data we initially wrote this function for has the unit t/ha. I'll check and add the unit option ha.

sameberenz:
> @bguillod thanks a lot for that comment. Thinking about it, probably the crop production data we initially wrote this function for has the unit t/ha. I'll check and add the unit option ha.

Sounds reasonable, in that case you should check how the funciton is used in teh crop module and if this needs to be aligned

bguillod:
> > @bguillod thanks a lot for that comment. Thinking about it, probably the crop production data we initially wrote this function for has the unit t/ha. I'll check and add the unit option ha.
> 
> Sounds reasonable, in that case you should check how the funciton is used in teh crop module and if this needs to be aligned

Actually a quick search on all calls to that function might be worth it, to make sure you don’t miss other cases which might expect units other than km2 or m2.

ChrisFairless:
yeah it looks like it's called once in `climada_petals/entity/exposures/crop_production.py`

carmensteinmann:
Yes, thanks a lot everyone! I changed the default of the function to ha (as it is used in crop production) and added the calculation in km2.

PR #394, opened 2022-03-30
fix util.plot.geo_scatter_categorical to work with Matplotlib 3.5

blassr:
Colorbar was continuous with the new Matplotlib 3.5 version. This is fixed and categorical again with this fix.


emanuel-schmid:
Thanks a lot for the PR, fixed bugs are always welcome!
I've declined it nevertheless, as it targets main and not develop.
Please create another one, targeting develop.

PR #396, opened 2022-03-30
Update climada_entity_MeasureSet.ipynb

frqqyy:
- Corrected a spelling mistake
- Put two impact functions (before and after the measure) on the same plot, because it was hard to find the differences if they are plotted separately


emanuel-schmid:
@frqqyy : I think in general it's enough to append ';' to the last line of a cell. Otherwise they have no impact.

PR #397, opened 2022-03-31
adapt categorical plotting function to new matplotlib version

emanuel-schmid:
cit. @blassr "Colorbar was continuous with the new Matplotlib 3.5 version. This is fixed and categorical again with this fix."

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -426,7 +426,6 @@ def geo_scatter_categorical(array_sub, geo_coord, var_name, title,
         cmap = kwargs['cmap']
         if isinstance(cmap, str):
             cmap_name = cmap
-            cmap = mpl.cm.get_cmap(cmap)
         else:
             cmap_name = 'defined by the user'
     else:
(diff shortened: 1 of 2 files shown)


emanuel-schmid:
@blassr : can you give an example of a colorbar that's continuous without this fix?

blassr:
'wistia' or 'viridis'

blassr (approved):

chahank:
Quick question: does it also work with Matplotlib < 3.5?

emanuel-schmid:
> Quick question: does it also work with Matplotlib < 3.5?

Very good question! Yes it works, at least for matplotlib 3.2.
(But, completely unrelated to this PR, it ain't pretty.)

chahank:
What is not pretty :D ?

chahank:
I think you probably have run the method in a way that it is not intended to be used as (and probably should not be used as). You should not have more than a few categories in general.

emanuel-schmid:
> I think you probably have run the method in a way that it is not intended to be used as (and probably should not be used as). You should not have more than a few categories in general.

Nah, it's just the font size which is way over the top. ;)

PR #404, opened 2022-04-07
Change ignore_zero to False as in docstring

chahank:
The docstring said the default for `ignore_zero` is `False`, but it was set to `True`. This pull request corrects this and sets it to `False`.

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -270,7 +270,7 @@ class Impact():
 
         return new_imp, Impact()
 
-    def plot_hexbin_eai_exposure(self, mask=None, ignore_zero=True,
+    def plot_hexbin_eai_exposure(self, mask=None, ignore_zero=False,
                                  pop_name=True, buffer=0.0, extend='neither',
                                  axis=None, adapt_fontsize=True, **kwargs):
         """Plot hexbin expected annual impact of each exposure.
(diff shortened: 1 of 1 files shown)


climada/engine/impact.py
```diff
@@ -932,14 +932,14 @@ def run(i_time):
             haz_list[i_time].plot_intensity(1, axis=axis, cmap='Greys', vmin=v_lim[0],
                                             vmax=v_lim[1], alpha=0.8)
             if plot_raster:
-                exp.plot_hexbin(axis=axis, mask=exp_list[i_time], ignore_zero=True,
+                exp.plot_hexbin(axis=axis, mask=exp_list[i_time], ignore_zero=False,
```
emanuel-schmid, line 935: This and the following change are unrelated to the Doc-String. Unlike in the other cases, ignore_zero=True sets an argument and not a default value. Hence the PR changes the result.

However - I couldn't tell which value for `ignore_zero` is more appropriate here.

emanuel-schmid: Well, it does not really _change_ input data, it just ignores it. I don't feel confident on setting it to False here without having seen the impact of this in a real life example.

I suggest to either introduce ignore_zero as a parameter or the undo this two last changes.

More generally and stepping back - since we could change the Pydoc string instead of the default value, do we have examples to have a look on the impact? For all I know, ignorance is bliss. 😉

chahank: So, in general it is GOOD that `ignore_zero=False` as is stated in the docstring. We should keep that. But or the above method for the video making, I agree the best is probably just to carry over the variable. Right?

Issue #423, opened 2022-04-13
A future warning in the impact_data tutorial


frqqyy:
My version of pandas (if it's useful): 1.4.1
I was running the following code block and got this warning:
![impact_data 2 1](https://user-images.githubusercontent.com/71314575/163278077-69f88879-3107-42d6-aa09-81847fe6b953.png)
![impact_data 2 2](https://user-images.githubusercontent.com/71314575/163278083-a278a0e7-281d-4592-a93b-b54db2932bce.png)
![impact_data 2 3](https://user-images.githubusercontent.com/71314575/163278088-5dce70b3-8067-4326-bec9-91251b0f0eb1.png)

peanutfun:
@frqqyy The FutureWarning should have been fixed by #443. Could you confirm that?

frqqyy:
> @frqqyy The FutureWarning should have been fixed by #443. Could you confirm that?

Yes :)

PR #425, opened 2022-04-15
Fixed warning from issue 423 

Abedq:
https://github.com/CLIMADA-project/climada_python/issues/423

climada/engine/impact_data.py 768 from out = out.append(data_out) to out = pd.concat([out, data_out])


peanutfun:
@Abedq Thank you for your contribution! I'm not sure if `main` is the correct target branch for this fix. You based your change off `develop`, which is why all other commits from this branch appear in the PR. I suggest changing the target branch to `develop`, and then cherry picking the merge if we want to move this fix into `main` quickly.

Abedq:
@peanutfun Makes perfect sense! I'll do that over the weekend.

PR #427, opened 2022-04-19
Enable offline mode for the API client

emanuel-schmid:
With this PR, the api_client's `Client` is enhanced with a caching mechanism that can be activated to allow a fallback to previously received results from the data API server.

The mechanism is turned on by the config value `data_api.cache_enabled`. The system default is `false`. But the config file for the unit-tests in the root github directory has set it to `true`, so unit-tests can be repeated off-line if they were executed on-line at some time in the past.
Furthermore, the `Client`'s constructor has an attribute for enabling the caching mechanism.

As long as the caching mechanism is enabled, all raw-results from the climada data api are stored as json files in the directory configured by `data_api.cache_dir`. If there is no internet connection and caching is enabled, results are read from these files, if present. The caching files' names are hash sums from normalized http requests. In this way it is easy to figure out whether a request has been submitted before: the client just looks for a file by the according name. Figuring out what request was submitted for a particular file in the cache is not possible on the other hand. But that is also not required.

The caching mechanism should be - to some extent - multi-processing capable. Racing conditions will be handled by the file system - although possibly not strictly gracefully. Again, that is probably not a requirement here.


climada/util/api_client.py
```diff
@@ -179,18 +251,45 @@ class AmbiguousResult(Exception):
...
-    def __init__(self):
+    class NoConnection(Exception):
+        """To be raised if there is no internet connection and no cached result."""
+
+
+    @staticmethod
+    def _is_online(url):
+        host = [x for x in url.split('/')
+                if x not in ['https:', 'http:', '']][0]
+        port = 80 if host.startswith('http://') else 443
```
ChrisFairless, line 262:
```suggestion
        port = 80 if url.startswith('http://') else 443
```
i think using `host` here will always return port 80, since the http has been stripped from it

climada/util/api_client.py
```diff
@@ -179,18 +251,45 @@ class AmbiguousResult(Exception):
...
+            local file system (location defined by CONFIG.data_api.cache_dir).
+            If set to true, the client can reload the results from the cache in case there is no
+            internet connection and thus work in offline mode.
+            Default: None, in this case the value is taken from CONFIG.data_api.cache_enabled.
         """
         self.headers = {"accept": "application/json"}
         self.url = CONFIG.data_api.url.str().rstrip("/")
         self.chunk_size = CONFIG.data_api.chunk_size.int()
+        self.cache = Cacher(cache_enabled)
+        self.online = Client._is_online(self.url)
```
ChrisFairless, line 290: how expensive is `_is_online` to call in terms of API load? would it make sense to evaluate `Client.is_online` whenever an API call is made (and `cache_enabled = True`), rather than when the client is created?

if this is more than ten minutes to implement it's probably not worth it. i'm thinking about it because my train wifi is disconnecting and reconnecting as i write.

climada/util/api_client.py
```diff
@@ -179,18 +251,45 @@ class AmbiguousResult(Exception):
...
+    def _is_online(url):
+        host = [x for x in url.split('/')
+                if x not in ['https:', 'http:', '']][0]
+        port = 80 if host.startswith('http://') else 443
+        socket.setdefaulttimeout(1)
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as skt:
+            try:
+                skt.connect((host, port))
+                return True
+            except socket.error:
```
ChrisFairless, line 268: will this distinguish between errors from 'no internet connection' and 'the data API is down'?
and do we want to?

emanuel-schmid: No, not really. Response time may be different. And no, I don't think we want to.

climada/engine/test/test_cost_benefit.py
```diff
@@ -34,7 +34,10 @@
 
 
 def get_test_file(ds_name):
-    client = Client()
+    # As this module is part of the installation test suite, we want tom make sure it is running
+    # also in offline modues even when installing from pypi, where there is no test configuration.
```
emanuel-schmid, line 38:
```suggestion
    # also in offline mode even when installing from pypi, where there is no test configuration.
```

climada/engine/test/test_impact.py
```diff
@@ -35,7 +35,10 @@
 
 
 def get_haz_test_file(ds_name):
-    client = Client()
+    # As this module is part of the installation test suite, we want tom make sure it is running
+    # also in offline modues even when installing from pypi, where there is no test configuration.
```
emanuel-schmid, line 39:
```suggestion
    # also in offline mode even when installing from pypi, where there is no test configuration.
```

PR #428, opened 2022-04-19
Suggest using data API for LitPop exposures

### LINKED ISSUE #426: LitPop: Suggest using data API wrapper instead of `from_countries`
On my system, the following code causes a slow memory overflow over several minutes although not much data is pulled from the internet. The process cannot be aborted via keyboard interrupt and is only killed by the OS once the RAM is full.
```python
from climada.entity import LitPop
lp = LitPop.from_countries(["BRA"], fin_mode="pc")
```
Can somebody replicate this behavior? Here's a screenshot from my system (note the slow memory increase and basically no network data exchange):
Screenshot from 2022-04-19 12-08-50

peanutfun:
Add warning that computing LitPop exposures may require excessive amounts of RAM. Suggest using the data API instead.

~~*Note*: I have trouble building the docs locally, so I cannot tell yet if everything works out.~~ *Edit: Looks good on my local build now*

Closes #426

### DIFF
--- a/doc/tutorial/climada_entity_LitPop.ipynb
+++ b/doc/tutorial/climada_entity_LitPop.ipynb
@@ -29,6 +29,8 @@
     "\n",
     "*Note*: All required data except for the population data from Gridded Population of the World (GPW) is downloaded automatically when an `LitPop.set_*` method is called.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


peanutfun:
Looks good on my local build

emanuel-schmid:
Thanks! 😃

Issue #410, opened 2022-04-11
A performance warning in the tutorial for exposures


frqqyy:
![3](https://user-images.githubusercontent.com/71314575/162750025-c6345c9d-b7b6-4f7f-9059-27d0d2e561ac.png)

peanutfun:
The Exposure contains an attribute that has no C-type equivalent and therefore cannot be stored efficiently. I think there is nothing that can be done from the CLIMADA side other than suppressing this warning. It just indicates that writing the file might take a bit longer.

frqqyy:
> The Exposure contains an attribute that has no C-type equivalent and therefore cannot be stored efficiently. I think there is nothing that can be done from the CLIMADA side other than suppressing this warning. It just indicates that writing the file might take a bit longer.

Thank you! I'll suppress the warnings.

emanuel-schmid:
Since there is nothing that can be done about it - wouldn't it be more appropriate to suppress it where it's raised?

Something along these lines in the exposures.base module:
```python
import warnings
with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    # code that produces a warning
```
(from https://stackoverflow.com/questions/54768539/suppress-a-python-warning)

PR #429, opened 2022-04-19
Suppressed the performance warning for writing the hdf5 file

frqqyy:


emanuel-schmid:
@frqqyy : ideally, the changes would be small enough so the diff can be shown. Since you provide no further description of the PR - could you provide some details of the essential changes?

frqqyy:
> @frqqyy : ideally, the changes would be small enough so the diff can be shown. Since you provide no further description of the PR - could you provide some details of the essential changes?

https://github.com/CLIMADA-project/climada_python/issues/410
The change aims to solve this issue. 
And here are the codes that were added:
image

frqqyy:
> Since there is nothing that can be done about it - wouldn't it be more appropriate to suppress it where it's raised?
> 
> Something along these lines in the exposures.base module:
> 
> ```python
> import warnings
> with warnings.catch_warnings():
>     warnings.simplefilter('ignore')
>     # code that produces a warning
> ```
> 
> (from https://stackoverflow.com/questions/54768539/suppress-a-python-warning)

Thank you, I updated the tutorial again!

peanutfun:
@emanuel-schmid suggested that we suppress the warning in the module where `Exposures.write_hdf5` is defined, not just in the tutorial. The proposed change avoids the warning in the tutorial, but it then will appear again if `write_hdf5` is called in another script. I'm closing this PR in favor of the fix suggested in #410.

PR #430, opened 2022-04-20
TropCyclone: lower memory requirements of compute_windfield

tovogt:
While playing around with higher resolutions in time and space (i.e., smaller timesteps for TC tracks, and more centroids), I found that the current implementation of `compute_windfield` has quite high memory requirements because there are several temporary numpy arrays that can be avoided with a comparably small code refactor.

The critical arrays are those that store information for each centroid and each track position. This PR reduces the number of such "critical arrays" from 16 to 8. The remaining critical arrays are:
```python
d_centr[:, :]
v_centr_normed[:, :, 0]
v_centr_normed[:, :, 1]
close_centr_msk[:, :]
v_ang_norm[:, :]
v_trans_corr[:, :]
windfields[:, :, 0]
windfields[:, :, 1]
```

For example, when there are 100 track positions, and 5,000,000 centroids, then the memory requirements of one such 64-bit float array is 4 GB. Hence, the total memory requirements are reduced from 64 GB down to 32 GB with this PR. (This is quite convenient on my cluster setup, where I have at most 64 GB of memory for each task.)

### DIFF
--- a/climada/hazard/trop_cyclone.py
+++ b/climada/hazard/trop_cyclone.py
@@ -627,16 +627,16 @@ def compute_windfields(track, centroids, model, metric="equirect"):
         return windfields, reachable_centr_idx
 
     # compute distances and vectors to all centroids
-    [d_centr], [v_centr] = u_coord.dist_approx(t_lat[None], t_lon[None],
-                                               track_centr[None, :, 0], track_centr[None, :, 1],
-                                               log=True, normalize=False, method=metric)
+    [d_centr], [v_centr_normed] = u_coord.dist_approx(
+        t_lat[None], t_lon[None], track_centr[None, :, 0], track_centr[None, :, 1],
+        log=True, normalize=False, method=metric)
 
     # exclude centroids that are too far from or too close to the eye
     close_centr_msk = (d_centr < CENTR_NODE_MAX_DIST_KM) & (d_centr > 1e-2)
     if not np.any(close_centr_msk):
         return windfields, reachable_centr_idx
-    v_centr_normed = np.zeros_like(v_centr)
-    v_centr_normed[close_centr_msk] = v_centr[close_centr_msk] / d_centr[close_centr_msk, None]
+    v_centr_normed[~close_centr_msk] = 0
+    v_centr_normed[close_centr_msk] /= d_centr[close_centr_msk, None]
 
     # make sure that central pressure never exceeds environmental pressure
     pres_exceed_msk = (t_cen > t_env)
@@ -646,8 +646,7 @@ def compute_windfields(track, centroids, model, metric="equirect"):
     t_rad[:] = estimate_rmw(t_rad, t_cen) * NM_TO_KM
 
     # translational speed of track at every node
-    v_trans = _vtrans(t_lat, t_lon, t_tstep, metric=metric)
-    v_trans_norm = v_trans[0]
+    [v_trans_norm, v_trans] = _vtrans(t_lat, t_lon, t_tstep, metric=metric)
 
     # adjust pressure at previous track point
     prev_pres = t_cen[:-1].copy()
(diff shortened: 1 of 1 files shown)


climada/hazard/trop_cyclone.py
```diff
@@ -627,16 +627,16 @@ def compute_windfields(track, centroids, model, metric="equirect"):
         return windfields, reachable_centr_idx
 
     # compute distances and vectors to all centroids
-    [d_centr], [v_centr] = u_coord.dist_approx(t_lat[None], t_lon[None],
-                                               track_centr[None, :, 0], track_centr[None, :, 1],
-                                               log=True, normalize=False, method=metric)
+    [d_centr], [v_centr_normed] = u_coord.dist_approx(
+        t_lat[None], t_lon[None], track_centr[None, :, 0], track_centr[None, :, 1],
+        log=True, normalize=False, method=metric)
```
tovogt, line 632: The array `v_centr` used to store the vectors from the track positions to the centroids. However, this array was never used, but only the normed version of it `v_centr_normed`. Only having a single array `v_centr_normed` instead reduces the number of critical arrays by 2 (since vectors have an x and a y component).

climada/hazard/trop_cyclone.py
```diff
@@ -646,37 +646,37 @@ def compute_windfields(track, centroids, model, metric="equirect"):
     t_rad[:] = estimate_rmw(t_rad, t_cen) * NM_TO_KM
 
     # translational speed of track at every node
-    v_trans = _vtrans(t_lat, t_lon, t_tstep, metric=metric)
-    v_trans_norm = v_trans[0]
+    [v_trans_norm, v_trans] = _vtrans(t_lat, t_lon, t_tstep, metric=metric)
```
tovogt, line 649: This does not affect the performance, but only makes the code a bit clearer.

climada/hazard/trop_cyclone.py
```diff
@@ -646,37 +646,37 @@ def compute_windfields(track, centroids, model, metric="equirect"):
...
 
     # adjust pressure at previous track point
     prev_pres = t_cen[:-1].copy()
     msk = (prev_pres < 850)
     prev_pres[msk] = t_cen[1:][msk]
 
     # compute b-value and derive (absolute) angular velocity
+    v_ang_norm = np.zeros((npositions, nreachable), dtype=np.float64)
```
tovogt, line 657: This does not affect the performance, but only avoids arrays with `npositions - 1` entries. It also improves the clarity of the code since the array is initialized before the `if ... else ...` block below.

climada/hazard/trop_cyclone.py
```diff
@@ -685,10 +685,8 @@ def compute_windfields(track, centroids, model, metric="equirect"):
     if np.count_nonzero(t_lat < 0) > np.count_nonzero(t_lat > 0):
         hemisphere = 'S'
     v_ang_rotate = [1.0, -1.0] if hemisphere == 'N' else [-1.0, 1.0]
-    v_ang_dir = np.array(v_ang_rotate)[..., :] * v_centr_normed[1:, :, ::-1]
-    v_ang = np.zeros_like(v_ang_dir)
-    v_ang[close_centr_msk[1:]] = v_ang_norm[close_centr_msk[1:], None] \
-                                 * v_ang_dir[close_centr_msk[1:]]
+    windfields = np.array(v_ang_rotate)[..., :] * v_centr_normed[:, :, ::-1]
+    windfields[close_centr_msk] *= v_ang_norm[close_centr_msk, None]
```
tovogt, line 689: Directly writing to the final `windfields` array avoids the temporary arrays `v_ang_dir`, `v_ang` and `v_full` (see below). That means we reduce the number of critical arrays by 6 (each of those arrays stores vectors, so has x and y component).

tovogt (commented):
I added comments to clarify which changes affect those "critical arrays".

emanuel-schmid:
🙌 Thanks a lot for saving! Looks very good. 👍

PR #436, opened 2022-05-03
Feature/refactor impact calc

chahank:
This is a draft pull request to refactor the impact class. It now consists of two classes: `Impact` and `ImpactCalc`. The `Impact` class has a proper `__init__`. Back-compatibility is 100% kept so far.


tovogt (commented):
Regarding the performance: Adding the ImpactCalc class as an intermediate didn't affect the performance at all for my use cases. So that's fine.

Overall, the situation of code structure and clarity improved with the new ImpactCalc class. However, I feel like it's still more fragmented than it should be. Why do we have the `from_imp_mat` and `from_imp_metrics` as part of the `Impact` class? I would say that we don't expect anyone to use this outside of `ImpactCalc`. So, I would make this functionality a part of `ImpactCalc` from the start. I would also incorporate the remaining top-level functions (risk_metrics_from_mat etc.) into `ImpactCalc`. Furthermore, `imp_mat` and `imp_mat_list` could be attributes of `ImpactCalc`, so that it would also make sense to remove the `@staticmethod` decorators. However, I understand that you might not want to convert all the function arguments into attributes because it can be considered better for clarity when you can see from a function's signature which data serves as an input and which output is generated. Still, in a context where each function is called only once or twice and is not expected to be used by external users, and where there are no individual unit tests for each function, it is probably okay to access all input/output through attributes of `self`.

Furthermore, I'm wondering whether we really want the user to start dealing with `ImpactCalc` objects that they most commonly won't need for their purposes, but they would discard it after generating the `Impact` object (which is what they want, ultimately). I would prefer to have some constructor `Impact.from_hazard_exposure` that internally generates an `ImpactCalc` object etc so that the user isn't bothered with these internals.

Eventually, I think that the impact calculation should be a one liner for the typical user `imp = Impact.from_something(arg1, arg2, arg3)`.

chahank:
> Overall, the situation of code structure and clarity improved with the new ImpactCalc class. However, I feel like it's still more fragmented than it should be. Why do we have the from_imp_mat and from_imp_metrics as part of the Impact class?

Excellent question. Actually, quite a few classes will use that without using the `ImpactCalc` class at the moment. Basically, the `lines_polygons` methods, the `multi-hazard` methods, and probably the `CostBenefit` methods once refactored.

chahank:
> Furthermore, imp_mat and imp_mat_list could be attributes of ImpactCalc, so that it would also make sense to remove the @staticmethod decorators. However, I understand that you might not want to convert all the function arguments into attributes because it can be considered better for clarity when you can see from a function's signature which data serves as an input and which output is generated. Still, in a context where each function is called only once or twice and is not expected to be used by external users, and where there are no individual unit tests for each function, it is probably okay to access all input/output through attributes of self.

Interesting points. However, I would like to be able to obtain an impact matrix without creating an impact object for use in other classes like `lines_polygons`, `multi-impact`, `cost-benefit` and possibly more in the future.

chahank:
> Regarding the performance: Adding the ImpactCalc class as an intermediate didn't affect the performance at all for my use cases. So that's fine.

Great!

tovogt:
> I agree with that. This is why the current implementation is 100% backwards compatible. The user can simply use the `Impact.calc()` as before. Is this not enough?

We want to get rid of these kinds of functions because they have this two-step anti-pattern of first initializing an empty object and then overwriting its contents.

> Excellent question. Actually, quite a few classes will use that without using the `ImpactCalc` class at the moment. Basically, the `lines_polygons` methods, the `multi-hazard` methods, and probably the `CostBenefit` methods once refactored.

Good to know, but then it might still make sense to have these functions be part of `ImpactCalc` instead of `Impact`.

> Interesting points. However, I would like to be able to obtain an impact matrix without creating an impact object for use in other classes like `lines_polygons`, `multi-impact`, `cost-benefit` and possibly more in the future.

That would be a perfect use case for the `ImpactCalc` class and it makes it even more useful to have `imp_mat` as an attribute of that class. The classes you mentioned could generate an instance of ImpactCalc and use its methods and attributes (like imp_mat) for their purposes.

chahank:
Sounds all good indeed. Thanks for the feedback. If you have a concrete implementation already in mind, do you have to time to do it? Otherwise, I will try it.

chahank:
This might also be a good moment to get further feedback. @emanuel-schmid could you please have a look and give some further feedback?

climada/engine/impact.py
```diff
@@ -19,7 +19,7 @@
 Define Impact and ImpactFreqCurve classes.
 """
 
-__all__ = ['ImpactFreqCurve', 'Impact']
+__all__ = ['ImpactFreqCurve', 'Impact', 'ImpactCalc']
```
peanutfun, line 22: I think this file is too large. I suggest splitting it into two files `impact.py` and `impact_calc.py` inside a submodule `engince/impact/`. The respective `__init__.py` file could already load the resepctive classes, so that this change is fully backwards compatible:
```python
# climada/enginve/impact/__init__.py

from .impact import Impact, ImpactFreqCurve
from .impact_calc import ImpactCalc
```
Usage:
```python
# my_script.py

from climada.engine.impact import Impact, ImpactCalc
```

climada/engine/impact.py
```diff
@@ -48,6 +47,410 @@
...
+            matrix num_events x num_exp with impacts.
+            Default is an empty matrix.
+
+        Returns
+        -------
+        None.
+
+        """
+
+        imp_mat = sparse.csr_matrix(np.empty((0, 0))) if imp_mat is None else imp_mat
```
peanutfun, line 84: I think setting `imp_mat` to `None` instead of an empty zero-size matrix would be a much better choice here (and in the `Impact` class). There are several instances where `imp_mat` is checked for a sensible content, but nothing beats `if imp_mat is None`, imho.
```suggestion
```

climada/engine/impact.py
```diff
@@ -48,6 +47,410 @@
...
+            apply deductible to impact matrix
+        apply_cover_to_mat:
+            apply cover to impact matrix
+        """
+        impf_col = self.exposures.get_impf_column(self.hazard.haz_type)
+        exp_gdf = self.minimal_exp_gdf(impf_col)
+        LOGGER.info('Calculating impact for %s assets (>0) and %s events.',
+                    exp_gdf.size, self.hazard.size)
+
+        if self.imp_mat.size == 0:
```
peanutfun, line 187: See my other comment
```suggestion
        if self.imp_mat is None:
```

climada/engine/impact.py
```diff
@@ -84,164 +487,157 @@ class Impact():
...
+        self.date = np.array([], int) if date is None else date
+        self.coord_exp = np.ndarray([], float) if coord_exp is None else coord_exp
+        self.crs = crs
+        self.eai_exp = np.array([], float) if eai_exp is None else eai_exp
+        self.at_event = np.array([], float) if at_event is None else at_event
+        self.frequency = np.array([],float) if frequency is None else frequency
+        self.tot_value = tot_value
+        self.aai_agg = aai_agg
+        self.unit = unit
+        self.imp_mat = sparse.csr_matrix(np.empty((0, 0))) if imp_mat is None else imp_mat
```
peanutfun, line 517: Same as for `ImpactCalc`:
```suggestion
        self.imp_mat = imp_mat
```

climada/engine/impact.py
```diff
@@ -270,52 +666,131 @@ def calc_risk_transfer(self, attachment, cover):
...
+        -------
+        np.array
+        """
+        LOGGER.info('Computing exceedance impact map for return periods: %s',
+                    return_periods)
+        try:
+            self.imp_mat.shape[1]
+        except AttributeError as err:
+            raise ValueError('attribute imp_mat is empty. Recalculate Impact'
+                             'instance with parameter save_mat=True') from err
```
peanutfun, line 735: See other comments
```suggestion
        if self.imp_mat is None:
            raise ValueError('attribute imp_mat is empty. Recalculate Impact'
                             'instance with parameter save_mat=True')
```

climada/engine/impact.py
```diff
@@ -48,6 +47,410 @@
...
+        Initialize an ImpactCalc object.
+
+        The dimension of the imp_mat variable must be compatible with the
+        exposures and hazard objects.
+
+        Parameters
+        ----------
+        exposures : climada.entity.Exposures
+            exposure used to compute imp_mat
+        impf_set: climada.entity.ImpactFuncSet
```
peanutfun, line 70: `impfset` is actually written without underscore, I'm not sure what the intended notation is
```suggestion
        impfset: climada.entity.ImpactFuncSet
```

climada/engine/impact.py
```diff
@@ -84,164 +487,157 @@ class Impact():
...
+            #      "please use Impact.calc_insured_risk")
+            self.__dict__ = impcalc.insured_impact(save_mat).__dict__
+        else:
+            # LOGGER.warning("The use of Impact.calc() is deprecated."
+            #      "Please use Impact.calc_risk() or Impact.calc_risk_insured().")
+            self.__dict__ = impcalc.impact(save_mat).__dict__
+
+#TODO: new name
+    @classmethod
+    def from_eih(cls, exposures, impfset, hazard,
```
peanutfun, line 535: There were already several suggestions, so I'll just add another one for the name of this method:
`from_exp_impf_haz` 😅

climada/engine/impact.py
```diff
@@ -84,164 +487,157 @@ class Impact():
...
-        # Sort descendingly the impacts per events
-        sort_idxs = np.argsort(self.at_event)[::-1]
-        # Calculate exceedence frequency
-        exceed_freq = np.cumsum(self.frequency[sort_idxs])
-        # Set return period and imact exceeding frequency
-        ifc.return_per = 1 / exceed_freq[::-1]
-        ifc.impact = self.at_event[sort_idxs][::-1]
-        ifc.unit = self.unit
-        ifc.label = 'Exceedance frequency curve'
+        imp_mat = sparse.csr_matrix(np.empty((0, 0))) if imp_mat is None else imp_mat
```
peanutfun, line 556: Same as above, just leave it as `None`.
```suggestion
```

climada/engine/impact.py
```diff
@@ -48,6 +47,410 @@
...
+            self.imp_mat = self.stitch_impact_matrix(imp_mat_gen)
+            at_event, eai_exp, aai_agg = self.risk_metrics()
+        else:
+            at_event, eai_exp, aai_agg = self.stitch_risk_metrics(imp_mat_gen)
+        return Impact.from_eih(
+            self.exposures, self.impfset, self.hazard,
+            at_event, eai_exp, aai_agg, self.imp_mat
+            )
+
+    def insured_impact(self, save_mat=False):
```
peanutfun, line 154: This method shares a lot of instructions with `impact()`. I think it would be best to separate these parts into one or multiple private methods to avoid repeated code.

climada/engine/impact.py
```diff
@@ -48,6 +47,410 @@
...
+        """
+        impf_col = self.exposures.get_impf_column(self.hazard.haz_type)
+        exp_gdf = self.minimal_exp_gdf(impf_col)
+        LOGGER.info('Calculating impact for %s assets (>0) and %s events.',
+                    exp_gdf.size, self.hazard.size)
+
+        if self.imp_mat.size == 0:
+            imp_mat_gen = self.imp_mat_gen(exp_gdf, impf_col)
+        else:
+            imp_mat_gen = ((self.imp_mat, self.exposures.gdf.index.values) for n in range(1))
```
peanutfun, line 190: I'm not sure if this is on purpose but the definition of the impact matrix (generator) here is different from the one in `impact()`. I don't unterstand the matrix creation well enough to see if this is an issue. Consider specifically the use case where somebody calls the two methods after each other, with different settings for `save_mat`. Why would the calculation change if `self.imp_mat` already exists?

climada/engine/impact.py
chahank: I am not sure whether this is a good idea. Having changes in types for attributes makes it hard for methods to work with it, as you then always have to test for None. Furthermore, this would represent a change in the `Impact` class that is not fully backwards compatible.

What would be gained from having `None` instead of an empty matrix that outways the disadvantages?

climada/engine/impact.py
chahank: Agreed, that is planed.

climada/engine/impact.py
chahank: I don't have strong feelings about what is best. Note that this is a change compared to the previous code where `impact_funcs` was used. I found this very confusing since we explicitly require a set and not a single function. Thus either `impf_set` or `impfset`. What do you prefer?

climada/engine/impact.py
chahank: How would you do that? There are already separated into the simplest methods I could think of. Except maybe the blob from `if save_mat` on. How would you call such a method?

climada/engine/impact.py
chahank: This is a bit long for my taste :D But I have no strong opinion here either ^^.

climada/engine/impact.py
peanutfun: And already done, as far as I see. Thanks! 😊

climada/engine/impact.py
peanutfun: I think I prefer `impf_set`

climada/engine/impact.py
peanutfun: Great, exactly what I was thinking of 👌

climada/engine/impact.py
peanutfun: I guess I know understand the generator-thing better, but I still have two more questions 😅 

> Thus, if the impact matrix is already computed, we would like to avoid computing it again. 

But shouldn't the same logic apply for the `impact()` method as well? It currently always re-computes the impact matrix, unlike `insured_impact()`.

> The save_mat flag is then simple. Either the ouput Impact object with the insured impact has the imp_mat defined or not. This is independent on whether the imp_mat is defined for the ImpactCalc object.

Note that`save_mat` also determines if the matrix will be stored in the `ImactCalc` object. Is that an over-complication of the interface? If the matrix is computed anyway (and occupies memory during the computation), why not always store (i.e. keep) it?

climada/engine/impact.py
peanutfun: I replaced the original default argument with an empty matrix and adapted all checks to verify that `imp_mat.size > 0` in e0a6a0adb4c4f0dfb958a331ee333746a9ce9cf4

climada/engine/impact_calc.py
```diff
@@ -40,7 +40,7 @@ def __init__(self,
                  exposures,
                  impfset,
                  hazard,
-                 imp_mat=None):
+                 imp_mat=sparse.csr_matrix((0, 0))):
```
emanuel-schmid, line 43: A sparse.csr_matrix is mutable - right?
Hence the default argument should still be None and then reassigned first thing within the method.

chahank:
@emanuel-schmid : how sensible do you think are the consistency test for the `Impact` object initialization 01e4e272fd6b6206d56fd34003030d81fdecd513 ?

climada/engine/impact.py
```diff
@@ -95,7 +95,7 @@ def __init__(self,
                  tot_value=0,
                  aai_agg=0,
                  unit='',
-                 imp_mat=None,
+                 imp_mat=sparse.csr_matrix((0, 0)),
```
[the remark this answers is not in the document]
peanutfun: Indeed. I quickly fixed it again in 90536e4e96fd67756fcfc0584e4e5a8efcf02d46

climada/engine/impact_calc.py
peanutfun: See 90536e4e96fd67756fcfc0584e4e5a8efcf02d46

climada/engine/test/test_impact_calc.py
```diff
@@ -0,0 +1,407 @@
...
+            {
+                "impact_functions": [0, 11, 11],
+                "centr_col": [0, 10, 20],
+                "value": [0.0, 1.0, 2.0],
+            }
+        )
+
+    def tearDown(self):
+        """Reset the original config"""
+        CONFIG = self.CONFIG_COPY
```
peanutfun, line 341: Some other tests are currently failing with the error message
```
ValueError: Increase max_matrix_size configuration parameter to > 216
```
I guess this is due to this line not working correctly. I wanted to reset the global `CONFIG` after altering it for this test. Not sure what to do here...

peanutfun: @chahank @emanuel-schmid I need some help here. The problem seems to be that I cannot `deepcopy` the `CONFIG` object.

peanutfun: I just came up with the solution: Replacing `CONFIG.max_matrix_size` with a new `Config` instance.

The problem is not that `climada.CONFIG` cannot be set (it can), but that the `CONFIG_COPY` still points to the original `CONFIG` that I changed. It's not a real copy.

emanuel-schmid: No it doesn't. 😝 Makes sense too. One needs to re-import CONFIG from climada to make it work.

peanutfun: See 2a1122f9e18cecb6a8d8d43f8d104f06a0314ddf for the solution (i hope 🙏)

emanuel-schmid: Should work (still optimistic). 😄 
However I think deepcopy _does_ work but porting it into the namespace is tricky.

peanutfun: It works now, so I'll resolve this discussion. Thanks a lot for your input! 🙌

climada/engine/impact_calc.py
```diff
@@ -220,7 +216,10 @@ def minimal_exp_gdf(self, impf_col):
             name of the impact function column in exposures.gdf
 
         """
-        self.exposures.assign_centroids(self.hazard, overwrite=False)
+        # since the original exposures object will be "spoiled" through assign_centroids
+        # (a column cent_XY is added), overwrite=True makes sure the exposures can be reused
+        # for ImpactCalc with another Hazard object of the same hazard type.
+        self.exposures.assign_centroids(self.hazard, overwrite=True)
```
chahank, line 222: Why would you change that to True? This is a bad idea as this results in a lot of extra computation time, often for nothing. A typical workflow implies that the Exposures object given to ImpactCalc already has assigned centroids. This line here is just in case that none are defined (which is a sub-optimal workflow, but needs to stay due to backwards compatibility).

Thus, this should be reversed.

emanuel-schmid: I see. Then we have the following problem:

```python
agg1 = ImpactCalc(exp, impf, haz1).impact().aai_agg
agg2 = ImpactCalc(exp, impf, haz2).impact().aai_agg
print(agg1 - agg2)
```

does i.g. not print the same as 

```python
agg2 = ImpactCalc(exp, impf, haz2).impact().aai_agg
agg1 = ImpactCalc(exp, impf, haz1).impact().aai_agg
print(agg1 - agg2)
```

Simply, because in the first example `agg2` is likely wrong and in the second `agg1`.

tovogt: Good point Emanuel, I remember that I stumbled upon this before!

I would be in favor of moving the whole centroids assignment procedure to the ImpactCalc object because semantically it's not a property of the Exposures object, but of the combination of an exposure with a hazard. However, I see the problem that it might be desirable to be able to reuse the same assignment with several different Hazard objects that share the same Centroids... And of course, there is the problem of backwards compatibility when people are explicitly using the `Exposures.assign_centroids` API.

Maybe the easiest solution is to leave the API as it is, but have a boolean keyword argument `reassign_centroids=True` for `ImpactCalc.impact()` so the user can opt out of the reassignment if that's desired?

emanuel-schmid: Me too actually, I'd favor moving the centroids assignments out of Exposures. Also I'm not yet ready to go for the easiest solution, as this looks like it may be crucial.

Two questions:

- What _is_ a typical workflow?
- When two Hazard objects share the same Centroids, are they "aware" of it, i.e., do they share the same Centroids object or do their Centroids just have the same content and only the user knows that they are the same?

chahank: This is a rather complex discussion that will be very hard to follow clearly on a chat forum like here. But, a few remarks:
- the current API and workflow is clear: the user is responsible for making sensible hazard-exposures-impact functions assignments. This includes assigning for a given hazard type the centroids to the exposures. The same goes true for assigning the correct impact functions to the exposures points. We can modify this of course, but this would be a major change in API and workflow
- we 100% do NOT want to assign centroids for each impact computation as this is a step that needs to taken once only per exposures-hazard pair. We do not want to repeat this rather expensive operation (e.g. in the uncertainty computations).
- two hazards are not aware that they have the same centroids. It is unclear to me how that would work, as the idea is that any user can define their hazard (and thus their centroids) from whatever source they want and bring it into CLIMADA for computation. 
- My opinion would be that actually the `assign_centroids` is entirely removed from the `Impact.calc` as it is an under the hood assignment which we would like the user to have control over (for instance, does one always want to use nearest-neighbors?). This way, the user must think and prepare consistent data. 

I also agree that a quick solution is not optimal. The current solution is the one that keeps the same convention as in the `Impact.calc()` method.

chahank: My suggestion would be to decouple the two discussion. This PR is about making the impact class more readable, more modular, better tested and more efficient. It is not about changing the inner workings of the API fundamentally as is here suggested with the change with the `assign_centroids`.

PR #438, opened 2022-05-05
Bugfix/unsequa helper

chahank:
Fix a bug when events are sub-sampled by name which might not be unique. Events are now sub-sampled by id. For this, the method `hazard.select` now allows for `event_id` selection.


manniepmkam:
The updates works fine for my case study of TC displacement forecast. Ok to merge.

PR #441, opened 2022-05-06
Add instructions for committing Jupyter Notebook changes

peanutfun:
This adds a section to the "Git Development" tutorial on how to avoid excessive Git Diffs when changing Jupyter Notebook files. When editing Jupyter Notebooks through a renderer, metadata or entire image files that is/are unrelated to the actual code changes might be modified. Git reports these modifications as changes, which are hard to read as soon as they are committed. The only remedy I see for now is to open the notebooks with a text editor and modify the code then. Of course, the overall layout of the notebooks makes this very inconvenient.

I am unsure if this is the right place to put this information. Suggestions on how to convey this informatin clearly are very welcome!

As a side note, I found [ReviewNB](https://www.reviewnb.com/), a plugin for GitHub that might ease these problems.


emanuel-schmid:
👍 Thank you!

PR #447, opened 2022-05-10
Feature/tutorial engine forecast update

frqqyy:
Changes proposed in this PR:
- The second block will fail if there is not a folder named "hazards" under the path "...\climada_main\doc\tutorial\results\forecast", so I added a line to create this folder
- (In fact, I think it might be better to directly fix the function "generate_WS_forecast_hazard()", but I'm not sure if there is a plan to update this tutorial or this class)

This PR fixes (not an issue, but rather a failed test result): https://ied-wcr-jenkins.ethz.ch/job/climada_notebooks/lastBuild/testReport/junit/(root)/NBTest/climada_engine_Forecast/

### Pull Request Checklist

- [x] Source branch up-to-date with the target branch
- [ ] Docs updated: I don't think that there are any docs that should be updated, but please correct me if there are some.
- [ ] Tests updated: I don't think that there are any tests that should be updated, but please correct me if there are some.
- [ ] Tests passing: Needs to be found out
- [ ] No new linter issues: Needs to be found out


emanuel-schmid:
> (In fact, I think it might be better to directly fix the function "generate_WS_forecast_hazard()", but I'm not sure if there is a plan to update this tutorial or this class)

👍 I've brushed the method up a bit. It should work now. For the task at hand - make the notebook executable without failures - that's sufficient. But that method is poorly tested and could still be improved by disentangling reading, calculation and writing from each other.

frqqyy:
> 👍 I've brushed the method up a bit. It should work now. For the task at hand - make the notebook executable without failures - that's sufficient. But that method is poorly tested and could still be improved by disentangling reading, calculation and writing from each other.

Thanks a lot! Should I close this pull request then?

PR #448, opened 2022-05-10
Update climada_installation_step_by_step.ipynb

frqqyy:
Changes proposed in this PR:
- Changed the type of all the code blocks to markdown because otherwise, they will fail the tests

This PR fixes (not an issue, but a failed test result): https://ied-wcr-jenkins.ethz.ch/job/climada_notebooks/lastBuild/testReport/junit/(root)/NBTest/climada_installation_step_by_step/

### Pull Request Checklist

- [ ] Docs updated: I don't think that there are any docs that should be updated, but please correct me if there are some.
- [ ] Tests updated: I don't think that there are any tests that should be updated, but please correct me if there are some.
- [x] Tests passing: Jenkins Build is successful. Tests / Declarative: Post Actions are passed.
- [x] No new linter issues

### DIFF
--- a/doc/tutorial/climada_installation_step_by_step.ipynb
+++ b/doc/tutorial/climada_installation_step_by_step.ipynb
@@ -53,9 +53,11 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "id": "e4cf88ff",
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "74d80546-3e1a-4a59-8630-183d830b9c01",
    "metadata": {},
+   "outputs": [],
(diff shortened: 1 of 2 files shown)


emanuel-schmid:
In this particular case, I'd rather exclude the notebook from the files being tested, because the code cells nicely jut out and are clearly recognizable as command lines in the doc-files.

frqqyy:
> In this particular case, I'd rather exclude the notebook from the files being tested, because the code cells nicely jut out and are clearly recognizable as command lines in the doc-files.

I see, thank you very much!

Issue #415, opened 2022-04-11
A user warning in the litpop tutorial: Geometry is in a geographic CRS.


frqqyy:
Although multiple user warnings are shown in the screenshots, only the first one should be focused here.
I get the following outputs and warnings if I run this block:
![litpop 1 1](https://user-images.githubusercontent.com/71314575/162840249-2e9e53b9-8f8d-426c-b4bf-c552bb5a9a55.png)
![litpop 1 2](https://user-images.githubusercontent.com/71314575/162840258-63506f12-3050-4c3c-b14e-e9d81de16200.png)

frqqyy:
![litpop 1 3](https://user-images.githubusercontent.com/71314575/162840289-6d29dc96-be36-46b7-96ee-44330fc35aa4.png)

frqqyy:
Although this particular user warning has not been resolved yet, another user warning (You will likely lose important information) has already been suppressed by PR #459

peanutfun:
@emanuel-schmid What's the use of the "fixed" label? Can we close this issue?

Issue #413, opened 2022-04-11
A user warning in the tutorial about polygons and lines


frqqyy:
Although there are two UserWarnings shown in the pictures, only the first user warning should be further examined.
I get the following outputs and warnings if I run this block:
![1 1](https://user-images.githubusercontent.com/71314575/162752361-629869e1-e025-4fd8-b4d9-cd343af04956.png)
![1 2](https://user-images.githubusercontent.com/71314575/162752372-d025a098-6317-4760-a196-b9916444cf72.png)

PR #455, opened 2022-05-25
plot.get_transformation: suppress UserWarning about losing information

emanuel-schmid:
Changes proposed in this PR:
- Suppress UserWarning about losing "important projection information" when converting to a projection string
- To do that the whole get_transformation function is set in a `catch_warnings` block.

This PR addresses issues #413 and #415

### Pull Request Checklist

- [ ] Docs updated
- [ ] Tests updated
- [x] Tests passing
- [x] No new linter issues (the alleged new issues were really just overruled by higher severity issues)

### DIFF
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -738,7 +738,7 @@ def get_transformation(crs_in):
     units : str
     """
     with warnings.catch_warnings():
-    # Within this block we want to suppress this UserWarning from pyproj._crs._CRS.to_proj4:
+    # Within this function we want to suppress the UserWarning from pyproj._crs._CRS.to_proj4:
     # You will likely lose important projection information when converting to a PROJ string from
     # another format. See:
     # https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
@@ -752,8 +752,7 @@ def get_transformation(crs_in):
         except ValueError:
             crs_epsg = ccrs.PlateCarree()
         except requests.exceptions.ConnectionError:
-            LOGGER.warning('No internet connection.'
-                        ' Using projection PlateCarree in plot.')
+            LOGGER.warning('No internet connection. Using projection PlateCarree in plot.')
             crs_epsg = ccrs.PlateCarree()
 
         try:
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -18,6 +18,7 @@ with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
 
 Define auxiliary functions for plots.
 """
+# pylint: disable=abstract-class-instantiated
 
 __all__ = ['geo_bin_from_array',
            'geo_im_from_array',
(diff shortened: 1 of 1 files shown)


peanutfun:
Is there nothing we can do to resolve the warning itself? Losing "important projection information" sounds somewhat serious 😅

climada/util/plot.py
```diff
@@ -736,31 +737,37 @@ def get_transformation(crs_in):
...
+        try:
+            if CRS.from_user_input(crs_in) == CRS.from_user_input('EPSG:3395'):
+                crs_epsg = ccrs.Mercator()
+            else:
+                crs_epsg = ccrs.epsg(CRS.from_user_input(crs_in).to_epsg())
+        except ValueError:
+            crs_epsg = ccrs.PlateCarree()
+        except requests.exceptions.ConnectionError:
+            LOGGER.warning('No internet connection. Using projection PlateCarree in plot.')
+            crs_epsg = ccrs.PlateCarree()
```
peanutfun, line 756: I suggest setting the overall fallback value first (this avoids setting it twice in differenct `catch` statements), and adding a warning in case of a parsing error:
```suggestion
        crs_epsg = ccrs.PlateCarree()  # Fallback value
        try:
            if CRS.from_user_input(crs_in) == CRS.from_user_input('EPSG:3395'):
                crs_epsg = ccrs.Mercator()
            else:
                crs_epsg = ccrs.epsg(CRS.from_user_input(crs_in).to_epsg())
        except ValueError:
            LOGGER.warning(f"Error parsing coordinate system '{crs_in}'."
                "Using projection PlateCarree in plot.")
        except requests.exceptions.ConnectionError:
            LOGGER.warning('No internet connection. Using projection PlateCarree in plot.')
```

peanutfun (changes requested):

climada/util/plot.py
emanuel-schmid: Thanks for the suggestions!

I agree to the warning, but to the prior fallback value setting I'm actually quite opposed. In general it does not help with readability and it does not help with avoiding mistakes. The only advantage I can see is that it saves some lines of code. While in the mean time code will be executed unnecessarily.

peanutfun: Fair enough. I wanted to avoid doubled lines of code but I see your point. I'll resolve this conversation and commit the warning separately. Then I'll merge ✌️

peanutfun:
@emanuel-schmid, I'm not familiar with the projections used in CLIMADA. Could you quickly comment on why we suppress the warning? Is there no way to resolve the underlying problem? (c.f. [my comment above](https://github.com/CLIMADA-project/climada_python/pull/455#issuecomment-1138517669))

emanuel-schmid:
> Is there nothing we can do to resolve the warning itself? Losing "important projection information" sounds somewhat serious 😅

True - but having exotic projections where essential information is lost for good is probably anyways out of scope for climada.

Issue #457, opened 2022-05-30
write/to_hdf5 methods increase size of existing files


emanuel-schmid:
When `Exposures.write_hdf5` is called with a path to an existing file, the result is not the same as when it's called with a path that a priori leads nowhere. Calling it several times in a row leads to a steadily increased file size.
Looks like the `pandas.HDFStore` object is not by default in the mood for cleaning up and freeing diskspace.

Therefore I suggest to remove any existing file before calling `Exposures.write_hdf5` or `UncOutput.to_hdf5`

chahank:
I think by default, if the file already exists, it appends to it whatever is being written. Instead of removing files by default, should there be a flag? Or maybe an error if the file already exists? I am always a bit worried when files get overwritten by default without warning.

emanuel-schmid:
Yes there is a flag, see #458. 
I'm not really worried about overwriting an existing file. None of the standard file writing methods (e.g. stdio, pandas, ...) do that and we'd have a hard time to raise such warnings consequently.

peanutfun:
Resolved by #458

PR #458, opened 2022-05-30
HDFStore always initialize with explicit mode

emanuel-schmid:
Changes proposed in this PR:
- explicitly set the mode for `pandas.HDFStore` everywhere to 'w' or 'r'. The default is 'a' which causes files to continuously grow in size when `write_hdf5` is called for the same file again.

This PR fixes issue #457

### Pull Request Checklist

- [ ] Docs updated
- [ ] Tests updated
- [x] Tests passing
- [x] No new linter issues

### DIFF
--- a/climada/engine/unsequa/unc_output.py
+++ b/climada/engine/unsequa/unc_output.py
@@ -1013,7 +1013,7 @@ class UncOutput():
         save_path = save_path.with_suffix('.hdf5')
 
         LOGGER.info('Writing %s', save_path)
-        store = pd.HDFStore(save_path)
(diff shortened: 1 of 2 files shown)


chahank:
The docstring could be updated to reflect this change maybe?

Is the `Exposure` class the only one with this behavior? The `Hazard` class seems to already use the `w` mode. I didn't check all the others. 

Thanks!

emanuel-schmid:
At least for exposures, the behavior has not really changed, for the user the result should be exactly the same - just the files do not grow anymore when the method is called twice.

And yes, all instances of `HDFStore` initializations have been updated throughout climada_python.

PR #469, opened 2022-06-13
LitPop: Fix GPW download path

tovogt:
Changes proposed in this PR:
- Fix the download URL in the error output of LitPop when the required GPW file cannot be found.
- Some minor code style improvements in the LitPop/GPW code.

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated

### DIFF
--- a/climada/entity/exposures/litpop/gpw_population.py
+++ b/climada/entity/exposures/litpop/gpw_population.py
@@ -31,11 +31,11 @@ LOGGER = logging.getLogger(__name__)
 
 
 def load_gpw_pop_shape(geometry, reference_year, gpw_version,
-                       data_dir=SYSTEM_DIR, layer=0, verbatim=True):
+                       data_dir=SYSTEM_DIR, layer=0, verbose=True):
     """Read gridded population data from TIFF and crop to given shape(s).
 
     Note: A (free) NASA Earthdata login is necessary to download the data.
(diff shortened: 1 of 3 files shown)


climada/entity/exposures/litpop/gpw_population.py
```diff
@@ -112,49 +115,39 @@ def get_gpw_file_path(gpw_version, reference_year, data_dir=SYSTEM_DIR, verbatim
...
-                              + 'viewer.')
-    raise FileExistsError(f'The file {file_path} could not '
-                          + 'be found. Please download the file '
-                          + 'first or choose a different folder. '
-                          + 'The data can be downloaded from '
-                          + 'http://sedac.ciesin.columbia.edu/'
-                          + 'data/collection/gpw-v4/sets/browse, '
-                          + 'e.g., https://sedac.ciesin.columbia.edu/downloads/data/'
-                          + f'gpw-v4/gpw-v4-population-count-rev{gpw_version}/'
-                          + f'gpw-v4-population-count-rev{gpw_version}_2020_30_sec_tif.zip '
```
tovogt, line 158: This is the main reason for this PR: Before this PR, the download URL in the error message would always point to the 2020 data of GPW instead of the year that is actually missing.

emanuel-schmid:
🙌

PR #471, opened 2022-06-16
u_coord.dist_approx: fix several points' distance to single reference point

chahank:
Changes proposed in this PR:
- Change one line from `np.concatenate` to `np.stack` in `util.coordinate.dist_approx`. This allows for computing dist_approx also between an array of points and a single point.
- Test were not changed.

This PR fixes issue #

### Pull Request Checklist

- [x ] Correct target branch selected (if unsure, select `develop`)
- [ ] Docs updated
- [ ] Tests updated
- [ ] Tests passing
- [ ] No new linter issues


tovogt:
I don't understand what you are trying to achieve, but maybe there is a misunderstanding about the syntax of the `dist_approx` function. So I clarified the syntax a bit in the following commit: https://github.com/CLIMADA-project/climada_python/commit/c59b1d0a674293f0b5013d8af6d50138593018bb

This explanation might help:
https://github.com/CLIMADA-project/climada_python/blob/c59b1d0a674293f0b5013d8af6d50138593018bb/climada/util/coordinates.py#L250-L257

chahank:
The reason is that the following query will fail:

```
lat1, lon1 = np.array([[1.0, 2.0]]), np.array([[1.0, 2.0]])
lat2, lon2 = np.array([[0.0]]), np.aarray([[0.0]])
dist_approx(lat1, lon1, lat2, lon2)
```

with 

```
*** ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 2 and the array at index 1 has size 1
```

So, basically, the method fails to compute the distance from all points to one single point (`k=1`, `i>1` `j=1`).

Changing to `hstack` solves the problem, and no tests fail. 

However, I maybe did not think long enough whether it is the right thing to do. In any case, the case with `j=1` should be addressed and probably the tests would have to be updated (if they did not fail when they should have).

chahank:
The update of the docstring is very good, thanks!

tovogt:
The mentioned commit 3f6ab8143a3503443f4803add08aaf3826886b45 also fixes the issue you mentioned and adds a regression test for it. The fix looks like this:
https://github.com/CLIMADA-project/climada_python/blob/3f6ab8143a3503443f4803add08aaf3826886b45/climada/util/coordinates.py#L309
And the new test case is this one:
https://github.com/CLIMADA-project/climada_python/blob/3f6ab8143a3503443f4803add08aaf3826886b45/climada/util/test/test_coordinates.py#L201-L231

tovogt:
I decided to merge my branch into this one. Feel free to review the changes again and merge this into develop if you are happy with the proposed changes.

chahank:
Great, thanks!

PR #484, opened 2022-06-20
Fixing the warning #413 in centr.py

frqqyy:
Apparently, the calculation of the buffer is usually done in the projected meters system because it takes a distance as an argument. So, we might need to project the coordinate system from geodetic coordinates to meters.

Changes proposed in this PR:
- Projected the coordinate system from geodetic coordinates to meters to calculate the buffer, then projected it back

This PR fixes issue #413 

### Pull Request Checklist

- [ ] Docs updated: No doc seems to need to be updated.
- [ ] Tests updated: No test seems to need to be updated. There is apparently no test for the relevant method
- [ ] Tests passing: Waiting to be found out.
- [ ] No new linter issues: Waiting to be found out.


climada/hazard/centroids/centr.py
```diff
@@ -819,7 +819,10 @@ def set_area_pixel(self, min_resol=1.0e-8, scheduler=None):
             res = np.abs(res).min()
         self.set_geometry_points(scheduler)
         LOGGER.debug('Setting area_pixel %s points.', str(self.lat.size))
-        xy_pixels = self.geometry.buffer(res / 2).envelope
+        xy_pixels = self.to_crs(crs=3857)
+        xy_pixels = self.geometry.buffer(res / 2)
+        xy_pixels = self.to_crs(crs=4326)
```
peanutfun, line 824: `Centroids` has no method `to_crs`. You need to call this on `geometry`. Also, what do the argument mean? Which coordinate system(s) have you chosen to project the data and why?
```suggestion
        xy_pixels = self.geometry.to_crs(crs=3857)
        xy_pixels = self.geometry.buffer(res / 2)
        xy_pixels = self.geometry.to_crs(crs=4326)
```

peanutfun (changes requested):

climada/hazard/centroids/centr.py
frqqyy: 3857 is apparently a crs that has the units meters, with which the buffer can be calculated. 4326 is our default crs, I suppose? But if I understood it wrong and it doesn't make sense, I'll just close this PR. Thank you for your suggestions!

chahank:
3857 is not a good projection to compute areas, although it is in meters.

frqqyy:
> 3857 is not a good projection to compute areas, although it is in meters.

Thank you for the advice! Do you have an idea what we should use?

climada/hazard/centroids/centr.py
```diff
@@ -819,7 +819,10 @@ def set_area_pixel(self, min_resol=1.0e-8, scheduler=None):
             res = np.abs(res).min()
         self.set_geometry_points(scheduler)
         LOGGER.debug('Setting area_pixel %s points.', str(self.lat.size))
-        xy_pixels = self.geometry.buffer(res / 2).envelope
+        xy_pixels = self.geometry.to_crs(crs=3857)
```
emanuel-schmid, line 822: perhaps define a constant?

climada/hazard/centroids/centr.py
```diff
@@ -819,7 +819,10 @@ def set_area_pixel(self, min_resol=1.0e-8, scheduler=None):
...
+        xy_pixels = self.geometry.buffer(res / 2)
+        xy_pixels = self.geometry.to_crs(crs=4326)
```
emanuel-schmid, line 824: I'd rather reset it to what it was before.

chahank:
Commit 6bd27f8de04f80f7a52b829f0f1235dc360bc471 is a proposal: keep the already defined as default crs `cea` if it is not geometric.

chahank:
The solution seems to pass all tests, although one is failing now, but it looks unrelated.

climada/hazard/test/test_storm_europe.py
```diff
@@ -127,21 +127,21 @@ def test_set_ssi(self):
         storms = StormEurope.from_footprints(WS_DEMO_NC)
 
         storms.set_ssi(method='dawkins')
-        ssi_dawg = np.asarray([1.44573572e+09, 6.16173724e+08])
+        ssi_dawg = np.asarray([0.18808288, 0.08014435])
```
emanuel-schmid, line 130: @ThomasRoosli : are the changed values ok?

ThomasRoosli: The large change in magnitude seems unexpected to me. I will try to investigate somewhen soon.

ThomasRoosli: So, I think there is a mistake in the proposed implementation and the tests in StormEurope should not be changed this drastically.
Here is a small code example:
```
from climada.hazard import StormEurope
from climada.util.constants import WS_DEMO_NC

storms = StormEurope.from_footprints(WS_DEMO_NC)
storms.centroids.area_pixel
# current: array([12042728.52494331, 12042728.52494315, 12042728.52494331, ..., 11142431.39844684, 11142431.39844654, 11142431.39844684])
# newly proposed in pr484: array([0.00152588, 0.00152588, 0.00152588, ..., 0.00152588, 0.00152588, 0.00152588])
```
The centroids created with the lines of code above have a resolution of about 4 km at the equator which would mean around 16000000 m^2 area per pixel at the equator. The function `set_area_pixel` aims to return the area per pixel in m*m and to me the magnitude looks somewhat reasonable in the current implementation. With the newly proposed changes the area is below 1 m^2 per pixel which is weird.

@emanuel-schmid : please revert the changes

chahank:
Is this pull request still active or shall we close it?

peanutfun:
Closing due to inactivity

PR #496, opened 2022-06-29
API Client Tutorial: add a paragraph about offline mode

emanuel-schmid:
Changes proposed in this PR:
- Add a description of how one can use climada in offline mode.

This PR fixes issue #

### Pull Request Checklist

- [x] Docs updated
- [ ] Tests updated
- [x] Tests passing
- [x] No new linter issues

### DIFF
--- a/climada.conf
+++ b/climada.conf
@@ -24,8 +24,7 @@
         "test_data": "{test_directory}/util/test/data"
     },
     "data_api": {
-        "supported_exposures_types": ["litpop", "crop_production", "base"],
-        "cache_enabled": true
+        "supported_exposures_types": ["litpop", "crop_production", "base"]
     },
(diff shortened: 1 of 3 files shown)


chahank:
I think the description is fine, but I still have a question: it is confusing why the API would not be able to retrieve data that has already been downloaded. This creates a very awkward situation. For instance, if a user used the API a few times, a lot of data might have been downloaded (easily several Gb). However, to be able to work offline, a user would have to re-run all these commands with the caching enable in order to be able to retrieve this data that is already on the computer. Why not put the cache on by default? It seems that this would add at most a very small amount of stored data (a few mb at most).

emanuel-schmid:
The cache is enabled now by default. The description has changed accordingly.

PR #500, opened 2022-06-30
Feature/wrapper_functions_api

zeliest:
Add a wrapper function to get centroids from the API 
change get_litpop_default() to get_litpop()
add more information to error message when no dataset are found in the wrapper functions
update notebook


climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
             raise ValueError("country must be string or list of strings")
         return self.get_exposures(exposures_type='litpop', dump_dir=dump_dir, properties=properties)
 
+    def get_centroids(self, res_arcsec_land=150, res_arcsec_ocean=1800, extent=None, reg_id=None, country_iso3alpha=None,
```
chahank, line 847: I am not sure whether the attribute `reg_id` is very sensible here. This is a number that cannot really be known by a user of the API right? Anyway, the `reg_id` is the same as the `country_iso3alpha` correct? I would thus keep only one of the two, and only the one that is really understandable by a user.

climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+
+        Parameters
+        ----------
+        res_land_arcsec : int
+            resolution for land centroids in arcsec. Default is 150
+        res_ocean_arcsec : int
+            resolution for ocean centroids in arcsec. Default is 1800
+        reg_id : int
+            region to filter according to region_id values
+        extent : tuple
```
chahank, line 859: Should we put the default to have `min_lat = -60` and `max_lat = 60`? Also, the default is not explained. Is it the whole world?

climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+            Format (min_lon, max_lon, min_lat, max_lat) tuple.
+            If min_lon > lon_max, the extend crosses the antimeridian and is
+            [lon_max, 180] + [-180, lon_min]
+            Borders are inclusive.
+        dump_dir : str
```
chahank, line 864: The `country_iso3alpha` is missing (also the `reg_id`, but I would remove that one probably, or make it clear that it is simply the country iso3 numeric.

climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+            directory where the files should be downoladed. Default: SYSTEM_DIR
+        Returns
+        -------
+        climada.hazard.centroids.Centroids
+            global centroids from the api
```
chahank, line 869: It is not always global, only by default.

climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+        """
+
+        res_arcsec_land = str(res_arcsec_land)
+        res_arcsec_ocean = str(res_arcsec_ocean)
+        extent_property = '(-180, -90, 180, 90)'
+        dataset = self.get_dataset_info("centroids", properties={'res_arcsec_land':res_arcsec_land,
+                                                                 'res_arcsec_ocean':res_arcsec_ocean,'extent':extent_property})
```
chahank, line 876: Maybe a check whether the data exist would be useful? Because the error message from `get_dataset_info` is not very useful if for instance I say `res_arcsec_land=2500` and it does not exist...

chahank (changes requested):
Great addition! I suggest some minor changes to the available variables.

climada/util/api_client.py
emanuel-schmid, line 876:
```suggestion
        dataset = self.get_dataset_info("centroids", properties={'res_arcsec_land':str(res_arcsec_land),
                                                                 'res_arcsec_ocean':st(res_arcsec_ocean),'extent':extent_property})
```

climada/util/api_client.py
zeliest: okay, it was just to follow the centroids.select() arguments. but probably country_iso3alpha is sufficient

climada/util/api_client.py
zeliest: yes for now I put the whole world, I would keep it like that. Because for most hazards it makes sense to have the entire globe

climada/util/api_client.py
```diff
@@ -844,6 +844,44 @@ def get_litpop_default(self, country=None, dump_dir=SYSTEM_DIR):
...
+        target_dir = self._organize_path(dataset, dump_dir) \
+            if dump_dir == SYSTEM_DIR else dump_dir
+        centroids = Centroids.from_hdf5(self._download_file(target_dir, dataset.files[0]))
+        if country_iso3alpha:
+            reg_id = pycountry.countries.get(alpha_3=country_iso3alpha).numeric
```
emanuel-schmid, line 881: In `get_litpop_default`, there is also a country argument, but it's treated with more tolerance, it can be anything that pycountry.countries.lookup makes sense of, and even a list.
Perhaps the two methods could be aligned with regard to treating countries?

climada/util/api_client.py
chahank: But is it cut at high/low latitudes or not? This would be important to know. The default extent could then be with +-60degrees?

Issue #487, opened 2022-06-21
Add classmethod to Hazard for reading raster-like data from NetCDF file


peanutfun:
The `Hazard` class offers several options to instantiate it from data files, e.g. `from_raster`, `from_excel`, etc. The classmethod `from_raster`, in particular, uses `rasterio` to open datasets and read their metadata, coordinates, and data. In this issue, I want to discuss if a general-purpose classmethod for reading data from a NetCDF file into a Hazard object might be useful, and how such a method could look like. A method implementing such a functionality to some extent can be found at [`climada_petals/blob/feature/wildfire/climada_petals/hazard/wildfire.py#L2247`](https://github.com/CLIMADA-project/climada_petals/blob/feature/wildfire/climada_petals/hazard/wildfire.py#L2247).

### What the method should do

Use a single NetCDF file to load data for a consistent instance of `Hazard`, meaning that if data is missing, it will be set to a sensible default.

The minimal (i.e., essential) data supplied as variables in the file should be
* hazard intensity data (2D or 3D dataset)
* coordinates (1D dataset each)
* time (1D dataset, if applicable)

Optional data could include:
* hazard fraction data (same dimensions as intensity)
* event frequency (1D)
* event names (1D)
* event IDs (1D)
* coordinate system information (attributes/metadata)

### Method signature

`from_netcdf` should take the following arguments:
* `data` (path-like or `xarray.Dataset`, required): The dataset. Open the file if it is a path.
* `intensity_var` (string, required): The name of the hazard intensity variable in the dataset
* `fraction_var` (string, optional): The name of the hazard fraction variable in the dataset
* `coordinate_vars` (dict, optional): A mapping from default coordinate names to the variables used as coords in the dataset, e.g. `dict(longitude="lon", latitude="y")`
* *tbd*

### Method outline 

Suppose a netCDF file contains the following data:
* `intensity`: 3D dataset (dims: "time", "longitude", "latitude")
* 1D coordinate dataset for each dimension

Then the following code creates a consistent `Hazard` instance from this data:
```python
import xarray as xr
from scipy.sparse import csr_matrix
from climada.hazard import Hazard
from climada.hazard.centroids.centr import Centroids

data = xr.open_dataset("...")
hazard = Hazard()

# Transpose the data so we flatten it with longitude running "fastest"
intensity = data["intensity"].transpose("time", "latitude", "longitude")
[...]

PR #507, opened 2022-07-05
Add Hazard classmethod for loading xarray Datasets

peanutfun:
Changes proposed in this PR:
- Add classmethod `from_raster_xarray` that loads a dataset from a file and reads the appropriate data into a Hazard instance.
- Add appropriate tests.

This PR fixes issue #487 

### To-Do List
- [x] Add hazard type and unit *(only to signature. They are not read from the data)*
- [x] Read hazard frequency
- [x] Read event names
- [x] Add logger messages
- [x] Fix as many linter issues as possible

### Open Questions
* Do we want to read hazard type and units from dataset attributes or is passing them as parameters enough?

    *Reading them from data is difficult because they should be supplied as xarray Attributes and we currently cannot handle those separately*
* Should a list of input files be allowed (using [`xarray.open_mfdataset`](https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html) internally)?

    *Probably not, users can do that. There are too many possibilities for combining datasets*
* Is the interface sensible, especially w.r.t. the keyword-only arguments?
* Should we rename this method to `from_raster_xarray`, since we technically can load anything `xarray` is able to load? This would mean we should also check if loading `.grib` files works

   *Yes, we renamed it. But `.grib` files are not checked yet.*

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing
- [ ] No new linter issues


timschmi95:
Great!
I think it would be good to also make the method work for data that is on a different grid, but has lat/lon coordinates as 2D arrays. Then, instead of creating a meshgrid, one can directly take the lat/lon coordinates and flatten them. 
I could add this functionality, should I just push it to the branch or make a sub branch?

Also regarding the open_mfdatset, I guess we don't need to include it internally. The user can do it themselves and just pass the xr.Dataset into the climada function. Then he's sure that the different files are concatenated correctly.

climada/hazard/base.py
```diff
@@ -369,6 +371,104 @@ def set_vector(self, *args, **kwargs):
...
+        """Read raster-like data from a NetCDF file
+
+        Parameters
+        ----------
+        data : xarray.Dataset or str
+            The NetCDF data to read from. May be a opened dataset or a path to a NetCDF
+            file, in which case the file is opened first.
+        intensity : str
+            Identifier of the `xarray.DataArray` containing the hazard intensity data.
+        fraction : str or Callable
```
chahank, line 392: I think it would be simpler to have `str` with default `None`. If `None`, the fraction is set to one. I do not really see a use case with another callable right now?

climada/hazard/base.py
```diff
@@ -369,6 +371,104 @@ def set_vector(self, *args, **kwargs):
...
+            data = xr.open_dataset(data)
+        hazard = cls()  # TODO: Hazard type
+
+        # Update coordinate identifiers
+        coords = {"time": "time", "longitude": "longitude", "latitude": "latitude"}
+        if coordinate_vars is not None:
+            unknown_coords = [co not in coords for co in coordinate_vars]
+            if any(unknown_coords):
+                raise ValueError(
+                    f"Unknown coordinates passed: '{list(coordinate_vars.keys())}'. "
```
chahank, line 417: Maybe list the unknown coordinates?

peanutfun: Yeah, that sounds like a very reasonable idea 😅

climada/hazard/base.py
peanutfun: It would be more useful if the callable could also take the coordinates as arguments, but I feel like this complicates matters too much. With the current implementation, you can use the callable, e.g., for implementing a simple function based on the intensity, like a custom threshold. It would be more simple if we left it out, but it also feels like a rather low-hanging fruit. What's the harm in leaving it in?

peanutfun:
@timschmi95 Thanks for the feedback! I agree that concatenating multiple data files within this method would be a bit of overkill, as we would probably need to handle too many possible cases.

> I think it would be good to also make the method work for data that is on a different grid, but has lat/lon coordinates as 2D arrays. Then, instead of creating a meshgrid, one can directly take the lat/lon coordinates and flatten them.
I could add this functionality, should I just push it to the branch or make a sub branch?

How would such a dataset look like, exactly? Isn't that – in CLIMADA terms – a "vector" dataset, that one would read with a `from_vector` method?

climada/hazard/base.py
```diff
@@ -369,6 +371,104 @@ def set_vector(self, *args, **kwargs):
...
+        hazard.fraction = to_hazard_csr_matrix(fraction_arr)
+        hazard.fraction.eliminate_zeros()
+
+        # Fill hazard with required information
+        num_events = data.sizes[coords["time"]]
+        hazard.event_id = np.array(range(num_events)) + 1  # event_id starts at 1
+        hazard.frequency = np.ones(num_events)  # TODO: Optional read from file
+        hazard.event_name = list(data[coords["time"]].values)
+
+        def to_datetime(date: np.datetime64):
```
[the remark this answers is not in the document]
peanutfun: I missed that, thanks! I'm using the util functions now, they work like a charm ☺️

climada/hazard/base.py
peanutfun: Done

climada/hazard/test/test_base_netcdf.py
```diff
@@ -0,0 +1,261 @@
...
+        )
+        self._assert_intensity_fraction(hazard)
+
+    def test_coordinate_naming(self):
+        """Test if coordinates with different names than dimensions can be read"""
+        hazard = Hazard.from_raster_netcdf(
+            self.netcdf_path,
+            dimension_vars=dict(latitude="y", longitude="x"),
+            coordinate_vars=dict(latitude="lat", longitude="lon", time="years"),
+        )
```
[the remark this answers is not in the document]
timschmi95: Couldn't you ask the user only for coordinate vars, as originally planned (with default to 'lat', 'lon','time') and then get the dimensions from data['lat'].dims ? That return a tuple for the dimensions of the lat coordinate, so then you know over which variables to stack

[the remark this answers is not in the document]
peanutfun: Excellent idea!

peanutfun:
I think this is coming along quite nicely. For now, I think the method is capable of the most important things. The examples and unit tests should make the capabilities somewhat clear. Would somebody like to review? What's missing are probably some integration tests based on real data.

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+        Notes
+        -----
+        * Intensity and fraction data must be three-dimensional data that is interpreted
+          with the coordinates given by ``coordinate_vars`` (or the default). All other
+          data must be given in time coordinates only.
+        * To avoid confusion in the call signature, all parameters are keyword-only
+          arguments, except ``data``.
+        * The attributes ``Hazard.tag.haz_type`` and ``Hazard.unit`` currently cannot be
+          read from the Dataset. Use the method parameters to set these attributes or set
+          them in the resulting ``Hazard`` object by yourself.
```
chahank, line 414: I would force the user to set these parameters. So they should not be optional :D.
```suggestion
        * The attributes ``Hazard.tag.haz_type`` and ``Hazard.unit`` currently cannot be
          read from the Dataset. Use the method parameters to set these attributes.
```

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+        # Transform coordinates into centroids
+        hazard.centroids = Centroids.from_lat_lon(
+            data[coords["latitude"]].values, data[coords["longitude"]].values
+        )
+
+        def to_csr_matrix(array):
+            """Store an array as sparse matrix, optimizing storage space"""
+            mat = sparse.csr_matrix(array)
+            mat.eliminate_zeros()
+            return mat
```
chahank, line 592: I think with this method you cannot actually load sparse data (i.e. vector data). This is an issue imho.

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+        -----
+        * Intensity and fraction data must be three-dimensional data that is interpreted
+          with the coordinates given by ``coordinate_vars`` (or the default). All other
+          data must be given in time coordinates only.
+        * To avoid confusion in the call signature, all parameters are keyword-only
+          arguments, except ``data``.
+        * The attributes ``Hazard.tag.haz_type`` and ``Hazard.unit`` currently cannot be
+          read from the Dataset. Use the method parameters to set these attributes or set
+          them in the resulting ``Hazard`` object by yourself.
+        * This method **does not** read lazily. Single data arrays must fit into memory.
```
chahank, line 415: That is probably a big issue. The intensity and fraction are sparse, thus casting them to dense will most likely make this method not usable for most larger datasets (which are quite common in CLIMADA). Any idea how to solve this?

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+            mat = sparse.csr_matrix(array)
+            mat.eliminate_zeros()
+            return mat
+
+        # Read the intensity data and flatten it in spatial dimensions
+        LOGGER.debug("Loading Hazard intensity from DataArray '%s'", intensity)
+        hazard.intensity = to_csr_matrix(data[intensity])
+
+        # Define accessors for xarray DataArrays
+        def default_accessor(x: xr.DataArray):
```
chahank, line 599: Maybe?
```suggestion
        def _default_accessor(x: xr.DataArray):
```

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+            """By default, use the numpy array representation of data"""
+            return x.values
+
+        def strict_positive_int_accessor(x: xr.DataArray, report_key: str):
```
chahank, line 603:
```suggestion
        def _strict_positive_int_accessor(x: xr.DataArray, report_key: str):
```

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+                )
+
+            # Make 'identifier' an index for replacement, then make it a column again
+            data_ident = data_ident.set_index("identifier")
+            data_ident["user_key"].loc[list(data_vars.keys())] = list(
+                data_vars.values()
+            )
+            data_ident.reset_index(inplace=True)
+
+        def read_or_default(ident: pd.Series):
```
chahank, line 665: I do not understand the naming of this method. I also do not understand the docstring.

climada/hazard/base.py
```diff
@@ -369,6 +371,361 @@ def set_vector(self, *args, **kwargs):
...
+        LOGGER.debug("Loading Hazard intensity from DataArray '%s'", intensity)
+        hazard.intensity = to_csr_matrix(data[intensity])
+
+        # Define accessors for xarray DataArrays
+        def default_accessor(x: xr.DataArray):
+            """By default, use the numpy array representation of data"""
+            return x.values
+
+        def strict_positive_int_accessor(x: xr.DataArray, report_key: str):
+            """Only allow positive, non-zero integers"""
```
chahank, line 604: This is a quite long method, so maybe a real docstring is needed.

climada/hazard/base.py
chahank, line 665:
```suggestion
        def _read_or_default(ident: pd.Series):
```

climada/hazard/test/test_base_xarray.py
```diff
@@ -0,0 +1,353 @@
...
+
+THIS_DIR = os.path.dirname(__file__)
+
+
+class ReadDefaultNetCDF(unittest.TestCase):
+    """Test reading a NetCDF file where the coordinates to read match the dimensions"""
+
+    def setUp(self):
+        """Write a simple NetCDF file to read"""
+        self.netcdf_path = os.path.join(THIS_DIR, "default.nc")
```
chahank, line 40: Using `pathlib` instead of `os` is recommended.
```suggestion
        self.netcdf_path = Path.cwd() / "default.nc"
```
The you can remove the ugly 
`THIS_DIR = os.path.dirname(__file__)`
and import
`from pathlib import Path`

climada/hazard/test/test_base_xarray.py
```diff
@@ -0,0 +1,353 @@
...
+import unittest
+import datetime as dt
+import numpy as np
+from scipy.sparse import csr_matrix
+
+import xarray as xr
+
+from climada.hazard.base import Hazard
+
+THIS_DIR = os.path.dirname(__file__)
```
chahank, line 32:
```suggestion

```

climada/hazard/test/test_base_xarray.py
```diff
@@ -0,0 +1,353 @@
...
+import os
+import unittest
+import datetime as dt
+import numpy as np
+from scipy.sparse import csr_matrix
+
+import xarray as xr
+
+from climada.hazard.base import Hazard
+
```
chahank, line 31:
```suggestion
from pathlib import Path
```

climada/hazard/test/test_base_xarray.py
```diff
@@ -0,0 +1,353 @@
...
+        self.longitude = np.array([0, 1, 2])
+        dset = xr.Dataset(
+            {"intensity": (["time", "latitude", "longitude"], self.intensity),},
+            dict(time=self.time, latitude=self.latitude, longitude=self.longitude),
+        )
+        dset.to_netcdf(self.netcdf_path)
+
+    def tearDown(self):
+        """Delete the NetCDF file"""
+        os.remove(self.netcdf_path)
```
chahank, line 53: self.netcdf_path.unlink()

climada/hazard/test/test_base_xarray.py
```diff
@@ -0,0 +1,353 @@
...
+                Hazard.from_raster_xarray(dset)
+            self.assertIn(f"'{key}' data must be larger than zero", str(cm.exception))
+
+
+class ReadDimsCoordsNetCDF(unittest.TestCase):
+    """Checks for dimensions and coordinates with different names and shapes"""
+
+    def setUp(self):
+        """Write a NetCDF file with many coordinates"""
+        self.netcdf_path = os.path.join(THIS_DIR, "coords.nc")
```
chahank, line 217:
```suggestion
        self.netcdf_path = Path.cwd() / "coords.nc"
```

chahank (changes requested):
Without going into details, I have a few suggestions. Mainly using Pathlib in the tests, and reflecting on the problem of sparse data one more time before finalizing. Otherwise, cool stuff!

climada/hazard/base.py
peanutfun: I would say it depends. I stopped at some point of my own wishlist because it otherwise get's too much to take into account in a single PR. I personally would say that this can be solved in a follow-up PR. There is quite a lot one has to take into account here, in particular the chunking of dask arrays and a memory-efficient conversion of those into a CSR matrix. I haven't tried that out yet.

climada/hazard/base.py
peanutfun: In my understanding, xarray does not consider sparse data. The arrays returned by xarray are always dense. So I don't see how this can be an issue here. I'm calling `eliminate_zeros` to make sure the resulting matrix is as sparse as possible.

Apparently, there you can use sparse data types as chunks, but that needs to be done on the user side and is not something xarray handles for you automatically, see e.g. https://stackoverflow.com/a/60451126

climada/hazard/base.py
peanutfun: Can do, but the function is not visible outside the scope of the `from_raster_xarray` method anyway, so I see no point in marking it as "private"

climada/hazard/test/test_base_xarray.py
peanutfun: Yes, that looks better! Maybe we should even use [`tempfile`](https://docs.python.org/3/library/tempfile.html) to create a temporary file/directory for that? 🤔

climada/hazard/base.py
peanutfun: I removed the option to insert a callable and made `fraction` an optional variable that is set via `data_vars`.

climada/hazard/base.py
peanutfun: So should I also remove the default values? This would mean that users indeed need to set these values every time they call the method.

climada/hazard/base.py
chahank: Well, to make it obvious to the reader of the code that this is an internal method. Currently, the methods are quite long, so it is not obvious what the scope of the methods are. Besides, they have no docstrings, and their internal working is very opaque at the moment. Thus, a minimal solution is to make them obviously private. Better would be to have clear docstring, even better to make the code more readable, but the latter is very subjective.

climada/hazard/test/test_base_xarray.py
chahank: Could also be nice. I haven't used it often yet myself though. Feel free to choose what you prefer!

climada/hazard/base.py
chahank: Is it possible to store vector data in an xarray? So, basically saving/reading only the non-zero data points?

climada/hazard/base.py
peanutfun: Alright, I'll definitely update the docstrings!

climada/hazard/base.py
chahank: Ok, fair enough. This will probably very strongly limit the use of this method for CLIMADA, so we should definitely keep it on the radar.

climada/hazard/base.py
chahank: Let's see how the use of this method evolves, and see whether we come back on this decision later or not. I like it this way, as it starts with the simpler version.

climada/hazard/base.py
chahank: I tend to say yes. I do not really see a way to have useful defaults. In the future, it would be nice to read it directly from the dataset, in particular if the dataset was written from within CLIMADA.

climada/hazard/base.py
peanutfun: Maybe I'm not sure what "vector data" actually means 🤔 You could store an array with only two dimensions "time" and "space" (basically the structure of `Hazard.frequency`), and then have two coordinates "latitude" and "longitude" that map each point in "space" to them. `from_raster_xarray` would read that as well.

To know if any data points are non-zero (or non-`NaN`), one has to load them, right? Xarray either loads all data at once or uses dask to devide the data into chunks that are only loaded into memory when needed.

climada/hazard/test/test_base_xarray.py
peanutfun:
```suggestion
        self.netcdf_path.unlink()
```

climada/hazard/base.py
chahank: >Maybe I'm not sure what "vector data" actually means 🤔 You could store an array with only two dimensions "time" and "space" (basically the structure of Hazard.frequency), and then have two coordinates "latitude" and "longitude" that map each point in "space" to them. from_raster_xarray would read that as well.

Exactly, this is what I mean. But you are right, it first would have to be read... All right, let's see how we can solve this in the future.

peanutfun: I at least found out the following: Xarray typically uses `NaN` to represent missing values, but `csr_matrix` stores `NaN`s explicitly. So I think it makes a lot of sense to replace `NaN`s with zeros before constructing the sparse matrix.

chahank: Good idea! Although then I would suggest to use a more intelligent way to build the matrix. First, identify all the non-zero elements, and then use only these to construct the matrix. It is quite inefficient to first create a sparse row matrix and then alter its sparcity. Alternatively, first create a `coo_matrix`, change the sparcity, and then transform to a `csr_matrix`.

peanutfun: I think this is not necessary. I just noticed that when you construct a matrix from some dense array, the zeros will be already eliminated (so we can save this extra call I put in). Plus, it seems like the constructor of a `csr_matrix` actually constructs a `coo_matrix` internally when being initialized from a dense array, see https://github.com/scipy/scipy/blob/v1.8.1/scipy/sparse/_compressed.py#L85

chahank: All right! Thanks for checking. Let's just keep in mind that if we manage to actually deal with sparse data, it would be much more efficient to initialize the sparse matrix not from a dense array.

climada/hazard/base.py
peanutfun: I updated it to `load_data_or_default`, improved the signature and wrote a proper docstring, see https://github.com/CLIMADA-project/climada_python/pull/507/commits/9c68a9e67b60e71e10c811759de0ad506c0eaa6a

climada/hazard/base.py
peanutfun: See https://github.com/CLIMADA-project/climada_python/pull/507/commits/9c68a9e67b60e71e10c811759de0ad506c0eaa6a

climada/hazard/base.py
peanutfun: NaN handling fixed and veryfied in abc9d320a2e970f2ee72cbd9e5bda1ecbe6f01e9

climada/hazard/base.py
peanutfun: But shouldn't it behave like the other classmethods? The others set the default value to `None` and dispatch the hazard type selection to possible derived classes. And the unit is only used for plotting, right?

chahank: The hazard type though is fundamental. I would push towards changing this habit. A hazard without a type and without a unit should not exist.

peanutfun: Ok, will do.

peanutfun: Done in 29b74f3f4b5288da48d45ca964e91f8909bdbac7

climada/hazard/base.py
```diff
@@ -369,6 +377,415 @@ def set_vector(self, *args, **kwargs):
...
+
+            Default values are:
+            * ``fraction``: 1.0 where intensity is not zero, else zero
+            * ``hazard_type``: Empty string
+            * ``frequency``: 1.0 for every event
+            * ``event_name``: String representation of the event time
+            * ``event_id``: Consecutive integers starting at 1 and increasing with time
+        crs : str, optional
+            Identifier for the coordinate reference system to use. Defaults to
+            ``EPSG:4326`` (WGS 84), defined by ``climada.util.constants.DEF_CRS``.
```
peanutfun, line 459: @chahank I implemented the `crs` argument in ccec87f8, but I'm unsure how to test this argument properly. Do you have an idea or suggestion?

chahank: Great addition!

What do you mean exactly by testing? Is it about how to automatically retrieve it from the netcdf file or is it about how to write a unit test on that specifically?

climada/hazard/base.py
```diff
@@ -369,6 +377,415 @@ def set_vector(self, *args, **kwargs):
...
+        )
+
+        def to_csr_matrix(array: np.ndarray) -> sparse.csr_matrix:
+            """Store a numpy array as sparse matrix, optimizing storage space
+
+            The CSR matrix stores NaNs explicitly, so we set them to zero.
+            """
+            return sparse.csr_matrix(np.where(np.isnan(array), 0, array))
+
+        # Read the intensity data and flatten it in spatial dimensions
```
peanutfun, line 608: Data is already flattened
```suggestion
        # Read the intensity data
```

peanutfun:
@chahank @emanuel-schmid I think this is finished 🙌 Could you give it a final look?

climada/hazard/base.py
```diff
@@ -392,6 +400,417 @@ def set_vector(self, *args, **kwargs):
...
+
+        # Stack (vectorize) the entire dataset into 2D (time, lat/lon)
+        # NOTE: We want the set union of the dimensions, but Python 'set' does not
+        #       preserve order. However, we want longitude to run faster than latitude.
+        #       So we use 'dict' without values, as 'dict' preserves insertion order
+        #       (dict keys behave like a set).
+        data = data.stack(
+            event=dims["event"],
+            lat_lon=dict.fromkeys(dims["latitude"] + dims["longitude"]),
+        )
```
peanutfun, line 619: TODO: Catch `KeyError` from nonexisting dims and explain

climada/hazard/base.py
```diff
@@ -392,6 +400,417 @@ def set_vector(self, *args, **kwargs):
...
+        data_vars: Optional[Dict[str, str]] = None,
+        crs: str = DEF_CRS,
+    ):
+        """Read raster-like data from an xarray Dataset or a raster data file
+
+        This method reads data that can be interpreted using three coordinates for event,
+        latitude, and longitude. The data and the coordinates themselves may be organized
+        in arbitrary dimensions in the Dataset (e.g. three dimensions 'year', 'month',
+        'day' for the coordinate 'event'). The three coordinates to be read can be
+        specified via the ``coordinate_vars`` parameter.
```
peanutfun, line 421: TODO: Add information on how to read single-event data (without event dimension)

Issue #495, opened 2022-06-27
Major bug? get_land_geometry and tc_tracks.track_land_params


bguillod:
I think I found **a major bug** in the assignment of TC track points to land or ocean, which occurs at least in `tc_track_synth`. This seems to be related to how the land geom is retrieved from the track extents. However I am afraid this might have implications elsewhere, too.

In essence, in `climada.hazard.tc_tracks_synth`, land_geom is retrieved using:
```
extent = tracks.get_extent()
land_geom = climada.util.coordinates.get_land_geometry(
    extent=extent, resolution=10
)
```

However, if I read global data as follows
```
tracks = TCTracks.from_ibtracs_netcdf(
  provider = 'usa'
)
extent = tracks.get_extent()
land_geom = u_coord.get_land_geometry(
    extent=extent, resolution=10
)
```
Then I get the following value for `extent`: `(31.4, 373.6, -44.800000762939455, 70.79999694824218)`, and the following for land_geom:
<img width="163" alt="image" src="https://user-images.githubusercontent.com/13904718/175941691-bad13364-fbe1-464b-83c9-d200fe37bbca.png">

Hence the extent of the land geometry that is being loaded only covers longitude from 31.4 to 180 degrees East. Looks like the underlying function `get_land_geometry` does not handle spherical coordinates.

The implication is that any track westward of 31.4 degrees East will be said not to be over land. The issue probably won't show up if I had loaded data only for the, e.g. North Atlantic basin, because then the extent would probably be within the [-180, +180] longitude range.

This is a major issue and it might apply to many other code instances than just for the TC synthetic tracks generation. My questions are then:

- Is there a way for `get_land_geometry` to properly account for spherical coordinates? That would be the easiest way to fix. If not, then at least the function should raise an error if longitude values are outside of [-180, +180] (and since we're at it let's also make sure latitude never goes outside of [-90,90], although this case is very unlikely).
- Where else in the code might this issue lead to mistakes, too?

bguillod:
As an add-on, I think generally the function `coord_on_land` might not do what's expected if the tracks longitude are normalized and go outside of [-180, +180], which for instance is the case for tracks that cross the anti-meridian in `from_ibtracs_netcdf` I believe. If `get_land_geometry` cannot easily be fixed to account for this, then probably `coord_on_land` must re-normalize longitude values back to [-180, +180] degrees within this function.

chahank:
[...]

PR #524, opened 2022-07-29
Bugfix/get geometries extent antimeridian

bguillod:
Changes proposed in this PR:
- Ensure `climada.util.coordinates.get_land_geometry` deals correctly with extent. Previously, if an extent going beyond [-180, +180] range in longitude the geometry would be cut (e.g. passing `extent=(160, 200, -50,50)` would actually return the land geometry only for the extent `(160, 180, -50, 50)`.
- Apply the same correction to `get_country_geometry`.
- Removes the redundancy between these two functions by using the latter in the former.
- Adjust function `coord_on_land` to ensure longitude coordinates are also properly dealt with.

This PR fixes issue #495.

A first test has been implemented but this can be completed.

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing
- [x] No new linter issues

### DIFF
--- a/climada/hazard/test/test_tc_tracks.py
+++ b/climada/hazard/test/test_tc_tracks.py
@@ -905,21 +905,21 @@ class TestFuncs(unittest.TestCase):
         on_land = np.array([False, True, True])
         lon_shift = np.array([-360, 0, 360])
         # ensure both points are considered on land as is
-        self.assertEqual(
+        np.testing.assert_array_equal(
             u_coord.coord_on_land(lat = lat_test, lon = lon_test),
             on_land
         )
         # independently on shifts by 360 degrees in longitude
-        self.assertEqual(
+        np.testing.assert_array_equal(
             u_coord.coord_on_land(lat = lat_test, lon = lon_test + lon_shift),
             on_land
         )
-        self.assertEqual(
+        np.testing.assert_array_equal(
             u_coord.coord_on_land(lat = lat_test, lon = lon_test - lon_shift),
             on_land
         )
         # also when longitude is within correct range
-        self.assertEqual(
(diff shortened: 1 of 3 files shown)


bguillod:
Working on fixing the failed checks...

bguillod:
Ok so one last test is still failing: `test_get_land_geometry_country_pass`.

It looks as if the PR slightly changed the extent of some countries used in this test. I think this might be due to the fact that the filtering of the shapes in natural earth has been slightly adjusted within this PR.

Does anybody know who wrote these tests?

In my opinion we can update the expected value of these tests, so they pass. However I need someone else's opinion to confirm as I don't want to oversee something that also could be an issue.

tovogt:
From my side this is ready to be merged. I just added some code clean-up for the linter, but the rest is fine.

> Ok so one last test is still failing: `test_get_land_geometry_country_pass`.
> 
> It looks as if the PR slightly changed the extent of some countries used in this test. I think this might be due to the fact that the filtering of the shapes in natural earth has been slightly adjusted within this PR.
> 
> Does anybody know who wrote these tests?
> 
> In my opinion we can update the expected value of these tests, so they pass. However I need someone else's opinion to confirm as I don't want to oversee something that also could be an issue.

The tests were broken from the start because they passed the keyword arguments as positional arguments and the result was that the `110` (which was obviously supposed to be the resolution) was actually assigned to the `extent` keyword argument, so that the actual resolution used was `10` (the default value of that keyword argument). I fixed this by explicitly selecting `10` as the resolution.

emanuel-schmid:
🙌

bguillod:
Thanks a lot @tovogt and @emanuel-schmid for reviewing and merging!

Issue #515, opened 2022-07-13
Make reassign_centroids the default in impact calculation


emanuel-schmid:
The current behavior in impact calculation is to _not_ re-assign centroids in exposures if they have a centr_ column with a matching hazard type. This can lead to a faulty assignment when the same exposure is subject of impact calculation for two hazards of the same type but with different centroids.

Task: Change the default to reassign == True, so advanced users being aware of the risks can turn it off while everybody else can rely on CLIMADA doing what they mean - at the cost of decreased performance.

emanuel-schmid:
PR #527 solves the issue: centroids are (re-)assigned by default, but in uncertainty and cost/benefit calculations they are conserved as there are i.g. no conflicting hazards present when they are running and reassigning would basically just make them unusable.

PR #527, opened 2022-08-04
Feature/reassign centr bydflt

emanuel-schmid:
Changes proposed in this PR:
- As discussed in PR #436, not reassigning centroids may lead to faulty results through side effects
- As decided on the CLIMADA developer day, the easiest solution is to introduce an `ImpactCalc.impact` parameter that may be used to suppress re-assignment of centroids
- This PR proposes the parameter(s) `reassign_centroids` in both methods `impact` and `insured_impact`.
- The parameter were set to `True` wherever it seemed suitable throughout the code base

This PR fixes issue #515

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing
- [x] No new linter issues

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -184,7 +184,7 @@ class Impact():
 
 
 
-    def calc(self, exposures, impact_funcs, hazard, save_mat=False):
+    def calc(self, exposures, impact_funcs, hazard, save_mat=False, reassign_centroids=True):
         """This function is deprecated, use ImpactCalc.impact
         and ImpactCalc.insured_impact instead.
         """
@@ -198,11 +198,11 @@ class Impact():
                 " for insured impacts instead. For non-insured impacts "
                 "please use ImpactCalc().impact()"
                 )
-            self.__dict__ = impcalc.insured_impact(save_mat).__dict__
+            self.__dict__ = impcalc.insured_impact(save_mat, reassign_centroids).__dict__
         else:
             LOGGER.warning("The use of Impact().calc() is deprecated. "
                            "Use ImpactCalc().impact() instead.")
-            self.__dict__ = impcalc.impact(save_mat).__dict__
+            self.__dict__ = impcalc.impact(save_mat, reassign_centroids).__dict__
 
 #TODO: new name
     @classmethod
@@ -1029,7 +1029,7 @@ class Impact():
         imp_arr = np.zeros(len(exp.gdf))
         for i_time, _ in enumerate(haz_list):
             imp_tmp = Impact()
-            imp_tmp.calc(exp, impf_set, haz_list[i_time])
+            imp_tmp.calc(exp, impf_set, haz_list[i_time], reassign_centroids=False)
             imp_arr = np.maximum(imp_arr, imp_tmp.eai_exp)
             # remove not impacted exposures
             save_exp = imp_arr > imp_thresh
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -184,7 +184,7 @@ class Impact():
 
 
 
-    def calc(self, exposures, impact_funcs, hazard, save_mat=False, reassign_centroids=True):
+    def calc(self, exposures, impact_funcs, hazard, save_mat=False, assign_centroids=True):
         """This function is deprecated, use ImpactCalc.impact
         and ImpactCalc.insured_impact instead.
         """
@@ -198,11 +198,11 @@ class Impact():
                 " for insured impacts instead. For non-insured impacts "
                 "please use ImpactCalc().impact()"
                 )
-            self.__dict__ = impcalc.insured_impact(save_mat, reassign_centroids).__dict__
+            self.__dict__ = impcalc.insured_impact(save_mat, assign_centroids).__dict__
         else:
             LOGGER.warning("The use of Impact().calc() is deprecated. "
                            "Use ImpactCalc().impact() instead.")
-            self.__dict__ = impcalc.impact(save_mat, reassign_centroids).__dict__
+            self.__dict__ = impcalc.impact(save_mat, assign_centroids).__dict__
 
 #TODO: new name
     @classmethod
@@ -1027,9 +1027,11 @@ class Impact():
         imp_list = []
         exp_list = []
         imp_arr = np.zeros(len(exp.gdf))
+        # assign centroids once for all
+        exp.assign_centroids(haz_list[0])
         for i_time, _ in enumerate(haz_list):
             imp_tmp = Impact()
-            imp_tmp.calc(exp, impf_set, haz_list[i_time], reassign_centroids=False)
+            imp_tmp.calc(exp, impf_set, haz_list[i_time], assign_centroids=False)
             imp_arr = np.maximum(imp_arr, imp_tmp.eai_exp)
             # remove not impacted exposures
             save_exp = imp_arr > imp_thresh
--- a/climada/engine/impact_calc.py
+++ b/climada/engine/impact_calc.py
@@ -102,13 +102,20 @@ class ImpactCalc():
         if 'cover' in self.exposures.gdf.columns:
             return self.exposures.gdf['cover'].to_numpy()
 
-    def impact(self, save_mat=True):
+    def impact(self, save_mat=True, reassign_centroids=True):
         """Compute the impact of a hazard on exposures.
 
         Parameters
         ----------
-        save_mat : bool
+        save_mat : bool, optional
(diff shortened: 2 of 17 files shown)


climada/engine/impact_calc.py
```diff
@@ -238,18 +252,22 @@ def _return_empty(self, save_mat):
...
-
-        """
-        self.exposures.assign_centroids(self.hazard, overwrite=False)
+        impf_col : str
+            Name of the impact function column in exposures.gdf
+        reassign_centroids : bool
+            Indicates whether centroids are re-assigned to the self.exposures object
+            or kept from previous impact calculation with a hazard of the same hazard type.
+            Centroids assignment is an expensive operation, set this to true if you know that
+            the centroids are the same between two impact calculations.
```
peanutfun, line 268: Same as above
```suggestion
            Centroids assignment is an expensive operation; set this to ``False`` to save 
            computation time if the centroids have not changed since the last impact
            calculation.
```

climada/test/test_calibration.py
```diff
@@ -18,6 +18,7 @@
 
 Test Calibration class.
 """
+import re
```
peanutfun, line 21: Unused import
```suggestion
```

climada/engine/test/test_impact_calc.py
```diff
@@ -299,7 +299,7 @@ def test_calc_insured_impact_fail(self):
     def test_minimal_exp_gdf(self):
         """Test obtain minimal exposures gdf"""
         icalc = ImpactCalc(ENT.exposures, ENT.impact_funcs, HAZ)
-        exp_min_gdf = icalc.minimal_exp_gdf('impf_TC')
+        exp_min_gdf = icalc.minimal_exp_gdf('impf_TC', False)
```
peanutfun, line 302: Maybe make it explicit here as well?
```suggestion
        exp_min_gdf = icalc.minimal_exp_gdf('impf_TC', reassign_centroids=False)
```

peanutfun (changes requested):
Looking good! I mostly suggest cosmetic changes. In my suggestions for the comments I try to make clear to users what they gain by setting `reassign_centroids` to `False`, and when they can use it. Finally, by passing the `force_reassign` parameter to `reassign_centroids`, the method `Forecast.calc` can be simplified by quite a bit.

peanutfun:
Oh by the way, why is it called `reassign_centroids` and not simply `assign_centroids`? These methods behave exactly the same if centroids are assigned or not. So I think `assign_centroids` explains more exactly what's happening, and it's also equivalent to the `Exposures.assign_centroids` method. @chahank, @emanuel-schmid, what do you think?

chahank:
I agree that the name should rather be `assign_centroids`. It is confusing to have `reassign_centroids`, when often there are no assigned centroids to begin with.

chahank:
The methods in the module `unsequa` must also still be adjusted. I am not yet sure what is the right way to do it. In any case, the default should be to `not` reassign the centroids.

emanuel-schmid:
If centroids are not assigned for that kind of hazard, they are going to be assigned regardless of the flag, hence, to me, `reassign_centroids` seems to be more accurate.

emanuel-schmid:
> the methods in the module `unsequa` must also still be adjusted. I am not yet sure what is the right way to do it. In any case, the default should be to `not` reassign the centroids. The same is true for the `cost_benefit` module. It makes little sense to repeat this costly computation for repeated impact calculations for which the centroids do not change.

You are right, I've noticed that too. And I'm not sure yet either about the right way. But in any case, I think the proposed changes here are independent and therefore follow-up PRs are the better way to deal with `unsequa` and `cost_benefit`.

peanutfun:
> If centroids are not assigned for that kind of hazard, they are going to be assigned regardless of the flag, hence, to me, reassign_centroids seems to be more accurate.

True, but I think this is unexpected with regard to the new method signature. If I set `[re]assign_centroids=False`, I would expect the method to never assign centroids. I think in this case the method should throw an error like

```
RuntimeError: '[re]assign_centroids' is set to 'False' but no centroids are assigned.
```

chahank:
> > the methods in the module `unsequa` must also still be adjusted. I am not yet sure what is the right way to do it. In any case, the default should be to `not` reassign the centroids. The same is true for the `cost_benefit` module. It makes little sense to repeat this costly computation for repeated impact calculations for which the centroids do not change.
> 
> You are right, I've noticed that too. And I'm not sure yet either about the right way. But in any case, I think the proposed changes here are independent and therefore follow-up PRs are the better way to deal with `unsequa` and `cost_benefit`.

I disagree that this can be done separately. With the current change, we essentially make unsequa not usable, and cost benefit partially not usable.

climada/engine/impact.py
```diff
@@ -184,7 +184,7 @@ def __init__(self,
 
 
 
-    def calc(self, exposures, impact_funcs, hazard, save_mat=False):
+    def calc(self, exposures, impact_funcs, hazard, save_mat=False, reassign_centroids=True):
```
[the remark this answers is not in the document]
emanuel-schmid: [.pylintrc](https://github.com/CLIMADA-project/climada_python/blob/main/.pylintrc)
```bash
# Maximum number of characters on a single line.
max-line-length=100
```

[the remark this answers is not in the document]
peanutfun: Fair enough!

climada/engine/impact_calc.py
```diff
@@ -101,13 +102,21 @@ def cover(self):
...
         Parameters
         ----------
-        save_mat : bool
+        save_mat : bool, optional
             if true, save the total impact matrix (events x exposures)
+            Default: True
+        assign_centroids : bool, optional
+            indicates whether centroids are re-assigned to the self.exposures object
+            or kept from previous impact calculation with a hazard of the same hazard type.
+            Centroids assignment is an expensive operation; set this to ``False`` to save
```
peanutfun, line 116: I introduced trailing whitespace, which triggers new linter warnings. Sorry about that!
```suggestion
            Centroids assignment is an expensive operation; set this to ``False`` to save
```

climada/engine/impact_calc.py
```diff
@@ -147,6 +156,13 @@ def insured_impact(self, save_mat=False):
         ----------
         save_mat : bool
             if true, save the total impact matrix (events x exposures)
+        assign_centroids : bool, optional
+            indicates whether centroids are re-assigned to the self.exposures object
+            or kept from previous impact calculation with a hazard of the same hazard type.
+            Centroids assignment is an expensive operation; set this to ``False`` to save
```
peanutfun, line 162:
```suggestion
            Centroids assignment is an expensive operation; set this to ``False`` to save
```

climada/engine/impact_calc.py
```diff
@@ -238,18 +254,29 @@ def _return_empty(self, save_mat):
...
-            name of the impact function column in exposures.gdf
-
-        """
-        self.exposures.assign_centroids(self.hazard, overwrite=False)
+        impf_col : str
+            Name of the impact function column in exposures.gdf
+        assign_centroids : bool
+            Indicates whether centroids are re-assigned to the self.exposures object
+            or kept from previous impact calculation with a hazard of the same hazard type.
+            Centroids assignment is an expensive operation; set this to ``False`` to save
```
peanutfun, line 269:
```suggestion
            Centroids assignment is an expensive operation; set this to ``False`` to save
```

climada/engine/unsequa/calc_impact.py
```diff
@@ -261,8 +261,8 @@ def _map_impact_calc(self, sample_iterrows):
         haz = self.haz_input_var.evaluate(**haz_samples)
 
         imp = Impact()
-
-        imp.calc(exposures=exp, impact_funcs=impf, hazard=haz)
+        exp.assign_centroids(haz, overwrite=False)
+        imp.calc(exposures=exp, impact_funcs=impf, hazard=haz, assign_centroids=False)
```
chahank, line 265: This is now a bit awkward, but it probably works fine. We should check whether the computation time remains stable. Have you already checked that on the test code? Shall I do some more testing?

I think the same change is needed for `unsequa.calc_cost_benefit` right?

emanuel-schmid: Yes, same, same for `calc_cost_benefit`. I just checked it in.

And yes, I have checked the computational impact, not by computation time itself which is marginal for the test cases I can get my hands on, but by number of calls to `assign_centroids` with `overwrite=True`.

And no, thanks, that should be fine. You don't have to do more testing. Unless you object I'm going to merge.

PR #529, opened 2022-08-08
Add impact classmethod to concatenate several impact objects

raphael-portmann:
**UPDATES**
- now includes tutorial example
- Now includes test
- checks are defined as separate functions
- ISSUE: due to the many checks, the code is rather complex

Changes proposed in this PR:
- add a classmethod with two concatenation options to concatenate two or more impact objects with the same hazard type, frequency, and coordinate system that either (1) have the same exposure but contain events with different dates / time periods of the year (e.g. seasonal data) or (2) have the same event dates but a different exposure (e.g. two different countries)

### To Do list

- [ ] implement exposure concatenation option 
- [x] make tests

### Open questions
- Is option (2) really needed or would one just merge the exposure before the impact calculation?
- Is it a problem that, after concatenation, the data is not sorted by date (e.g. in the impact matrix and event ids)?

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing


climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        """Concatenate impact objects of the same hazard type, frequency, 
+        exposure, and coordinate system.
 
+        Parameters
+        ----------
+        imp_list : list of climada.engine.Impact objects
+                list of Impact object to concatenate
+        concat_type: string
+                string specifying the type of concatenation to be done.
+                    'time_window': concatenate impacts for different time windows / time slices within a year (e.g. May to June, July to August,...)
```
chahank, line 1386: I think this is a bit of a confusing naming. I would expect that with the value `time_window`, I would have to actually define a time window over which the algorithm will aggregate in some way.

How about `event` ?

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        --------
+        impact: climada.engine.Impact
+               New impact object which is a concatenation of all impacts
+        """
+       
+        if len(imp_list) == 0:
+           return cls()
+       
+        if concat_type not in ['time_window','exposure']:
+           raise ValueError(f"Concatenation option {concat_type} not available.")
```
chahank, line 1398: I suppose @emanuel-schmid would like you to use an [Enum](https://docs.python.org/3/library/enum.html) class here. C.f. for instance https://github.com/CLIMADA-project/climada_python/blob/main/climada/util/lines_polys_handler.py  .

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+       
+        #initialize new impact object
+        impact=cls()
```
chahank, line 1401: You should not first initialize an empty object, and then fill it. It is better to initialize the object with all relevant attributes from the get-go, as `Impact` are static objects. 
```suggestion
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+       
+        # check impacts consistency and fill impact object
+       
+        ### the following checks are common for both concat options
+       
+        # check hazard type
+        haz_types = {imp.tag['haz'].haz_type for imp in imp_list if imp.tag['haz'].haz_type != ''}
+        if len(haz_types) > 1:
+           raise ValueError(f"The impacts are based on hazards of different types: {haz_types}. "
```
chahank, line 1410: What if haz_type = {} ? I.e., when all impacts are wrongly defined and the hazard type is not set.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+                            "The impacts are incompatible and cannot be concatenated.")      
+       
+        # check unit
+        units = {imp.unit for imp in imp_list if imp.unit != ''}
+        if len(units) > 1:
+           raise ValueError(f"The given impacts have different units: {units}. "
```
chahank, line 1416: What if some units are defined, and some not?

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+                            "The impacts are incompatible and cannot be concatenated.")
+        if len(units) == 0:
+           units = {''}
+        impact.unit = units.pop()
```
chahank, line 1420:
```suggestion
```
Initialize a proper object in the end.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+        #check coordinate reference system
+        crss={imp.crs for imp in imp_list if imp.crs != ''}
```
chahank, line 1423:
```suggestion
        crs={imp.crs for imp in imp_list if imp.crs != ''}
```

chahank: Why `crss` ?

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        if len(crss)>1:
+           raise ValueError(f"The given impacts have different crs: {crss}. "
```
chahank, line 1425:
```suggestion
        if len(crss)>1:
           raise ValueError(f"The given impacts have different crs: {crs}. "
```
What if some crs are defined and some not?

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          dates=[date for imp in imp_list for date in imp.date]
+          
+          if len(dates)!=len(set(dates)): 
+              raise ValueError("There is at least one date in more than one impact. Please make sure impacts do not contain same dates."
+                               "The impacts are incompatible and cannot be concatenated.")
+          impact.date = np.array(dates)
+          
+          
+          #check if exposure is consistent
+          isin=[np.isin(imp.coord_exp,imp_list[0].coord_exp) for imp in imp_list[0::]]
```
chahank, line 1458: This will fail. You cannot check 2-d arrays with `np.isin`, as it will consider each number as independent. What you really want to do here is to compare coordinates, so pair of numbers. 

If you want to check that they are equal, use `numpy.array_equal`.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          tot_vals=set([imp.tot_value for imp in imp_list])
+          if (not np.all(isin)) or (len(tot_vals)>1):
+               raise ValueError("Impacts are not based on the same exposure. Please make sure exposures are identical."
+                                "The impacts are incompatible and cannot be concatenated.")
+          exp_descriptions= {imp.tag['exp'].description for imp in imp_list if imp.tag['exp'].description != ''}
+          if len(exp_descriptions) > 1:
+               raise ValueError(f"The given impacts are based on exposures with different descriptions: {exp_descriptions}. "
+                            "The impacts are incompatible and cannot be concatenated.")      
+          impact.coord_exp=imp_list[0].coord_exp
```
chahank, line 1467:
```suggestion
          coord_exp = imp_list[0].coord_exp
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          event_ids=[event_id for imp in imp_list for event_id in imp.event_id]
+          if len(event_ids)!=len(set(event_ids)):
+              raise ValueError("Duplicate event IDs found.")
+          impact.event_id=np.array(event_ids)
+          
+          #event names
+          event_names=[event_name for imp in imp_list for event_name in imp.event_name]
+          if len(event_names)!=len(set(event_names)):
+              raise ValueError("Duplicate event names found.")   
+          impact.event_name = np.array(event_names,dtype=object)
```
chahank, line 1483:
```suggestion
          event_name = event_names
```
Event name is a list.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+          #impact matrix
+          imp_mats=[imp.imp_mat for imp in imp_list]
+          impact.imp_mat=sparse.vstack(imp_mats)
```
chahank, line 1487:
```suggestion
          imp_mat = sparse.vstack(imp_mats)
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          
+          #concatenate remaining attributes
+          at_events=[imp.at_event for imp in imp_list]
+          impact.at_event = np.concatenate(at_events,axis=0)
```
chahank, line 1491:
```suggestion
          at_event = np.concatenate(at_events,axis=0)
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          eai_exps=[imp.eai_exp for imp in imp_list]
+          impact.eai_exp = np.nansum(eai_exps,axis=0)
```
chahank, line 1493:
```suggestion
          eai_exp = np.nansum(eai_exps,axis=0) 
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          imp_mats=[imp.imp_mat for imp in imp_list]
+          impact.imp_mat=sparse.vstack(imp_mats)
+          
+          #concatenate remaining attributes
+          at_events=[imp.at_event for imp in imp_list]
+          impact.at_event = np.concatenate(at_events,axis=0)
+          eai_exps=[imp.eai_exp for imp in imp_list]
+          impact.eai_exp = np.nansum(eai_exps,axis=0) 
+          aai_aggs=[imp.aai_agg for imp in imp_list]
+          impact.aai_agg = np.nansum(aai_aggs)
```
chahank, line 1495:
```suggestion
          aai_agg = np.nansum(aai_aggs)
```

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        
+          raise ValueError("Concatenation of impacts with different exposures not yet implemented."
+                             "Feel free to do so!")
+     
+
+          
+
+
+
+        return impact
```
chahank, line 1514:
```suggestion
                return cls(
            event_id = event_id,
            event_name = event_name,
            date = date,
            frequency = frequency,
            coord_exp = coords,
            crs = crs,
            unit = value_unit,
            tot_value = total_value,
            eai_exp = eai_exp,
            at_event = at_event,
            aai_agg = aai_agg,
            imp_mat = imp_mat if imp_mat is not None else sparse.csr_matrix((0, 0)),
            tag = tag
      )

```
Something along these lines, you will have to update the variables.

chahank (changes requested):
Thanks @raphael-portmann ! This will come in very handy :D

For the moment, the code is still a bit raw. Please update the cosmetics (white spaces, lines, etc.) to make the code cleaner.

In addition, I suggest that you group all the checks into individual functions that can then be called. This will make the code much more readable and modular. 

I also suggest that you now write test methods. Try also to do some testing with 'real data' from colleagues to find the hidden issues quickly.

climada/engine/impact.py
raphael-portmann: I used `crss` as a grammatically awful expression of the plural of `crs`, because it can be a dictionary with more than one crs, similarly to the `units` dictionary.

climada/engine/impact.py
raphael-portmann: At the moment, the method assumes in this case that the missing crs are the same as the ones that are defined. If this is not the case, this will lead to an error when equality of exposure coordinates is checked. So I would suggest to leave this check as it is. Ok?

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+        #check coordinate reference system
+        crss={imp.crs for imp in imp_list if imp.crs != ''}
+        if len(crss)>1:
+           raise ValueError(f"The given impacts have different crs: {crss}. "
+                        "The impacts are incompatible and cannot be concatenated.")
+        impact.crs=crss.pop()
+ 
+        #check attributes
+        attributes = sorted(set.union(*[set(vars(imp).keys()) for imp in imp_list]))
```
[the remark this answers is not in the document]
raphael-portmann: Not one that I could remember :D

climada/engine/impact.py
chahank: I then suggest to drop the plural form. This just creates confusion.

climada/engine/impact.py
chahank: I would then raise a warning.

climada/engine/impact.py
raphael-portmann: I now issue a warning and continue nonetheless

climada/engine/impact.py
raphael-portmann: I now follow the same approach as for hazard type, and units. Crs has to be defined and the same in all impacts. Otherwise I issue a value error.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        for attr_name in attributes:
+           if not all(hasattr(imp, attr_name) for imp in imp_list):
+               raise ValueError(f"Attribute {attr_name} is not shared by all impacts. "
+                                "The impacts are incompatible and cannot be concatenated.")
+
+        #check if impacts have same frequency values (only first entries are checked)
+        freqs=set([imp.frequency[0] for imp in imp_list])
+        if len(freqs)>1:
+           raise ValueError(f"Impacts do not have the same frequency: {freqs}."
```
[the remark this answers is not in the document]
raphael-portmann: Hmm...It does work for time_window if the impacts are computed from the same original hazard set, right? Because then they all have the same frequency. I would argue that this is needed for concatenation because adding impacts with different frequencies does not seem to be reasonable if we don't include some kind of weighting. And numpy.array_equal does not work because impacts can have a different number of events.

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+        if len(imp_list) == 0:
+           return cls()
+       
+        if concat_type not in ['time_window','exposure']:
+           raise ValueError(f"Concatenation option {concat_type} not available.")
+       
+        #initialize new impact object
+        impact=cls()
+       
+        # check impacts consistency and fill impact object
```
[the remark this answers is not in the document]
raphael-portmann: done!

climada/engine/impact.py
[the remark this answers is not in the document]
chahank: Events in the same hazard in general do not have to have the same frequencies.

climada/engine/impact.py
raphael-portmann: agreed, I changed it using numpy.array_equal

climada/engine/impact.py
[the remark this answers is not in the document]
raphael-portmann: ok then I have to think about if in such a case all the rest of the concatenation is still correct...

climada/engine/impact.py
raphael-portmann: Hm...not sure if I understand what you mean :-)

climada/engine/impact.py
```diff
@@ -1372,7 +1372,147 @@ def _selected_exposures_idx(self, coord_exp):
...
+          impact.coord_exp=imp_list[0].coord_exp
+          impact.tot_value=imp_list[0].tot_value
+          
+          #fill remaining attributes
+          impact.tag=imp_list[0].tag
+
+          #event IDs
+          event_ids=[event_id for imp in imp_list for event_id in imp.event_id]
+          if len(event_ids)!=len(set(event_ids)):
+              raise ValueError("Duplicate event IDs found.")
```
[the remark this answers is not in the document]
chahank: If I create hazard A and hazard B, both will have their `event_ids` integers starting from 1, until the number of events `NA` and NB`. Thus, if I now want to concatenate the impacts corresponding to these two hazards, the `event_ids` are not the same length and are not unique.

climada/engine/impact.py
chahank: When you have a function that uses different choices, it is best to use an `Enum` class object to define these choices. Thus, instead of using strings to label your choices, you use objects of an `Enum` class.

raphael-portmann: Unfortunately, I'm still not sure if it makes sense in this case an how to use it...

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+                if not all(hasattr(imp, attr_name) for imp in imp_list):
+                    raise ValueError(f"Attribute {attr_name} is not shared by all impacts. "
+                                    "The impacts are incompatible and cannot be concatenated.")
+            return True
+
+        def check_frequency(imp_list):
+            freqs = {imp.frequency[0] for imp in imp_list}
+            if len(freqs) > 1:
+                raise ValueError(f"Impacts do not have the same frequency: {freqs}."
+                                 "The impacts are incompatible and cannot be concatenated.")
```
chahank, line 1470: Impacts do not have a frequency... do you mean the hazard events associated with the given impact? Or what is meant here? 

Also, I am a still bit confused why you need the same frequencies to concatenate. You just need the full impact matrices if the frequencies of the events are different right?

climada/engine/impact.py
```diff
@@ -1390,17 +1390,25 @@ def _selected_exposures_idx(self, coord_exp):
...
+                that is split into time windows using hazard.select()
 
         Parameters
         ----------
         imp_list : list of climada.engine.Impact objects
                 list of Impact object to concatenate
         concat_type: string
                 string specifying the type of concatenation to be done.
-                    'time_window': concatenate impacts for different events / time slices
+                    'time': concatenate impacts for different events / time slices
```
chahank, line 1411: I suggest to call it `event` then.

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+        def check_dates(imp_list):
+
+            #compare dates of all impact pairs
+            for i, imp1 in enumerate(imp_list):
+                for imp2 in imp_list[i+1::]:
+                    dates_a_set = set(imp1.date)
+                    dates_b_set = set(imp2.date)
+                    if len(dates_a_set.intersection(dates_b_set)) > 0:
+                        raise ValueError("Found at least one duplicate date. "
+                                         "Impacts can not contain same dates."
```
chahank, line 1483:
```suggestion
                        raise ValueError("Found at least one duplicate date. "
                                         "Impacts cannot contain same dates."
```

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+            return date
+
+        def check_exposure(imp_list):
+
+            #check total value
+            tot_vals = {imp.tot_value for imp in imp_list}
+            if len(tot_vals) > 1:
+                raise ValueError("The impacts are not based on same exposure."
+                                 "Total exposure values are different."
+                                 "The impacts are incompatible and cannot be concatenated.")
```
chahank, line 1495: I suggesst to make complete sentences as it reads much better.
```suggestion
                raise ValueError("The impacts are not based on the same exposure."
                                 "The total exposure values are different."
                                 "The impacts are incompatible and cannot be concatenated.")
```

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+
+        #if impact list is empty, return cls()
         if len(imp_list) == 0:
             return cls()
-       
-        if concat_type not in ['time_window','exposure']:
+
+        #raise error if concat_type is not available
+        #NOTE: not sure how to use an ENUM class in this context
+        if concat_type not in ['time','exposure']:
```
chahank, line 1518: I suggest to use an `Enum` class instead.

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
-        freqs=set([imp.frequency[0] for imp in imp_list])
-        if len(freqs)>1:
-            raise ValueError(f"Impacts do not have the same frequency: {freqs}."
-                            "The impacts are incompatible and cannot be concatenated.")
-        frequency = np.concatenate([imp.frequency for imp in imp_list], axis=0)
-       
+        #unclear if this check can be avoided to allow concatenating impacts
+        #with different frequencies
+        #but this affects other attributes such as eai
+        frequency = check_frequency(imp_list)
```
chahank, line 1536: This should only be done if the impact matrix is not defined. Otherwise, there is no need for identical frequencies.

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+            check_exposure(imp_list)
+
+            #set attributes related to exposure
             coord_exp = imp_list[0].coord_exp
             tot_value = imp_list[0].tot_value
-          
+
             #fill remaining attributes
-            tag=imp_list[0].tag
+            tag = imp_list[0].tag
```
chahank, line 1553: Why are you discarding all the tags? Is this intended?

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
-     
-
-          
-
+            #ad_coords=[impact.coords[ind] for ind,x in enumerate(compare_coords) \
+            #                           if not np.all(x)]
+            #imp.coord_exp=np.concatenate(self.coord_exp,ad_coords)
 
+            raise ValueError("Concatenating impacts with different exposures not yet implemented."
+                             "Feel free to do so!")
```
chahank, line 1590: If not implemented, one can simply remove it. Or, then raise a NotImplementedError.

climada/engine/test/test_impact.py
```diff
@@ -69,6 +69,119 @@ def test_from_eih_pass(self):
             np.stack([exp.gdf.latitude.values, exp.gdf.longitude.values], axis=1)
             )
 
+    def test_concat_pass(self):
```
chahank, line 72: I think it is better to separate these test into sub methods. Otherwise, if something breaks, it is hard to see what from the report.

chahank:
Great Stuff! Almost done, just a few small comments. Thanks.

chahank:
Oh, and I could not see it, but have you added an example to the tutorial?

climada/engine/impact.py
raphael-portmann: Ok, I will remove it  that will simplify some other parts of the code as well

climada/engine/impact.py
raphael-portmann: resolved because I removed the concatenation option "exposure" as it is not implemented anyways.

climada/engine/impact.py
raphael-portmann: resolved as the "exposure" concatenation option is not implemented anyways the user does not have specify a  `concat_type`  now

climada/engine/impact.py
raphael-portmann: You are probably right regarding your point with the frequencies. I remove the frequency check.

climada/engine/impact.py
raphael-portmann: ok, you are right. I remove the frequency check.

climada/engine/impact.py
```diff
@@ -1409,123 +1417,177 @@ def concat(cls, imp_list,concat_type='time_window'):
...
+            units = [imp.unit for imp in imp_list if imp.unit != '']
+            if len(units) == 0:
+                LOGGER.warning("No units specified. Assume unitless impact measure.")
+            elif len(units) < len(imp_list):
+                raise ValueError("Some impacts have a unit, some not. Same units needed."
+                                 "The impacts are incompatible and cannot be concatenated.")
+            if len(set(units)) > 1:
+                raise ValueError(f"The given impacts have different units: {units}. "
+                                 "The impacts are incompatible and cannot be concatenated.")
+            unit = set(units[0])
```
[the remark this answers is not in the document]
raphael-portmann: me neither, I removed the set()

climada/engine/impact.py
```diff
@@ -1389,6 +1389,184 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+        NOTES
+        -----
+            - Concatenation of impacts with different exposure (e.g. different countries)
+                could also be implemented here in the future.
+
+        Parameters
+        ----------
+        imp_list : list of climada.engine.Impact objects
+                list of Impact object to concatenate
```
chahank, line 1406: I think your spacing for the docstrings are off.

climada/engine/impact.py
```diff
@@ -1389,6 +1389,184 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+        # checks
+        def check_haz_types(imp_list):
+            haz_types = [imp.tag['haz'].haz_type for imp in imp_list if \
+                         imp.tag['haz'].haz_type != '']
+            if len(haz_types) < len(imp_list):
+                raise ValueError(
+                    "Not all impacts have a hazard type."
+                    "Please specify the same hazard type for each impact."
+                    "The impacts are incompatible and cannot be concatenated.")
```
chahank, line 1422: I do not really like the sentence you added to all errors `The impacts are incompatible and cannot be concatenated.` What is that supposed to mean?

climada/engine/impact.py
```diff
@@ -1389,6 +1389,184 @@ def _selected_exposures_idx(self, coord_exp):
...
+
+        #get hazard frequencies
+        frequency = np.concatenate([imp.frequency for imp in imp_list], axis=0)
+
+        #set attributes related to exposure
+        coord_exp = imp_list[0].coord_exp
+        tot_value = imp_list[0].tot_value
+
+        #fill remaining attributes
+        tag = imp_list[0].tag
```
chahank, line 1528: I still think it is not ok to discard all tags other than for the first instance of impact.

climada/engine/test/test_impact.py
```diff
@@ -69,6 +69,123 @@ def test_from_eih_pass(self):
...
+        """test concatenation of impacts"""
+
+        def test_haztype_unit_crs_attrs(self,imp1,imp2,imp3,imp3_3,imp3_2):
```
chahank, line 75:
```suggestion
        def test_haztype_unit_crs_attrs(self, imp1, imp2, imp3, imp3_3, imp3_2):
```

climada/engine/test/test_impact.py
```diff
@@ -69,6 +69,123 @@ def test_from_eih_pass(self):
...
+                imp3.crs = ''
+                Impact.concat([imp1, imp2, imp3])
+            with self.assertRaises(ValueError):
+                imp3.crs = 'OTHER'
+                Impact.concat([imp1, imp2, imp3])
+            imp3.crs = imp2.crs
+            with self.assertRaises(ValueError):
+                imp3.new_attr = 'test_attr'
+                Impact.concat([imp1, imp2, imp3])
+            delattr(imp3, "new_attr")
```
chahank, line 97: This test is very very cryptic. What are the errors that raised and for what?

You should test the specific error raised (check the error message). Also, the code is not readable, so please choose better names, add some comments, and maybe breakup in smaller bits.

climada/engine/impact.py
peanutfun: I second this. The error traceback shows that it occurs within the `concat` method, so there is no need to mention that it did not work as expected 😉

climada/engine/impact.py
```diff
@@ -1389,6 +1389,184 @@ def _selected_exposures_idx(self, coord_exp):
...
+        event_ids = [event_id for imp in imp_list for event_id in imp.event_id]
+        if len(event_ids) != len(set(event_ids)):
+            raise ValueError("Duplicate event IDs found.")
+        event_id = np.array(event_ids)
+
+        #event names
+        event_names = [event_name for imp in imp_list for event_name in imp.event_name]
+        if len(event_names) != len(set(event_names)):
+            raise ValueError("Duplicate event names found.")
+        event_name = event_names
```
peanutfun, line 1540: Again, this is the same operation just on a different impact attribute. We could write one function for that.

climada/engine/impact.py
```diff
@@ -1389,6 +1389,184 @@ def _selected_exposures_idx(self, coord_exp):
...
+                for imp2 in imp_list[i + 1::]:
+                    if not np.array_equal(imp1.coord_exp, imp2.coord_exp):
+                        raise ValueError("The impacts are not based on same exposure."
+                                         "Impacts are incompatible and cannot be concatenated.")
+            #compare exposure descriptions
+            exp_desc = {imp.tag['exp'].description for imp in imp_list
+                        if imp.tag['exp'].description != ''}
+            if len(exp_desc) > 1:
+                LOGGER.warning("Impacts are based on exposures with different descriptions."
+                               "Concatenate anyways.")
```
peanutfun, line 1494: Shouldn't the user be aware of that? I don't see how this information would help anybody 😅

climada/engine/test/test_impact.py
peanutfun: Check for specific errors like this:
```python
with self.assertRaises(SomeError) as cm:
    raise_error()
self.assertIn("part of a specific error message I was expecting", str(cm.exception))
```

climada/engine/test/test_impact.py
```diff
@@ -69,6 +69,123 @@ def test_from_eih_pass(self):
             np.stack([exp.gdf.latitude.values, exp.gdf.longitude.values], axis=1)
             )
 
+    def test_concat_pass(self):
```
peanutfun, line 72: Maybe define a new test class instead of a single test? This would make it easier to narrow down the issue if only a single one of the tests is failing

peanutfun (changes requested):

climada/engine/test/test_impact.py
```diff
@@ -69,6 +69,134 @@ def test_from_eih_pass(self):
...
+        self.assertIn("Impact matrices do not have the same number of exposure points", str(cm.exception))
+
+
+    def test_results(self):
+        """test results of impact.concat"""
+
+        # get test data
+        ent = Entity.from_excel(ENT_DEMO_TODAY)
+        ent.check()
+        hazard = Hazard.from_hdf5(HAZ_TEST_TC)
```
peanutfun, line 169: These are unit tests, so I think you should create your own impact objects with controlled data instead of some random input. You do not need hazards or exposures for that. Simply fill your impact objects with data you know and test the data in the resulting impact object explicitly. Note that `assert_array_equal` also compares the array shapes/lengths. Please write tests allong these lines (with the other Impact attributes added as well):
```python
impact_1 = Impact(frequency=[0.1])
impact_2 = Impact(frequency=[0.2])

impact = Impact.concat(impact_1, impact_2)
np.testing.assert_array_equal(impact.frequency, [0.1, 0.2])
```
This goes for all tests. If you want to check that concatenation works with "real world" examples, this should go into an integration test

climada/engine/impact.py
```diff
@@ -1389,6 +1389,114 @@ def _selected_exposures_idx(self, coord_exp):
...
+        else:
+            if len(event_ids) != len(set(event_ids)):
+                raise ValueError("Duplicate event IDs found.\
+                Consider to set reset_event_id = True.")
+            event_id = np.array(event_ids)
+
+        # concatenate impact matrices
+        imp_mats =[imp.imp_mat for imp in imp_list]
+        if None in imp_mats:
+            raise ValueError("Passed an impact object with an undefined impact matrix.")
```
peanutfun, line 1457: I just checked this: The [Impact constructor](https://github.com/CLIMADA-project/climada_python/blob/develop/climada/engine/impact.py#L190) sets the `imp_mat = sparse.csr_matrix(np.empty((0, 0)))` if no impact matrix is passed, so this `Impact.imp_mat` attribute should never be `None`!
```suggestion
```

climada/engine/impact.py
```diff
@@ -1389,6 +1389,114 @@ def _selected_exposures_idx(self, coord_exp):
             LOGGER.warning("No exposure coordinates match the selection.")
         return sel_exp
 
+    @classmethod
+    def concat(cls, imp_list, reset_event_ids = False):
+        """
+        Concatenate impact objects with the same exposure.
```
peanutfun, line 1395: Single line as docstring "title"
```suggestion
        """Concatenate impact objects with the same exposure

```

climada/engine/impact.py
```diff
@@ -1389,6 +1389,114 @@ def _selected_exposures_idx(self, coord_exp):
...
+        This function is useful if, e.g. different impact functions
+        have to be applied for different seasons (e.g. for agricultural impacts).
+
+        It concatenates the following attributes:
+            event_id, event_name, date, frequency, imp_mat, at_event, eai_exp, aai_exp
+        and takes the following attributes from the first impact object in the passed imp_list
+            coord_exp, crs, unit, tot_value, tag
```
peanutfun, line 1402: Please add some explanations on how this works:
1. What happens if all impact matrices are empty?
2. Several values are *not* concatenated but *summed up*
3. What happens to the event IDs?

doc/tutorial/climada_engine_Impact.ipynb
```diff
@@ -341,12 +509,12 @@
...
-       "1  2.990876e+10  POINT (-82.29167 23.12500)    23.125 -82.291667        192   \n",
-       "2  6.839380e+09  POINT (-82.20833 23.12500)    23.125 -82.208333        192   \n",
-       "3  1.925246e+09  POINT (-82.12500 23.12500)    23.125 -82.125000        192   \n",
-       "4  2.422598e+08  POINT (-82.04167 23.12500)    23.125 -82.041667        192   \n",
+       "          value                    geometry   latitude  longitude  region_id  \\\n",
+       "0  1.077368e+05  POINT (-81.37500 21.70833)  21.708333 -81.375000        192   \n",
+       "1  1.671873e+06  POINT (-81.54167 21.62500)  21.625000 -81.541667        192   \n",
+       "2  3.421208e+06  POINT (-82.95833 21.87500)  21.875000 -82.958333        192   \n",
+       "3  1.546590e+07  POINT (-82.87500 21.87500)  21.875000 -82.875000        192   \n",
+       "4  7.168305e+07  POINT (-82.79167 21.87500)  21.875000 -82.791667        192   \n",
```
peanutfun, line 517: What's happening here? Why do exposure values change by several orders of magnitude, along with the coordinates? I get the same result on my machine. Seems like these results were from early 2021. But what changed since then? All parameters in the function calls are the same.

Any ideas, @emanuel-schmid @tovogt?

doc/tutorial/climada_engine_Impact.ipynb
```diff
@@ -1016,11 +1237,11 @@
...
       "text/plain": [
-       "<GeoAxesSubplot:title={'center':'Expected annual impact'}>"
+       "<Figure size 900x1300 with 2 Axes>"
       ]
      },
-     "execution_count": 14,
```
peanutfun, line 1023: Something is off with this notebook. The documentation cannot be built because one of the execution counts is missing somewhere. Please do the following:
1. Open a new jupyter notebook instance
2. Restart the kernel
3. Open this file and clear *all* outputs
4. Run all cells from start to finish
5. Save the file
6. Commit the file

peanutfun (changes requested):
Almost there! Please use "handmade" test data and check that concetation works based on the resulting *values*  and not just the lengths of the arrays. Also, the docstring could explain a little bit better what's happing internally. Please make sure that the docs can be built and check the appearance of the docstring in the preview

doc/tutorial/climada_engine_Impact.ipynb
tovogt: This is not related to this PR. It even happens when rolling back the source code to the last time that this notebook has been changed (in early 2022). I don't have time to dig any deeper, unfortunately. But when looking at the plot output below this cell, the changes are quite moderate (but at least big enough to spot differences with your bare eyes). I would suggest to continue this discussion somewhere else, on Slack or in a separate GitHub issue.

peanutfun: Great, thanks for the quick feedback! 🙌

peanutfun (approved):

peanutfun:
I reworked this a bit, tidying up the code and the docstring and improving the tests. Now even the linter is happy! Merging...

peanutfun:
@raphael-portmann Thanks for your contribution! 🥳

PR #531, opened 2022-08-15
Update tc_tracks.py

chahank:
Proposal: start numbering the year as in the data file. This way, the years can be reconstructed, a process needed to for instance study the time-correlations of storms or study changing vulnerabilities.

This PR fixes issue with STORM datasets having all years set to 1980.

### Pull Request Checklist

- [ ] Docs updated
- [ ] Tests updated
- [ ] Tests passing
- [ ] No new linter issues

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -953,6 +953,20 @@ class TCTracks():
         -------
         tracks : TCTracks
             TCTracks with data from the STORM simulations.
+
+        Notes
+        -----
+        All tracks are set in the year 1980. The id of the year (starting from 0) is saved in the
+        attribute 'id_no'. To obtain the year of each track use e.g.
+        `years = [int(tr.attrs['id_no'] / 1000) for tr in tc_tracks.data]`
+
+        If a windfield is generated from these tracks using the method `TropCylcone.from_tracks()`,
+        the following should be considered:
+        1- the frequencies will be set to "1" for each storm. Thus, in order to compute annual values,
+           the frequencies of the TropCylone should be changed to "1/number of years".
+        2- the storm year and the storm id are stored in the 'TropCyclone.event_name' attribute
+
+
         """
         basins = ["EP", "NA", "NI", "SI", "SP", "WP"]
         tracks_df = pd.read_csv(path, names=['year', 'time_start', 'tc_num', 'time_delta',
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -963,13 +963,12 @@ class TCTracks():
         >>> # or, alternatively, 
         >>> years = [int(tr.attrs['sid'].split("-")[-2]) for tr in tc_tracks.data]
 
-        If a windfield is generated from these tracks using the method `TropCylcone.from_tracks()`,
+        If a windfield is generated from these tracks using the method ``TropCylcone.from_tracks()``,
         the following should be considered:
+
         1. The frequencies will be set to ``1`` for each storm. Thus, in order to compute annual
-        values, the frequencies of the TropCylone should be changed to ``1/number of years``.
+           values, the frequencies of the TropCylone should be changed to ``1/number of years``.
         2. The storm year and the storm id are stored in the ``TropCyclone.event_name`` attribute.
-
-
         """
         basins = ["EP", "NA", "NI", "SI", "SP", "WP"]
         tracks_df = pd.read_csv(path, names=['year', 'time_start', 'tc_num', 'time_delta',
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -962,9 +962,9 @@ class TCTracks():
 
         If a windfield is generated from these tracks using the method `TropCylcone.from_tracks()`,
         the following should be considered:
-        1- the frequencies will be set to "1" for each storm. Thus, in order to compute annual values,
-           the frequencies of the TropCylone should be changed to "1/number of years".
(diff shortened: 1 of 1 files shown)


tovogt:
This won't work because years are numbered starting from 0 and this is not an actual year. Furthermore, the convenient datetime-accessors of pandas are only available for columns of dtype `datetime64` and that dtype does only cover years between 1677 and 2262 (https://pandas.pydata.org/docs/user_guide/timeseries.html#timestamp-limitations).

However, the current solution is to store the year information in the `id_no` attribute of each track. That allows to do any kind of statistics that you might be interested in. You obtain the years of all tracks in a `TCTracks` object like this:
```python
years = [int(tr.attrs['id_no'] / 1000) for tr in tracks.data]
```

chahank:
I see, thanks for the update!
Then my follow-up thoughts:
1- we should state this very clearly in the docstring. it is very hidden now 
2- why is the id number multiplied by 1000? 
3- how do we then produce usable `TropCyclone` objects? The method `TropCyclone.from_tracks` reads the `track.time` attribute, which is unfortunately set to always be in `1980`. Do you propose to manually re-place the date afterward? Also, the frequency is obtained from the `track.time` attribute, which will be wrongly set. Should then the frequency also be manually reset?

tovogt:
Yes, we should mention this in the docstring of `from_simulations_storm`. Feel free to propose this as part of this PR. 

The year information in STORM is no actual time information. It should probably be named `year_id`, because it's just numbering the years starting at 0. It is not meant to be an actual time information which would actually give an *order* to the tracks. Compared to that, the month and day information provided in the STORM data is an actual time informatio, so it's stored in the actual time variable.

The year information is also stored in the `sid` attribute, so it can be recovered from the `event_name` in the Hazard object. Like this:
```python
storm_year_ids = [n.split("-")[-2] for n in haz.event_name]
```

tovogt:
The problem of the frequency being wrong is really inconvenient. However, one set of STORM tracks might contain much more than the maximum allowed ~600 years that are supported by pandas. So, there is really no way to store that information in the datetime information in a consistent way.

chahank:
Ok, thanks for the clarifications! I think then we can set the frequency manually in the date provide in the API. That the year is stored in the event_name is convenient then. I can update the docstrings too. Thanks!

chahank:
I just removed the incorrect code addition and updated the docstrings. Do you think this is clear/correct enough?

climada/hazard/tc_tracks.py
```diff
@@ -953,6 +953,20 @@ def from_simulations_storm(cls, path, years=None):
         -------
         tracks : TCTracks
             TCTracks with data from the STORM simulations.
+
+        Notes
+        -----
+        All tracks are set in the year 1980. The id of the year (starting from 0) is saved in the
+        attribute 'id_no'. To obtain the year of each track use e.g.
+        `years = [int(tr.attrs['id_no'] / 1000) for tr in tc_tracks.data]`
```
tovogt, line 961:
```suggestion
        attribute 'id_no'. To obtain the year of each track use
        
        >>> years = [int(tr.attrs['id_no'] / 1000) for tr in tc_tracks.data]
        >>> # or, alternatively, 
        >>> years = [int(tr.attrs['sid'].split("-")[-2]) for tr in tc_tracks.data]
```

climada/hazard/tc_tracks.py
```diff
@@ -953,6 +953,20 @@ def from_simulations_storm(cls, path, years=None):
...
+
+        If a windfield is generated from these tracks using the method `TropCylcone.from_tracks()`,
+        the following should be considered:
+        1- the frequencies will be set to "1" for each storm. Thus, in order to compute annual values,
+           the frequencies of the TropCylone should be changed to "1/number of years".
+        2- the storm year and the storm id are stored in the 'TropCyclone.event_name' attribute
```
tovogt, line 967: Use proper reST formatting:
```suggestion
        1. the frequencies will be set to ``1`` for each storm. Thus, in order to compute annual values,
           the frequencies of the TropCylone should be changed to ``1/number of years``.
        2. the storm year and the storm id are stored in the ``TropCyclone.event_name`` attribute.
```
Use double backticks for monospace literals.

tovogt (commented):
See https://docutils.sourceforge.io/docs/user/rst/demo.txt for instructions on how to use reST formatting.

climada/hazard/tc_tracks.py
chahank: Cool, thanks!

climada/hazard/tc_tracks.py
```diff
@@ -953,6 +953,23 @@ def from_simulations_storm(cls, path, years=None):
...
+        >>> # or, alternatively, 
+        >>> years = [int(tr.attrs['sid'].split("-")[-2]) for tr in tc_tracks.data]
+
+        If a windfield is generated from these tracks using the method `TropCylcone.from_tracks()`,
+        the following should be considered:
+        1. The frequencies will be set to ``1`` for each storm. Thus, in order to compute annual
+        values, the frequencies of the TropCylone should be changed to ``1/number of years``.
+        2. The storm year and the storm id are stored in the ``TropCyclone.event_name`` attribute.
+
+
```
tovogt, line 972: I played around with the formatting in https://livesphinx.herokuapp.com/ and found that the spacing has to be a bit different, actually:
```suggestion
        If a windfield is generated from these tracks using the method ``TropCylcone.from_tracks()``,
        the following should be considered:

        1. The frequencies will be set to ``1`` for each storm. Thus, in order to compute annual
           values, the frequencies of the TropCylone should be changed to ``1/number of years``.
        2. The storm year and the storm id are stored in the ``TropCyclone.event_name`` attribute.
```

Issue #516, opened 2022-07-13
Frequency Unit


emanuel-schmid:
The frequency unit is globally set to 'annual' (y<sup>-1</sup>).
We need to make this a variable.
Task: introduce a `frequency_unit` being part of the meta data of the affected CLIMADA objects.

chahank:
Can it be a property instead of a meta data?

emanuel-schmid:
Sure, the term "meta data" is used in a general sense, without implementation bias.

PR #532, opened 2022-09-05
Feature/frequency unit

emanuel-schmid:
Changes proposed in this PR:
- Introduce a field `frequency_unit` in classes `Impact`, `ImpactFreqCurve`
  The fields have no impact on the calculation whatsoever, they are purely informative.
  Nevertheless concatenation of `Hazard` objects is inhibited if they have different `frequency_unit` values.

This PR fixes issue #516

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing
- [x] No new linter issues


climada/engine/impact.py
```diff
@@ -842,11 +859,12 @@ def write_col(i_col, imp_ws, xls_data):
         write_col(7, imp_ws, self.event_name)
         write_col(8, imp_ws, self.date)
         write_col(9, imp_ws, self.frequency)
-        write_col(10, imp_ws, self.at_event)
-        write_col(11, imp_ws, self.eai_exp)
-        write_col(12, imp_ws, self.coord_exp[:, 0])
-        write_col(13, imp_ws, self.coord_exp[:, 1])
-        write_col(14, imp_ws, [str(self.crs)])
+        write_col(10, imp_ws, [self.frequency_unit])
```
chahank, line 862: Would it not be good to have the frequency unit at the end, such that files that have been created so far are still readable by custom scripts?

climada/engine/test/test_impact.py
```diff
@@ -171,7 +176,6 @@ def test_impact_per_year_sum(self):
         imp.at_event[7] = 0.381063674256423e9
         imp.at_event[8] = 0.569142464157450e9
         imp.at_event[9] = 0.467572545849132e9
-        imp.unit = 'USD'
```
chahank, line 174: What is the reason to remove the unit here? Is it not needed?

climada/engine/impact.py
```diff
@@ -445,29 +453,39 @@ def calc_freq_curve(self, return_per=None):
...
+            frequency_unit=self.frequency_unit,
+            label='Exceedance frequency curve'
+        )
 
-        return ifc
+    def _eai_title(self):
+        if self.frequency_unit in ['1/year', 'annual', '1/y', '1/a']:
+            return 'Expected annual impact'
+        if self.frequency_unit in ['1/day', 'daily', '1/d']:
+            return 'Expected daily impact'
```
chahank, line 482:
```suggestion
        if self.frequency_unit in ['1/day', 'daily', '1/d']:
            return 'Expected daily impact'
        if self.frequency_unit in ['1/month', 'monthly', '1/m']:
            return 'Expected monthly impact'
```

climada/hazard/base.py
```diff
@@ -197,6 +201,23 @@ def __init__(self, haz_type='', pool=None):
...
+        attribute : str
+            attribute name
+
+        Returns
+        ------
+        Any
+        """
+        return {
+            'frequency_unit': DEF_FREQ_UNIT,
+        }.get(attribute)
```
chahank, line 219: I am not sure to understand why this is needed, and if yes, then why is not needed for the other classes with a frequency unit?

climada/hazard/base.py
```diff
@@ -773,6 +798,10 @@ def select(self, event_names=None, event_id=None, date=None, orig=None,
 
         # reset frequency if date span has changed (optional):
         if reset_frequency:
+            if self.frequency_unit not in ['1/year', 'annual', '1/y', '1/a']:
+                LOGGER.warning("resetting the frequency on a hazard object who's frequency unit"
+                    "is %s and not %s will most likely lead to unexpected results",
```
chahank, line 803: I do not really like the "will lead to unexpected results". This very vague. Can we make it more precise you think?

doc/tutorial/climada_engine_Impact.ipynb
```diff
@@ -66,11 +66,12 @@
...
     "| :- | :- | :- |\n",
     "| tag |(dict)| dictionary storing the tags of the inputs (Exposure.tag, ImpactFuncSet.tag Hazard.tag)|\n",
-    "| even_id |list(int)| id (>0) of each hazard event (Hazard.event_id) |\n",
+    "| event_id |list(int)| id (>0) of each hazard event (Hazard.event_id)|\n",
     "| event_name |(list(str))| name of each event (Hazard.event_name)|\n",
     "| date |np.array| date of events (Hazard.date)|\n",
     "| coord_exp |np.array| exposures coordinates [lat, lon] (in degrees) (Exposure.gdf.latidues, Exposure.gdf.longitude)|\n",
-    "| frequency |np.array| annual frequency of events (Hazard.frequency)|\n",
+    "| frequency |np.array| frequency of events (Hazard.frequency)|\n",
+    "| frequency_unit |str| unit of event frequency, by default '1/year', i.e., annual, this attribute is purely informative and does not effect calculations (Hazard.frequency_unit)|\n",
```
chahank, line 74: Is it really needed to the "this attribute is purely informative..." ? For instance, this is also the case for the `event_id`, the `unit` or the `date`. And this can change in the future, in particular if time is integrated more deeply into CLIMADA.

chahank:
Thanks for the changes! Very useful. A few small comments.

climada/engine/impact.py
```diff
@@ -21,6 +21,7 @@
 
 __all__ = ['ImpactFreqCurve', 'Impact']
 
+from dataclasses import dataclass, field
```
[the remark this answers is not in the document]
emanuel-schmid: In general, `@dataclass` can be seen as a code generator for `__init__()` and other methods. Using it avoids boilerplate code ans synchronization failures in future modifications. And pydoc gets simpler.

In particular, it was introduced for `ImpactFreqCurve`. As it provides an `__init__` method for this class which goes along with general efforts to initialize objects properly from the start.

climada/engine/test/test_impact.py
emanuel-schmid: exactly.

climada/hazard/base.py
emanuel-schmid: Good question! Can we? The thing is that the user is sending contradicting signals at this point. Why do they set a frequency_unit other than 'annual' and then still use the method in a way that is strongly related to 'annual' frequencies.
I'd say they have to look and see themselves. 🤷

doc/tutorial/climada_engine_Impact.ipynb
emanuel-schmid: Just wanted to emphasize that the frequency_unit is not being considered at any point in calculations. All it really does is change labels in plots. 
As a naive user, when I can set a frequency_unit in an object and I get a frequency, I might naturally suspect the two to be somehow correlated. Which is not the case. Therefore the warning.

emanuel-schmid:
@chahank  Thanks for reviewing! 🎉

climada/hazard/base.py
emanuel-schmid: That's just for the `clear` method, to be able to restore the default. It could have been integrated right there, but I somehow liked it better outside. In case another Hazard is by default on a 'daily' base, it seems to be more natural to overwrite `get_default` then `clear`.
The other classes with a `frequency_unit` don't have a `clear` method. That's why.

climada/engine/impact.py
chahank: Ok, fine for me.

doc/tutorial/climada_engine_Impact.ipynb
chahank: But this is also the case for other entries. Thus, I would remove it as it is confusing why this is said for that entry, but not for others. Furthermore, there will likely be methods in the future that will actually care about the frequency unit.

climada/hazard/base.py
chahank: I think we could say something like:

"The frequency unit will be changed from %s to %s. The resulting frequency will thus be most likely incorrect. Consider setting reset_frequency=False and set the frequency manually for the selection."

PR #533, opened 2022-09-05
Update coordinates.py

chahank:
It is important to note that for the method `latlon_bounds` which calls the method `lon_bounds` the `lon` data is modified inplace. This is very unexpected behaviour and led to errors in different CLIMADA scripts.


chahank:
Mistake in the pull request. Deleted.

PR #538, opened 2022-09-14
Feature/tc equal timestep check

bguillod:
Changes proposed in this PR:
- When TCTracks.equal_timestep is called, the method first checks is the instance is already at the requested temporal resolution. If it is, this is communicated to the user and nothing is done.
- This avoid spending resources to equalize timesteps when they are already equalized.

### Pull Request Checklist

- [x] Docs updated
- [x] Tests updated
- [x] Tests passing
- [x] No new linter issues


bguillod:
Note that I think tests pass, but on Jenkins some tests fail. I do not thing this is related to this PR though.
I am not aware of any documentation to be updated.

tovogt:
Good idea!

Why don't we check this for each track separately?

Furthermore, how about mentioning this in the docstring of `equal_timestep`? "Note that nothing is done for tracks whose current time step size already agrees with the requested time step size."

ChrisFairless:
Will this fail when processing tracks that have an extra timestep added for landfall? e.g. IBTrACS does this.

tovogt:
@ChrisFairless What do you mean by "fail"? My guess is that it would find that not all time steps are equal or that some disagree with the given step size and then it would interpolate all tracks.

ChrisFairless:
Hey! Reviving this thread to get the request merged.

I've checked the code and my earlier question about landfall frames didn't make sense. I was worried about its behaviour with irregular timesteps (e.g. when an extra frame is added for landfall) but the interpolation method just removes them.

So this works fine. @tovogt yes it might make sense to run the checks on individual tracks, but it's not often you'll have a TCTracks object where some tracks are interpolated and others aren't.

tovogt:
I pushed a change that makes the check for each track individually: 780f731c

@ChrisFairless As you already know, IBTrACS inserts an extra step at each time of landfall. That means that raw data from IBTrACS will be "almost" 3-hourly. For many tracks, an interpolation to 3-hourly resolution can be skipped. For example:

```python
>>> tc_track = TCTracks.from_ibtracs_netcdf(year_range=(2005, 2005))
>>> tc_track.equal_timestep(time_step_h=3)
2022-12-15 14:17:14,426 - climada.hazard.tc_tracks - INFO - 93 tracks are already at the requested temporal resolution.
2022-12-15 14:17:14,427 - climada.hazard.tc_tracks - INFO - Interpolating 21 tracks to 3h time steps.
```

ChrisFairless:
@tovogt brilliant, thanks!!

And yeah, hopefully this method is only used to interpolate to timesteps << 1 hour. You risk losing important landfall information interpolating IBTrACS to e.g. 3 hours.

If we think people might unintentionally be dropping landfall frames, maybe we should add an option where the interpolated frames are _added_ to the existing track, instead of replacing it.

But for now I think we're ready to merge :)

climada/hazard/tc_tracks.py
```diff
@@ -1056,7 +1068,24 @@ def equal_timestep(self, time_step_h=1, land_params=False, pool=None):
...
+            None if np.allclose(np.unique(tr['time_step'].values), time_step_h)
+            else time_step_h
+            for tr in self.data
+        ]
+
+        n_skip = np.sum([ts is None for ts in l_time_step_h])
+        if n_skip == self.size:
+            LOGGER.info('All tracks are already at the requested temporal resolution.')
+            return
+        elif n_skip > 0:
```
emanuel-schmid, line 1083:
```suggestion
        if n_skip > 0:
```

emanuel-schmid:
Thank you all! 😃

PR #546, opened 2022-10-14
Draft: Set all Hazard attributes in __init__

peanutfun:
*WIP: This serves as an example draft*

Changes proposed in this PR:
* Update `Hazard.__init__` to take all data as arguments, skipping the null-initialization.
* Default-initialize all of the attributes that do not require user input.
* Perform a self-check after initialization by default. Users can opt to skip the check by setting `fast=True`.

### To Do
- [ ] Remove superfluous checks from `Hazard.from_raster_xarray`

### PR Author Checklist

### PR Reviewer Checklist

[contrib]: /CONTRIBUTING.md

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -159,10 +159,10 @@ class Hazard():
     scalar, string, list, 1dim np.array of size num_events."""
 
     def __init__(self,
-                 haz_type: str,
                  intensity: sparse.csr_matrix,
                  centroids: Centroids,
                  date: np.ndarray,
+                 haz_type: str = "",
                  units: str = "",
                  frequency: Optional[np.ndarray] = None,
                  frequency_unit: str = DEF_FREQ_UNIT,
@@ -182,14 +182,15 @@ class Hazard():
 
         Parameters
         ----------
-        haz_type : str
-            String indicating the hazard type (e.g., ``TC`` for tropical cyclone)
         intensity : sparse.csr_matrix
             The intensity of the hazard for each event. Shape ``(n_event, n_centroids)``
         centroids : Centroids
             The centroids on which this hazard is defined.
         date : np.array of int
             The ordinal representation of the date of each event.
+        haz_type : str (optional)
+            String indicating the hazard type (e.g., ``TC`` for tropical cyclone). This
+            parameter is passed to the hazard ``Tag``.
         units : str (optional)
             The physical units of the intensity (used for plotting)
         frequency : np.array of float (optional)
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1504,6 +1504,8 @@ class Hazard():
         num_cen = self.centroids.size
         if np.unique(self.event_id).size != num_ev:
             raise ValueError("There are events with the same identifier.")
+        if np.any(self.event_id < 1):
+            raise ValueError("Event IDs must be larger than zero")
 
         u_check.check_oligatories(self.__dict__, self.vars_oblig, 'Hazard.',
                                   num_ev, num_ev, num_cen)
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -27,6 +27,7 @@ import itertools
 import logging
 import pathlib
 import warnings
+from typing import List, Optional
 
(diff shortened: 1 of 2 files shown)


climada/hazard/base.py
```diff
@@ -157,49 +158,88 @@ class Hazard():
     """Name of the variables that aren't need to compute the impact. Types:
     scalar, string, list, 1dim np.array of size num_events."""
 
-    def __init__(self, haz_type='', pool=None):
-        """
-        Initialize values.
+    def __init__(self,
+                 haz_type: str,
```
peanutfun, line 162: This is the only (?) major change of the interface so far. Here, I'm requiring users to specify the hazard type. I opted for this because I feel like the hazard type is a very important bit of information. I could give it an empty string as default value, but then I would need to move the argument behind the parameters without default types. That would make it look "less important" to me.

@chahank, what do you think about this?

chahank: Actually, to be consistent with the previous API, all attributes must have the same defaults as before. I think changing the defaults (and removing some) is a major change, and should not be done lightly. I would, although it is still sub-optimal, for this update, just keep all the defaults.

climada/hazard/base.py
```diff
@@ -157,49 +158,88 @@ class Hazard():
...
+                 date: np.ndarray,
+                 units: str = "",
+                 frequency: Optional[np.ndarray] = None,
+                 frequency_unit: str = DEF_FREQ_UNIT,
+                 orig: Optional[np.ndarray] = None,
+                 event_id: Optional[np.ndarray] = None,
+                 event_name: Optional[List[str]] = None,
+                 fraction: Optional[sparse.csr_matrix] = None,
+                 pool: Optional[Pool] = None,
+                 fast: bool = False,
```
chahank, line 174: I am not convinced by the addition of this variable. We have this no-where in the code, and I also have not seen that in other Python modules.

climada/hazard/base.py
peanutfun: Then I have a follow-up issue: The `haz_type` is actually an attribute of the `Tag`, not the Hazard. So for keeping the previous defaults, we could simply remove the `haz_type` from the argument list. It will then become part of the `tag_kwargs`. However, that obscures this very important variable. What to do?

climada/hazard/base.py
peanutfun: Is your issue the existence of the variable or its name? With the full initialization of the object we now have the chance to perform an automated check of the resulting object. I think this is very important and should only be skipped if a user is absolutely sure that the data is consistent.

I don't think the current checks are very expensive. So if you opt for removing this variable, I would be in favor of *always* performing the check.

climada/hazard/base.py
chahank: I do not understand. Currently, the default is `haz_type=''`, and it is a variable of the `__init__`... why would that be a problem now? You can just leave it as it is right?

climada/hazard/base.py
chahank: I agree. This is the route we took so far with `Impact`.

climada/hazard/base.py
peanutfun: This is the [same default as in `Tag`](https://github.com/CLIMADA-project/climada_python/blob/main/climada/hazard/tag.py#L42). Such double-defaults are dangerous because the default in `Hazard` would ignore any changes to the default in `Tag`.

Another viable option I see is to have users insert a proper `Tag` object as optional parameter. If none is passed, we initialize a default one (this also default-inits the `haz_type`).

chahank: Yes, but this is a problem not to be solved here. The whole `Tag` needs a complete overhaul. This is also in the pipeline ;).

peanutfun: See 06c40654146bc41134af06a99b673f7023bf3e32

chahank:
I do not dislike the idea per se, but I do not know of other larger Python packages that do this. Is there a reason maybe?

peanutfun:
Superseded by #549, which is a simpler implementation. We'll come back to this

PR #551, opened 2022-10-25
read_raster_sample: fix antimeridian and implement raster gradients

tovogt:
Changes proposed in this PR:
- Fix the antimeridian problem in the utility function `u_coord.read_raster_sample` by making use of the already existing and antimeridian-proof function `u_coord.read_raster_bounds`.
- Implement a version of `u_coord.read_raster_sample` (called `u_coord.read_raster_sample_with_gradients`) that also returns the gradient of the raster at the given points.

The raster gradient feature will be used in the upcoming new TC rainfall model, since the rainfall depends on the slope of the terrain.

### PR Author Checklist

- [x] [Tests][testing] passing (note that there are some tests failing that are not related to this PR)

### PR Reviewer Checklist

[contrib]: /CONTRIBUTING.md


climada/util/coordinates.py
```diff
@@ -1822,6 +1821,39 @@ def _read_raster_reproject(src, src_crs, dst_meta, band=None, geometry=None, dst
...
+    supported_suffixes = {
+        ".gz": "gzip",
+        ".zip": "zip",
+        ".tar": "tar",
+        ".tgz": "tar",
+    }
+    suffix = Path(path).suffix
+    if suffix in supported_suffixes:
+        path = f"/vsi{supported_suffixes[suffix]}/{path}"
+    return str(path)
```
tovogt, line 1855: Previously, the prefixing was manually applied for `gz`-files in three places. But we can easily add support for other archive formats and we can avoid redundant code by introducing the helper function.

Note that the prefix "looks like a path". Since the result with the added prefix is not really a path, I would not convert it to a `Path` object. And for consistency, I would also convert to a `str` in the cases where no prefix is added at all.

tovogt (commented):
I also refactored a different aspect of the raster IO code which is related to input files that come in compressed form. See my comment below.

climada/util/coordinates.py
```diff
@@ -1996,105 +2024,165 @@ def read_raster_bounds(path, bounds, res=None, bands=None, resampling="nearest",
...
+        to meters if the raster's CRS is EPSG:4326 (lat/lon). Default: False
 
     Returns
     -------
-    values : np.array of same length as lat
+    values : np.array of shape (npoints,)
         Interpolated raster values for each given coordinate point.
+    gradient : np.array of shape (npoints, 2), optional
+        If grad=True, the raster gradient at each of the given coordinate points is returned.
+        The first/second value in each row is the derivative in lat/lon direction (lat is first!).
```
emanuel-schmid, line 2108: I prefer methods returning unambiguous data types, or if need be an Option thereof. So that after `y = f(x)` `y` is always of type `T` (or `None`). Here `y` could be an array or a pair of arrays.

In this particular case I would have preferred two methods sharing the job, something like this:

```python
values = read_raster_sample_without_gradient(...)
gradient = read_raster_sample_gradient(values, ...)
```

Since I assume that would be a waste of cpu in some way, I'd suggest to always return a pair 
```python 
def read_raster_sample(...) -> Tuple[np.array, Option[np.array]]:
    ...
```

emanuel-schmid:
Looks good to me. 😃 Although I went over it only superficially.

climada/util/coordinates.py
tovogt: I solved this problem by leaving the old functionality of `read_raster_sample` untouched, and adding a new function `read_raster_sample_with_gradients`. There is almost no redundant code, because I moved some of the code to a helper function that is used by both functions.

emanuel-schmid: 👍 _much_ better than my suggestion as it keeps backward compatibility!

emanuel-schmid:
🙌

PR #578, opened 2022-11-07
Load hazard raster data lazily

### LINKED ISSUE #544: Follow-up to #507: Lazy loading of chunked xarray Datasets
During the review of #507 the missing support for chunked (lazily-evaluated) datasets was considered a major issue because the entire dataset has to fit into memory (see https://github.com/CLIMADA-project/climada_python/pull/507#discussion_r931212486).

I am currently investigating a way to load the data lazily with dask arrays (a feature automatically supported by xarray, see https://docs.xarray.dev/en/stable/user-guide/dask.html). Dask arrays are distributed into smaller blocks or "chunks", which are designed to fit into memory. By default, such a block is implemented as `numpy.ndarray`.

Xarray offers functions that are agnostic to whether the underlying array type is a `dask.array` or a `numpy.ndarray`. This enables writing a function that takes the dask chunks and returns "sparse" array chunks:
```python
import xarray as xr
import scipy
da = xr.open_dataarray("file.nc", chunks="auto")  # Chunks are dask.array
da = xr.apply_ufunc(scipy.sparse.csr_matrix, da, dask="parallelized", output_dtypes=[da.dtype])
da = da.compute().data  # Load into memory, da is now a scipy.sparse.csr_matrix
```

However, the above example **does not** work because `scipy.sparse.csr_matrix` does not support the required `numpy.ndarray` API for this operation. One workaround would be to use the [sparse](https://github.com/pydata/sparse) library:
```python
import xarray as xr
import sparse
da = xr.open_dataarray("file.nc", chunks="auto")  # Chunks are dask.array
da = xr.apply_ufunc(sparse.COO.from_numpy, da, dask="parallelized", output_dtypes=[da.dtype])
da = da.compute().data  # Load into memory, da is now a sparse.COO array
da = da.tocsr()  # Convert sparse.COO to scipy.sparse.csr_matrix
```

If we don't want a new dependency, we would need to operate on the xarray data types directly, which probably means stitching the chunks together ourselves. Is it okay to add [sparse](https://github.com/pydata/sparse) as new dependency only for this particular use case or should I dig further to see if we can work around it?

peanutfun:
Changes proposed in this PR:
- Use dask array chunks to load hazard raster data lazily
- Add [`sparse`](https://sparse.pydata.org/en/stable/index.html) as new dependency for that task 

This PR fixes #544, an issue raised during the review of #507

### PR Author Checklist

### PR Reviewer Checklist


timschmi95:
This change does enable to use the .from_raster_xarray method instead of my custom method for my hazard file without running into memory issues 👍 
My hazard is a xarray dataset created from multiple netcdf files, and the following dimensions: time: 3660, chy: 225, chx: 352

timschmi95:
> This change does enable to use the .from_raster_xarray method instead of my custom method for my hazard file without running into memory issues 👍 My hazard is a xarray dataset created from multiple netcdf files, and the following dimensions: time: 3660, chy: 225, chx: 352

Actually, I realize that depending on the previous memory allocation I still run into a memory error. I'm not sure if the chunking could be made more efficient, or if it's just too much for my laptop anyways and I either pre-slice the data to a smaller extent (like in my custom function), or run it on euler if I need the full extent.

peanutfun:
> Actually, I realize that depending on the previous memory allocation I still run into a memory error. 

Xarray does not check how much memory in your system is already in use. I currently set `chunks="auto"`, which simply ensures that one chunk can fit into the entire system RAM. However, `from_raster_xarray` explicitly supports passing Datasets that are already opened, so you can open it with custom chunking. The following would be a somewhat conservative choice:

```python
ds = xr.open_dataset("file.nc", chunks=dict(time=100, chy=-1, chx=-1))
hazard = Hazard.from_raster_xarray(ds, **kwargs)
```

climada/hazard/base.py
```diff
@@ -725,8 +729,10 @@ def strict_positive_int_accessor(array: xr.DataArray) -> np.ndarray:
                 default_value=[
                     to_csr_matrix(
                         xr.apply_ufunc(
-                            lambda x: np.where(x != 0, 1, 0), data[intensity]
-                        ).values
+                            lambda x: np.where(x != 0, 1, 0), data[intensity],
+                            dask="parallelized",
```
emanuel-schmid, line 733: so the default is multiprocessing - right?
silly question: does this work in a Jupyter Notebook? multiprocessing used to be cumbersome on Jupyter...
(and if so - is it _really_ multiprocessing in there?)

peanutfun: Woops, this is left over from before the new fraction handling. With the new constructor (#541), we do not need this anymore and I actually already adapted this in #589. So I'll merge `develop` into this branch again and resolve this.

About your question: Without explicitly constructing a `dask.distributed` cluster in the background, dask tries to parallelize using multithreading. So the parallelization here would come at no cost but with the benefit of potentially exploiting multithreading capabilities of the machine

peanutfun:
@emanuel-schmid The `sparse` package still seems to be missing from the testing environment

emanuel-schmid:
> @emanuel-schmid The `sparse` package still seems to be missing from the testing environment

rather again than still ;)
But it's been re-installed just now.

climada/hazard/base.py
```diff
@@ -713,12 +714,20 @@ def from_raster_xarray(
...
-            return sparse.csr_matrix(np.where(np.isnan(array), 0, array))
+            array = array.where(array.notnull(), 0)
+            array = xr.apply_ufunc(
+                sp.COO.from_numpy,
+                array,
+                dask="parallelized",
+                output_dtypes=[array.dtype]
+            )
+            array = array.compute().data  # Load into memory, now it's a sparse.COO array
+            return array.tocsr()  # Convert sparse.COO to scipy.sparse.csr_matrix
```
emanuel-schmid, line 730:
```suggestion
            sparse_coo = array.compute().data  # Load into memory
            return sparse_coo.tocsr()  # Convert sparse.COO to scipy.sparse.csr_matrix
```
Just because variables changing their type in the middle of a method make me uneasy. Or is there a reason for it?

peanutfun: No, not at all 😁

climada/hazard/base.py
```diff
@@ -788,7 +797,7 @@ def maybe_repeat(values: np.ndarray, times: int) -> np.ndarray:
                 ],
                 # The accessor for the data in the Dataset
                 accessor=[
-                    lambda x: to_csr_matrix(default_accessor(x)),
+                    to_csr_matrix,
                     lambda x: maybe_repeat(default_accessor(x), num_events),
                     lambda x: strict_positive_int_accessor(x),
```
peanutfun, line 802: @emanuel-schmid The superfluous lambda caused a `NORMAL` linter warning: `unnecessary-lambda` 😬

emanuel-schmid, line 802:
```suggestion
                    strict_positive_int_accessor,
```
🙄

emanuel-schmid: pylint is a philistine!

emanuel-schmid:
looks good to me! shall we merge? 😃

emanuel-schmid:
Let me quickly come back to a topic related to a previous PR, not really related to this one but by the method concerned:
Why did we chose 'time' as the default 'event' coordinate name, and not 'event'? 🧐

emanuel-schmid:
Yes, that makes sense. 👍 I've added a note to this respect in the pydoc string in case somebody else is wondering. Feel free to revert if it isn't appropriate.

peanutfun:
Very good! 🤩

peanutfun:
@emanuel-schmid Is this ready to go from your perspective?

requirements/env_climada.yml
```diff
@@ -31,7 +31,6 @@ dependencies:
   - salib>=1.3.0
   - scikit-learn>=1.0
   - scipy>=1.6
-  - sparse
```
peanutfun: From my experience, it is better to have `conda` handle as many packages as possible, because it tries to create an environment that is compatible to all packages. It seems to make a better job than `pip`. Also, where does the version restriction come from? I did not check explicitly but from the [changelog](https://sparse.pydata.org/en/stable/changelog.html) I do not really see a reason why lower versions should not be supported

emanuel-schmid: As for letting `conda` handle it: agreed! (I thought it had eventually been installed from the pypi channel even when listed above, but that was a mistake.)

As for the version restrictions: they are always made quite blindly without further thought. The reason for them is that `conda` i.g. cannot cope well with large possibility spaces and tends to get lost there forever or runs out of memory when versions are defined too liberally. (The situation is better with `mamba` though but that's not the officially suggested environment tool.)

emanuel-schmid:
:tada: Ready to go, from my perspective.

requirements/env_climada.yml
peanutfun: I see, that makes sense! Thanks a lot for the review! Then I'll go ahead and merge 😇

emanuel-schmid:
🙇‍♂️ many thanks!

PR #594, opened 2022-11-30
Feature/coverage

NicolasColombi:
Changes proposed in this PR:
-  Coverage test increase in Exposures/base.py
- Delete 2 lines of Exposures/base.py

This PR add:

1) The first test, test line 497-498 of climada_python/entity/exposures/base.py. It check that if a geometry column is already present in the gdf, a specifc error message is displayed. 

2) The second test, test  lines 179-185 of climada_python/entity/exposures/base.py. It checks that if metadata is in kwargs it will be removed from it, and if it's present in meta, it will be added to the dictionary meta. Done by Lukas.

3) Remove a half finished and wrong test that was pushed in the first commit.

4) Remove 2 lines of Exposures/base.py that are no longer useful

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/entity/exposures/test/test_base.py
+++ b/climada/entity/exposures/test/test_base.py
@@ -98,7 +98,8 @@ class TestFuncs(unittest.TestCase):
                          str(cm.exception))
 
     def test__init__mda_in_kwargs(self):
-        """Check if `_metadata` attributes are instantiated correctly for sub-classes of `Exposures`"""
+        """Check if `_metadata` attributes are instantiated correctly for sub-classes of
+        ``Exposures``"""
         litpop = LitPop(exponents=2)
         self.assertEqual(litpop.exponents, 2)
         litpop = LitPop(meta=dict(exponents=3))


climada/entity/exposures/base.py
```diff
@@ -493,9 +493,7 @@ def from_raster(cls, file_name, band=1, src_crs=None, window=False,
         --------
         Exposures
         """
-        exp = cls()
-        if 'geometry' in exp.gdf:
-            raise ValueError("there is already a geometry column defined in the GeoDataFrame")
+        exp = cls()
```
emanuel-schmid, line 496:
```suggestion
        exp = cls()
```

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
         self.assertEqual("meta must be a dictionary",
                       str(cm.exception))
 
+    def test__init__geometry_type(self):
+        """ Check if the column name of geometry of GDF was customized"""
```
emanuel-schmid, line 94:
```suggestion
        """Check that initialization fails when `geometry` is given as a `str` argument"""
```

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
+        with self.assertRaises(ValueError) as cm:
+            Exposures(geometry = 'myname')
```
emanuel-schmid, line 96:
```suggestion
            Exposures(geometry='myname')
```

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
+        self.assertEqual("Exposures is not able to handle customized 'geometry' column names.",
+                      str(cm.exception))
```
emanuel-schmid, line 98:
```suggestion
                         str(cm.exception))
```

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
+
+    def test__init__mda_in_kwargs(self):
+        """ Check if mda is not present in Kwargs after being removed"""
```
emanuel-schmid, line 101:
```suggestion
        """Check if mda is not present in kwargs after being removed"""
```
That descriptions is a bit hard to understand.

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
+        litpop = LitPop(exponents=2)
+        self.assertEqual(litpop.exponents, 2)
+        LitPop(meta=dict(exponents=2))
+        self.assertEqual(litpop.exponents, 2)
```
emanuel-schmid, line 105: In line 104 you create a `LitPop` object but do not assign it to any variable, you just leave it for the garbage collector to be removed.\
This action should in theory have no effect at all with the exception of some Logging messages that are written out.\
In line 105 you are checking again whether the object created in line 102 has an `exponents` value of `2`. But that was already checked in line 103. (If it really has changed since than we are in deep trouble! :scream:)

NicolasColombi: Hi @emanuel-schmid , that was the tricky test we saw in the meeting with you and Chahan. I asked @peanutfun  help on it before the meeting and to explain it to me, he had to do it first. Then, since I thought it was finished, I decided to commit it with my other tests. I suggest that we cancel it and when I will know how to do it I will. Next time I won't commit a test that I don't understand and that I didn't do, even though I specified that in the commit. It would have been probably a better idea to not commit that, but I didn't knew what to do with it else then commit it.

peanutfun: @NicolasColombi No worries! Even making the effort of raising a PR and explaining the test already helps us 🙌 

@emanuel-schmid You are right, this is a bug in the test. Here's my suggestion to fix it:

```suggestion
        litpop = LitPop(meta=dict(exponents=3))
        self.assertEqual(litpop.exponents, 3)
```

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
+            Exposures(geometry='myname')
+        self.assertEqual("Exposures is not able to handle customized 'geometry' column names.",
+                         str(cm.exception))
+
+    def test__init__mda_in_kwargs(self):
+        """Check if mda is not present in kwargs after being removed"""
+        litpop = LitPop(exponents=2)
+        self.assertEqual(litpop.exponents, 2)
+        litpop = LitPop(meta=dict(exponents=3))
+        self.assertEqual(litpop.exponents, 3)
```
emanuel-schmid, line 105: much better now 😉 though I still cannot make sense out of the test description. What is "mda", where is it removed and how is that related to the the `exponents`?

emanuel-schmid (changes requested):

climada/entity/exposures/test/test_base.py
peanutfun: This is due to the `Entity` class being written quite badly, if I may say so. We try to test [`climada/entity/exposures/base.py#L178`](https://github.com/CLIMADA-project/climada_python/blob/main/climada/entity/exposures/base.py#L178) and following. There is a class attribute `_metadata` which is supposed to be extended by derived classes. The base class `__init__`, however, is built to handle these extensions. L178 and L179 check if there are keywords in the *current* class type `_metadata` that is not part of the `Exposures._metadata`. This can only be the case if `self` is a derived class, and not `Exposures` itself. The following code will then extract these keywords from the `kwargs` before the remainder of the `kwargs` is passed to the GeoDataframe.

See [`Exposures._metadata`](https://github.com/CLIMADA-project/climada_python/blob/main/climada/entity/exposures/base.py#L125) and [`LitPop._metadata`](https://github.com/CLIMADA-project/climada_python/blob/main/climada/entity/exposures/litpop/litpop.py#L69)

This is all very convoluted, and the only advantage that I see is that you do not need a dedicated `__init__` for derived classes anymore, at the expense of code that is very difficult to understand.

peanutfun: Oh, what's even worse is that the `Exposures.__init__` will thus instantiate attributes with names listed in `_metadata`, even for derived classes.

emanuel-schmid: I see. Then the suggested description would about describe the purpose of this test?

climada/entity/exposures/test/test_base.py
```diff
@@ -89,6 +90,20 @@ def test__init__meta_type(self):
...
 
+    def test__init__geometry_type(self):
+        """Check that initialization fails when `geometry` is given as a `str` argument"""
+        with self.assertRaises(ValueError) as cm:
+            Exposures(geometry='myname')
+        self.assertEqual("Exposures is not able to handle customized 'geometry' column names.",
+                         str(cm.exception))
+
+    def test__init__mda_in_kwargs(self):
+        """Check if `_metadata` attributes are instantiated correctly for sub-classes of `Exposures`"""
```
emanuel-schmid, line 101:
```suggestion
        """Check if `_metadata` attributes are instantiated correctly for sub-classes of
        ``Exposures``"""
```

PR #602, opened 2022-12-09
added assign_centroids for impact objects

timschmi95:
Changes proposed in this PR:
- added util function assign_gdf_centroids to assign hazard centroids to any geodataframe 
- wrapper for impact object to assign hazard centroids to impact objects

This PR supports the calibration of impact functions based with damage data from impact objects

### PR Author Checklist

### PR Reviewer Checklist


aleeciu:
The code is quite readable. I think a more thorough look a the doc string is needed and then it can be merged (I pushed few suggestions and see also comments).

timschmi95:
@aleeciu I added a test for the imp.assign_centroids() method, and all tests are passing. Shall I covert it to a 'real' pull request?

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,79 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+    """Assign to each exposure's coordinate point its closest hazard's coordinate.
+    If disatances > threshold in points' distances, -1 is returned.
+    If raster hazard and coordinate point outside raster, -1 is returned.
+
+    Parameters
+    ----------
+    coord_df : pd.DataFrame
+        Dataframe with defined latitude/longitude column
+    hazard : Hazard
+        Hazard to match (with raster or vector centroids).
```
emanuel-schmid, line 1072: This method that shouldn't care about CLIMADA data structures unnecessarily. 
I suggest to use `centroids` as an argument instead.

The only other attribute needed is `tag.haz_type` and that's used "in the wrong way" to check whether the centroids have been assigned before. Which is misleading as it's thinkable to have two different hazards with the same type. When returning output data instead of overwriting input data, `tag.haz_type` will be obsolete anyways. (see comment to `overwrite`)

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,79 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+    crs : pyproj.crs.crs.CRS
+        Coordinate system of the dataframe
```
emanuel-schmid, line 1074: why? if it's not the same as the one from `centroids` it just leads to an error?

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,79 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+        Possible values are "euclidean", "haversine" and "approx".
+        Default: "euclidean"
+    threshold : float
+        If the distance (in km) to the nearest neighbor exceeds `threshold`,
+        the index `-1` is assigned.
+        Set `threshold` to 0, to disable nearest neighbor matching.
+        Default: 100 (km)
+    overwrite: bool
+        If True, overwrite centroids already present. If False, do
+        not assign new centroids. Default is True.
```
emanuel-schmid, line 1086: I'd rather return a data frame or even just a column of a data frame instead of messing around with (= modifying) input data. It will just cause headaches.

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,79 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+                        threshold=NEAREST_NEIGHBOR_THRESHOLD,
+                        overwrite=True):
+    """Assign to each exposure's coordinate point its closest hazard's coordinate.
+    If disatances > threshold in points' distances, -1 is returned.
+    If raster hazard and coordinate point outside raster, -1 is returned.
+
+    Parameters
+    ----------
+    coord_df : pd.DataFrame
+        Dataframe with defined latitude/longitude column
```
emanuel-schmid, line 1070: shouldn't that rather be a `GeoDataFrame`? because then the `crs` argument would become obsolete and the check for equality would make sense.

emanuel-schmid:
Thanks a lot! The time for refactoring centroids assignment is just right I think. 😁 
I suggest to first change the signature and behavior of the `util.coordinates.assign_haz_centroids` function and add some unittests that take its whole argument space into account.

emanuel-schmid (changes requested):

climada/util/coordinates.py
timschmi95: Originally it was a geoDataFrame. But since impact object don't have a geodataframe by default I thought I thought I make the method more flexible and for the impact.assign_centroids() I don't create a GeoDataFrame first, but directly pass the coord_df since the crs information is already saved in the impact object.
The most common use case (also when using the function on its own) will still be a GeoDataFrame. 
I could change it back again to only allow GeoDataFrames as input, if you think this is better?

climada/util/coordinates.py
timschmi95: I did the method so the behaviour of exp.assign_centroid() remains unchanged, since this operation is used all the time in climada. I think if we change this behaviour, we would have to have bit bigger discussion?

Could it be an option to change the default to overwrite=False in this function, but always pass overwrite=True from the exp.assign_centroids() method?

emanuel-schmid: I agree, `exp.assign_centroid()` should remain unchanged in behavior atm. But nevertheless should this method be free from side effects. The solution is to deal with the `overwrite` argument in `assign_centroid()`. 
Little more complexity there - a little less here - better distributed. 😁

climada/util/coordinates.py
timschmi95: I resolved it by passing an additonal test_crs argument, which is True by default. Only when it's called from exp.assign_centroids(), the crs match is checked outside of the u_coord.assign_gdf_centroids() function, because the exp.gdf does not require an active geometry column, without which the equal crs cannot be tested

climada/util/coordinates.py
timschmi95: Is changed to centroids now

climada/engine/impact.py
```diff
@@ -1736,6 +1737,72 @@ def stack_attribute(attr_name: str) -> np.ndarray:
...
+        geometry = gpd.points_from_xy(coord_df.longitude,coord_df.latitude,crs = self.crs)
+        coord_gdf = gpd.GeoDataFrame(coord_df,geometry = geometry)
+
+        #call the assign_gdf_centroids util function
+        coord_gdf[centr_haz] = u_coord.assign_gdf_centroids(coord_gdf, hazard.centroids,
+                                                distance=distance,
+                                                threshold=threshold)
+
+        #create new impact attribute with hazard centroids
+        setattr(self,f'centr_{hazard.tag.haz_type}' , coord_df[f'centr_{hazard.tag.haz_type}'])
```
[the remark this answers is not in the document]
emanuel-schmid: Introducing an attribute is cumbersome to some extent. There are read/write methods to be considered and tested,. All methods that use impact as argument may in theory be affected, etc, etc,.
With exposures there is nothing we can do about it, the damage is done, but with impact we still have a chance to stay side-effect free. 🥺

climada/engine/test/test_impact.py
```diff
@@ -816,6 +816,21 @@ def test__exp_build_event(self):
...
+        fake_at_event = np.arange(HAZ.size)
+        fake_aai_agg = np.sum(fake_eai_exp)
+        imp = Impact.from_eih(exp, ENT.impact_funcs, HAZ,
+                              fake_at_event, fake_eai_exp, fake_aai_agg)
+        imp.assign_centroids(HAZ)
+        imp_centr = getattr(imp,f'centr_{HAZ.tag.haz_type}')
+        np.testing.assert_array_equal(imp_centr,exp.gdf.centr_TC)
+
 
 class TestImpactH5IO(unittest.TestCase):
```
emanuel-schmid, line 835: I'd also make another test for the utility function that checks all three main cases:
- the centroids are a raster
- the centroids are not a raster
- the two crs don't match

emanuel-schmid (changes requested):
Thanks for the effort! looking good already and getting better and better 😄 
These are my remaining objections:
- I'd rather don't store the result in the impact object
- fix the crs within the exposures and dismiss the test_crs flag
- refer to the utility function in the other two doc strings
- add a unit test for the utility function with three main cases raster, no-raster, crs-clash

climada/entity/exposures/base.py
```diff
@@ -424,16 +426,12 @@ def assign_centroids(self, hazard, distance='euclidean',
...
-                hazard.centroids.meta['transform'])
-        else:
-            assigned = u_coord.assign_coordinates(
-                np.stack([self.gdf.latitude.values, self.gdf.longitude.values], axis=1),
-                hazard.centroids.coord, distance=distance, threshold=threshold)
-        self.gdf[centr_haz] = assigned
+
+        assigned_centr = u_coord.assign_gdf_centroids(self.gdf, hazard.centroids,
+                        distance=distance,
+                        threshold=threshold,test_crs=False)
```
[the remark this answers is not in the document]
timschmi95: The problem is that climada doesn't require exp.gdf to have a valid geometry column (only a latitude and longitude column). Thus calling exp.gdf.crs throws an Error in many of the tests. (See my previous [comment](https://github.com/CLIMADA-project/climada_python/pull/602#discussion_r1093375056)).
That's why I came up with the test_crs flag. It may not be the most elegant, but do you think we should keep it? Or do you have an alternative idea?

climada/engine/impact.py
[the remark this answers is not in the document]
timschmi95: For our use case (impact function calibration with impact objects that store the recorded damages) it was very convenient to have the assigned centroids stored within the impact object. Then we don't always have to pass the assigned centroids seperately to other functions. 

If it is strongly discouraged, we could also just let the method return the ``coord_gdf`` and we would just manually set it as an attribute of the impact object. Would this be preferred?

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,67 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+def assign_gdf_centroids(coord_gdf, centroids, distance='euclidean',
+                        threshold=NEAREST_NEIGHBOR_THRESHOLD,test_crs=True):
+    """Assign to each gdf coordinate point its closest centroids's coordinate.
+    If disatances > threshold in points' distances, -1 is returned.
+    If raster centroids and coordinate point outside raster, -1 is returned.
+
+    Parameters
+    ----------
+    coord_gdf : gpd.GeoDataFrame
+        GeoDataframe with defined latitude/longitude column and crs
```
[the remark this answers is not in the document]
timschmi95: If I understand the question correct, the answer is no. The coord_gdf is not altered, but the method just returns an array. 
Only in the exp.assign_centroids() the exp.gdf is altered (which is mentioned there now)

climada/entity/exposures/base.py
```diff
@@ -368,7 +368,8 @@ def get_impf_column(self, haz_type=''):
     def assign_centroids(self, hazard, distance='euclidean',
                          threshold=u_coord.NEAREST_NEIGHBOR_THRESHOLD,
                          overwrite=True):
-        """Assign for each exposure coordinate closest hazard coordinate.
+        """Wrapper for u_coord.assign_haz_centroids():
+        Assign for each exposure coordinate closest hazard coordinate.
```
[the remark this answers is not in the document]
timschmi95: I shortened it a bit (especially in the imp.assign_centroids method). 
Since the parameter docstrings are often useful when using the function (as the editor directly shows you the documentation) I thought it may be useful to have them also in the wrapper methods. 
But of course it could also be further shortened

climada/engine/impact.py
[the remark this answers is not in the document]
emanuel-schmid: Yes, sorry, but this would actually be preferred.
Generally, in Python adjusting data types (classes) is very easy and seems to come in handy at many occasions. However the ease is much more of a curse than a blessing.
In this particular case it's most likely even justified to store the centroids assignment within the impact class, but nevertheless: patching a class with an attribute on the fly is not best practice, strictly speaking, and would require a thorough analysis of the introduced side effect and, as stated, _lots_ of tests.

climada/engine/test/test_impact.py
timschmi95: I'm not so familiar with best practices for testing.
I thought since the function is basically a wrapper around 2 functions (for ``u_coord.assign_grid_points`` and ``u_coord.assign_coordinates``) which both are tested, and it is called from exp.assign_centroids() and imp.assign_centroids() which also both are tested, it would be sufficient.

So would the idea be that I replicate the tests from exp.assign_centroids() for [points](https://github.com/CLIMADA-project/climada_python/blob/e26f900ca0e7e3ea627555e9345c5ae76f911428/climada/entity/exposures/test/test_base.py#L58) and [rasters ](https://github.com/CLIMADA-project/climada_python/blob/e26f900ca0e7e3ea627555e9345c5ae76f911428/climada/entity/exposures/test/test_base.py#L126) except that I directly use the u_coord.assign_centroids_to_gdf instead of the exp.assign_centroids() ?

climada/entity/exposures/base.py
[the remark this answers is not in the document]
emanuel-schmid: I see. The case of an exposures that does not have a geometry column in its gdf but who's centroids are assigned - is it a common use case? or at least valid? or just a test artefact?

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,69 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+    Both cases can introduce innacuracies for coordinates in lat/lon
+    coordinates as distances in degrees differ from distances in meters
+    on the Earth surface, in particular for higher latitude and distances
+    larger than 100km. If more accuracy is needed, please use 'haversine'
+    distance metric. This however is slower for (quasi-)gridded data,
+    and works only for non-gridded data.
+    """
+
+    if test_crs and (not equal_crs(coord_gdf.crs, centroids.crs)):
+        raise ValueError('Set hazard and GeoDataFrame to same CRS first!')
```
emanuel-schmid, line 1110:
```suggestion
    try:
        if not equal_crs(coord_gdf.crs, centroids.crs):
            raise ValueError('Set hazard and GeoDataFrame to same CRS first!')
    except AttributeError:
        pass
```

climada/engine/test/test_impact.py
emanuel-schmid: Ideally the actual points and rasters assignment tests from exposures rather would be rather _moved_ to the test_coordinates module, And in the exposures tests only the impact on the exposures object is tested, like: the gdf has a new/modified column now.

climada/engine/impact.py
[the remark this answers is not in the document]
timschmi95: That's fine it can easily be changed. And I'm happy to learn from you regarding python best practices ;)

climada/util/coordinates.py
timschmi95: I assume this is a suggestion to avoid the test_crs=True flag? 
Why is this better than the test-crs flag? I thought such pass statements should be avoided according to the [climada guide](https://github.com/CLIMADA-project/climada_python/blob/feature/assign_centroids_impact/doc/guide/Guide_PythonDos-n-Donts.ipynb), but I'm happy to learn :)

climada/engine/impact.py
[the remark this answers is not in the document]
emanuel-schmid: 🤓

climada/engine/test/test_impact.py
timschmi95: I added the suggested tests [here](https://github.com/CLIMADA-project/climada_python/pull/602/commits/58056448120691e87d23a6a4da9567421d5cae3b). Is this suitable?

climada/util/coordinates.py
emanuel-schmid: > ... such pass statements should be avoided ...

Like in " 2. Do not catch exception if you do not handle them."?
 But you totally handle it: if there is no crs in coord_gdf there is nothing to compare and no reason to fail.

Why I don't like the `test_crs` flag?
 Because one is not free to give it a value. There doesn't seem to be a situation where you could set it to `True` _or_ `False` deliberately. The argument is strongly coupled to the `coord_gdf` argument. It increases complexity unnecessarily. It's worse than `pass`ing.

climada/engine/test/test_impact.py
emanuel-schmid: 😍 Thanks!

climada/util/test/test_coordinates.py
```diff
@@ -515,6 +517,103 @@ def test_assign_grid_points(self):
...
+        centroids = Centroids(
+            lat=[1100000,1200000],
+            lon=[2500000,2600000],
+            geometry = gpd.GeoSeries(crs='EPSG:2056')
+        )
+        try:
+            assigned = u_coord.assign_centroids_to_gdf(
+                gdf, centroids, distance=distance, threshold=thresh)
+        except ValueError as err:
+            assert(str(err)=='Set hazard and GeoDataFrame to same CRS first!')
```
emanuel-schmid, line 614: This test passes when it ought to fail, doesn't it? You could use `self.assertRaises` here.

climada/util/coordinates.py
emanuel-schmid: _However_ - The `try`/`pass` approach is due to the `crs` being a property (I think) and the canonical way to check for it being there. If there was another way like `gdf.has_crs()` or something that would be preferable. But I'm not aware of such a way...

timschmi95: Hm, yeah I haven't looked at it this way. So when no crs is defined, we assume that the user selected the correct crs, right?
And furthermore, we assume that there is no other cause of an ``AttributeError`` within the try statement, which I guess is reasonably since it's a small function. 

Shall I add a comment to the ``pass`` statement: 
"If the coord_gdf has no crs defined, no Error is raised and it is assumed that the user set the crs correctly"

timschmi95: For the case of exp.gdf without a valid 'geometry' column, it would then be as follows:
in exp.assign_centroids() one checks that ``exp.crs`` equals ``haz.centroids.crs``, then within the u_coord.assign_centroids_to_gdf() one tries and fails to check that ``exp.gdf.crs`` equald ``haz.centroids.crs``, and it is cought by the ``except AttributeError`` statement, because exp.gdf does not have a valid geometry column.

climada/util/test/test_coordinates.py
timschmi95: I adjusted it to use assertRaises

timschmi95:
@emanuel-schmid I think everything is resolved now. Do you want to have a last look over it before I merge it?

climada/engine/impact.py
```diff
@@ -1736,6 +1737,48 @@ def stack_attribute(attr_name: str) -> np.ndarray:
...
+            Set `threshold` to 0, to disable nearest neighbor matching.
+            Default: 100 (km)
+        """
+
+        haz_type = hazard.tag.haz_type
+        centr_haz = 'centr_' + haz_type
+
+        #Assert that the imp crs is epsg:4326, as it is required by the u_coord methods
+        if not u_coord.equal_crs(self.crs, 'EPSG:4326'):
+            raise ValueError('Set Impact to lat/lon crs (EPSG:4326)!')
```
emanuel-schmid, line 1768: One more question (sorry 😁) - which u_coord methods require epsg:4326?
And why can't they speak for themselves, i.e. raise an exception if the crs is wrong?

climada/util/coordinates.py
```diff
@@ -1057,6 +1057,70 @@ def assign_coordinates(coords, coords_to_assign, distance="euclidean",
...
+    distance : str, optional
+        Distance to use in case of vector centroids.
+        Possible values are "euclidean", "haversine" and "approx".
+        Default: "euclidean"
+    threshold : float, optional
+        If the distance (in km) to the nearest neighbor exceeds `threshold`,
+        the index `-1` is assigned.
+        Set `threshold` to 0, to disable nearest neighbor matching.
+        Default: ``NEAREST_NEIGHBOR_THRESHOLD`` (100km)
+
```
emanuel-schmid, line 1081:
```suggestion

    Returns
    -------
    np.array
        array of closest centroids, aligned with the index of the `coord_gdf`
```

emanuel-schmid (commented):
Since there is no real assignment anymore, the method name is kind of misleading.
My favorite would be `closest_centroids`. But then, the same argument is also valid for the `coordinates` methods `assign_centroids_to_gdf`, `assign_gridpoints` and `assign_coordinates`. Consequently they should be renamed too. Therefore I propose a wimp compromise: `assigned_centroids` as in "if there _was_ an assignment the assigned centroids _would be assigned_ like this:"

And CHANGELOG.md needs to be updated ... 😁

chahank:
Adding a method `assigned_centroids` sounds like a terrible idea to me, sorry. This makes just a very confusing naming. What is the purpose of this new method?

emanuel-schmid:
With regard to `Impact.assign_centroids` - it's for those cases where you still have the impact, but not the exposures anymore I reckon. (?)
Apart from the `Impact.asssign_centroids` method it's a nice decomposition of `Exposures.assign_centroids`.

chahank:
@timschmi95 : could you please clarify the use case of the methods `impact.assign_centroids` ?

timschmi95:
@chahank The ``impact.assign_centroids`` is used for the spatially explicit calibration, which we do for hail damages. I.e. the reported damages are read into an impact object, and assigned hazard intensity values for each reported impact allow for a spatially explicit calibration.

timschmi95:
@emanuel-schmid I am also not sure about the proposed name ``impact.assigned_centroids``. 
As you pointed out with the newest change the assigned centroids are not stored in the impact object anymore. Nevertheless, they are assigned to each impact coordinate point (just not stored within the imp object), thus I would vote to keep the name as ``impact.assign_centroids``. Or if preferred I think ``impact.closest_centroids`` also works well, even if within the function it then uses the ``assign_centroids_to_gdf``` function.

emanuel-schmid:
Many thanks @timschmi95 
I've changed all the misleading `assign_` names to `match_` (with the exception of `Exposures.assign_centroids`, of course) and simplified the `Impact.match_centroids` method somewhat.
Now I'm happy. If @chahank approves, we can merge.

chahank:
I still think that there should not be any method `impact.macht_centroids`. Since it is a single liner and it was only introduced to be used in the calibration module, I would remove it entirely. The pull request was nevertheless very useful as the utility methods were much improved.

chahank:
That being said, if you really think it is useful, please feel free to merge.

PR #606, opened 2022-12-14
Feature/impact write from hdf5

ThomasRoosli:
Changes proposed in this PR:
- introduce write_hdf5 and read_hdf5 methods in Impact
- so far Impact.event_name does not have to be of type list(str) but just of type list. This is a requirement in Hazard, where this attribute generally originates from. I propose to make this the required type also in Impact. This would mean we have to change the docstring and some tests.

This PR fixes # 

### PR Author Checklist

### PR Reviewer Checklist


climada/engine/impact.py
```diff
@@ -862,6 +865,77 @@ def write_col(i_col, imp_ws, xls_data):
...
+        todense: bool
+            if True write the sparse matrices as hdf5.dataset by converting them to dense format
+            first. This increases readability of the file for other programs. default: False
+        """
+        if isinstance(file_data, (str, Path)):
+            LOGGER.info('Writing %s', file_data)
+            with h5py.File(file_data, 'w') as data:
+                self._write_hdf5(data, todense)
+        else:
+            self._write_hdf5(file_data, todense)
```
peanutfun, line 884: I feel like the split in `write_hdf5` and `_write_hdf5` is cumbersome and supporting an already opened file is not necessary. I suggest renaming `file_data` to `file_path` and just allowing paths/strings here. Then both functions can be merged into one

climada/engine/impact.py
```diff
@@ -862,6 +865,77 @@ def write_col(i_col, imp_ws, xls_data):
...
+
+    def _write_hdf5(self, data, todense):
+        str_dt = h5py.special_dtype(vlen=str)
+        for (var_name, var_val) in self.__dict__.items():
+            if var_name == 'tag':
+                # write the dictionary of tags of exposures, impact functions set and hazard
+                # write hazard tag
+                hf_str = data.create_dataset('haz_tag_haz_type', (1,), dtype=str_dt)
+                hf_str[0] = var_val['haz'].haz_type
```
peanutfun, line 893: I think [Attributes](https://docs.h5py.org/en/stable/high/attr.html) are much more suitable to store single-value string (meta-)data. They are created automatically through a dictionary-type lookup:
```suggestion
                data.attrs["haz_tag_haz_type"] = var_val["haz"].haz_type
```
(Same goes for all other stored strings). See also https://docs.h5py.org/en/stable/quick.html#attributes

climada/engine/impact.py
```diff
@@ -862,6 +865,77 @@ def write_col(i_col, imp_ws, xls_data):
...
+                hf_str[0] = str(var_val['exp'].file_name)
+                hf_str = data.create_dataset('exp_tag_description', (1,), dtype=str_dt)
+                hf_str[0] = str(var_val['exp'].description)
+                # write impact functions tag
+                hf_str = data.create_dataset('impf_set_tag_file_name', (1,), dtype=str_dt)
+                hf_str[0] = str(var_val['impf_set'].file_name)
+                hf_str = data.create_dataset('impf_set_tag_description', (1,), dtype=str_dt)
+                hf_str[0] = str(var_val['impf_set'].description)
+            elif isinstance(var_val, sparse.csr_matrix):
+                if todense:
```
peanutfun, line 909: `todense=True` is not tested in the unit test

climada/engine/impact.py
```diff
@@ -994,6 +1068,81 @@ def read_excel(self, *args, **kwargs):
...
+                        (hf_csr['data'][:], hf_csr['indices'][:], hf_csr['indptr'][:]),
+                        hf_csr.attrs['shape'])
+            elif isinstance(var_val, str):
+                impact_kwargs[var_name] = u_hdf5.to_string(
+                    data.get(var_name)[0])
+            elif isinstance(var_val, (float, int)):
+                impact_kwargs[var_name] = data.get(var_name)[0]
+            elif isinstance(var_val, list):
+                impact_kwargs[var_name] = [x for x in map(
+                    u_hdf5.to_string, np.array(data.get(var_name)).tolist())]
```
peanutfun, line 1139: Fix linter issue
```suggestion
                impact_kwargs[var_name] = list(map(
                    u_hdf5.to_string, np.array(data.get(var_name)).tolist()))
```

climada/engine/impact.py
```diff
@@ -994,6 +1068,81 @@ def read_excel(self, *args, **kwargs):
...
+        -------
+        imp : Impact
+            Impact with data from the given file
+        """
+        if isinstance(file_data, (str, Path)):
+            LOGGER.info('Reading %s', file_data)
+            with h5py.File(file_data, 'r') as data:
+                return cls._from_hdf5(data)
+        else:
+            return cls._from_hdf5(file_data)
```
peanutfun, line 1090: Same here, only allow paths as argument

peanutfun (changes requested):
It gets the job done, but I have to say that I find the overall structure of the functions a bit hard to understand. It currently iterates over all `__dict__` entries, and then uses a large if/elif/else tree to do specific things with specific items. I would much prefer to instead define local "writer functions" and a mapping from attribute name or type to writer function. Then we can iterate over all attributes, check for a writer function or fall back to a default one. This would also help with some code duplications, specifically for writing all the tags. I'm thinking of something like this:
```python
def write_str(name, value):
    data.attrs[name] = value
def write_default(name, value):
    data.create_dataset(name, data=value.toarray())

attr_writers = {str: write_str, sparse.csr_matrix: write_default}
    # Fetch the write function if it exists for the type. If not, use the default one.
    write_func = attr_writers.get(type(var_val), write_default)
    # Call it with the attr name and value
    write_func(var_name, var_val)
```
(The same goes for the reader)

peanutfun:
@ThomasRoosli I think I am finished. Would you have a look?

climada/engine/impact.py
ThomasRoosli: I agree and thanks for adapting accordingly.

climada/engine/impact.py
ThomasRoosli: Good suggestion and thanks for implementing.

climada/engine/impact.py
ThomasRoosli: Thanks for adding the test.

ThomasRoosli:
Thanks @peanutfun  for restructuring especially the write_hdf5 function using type specific writers. And thanks for extending the tests. I am now very happy with the status of the code and the functionality it adds to the impact class. I am happy if we can merge this.

peanutfun:
@emanuel-schmid This new `Impact.write_hdf5` uses generic writer functions. @ThomasRoosli and I talked about making them part of a utility module so that we can use these functions in any other `<Class>.write_hdf5` method. We opted against it to simplify this PR. However, defining all the writer functions inside the `write_hdf5` method makes the linter unhappy. What would you recommend: Split the generic writer functions into a utility module or keep them inside `write_hdf5` at the expense of some added linter issues?

peanutfun (approved):

climada/engine/impact.py
```diff
@@ -862,6 +866,120 @@ def write_col(i_col, imp_ws, xls_data):
...
+
+        # Set up writers based on types
+        # NOTE: Many things are 'Collection', so make sure that precendence fits!
+        type_writers = {
+            str: write_attribute,
+            Tag: write_tag,
+            TagHaz: write_tag,
+            dict: write_dict,
+            sparse.csr_matrix: write_csr,
+            Collection: write_dataset,
```
emanuel-schmid, line 973:
```suggestion
            Collection: write_dataset,
            object: write_attribute,
```

emanuel-schmid:
Very nice! 😁 

> Split the generic writer functions into a utility module or keep them inside `write_hdf5` at the expense of some added linter issues?

My preference would be to make it a `TypeWriter` class that takes the `type_writers` dict as `__init__` argument and has one method, `write`. Could be part of the `hdf4_handlers` module or have a module of its own. I'm positive we'll be using it elsewhere, eventually.
However this can also be a PR of its own. Up to you. 😎

climada/engine/impact.py
```diff
@@ -862,6 +866,120 @@ def write_col(i_col, imp_ws, xls_data):
...
+            Parameters
+            ----------
+            group : h5py.Group
+                The group in the H5 file to write into
+            name : str
+                The identifier of the value
+            value : scalar or array
+                The value/data to write
+            default_writer
+                Fallback writer if no writer in ``type_writers`` matches
```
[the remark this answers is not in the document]
peanutfun: Excellent proposal!

peanutfun:
@emanuel-schmid I don't want to over-complicate this PR. I like your proposal a lot, but I would want to discuss it first and then have a follow-up PR.

climada/engine/test/test_impact.py
```diff
@@ -314,6 +318,29 @@ def test_excel_io(self):
...
+        np.testing.assert_array_almost_equal_nulp(imp_write.at_event, imp_read.at_event, nulp=5)
+        np.testing.assert_array_equal(imp_write.frequency, imp_read.frequency)
+        self.assertEqual(imp_write.tot_value, imp_read.tot_value)
+        self.assertEqual(imp_write.aai_agg, imp_read.aai_agg)
+        self.assertEqual(imp_write.unit, imp_read.unit)
+        self.assertEqual(imp_write.frequency_unit, imp_read.frequency_unit)
+        self.assertEqual(
+            0, len([i for i, j in zip(imp_write.event_name, imp_read.event_name) if i != j]))
+        self.assertIsInstance(imp_read.crs, str)
+
```
peanutfun, line 343: Remove leftover test

PR #612, opened 2023-01-05
trop_cyclone: convert module constants into kwargs

tovogt:
Changes proposed in this PR:
- Get rid of module constants in the `climada.hazard.trop_cyclone` module in favor of more user-friendly kwargs for the `TropCyclone.from_tracks` function.
- When applying thresholds that are called `MAX_...` make sure to use `<=` instead of `<`, because when you say that a certain value is the maximum allowed value then there is a general understanding that the exact value should also be permissible.

This does not change the (default) behavior of the wind field computations unless in rare boundary cases where the thresholds are exactly achieved.

### PR Author Checklist

### PR Reviewer Checklist


ChrisFairless:
I think it makes sense to add these as kwargs, but my feeling is that we shouldn't get rid of the module constants `INLAND_MAX_DIST_KM` and `CENTR_NODE_MAX_DIST_KM`

My opinion is that these default values count as 'data' and so should be treated as such. Especially since they might be used in multiple places (as with `DEF_INTENSITY_THRES`).

(`CENTR_NODE_MAX_DIST_DEG` should be removed since it's now unnecessary. I like what you did there. Much less likely to do something unexpected!)

tovogt:
I agree that it makes sense to define a constant for `CENTR_NODE_MAX_DIST_KM` since it's used in several places. I reintroduced it under the name `DEF_MAX_DIST_EYE_KM`.

However, I don't think it's very useful to have a constant for the default value of `max_dist_inland_km`. This value only appears once in the code and there is no reason to assume that it will be used in other places in the future. Just because it's an explicit "data" value doesn't justify to define a constant for it. There are dozens of explicit "data" values in the wind field implementation and, unless they are used in several places, I don't think that there would be any benefit from defining module constants for each of them.

ChrisFairless:
Yes OK.

Have we got guidance on how we handle a constant the user might (in theory) want to change? I think what you're saying is it belongs to a (public) method, unless it's used multiple times in the class/module, when it becomes a module constant, unless it's used in multiple modules, in which case I guess it goes into a config or util? That makes sense to me.

And if so, this is ready to merge.

tovogt:
There is some guidance about constants here: https://climada-python.readthedocs.io/en/latest/guide/Guide_Configuration.html But I don't think that it can be read in a very strict way. Your summary sounds good though :)

Thanks for your review!

PR #613, opened 2023-01-09
TCTrack: Fix plotting of segments at antimeridian

tovogt:
Changes proposed in this PR:
- Following #611, this fixes three problems with the TC track plotting:
  1. Tracks that crossed 0 longitude were identified as crossing the antimeridian and had a gap in them when plotted.
  2. Tracks that cross the antimeridian had their wind speeds shifted by one frame.
  3. Tracks crossing the antimeridian have a line segment removed.
- The solution proposed in this PR is to truncate track segments crossing the antimeridian (instead of removing them). This fixes the first two problems, but it means for the third point that still at least part of the line segments crossing the antimeridian is removed. We discussed in #611 that it is not worth the effort to properly split the affected segments into two segments because it's complicated to implement and probably barely visible in the affected plots.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -1235,22 +1235,27 @@ class TCTracks():
         axis.set_extent(extent, crs=kwargs['transform'])
         u_plot.add_shapes(axis)
 
-        synth_flag = False
         cmap = ListedColormap(colors=CAT_COLORS)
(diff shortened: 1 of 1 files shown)


ChrisFairless:
Thanks for implementing this. It looks like a neat fix. Nothing to suggest!

I also took the chance to see how it handles self-intersecting tracks (since there's a warning in a comment), and it had no problem with my test example.

Ready to merge.

PR #635, opened 2023-01-30
Feature/coverage

NicolasColombi:
Changes proposed in this PR:

1) Add coverage test for function read_bm_file at line 422 of:
`climada/entity/exposures/litpop/nightlight.py`

2) Invert the arguments of the `Path()` function at line 441 of the file:
`climada/entity/exposures/litpop/nightlight.py`. The wrong order of the
arguments lead to the creation of a wrong path, which lead to an error
when opening the GeoTiff file as a Gdal DataSet using the function `gdal.Open()` at line 444.

3) Add file `BlackMarble_2016_C1_geo_gray.tif` to: `climada/entity/exposures/test/data`. The File is needed for the test.

### PR Author Checklist

### PR Reviewer Checklist


peanutfun:
Was this closed on purpose? 🤔

NicolasColombi:
@peanutfun  Yes, because it passes the test on my machine but not on Jenkins. I suspect that it is due to the fact that the test I created uses a file that I placed in exposure/data, then I created a path to that file using CONFIG. So it all works fine on my machine because it can find the file, but probably due to the gitignore file that specify that all files in that folder should be ignored by git, once I commit my test the file can't be found.

Do you know how to fix that ?

peanutfun:
If you want to add a file that would be ignored due to the rules of `.gitignore`, you have to use `git add -f <file>`. How large is the file? We have to be careful not to increase the repository size too much. Might there be a way to download the file automatically during the test instead of placing it in the repository?

Btw, loading "real" data should not be part of a unit test, because it typically takes much longer. You can add this test to the [corresponding integration test](https://github.com/CLIMADA-project/climada_python/blob/main/climada/test/test_nightlight.py)

Issue #595, opened 2022-12-01
Impact Aggregation by Region


emanuel-schmid:
There should be an easy way to get damages per event and region_id.

There are many ways to implement this, but all of them come with a price on calculation and/or memory. From a discussion on Slack a vague plan emerged to 
- introduce an additional method in `Impact` that provides this kind of aggregation 
- introduce an additional argument in the `ImpactCalc.impact` method that indicates how/whether to persist the aggregation result within the `Impact` object.

aleeciu:
- introduce an additional argument in the `ImpactCalc.impact` method (e.g., `at_region_id_event`) and we the current `at_event` becomes `at_region_id_event.sum(0)` or something

emanuel-schmid:
Er, actually, the case depending replacement of `at_event` was declined, I'm afraid 😬

aleeciu:
see proposed addition in #642

peanutfun:
Resolved by #642

PR #642, opened 2023-02-05
add at_reg_id_event

aleeciu:
This PR addresses #595


climada/engine/impact.py
```diff
@@ -150,6 +153,7 @@ def __init__(self,
         self.crs = crs
         self.eai_exp = np.array([], float) if eai_exp is None else eai_exp
         self.at_event = np.array([], float) if at_event is None else at_event
+        self.at_reg_id_event = np.array([], float) if at_reg_id_event is None else at_reg_id_event
```
peanutfun, line 156: I don't understand why this should become an attribute of `Impact`. If it can be calculated from `Impact`, is there any case where you could *create* an `Impact` with it? I would make the function *only* return the value, but not store it in the `Impact` object. The user than has to take care of storing it.

climada/engine/impact.py
```diff
@@ -390,6 +392,24 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+        Parameters
+        ----------
+        exposures : climada.entity.Exposures
+            exposure used to compute imp_mat
+        Returns
+        -------
+        np.matrix
+        """
+        self.at_reg_id_event = np.hstack([
+                       self.imp_mat[:, np.where(exposures.gdf.region_id == reg_id)[0]].sum(1)
```
peanutfun, line 408: Does this work if `imp_mat` is an empty matrix? Does it make sense to then return zeros here or should this method throw an error instead?

climada/engine/impact.py
```diff
@@ -390,6 +392,24 @@ def impact_per_year(self, all_years=True, year_range=None):
             year_set[year] = sum(self.at_event[orig_year == year])
         return year_set
 
+    def impact_per_exp_reg(self, exposures):
```
peanutfun, line 395: Please add a unit test for this method, using a self-made impact matrix and exposures, and checking the exact result.

peanutfun (changes requested):

climada/engine/impact.py
emanuel-schmid: As the PR proposes a change of the type `Impact`, all functions and methods with an `Impact` in their signature are possibly affected. A single unittest won't do, I'm afraid. In particular, `unsequa`, `cost_benefit` and read/write methods should probably be tested for consistency.

aleeciu: thanks for the clarification, if this is the case then this PR does indeed require more care... however, I still do not fully understand how can these classes (e.g., `unsequa`, `cost_benefit`) be affected if none of the methods therein use`Impact.impact_per_exp_reg`.

peanutfun: I think @emanuel-schmid refers to the change of the `Impact` signature. If we add a new attribute that can be set via the `Impact.__init__`, a lot of tests would need to be checked and updated. However, already from a design perspective, I think we should not do that (see my https://github.com/CLIMADA-project/climada_python/pull/642#discussion_r1107390523). If we just add a method that returns a value, but otherwise does not modify the attributes of `Impact`, we only need to test that method (and make sure that all other tests pass unchanged)

aleeciu: cool, then I will 1) just return a value and 2) make a unit test. Then we should be fine I guess.

climada/engine/impact.py
aleeciu: I am quite fine with that and will make the change. But you can't _create_ an `Impact` from `eai_exp`, `at_event` and many others either, can you?

aleeciu: removed from attributes

climada/engine/impact.py
aleeciu: done

climada/engine/impact.py
aleeciu: I am now returning `None` and throwing a warning when `imp_mat` is empty

climada/engine/impact.py
```diff
@@ -390,6 +388,42 @@ def impact_per_year(self, all_years=True, year_range=None):
             year_set[year] = sum(self.at_event[orig_year == year])
         return year_set
 
+    def impact_at_reg(self, exposures):
+        """Aggregate impact at the region_id level (if these are specified in exposure)
+        or at the admin 0 level (if they are not). This method works only if Impact.imp_mat
+        was stored during the impact calculation.
```
emanuel-schmid, line 394: The `exposures` object must be the very same Exposures that was given as argument to `ImpactCalc`, right?
That is a very tight constraint but not verified.
Makes me feel uneasy. I'd be happier if the argument wasn't an `Exposures` object but just an arbitrary aggregation column. So you would call it like this:

```python
impact.impact_at_reg(exposures.gdf.region_id)
```

instead of 

```python
impact.impact_at_reg(exposures)
```

This would transfer all the responsibility for being consistent to the caller.

In any case, the docstring could be more insisting about consistency and perhaps even provide an example.

aleeciu: it now works with user-defined aggregation regions. If `None`, `admin_0` levels are used.

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+        pd.DataFrame
+        """
+        if self.imp_mat.nnz == 0:
+            LOGGER.warning("No Impact.imp_mat was stored during the impact calculation,"
+                           "thus the impact per region cannot be computed")
+            return None
+
+        if agg_regions is None:
+            LOGGER.warning("Aggregation regions were not specified,"
+                           "impact is aggregated at the admin_0 level")
```
peanutfun, line 414: This is expected behavior, so it should not emit a warning
```suggestion
```

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+    def impact_at_reg(self, agg_regions=None):
+        """Aggregate impact on given aggregation regions. This method works 
+        only if Impact.imp_mat was stored during the impact calculation.
+
+        Parameters
+        ----------
+        agg_regions : np.array, list, pd.Series
+            It has the same lenght as the number of coordinates points in exposure 
+            and reports what macro-regions these points belong to. If no aggregation
+            regions are passed, the method aggregates impact at the admin_0 level.
```
peanutfun, line 400:
```suggestion
            regions are passed, the method aggregates impact at the country (admin_0) level.
```

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
             year_set[year] = sum(self.at_event[orig_year == year])
         return year_set
 
+    def impact_at_reg(self, agg_regions=None):
+        """Aggregate impact on given aggregation regions. This method works 
+        only if Impact.imp_mat was stored during the impact calculation.
+
+        Parameters
+        ----------
+        agg_regions : np.array, list, pd.Series
```
peanutfun, line 397:
```suggestion
        agg_regions : np.array, list, pd.Series (optional)
```

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+            It has the same lenght as the number of coordinates points in exposure 
+            and reports what macro-regions these points belong to. If no aggregation
+            regions are passed, the method aggregates impact at the admin_0 level.
+            Default is None.
+
+        Returns
+        -------
+        pd.DataFrame
```
peanutfun, line 405:
```suggestion
        pd.DataFrame
            Contains the aggregated data per event.
            Rows: Events. Columns: Unique aggregate regions.
```

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+        """
```
peanutfun, line 406: Maybe add a usage example? The input parameter `agg_regions` is a bit unintuitive.

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+        if self.imp_mat.nnz == 0:
+            LOGGER.warning("No Impact.imp_mat was stored during the impact calculation,"
+                           "thus the impact per region cannot be computed")
+            return None
```
peanutfun, line 410: Could also return an empty dataframe. Not sure what would be smartest here. I personally would prefer throwing an error because the user should be aware that using the method in this context makes no sense.

climada/engine/impact.py
```diff
@@ -390,6 +388,44 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+                                                             self.coord_exp[:,1]))
+        elif not isinstance(agg_regions, pd.Series):
+            agg_regions = pd.Series(agg_regions)
+
+        at_reg_event = np.hstack([
+                self.imp_mat[:, np.where(agg_regions == reg)[0]].sum(1)
+                for reg in agg_regions.unique()
+                ])
+
+        at_reg_event = pd.DataFrame(at_reg_event, columns=agg_regions.unique())
```
peanutfun, line 425: Maybe use event IDs as indices for this data frame? 🤔

climada/engine/test/test_impact.py
```diff
@@ -467,6 +467,48 @@ def test_local_exceedance_imp_pass(self):
...
+
+        # Read default hazard file
+        hazard = Hazard.from_hdf5(HAZ_TEST_TC)
+
+        # Read an exposure
+        ent = Entity.from_excel(ENT_DEMO_TODAY)
+        ent.check()
+
+        # Calculate impact
+        impact = ImpactCalc(ent.exposures, ent.impact_funcs, hazard).impact(save_mat=True)
```
peanutfun, line 483: I would much prefer if you could create a impact object from scratch here. All you need for this test is a very simple impact matrix (e.g., 2 events x 4 coords) and corresponding exposure coordinates (in this case, 4). Then you can check for the exact values you put into the matrix further below, instead of these seemingly random numbers based on the example data. If you get stuck please notify me and I'll give it a try.

peanutfun (changes requested):

climada/engine/impact.py
aleeciu: agreed, done.

climada/engine/impact.py
aleeciu: agree on throwing error. done.

climada/engine/impact.py
aleeciu: mhh, I am a bit reluctant to add a real example because this is not done in CLIMADA not event for much more complex functions. I added a thorough description of what these `agg_regions` are in the docstring.

climada/engine/impact.py
```diff
@@ -390,6 +388,47 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+            region A, whereas impact from the second centroid will be assigned
+            to area B. If no aggregation regions are passed, the method aggregates
+            impact at the country (admin_0) level.
+            Default is None.
+
+        Returns
+        -------
+        pd.DataFrame
+            Contains the aggregated data per event.
+            Rows: Hzard events. Columns: Aggregation regions.
```
chahank, line 411:
```suggestion
            Rows: Hazard events. Columns: Aggregation regions.
```

climada/engine/impact.py
```diff
@@ -390,6 +388,47 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+        """
+        if self.imp_mat.nnz == 0:
+            raise ValueError("The aggregated impact cannot be computed as no"
+                             "Impact.imp_mat was stored during the impact calculation")
+
+        if agg_regions is None:
+            agg_regions = pd.Series(u_coord.get_country_code(self.coord_exp[:,0],
+                                                             self.coord_exp[:,1]))
+        elif not isinstance(agg_regions, pd.Series):
+            agg_regions = pd.Series(agg_regions)
```
chahank, line 421: Why make it a pandas series instead of a numpy array?

climada/engine/test/test_impact.py
aleeciu: Tests have been rewritten. I am now using the dummy impact object also used in other impact tests.

climada/engine/impact.py
aleeciu: for two reasons:
- I find it cleaner when as you now what impact belongs to what region simply by looking at the dataframe
- I like the `pd.Series.unique()` method because it retains the order. Doing the same with `numpy` is more cumbersome, to my knowledge.

chahank: I would argue that this is a misuse of a pandas series. 

`
  I find it cleaner as you know what impact belongs to what region simply by looking at the dataframe
`
That is independent on using numpy arrays or pandas series. But using a pandas series when it is not needed is imho actually less clean.
`
  I like the pd.Series.unique() method because it retains the order. Doing the same with numpy is more cumbersome, to my knowledge.
`
I would argue that this is suboptimal as an important point is now hidden in the subtle internal working of the `pandas.series.unique()` vs the `np.unique()` methods.

@peanutfun : what do you think?

aleeciu: - on the first point you are right: having a `pd.Series` is not strictly needed to generate a `pd.DataFrame`

- on the second: I am transforming `agg_regions` into `pd.Series` only when the user supplies it as an `np.array' or `list` . But the user can supply a `pd.Series` in the first place. So I don't think it's a misuse, I just wrote the code in a way that it works with a `pd.Series`. 
So I think your doubt is rather: should we use `pandas` for things that can be done with `numpy`? Not sure I have a clear answer to that, as this would probably apply to any code that make use of `pandas`.

chahank: The point is that `pd.series` are objects that are made to be used in pandas dataframes, but they are not made to be used a standalone objects. They are essentially numpy arrays, with some more. Thus, instead of making an unclear use of the `pd.series`, I suggest to use numpy arrays, and to make the ordering requirement clear (as currently this is a hidden feature).

peanutfun: I don't really see how the order is important because the order in the `agg_regions` parameter is determined by the "order" of the exposure points. Also, the order of columns in the final data frame does not appear relevant to me 😅

I would actually argue that the unordered result of `pandas.Series.unique` does not need further explanation, whereas I feel we should add a note to the docstring that the unique items will be ordered when using `np.unique`. So, I think the easiest solution is to just stick to the current implementation.

@chahank: Following our discussion on the "clean" indexing, I just realized that `pandas.Series.unique` returns a numpy array, so everything is fine from that perspective.

climada/engine/impact.py
peanutfun: Great! Thank you

climada/engine/impact.py
```diff
@@ -390,6 +388,47 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+            Contains the aggregated data per event.
+            Rows: Hazard events. Columns: Aggregation regions.
+        """
+        if self.imp_mat.nnz == 0:
+            raise ValueError("The aggregated impact cannot be computed as no"
+                             "Impact.imp_mat was stored during the impact calculation")
+
+        if agg_regions is None:
+            agg_regions = pd.Series(u_coord.get_country_code(self.coord_exp[:,0],
+                                                             self.coord_exp[:,1]))
```
peanutfun, line 419: Maybe using the alpha-3 codes instead of the numeric codes would be more descriptive?

```suggestion
            agg_regions = pd.Series(
                u_coord.country_to_iso(
                    u_coord.get_country_code(self.coord_exp[:,0], self.coord_exp[:,1])
                )
            )
```

climada/engine/impact.py
aleeciu: You are right, the order does not count. I am now using only `numpy` arrays.

climada/engine/impact.py
aleeciu: Implemented.

climada/engine/impact.py
emanuel-schmid: > I would argue that this is a misuse of a pandas series.

Not necessarily. According to the docs a `Series` is a "one-dimensional `ndarray` with axis labels (including time series)". Sounds useful also outside of a `DataFrame`.
🤷

climada/engine/impact.py
peanutfun: Excellent!

climada/engine/impact.py
peanutfun: I very much agree with @emanuel-schmid here 😃

I'm fine with the current implementation. However, please make sure to only call `np.unique` **once**, and store the result for later use. It is currently executed twice and will be costly for large arrays. I can also take care of that when merging. Are we ready for that?

aleeciu:
I guess we are ready to merge

climada/engine/impact.py
```diff
@@ -390,6 +388,55 @@ def impact_per_year(self, all_years=True, year_range=None):
...
 
+    def impact_at_reg(self, agg_regions=None):
+        """Aggregate impact on given aggregation regions. This method works
+        only if Impact.imp_mat was stored during the impact calculation.
+
+        Parameters
+        ----------
+        agg_regions : np.array, list (optional)
+            It has a lenght equal to the number of coordinates points in exposure
+            and it reports what macro-regions these points belong to. For example,
```
emanuel-schmid, line 399:
```suggestion
            It has a lenght equal to the number of coordinates in exposures
            and it reports what macro-regions these coordinates belong to. For example,
```

climada/engine/impact.py
```diff
@@ -390,6 +388,55 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+            asuming there are three centroids and agg_regions = ['A', 'A', 'B']
+            then impact of the first and second centroids will be assigned to
+            region A, whereas impact from the second centroid will be assigned
```
emanuel-schmid, line 402:
```suggestion
            region A, whereas impact from the third centroid will be assigned
```

climada/engine/impact.py
```diff
@@ -390,6 +388,55 @@ def impact_per_year(self, all_years=True, year_range=None):
...
+
+        Returns
+        -------
+        pd.DataFrame
+            Contains the aggregated data per event.
+            Rows: Hazard events. Columns: Aggregation regions.
+        """
+        if self.imp_mat.nnz == 0:
+            raise ValueError(
+                "The aggregated impact cannot be computed as no Impact.imp_mat was "
```
emanuel-schmid, line 415:
```suggestion
                "The aggregated impact cannot be computed as no ``Impact.imp_mat`` was "
```

emanuel-schmid:
Yes, looks good. 😄 I did some cosmetics in the doc string. If your happy with them @aleeciu, let me know or just commit and merge.

peanutfun (approved):

emanuel-schmid:
Oh and one more thing we need to change the CHANGELOG.md

peanutfun:
Will do and handle the merge! 👌

PR #646, opened 2023-02-07
Feature/get stormtype from ibtracs

ChrisFairless:
Changes proposed in this PR:
- When creating a trackset from IBTrACS also read the 'Nature' of a storm. The nature gives information about whether a system is a disturbance, tropical storm, post-transition extratropical storm, or subtropical storm (or whether agencies disagree or don't report). This is potentially useful in visualisations and for generating synthetic tracks.

**Notes:**
- Not urgent! We don't need this in time for the next release!
- Do we need a discussion? There are lots of variables that we exclude when reading from IBTrACS and we can't include them all. A more flexible alternative would be to add a keyword for 'other columns' you want from the file.
- The 'nature' variable is only explicitly created for tracksets from current IBTrACS (not olders csvs), and not for CHAZ, STORM, etc. It looks like there are already other properties that are IBTrACS-only, e.g radius of outermost closed isobar, so hopefully this is ok. I added a check for existence of the variable where relevant.
- When interpolating with `equal_timestep` I chose a forward fill method, rather than the default nearest neighbour.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -178,6 +178,7 @@ class TCTracks():
             - central_pressure (in hPa/mbar)
             - environmental_pressure (in hPa/mbar)
             - basin (for each track position)
+            - nature (for each track position)
             - max_sustained_wind_unit (attrs)
             - central_pressure_unit (attrs)
             - name (attrs)
@@ -575,7 +576,7 @@ class TCTracks():
                         ibtracs_ds[tc_var].values[nonsingular_mask] = (
                             ibtracs_ds[tc_var].sel(storm=nonsingular_mask).interpolate_na(
                                 dim="date_time", method="linear"))
-        ibtracs_ds = ibtracs_ds[['sid', 'name', 'basin', 'time', 'valid_t']
+        ibtracs_ds = ibtracs_ds[['sid', 'name', 'basin', 'time', 'valid_t', 'nature']
                                 + phys_vars + [f'{v}_agency' for v in phys_vars]]
 
         if estimate_missing:
@@ -693,6 +694,7 @@ class TCTracks():
                 'central_pressure': ('time', track_ds.pres.data),
                 'environmental_pressure': ('time', track_ds.poci.data),
                 'basin': ('time', track_ds.basin.data.astype("<U2")),
+                'nature': ('time', track_ds.nature.data.astype("<U2")),
             }, coords={
                 'time': track_ds.time.dt.round('s').data,
                 'lat': ('time', track_ds.lat.data),
@@ -1500,6 +1502,8 @@ class TCTracks():
             track_int = track.resample(time=time_step, skipna=True)\
                              .interpolate('linear')
             track_int['basin'] = track.basin.resample(time=time_step).nearest()
+            if hasattr(track_int, 'nature'):
+                track_int['nature'] = track.nature.resample(time=time_step).ffill()
             track_int['time_step'][:] = time_step_h
             lon_int = lon.resample(time=time_step).interpolate(method)
             lon_int[lon_int > 180] -= 360
@@ -1879,6 +1883,7 @@ def _read_ibtracs_csv_single(file_name):
     tr_ds['central_pressure'] = ('time', cen_pres)
     tr_ds['environmental_pressure'] = ('time', dfr['penv'].values.astype('float'))
     tr_ds['basin'] = ('time', dfr['gen_basin'].values.astype('<U2'))
+    tr_ds['nature'] = ('time', dfr['nature'].values.astype('<U2'))
     tr_ds.attrs['max_sustained_wind_unit'] = max_sus_wind_unit
     tr_ds.attrs['central_pressure_unit'] = 'mb'
     tr_ds.attrs['name'] = name
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -1385,7 +1385,8 @@ class TCTracks():
             track = ds_dict[f'track{i}']
             track.attrs['orig_event_flag'] = bool(track.attrs['orig_event_flag'])
             # when writing '<U2' and reading in again, xarray reads as dtype 'object'. undo this:
-            track['basin'] = track['basin'].astype('<U2')
+            for varname in ['basin', 'nature']:
+                track[varname] = track[varname].astype('<U2')
             data.append(track)
         return cls(data)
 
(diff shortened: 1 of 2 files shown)


tovogt:
Thanks for the PR! I can imagine that this might be useful information for some people.


> * Do we need a discussion? There are lots of variables that we exclude when reading from IBTrACS and we can't include them all. A more flexible alternative would be to add a keyword for 'other columns' you want from the file.

Here is a list of all non-agency-specific variables that we currently ignore:
```python
[
    'numobs', 'season', 'number', 'subbasin', 'name', 'source_usa', 'source_jma',
    'source_cma', 'source_hko', 'source_new', 'source_reu', 'source_bom', 'source_nad',
    'source_wel', 'source_td5', 'source_td6', 'source_ds8', 'source_neu', 'source_mlc',
    'iso_time', 'nature', 'lat', 'lon', 'wmo_wind', 'wmo_pres', 'wmo_agency', 'track_type',
    'main_track_sid', 'dist2land', 'landfall', 'iflag', 'storm_speed', 'storm_dir'
]
```
When including agency-specific variables, the list is even longer. I tend to prefer supporting a keyword argument for "other variables" to include instead of adding explicit support for the "nature" variable. Maybe we can start by supporting non-agency-specific variables, because the agency-specific need to be treated a bit differently?

> * The 'nature' variable is only explicitly created for tracksets from current IBTrACS (not olders csvs), and not for CHAZ, STORM, etc. It looks like there are already other properties that are IBTrACS-only, e.g radius of outermost closed isobar, so hopefully this is ok. I added a check for existence of the variable where relevant.

In general, I think adding additional variables should be fine. We have to avoid breaking file read/write operations though.

> * When interpolating with `equal_timestep` I chose a forward fill method, rather than the default nearest neighbour.

Why would you do that?

climada/hazard/tc_tracks.py
```diff
@@ -1885,6 +1890,8 @@ def _read_ibtracs_csv_single(file_name):
     tr_ds.attrs['sid'] = name
     tr_ds.attrs['orig_event_flag'] = bool(dfr['original_data']. values[0])
     tr_ds.attrs['data_provider'] = dfr['data_provider'].values[0]
+    if hasattr(dfr, 'nature'):
+        tr_ds['nature'] = ('time', dfr['nature'].values.astype('<U2'))
```
tovogt, line 1894: I don't think it's necessary to support this. The legacy CSV format is a completely undocumented data type. I have never seen such a file apart from our test file. I don't know anyone who is using this functionality.

ChrisFairless: Good point! This was just to stop the tests failing :p

ChrisFairless:
Sorry, I missed your main response last week!

> When including agency-specific variables, the list is even longer. I tend to prefer supporting a keyword argument for "other variables" to include instead of adding explicit support for the "nature" variable. Maybe we can start by supporting non-agency-specific variables, because the agency-specific need to be treated a bit differently?

I think this makes sense. I'll update the pull request, but for non-agency-specific variables. A few of the variables you list will be useful in different situations.

> In general, I think adding additional variables should be fine. We have to avoid breaking file read/write operations though.

To be clear: are you suggesting always reading 'nature' _and_ allowing additional variables, or just allowing additional variables here?

> * When interpolating with `equal_timestep` I chose a forward fill method, rather than the default nearest neighbour.

Not sure why I chose this now ... I think it was from an overabundance of caution around landfalls. Let's go for nearest neighbour then.

tovogt:
> > In general, I think adding additional variables should be fine. We have to avoid breaking file read/write operations though.
> 
> To be clear: are you suggesting always reading 'nature' _and_ allowing additional variables, or just allowing additional variables here?

I think this remark is mostly about this section of code:
https://github.com/CLIMADA-project/climada_python/blob/cfebf28ab6a4fe1bed5938e530f959e54db5eb40/climada/hazard/tc_tracks.py#L1387-L1389
It's a bit unclear what to do when we support additional variables of varying dtype. We can't just assume that all "object" variables are actually supposed to be `"<U2"` like here:
```python
# when writing '<U2' and reading in again, xarray reads as dtype 'object'. undo this:
for varname in track.variables:
    if track[varname].dtype == "object":
        track[varname] = track[varname].astype('<U2')
```
Maybe we can at least assume that all "object" dtypes are actually strings. Then we could do something like this:
```python
# when writing '<U2' and reading in again, xarray reads as dtype 'object'. undo this:
for varname in track.variables:
    if track[varname].dtype == "object":
        strlen = track[varname].str.len().max().item()
        track[varname] = track[varname].astype(f"<U{strlen}")
```

climada/hazard/tc_tracks.py
tovogt: I don't understand why the tests would fail without this. Can you elaborate, please?

ChrisFairless:
Assuming a string seems reasonable. Is there a reason we always want to limit its length? What are the memory/efficiency gains there?

tovogt:
The most important aspect for me was that you should be able to cycle, i.e., storing data and reading it again should yield the identical object. I don't know about the memory requirements.

climada/hazard/tc_tracks.py
ChrisFairless: Oh lol the tests were failing due to a line I added and later removed. We can definitely drop this.
I'll put together a new pull request shortly.

peanutfun:
> I'll put together a new pull request shortly.

@ChrisFairless I don't think that's necessary, you can simply rewrite your changes here. But if you still want to, I would wait with my review until this or the new PR are ready. In general, I second @tovogt's suggesion of reading any additional variable through keyword arguments.

As a side note, I guess this discussion goes to show that it's helpful to discuss any additions *before* implementing them, although I very much appreciate the initiative of coming up with suggestions through a PR in the first place 😅

tovogt:
> As a side note, I guess this discussion goes to show that it's helpful to discuss any additions _before_ implementing them, although I very much appreciate the initiative of coming up with suggestions through a PR in the first place sweat_smile

In my opinion, it's good to create a draft PR right from the start when proposing a new feature, and have a discussion in the PR. I know that it somehow makes the PR hard to follow because it has so many posts. Because of that, it might make sense to create a new PR at some point. But the big advantage is that you avoid a lot of misunderstandings because you talk directly about concrete code changes. When talking about features at a very high level, I often discuss for a while before noticing that there was actually a misunderstanding about the kind of change we are actually discussing.

ChrisFairless:
Definitely definitely agree here!!

But in this case I needed the feature pretty urgently so I coded first, PR'd later 😅

peanutfun:
How to continue here? There seems to be at least one unresolved discussion. @ChrisFairless will you update this PR or should we close it in favor of a new one?

ChrisFairless:
Fine to close this.

I want to update things as discussed above, but it'll take some time I don't have right now.

bguillod:
Since nobody seems to have capacity to implement the flexible reading of any variable in IBTrACS, it looks to me like the requested addition to this PR is not very relevant or does not have any specific use case for the moment.
On the other hand, we urgently need the storm type information (and only that additional variable) for the new TC synthetic tracks methodology. Therefore I'd like this PR to be reconsidered for merging into the develop branch, to allow us to move ahead and finalize the new synthetic tracks model. Does this make sense?

chahank:
Please address the review comments in the code and then we can proceed. Thanks.

tovogt:
If nobody has time to do this "right", I will take this on immediately. We have discussed everything, the coding should be really easy. I don't see why we would go with the "nature" variable only. I will report back later today.

bguillod:
> If nobody has time to do this "right", I will take this on immediately. We have discussed everything, the coding should be really easy. I don't see why we would go with the "nature" variable only. I will report back later today.

Sure, you're most welcome to implement it if you think that's a worthwhile addition. Thanks a lot, very much appreciated!

emanuel-schmid:
🙌

tovogt:
Let's continue here: #728

Issue #665, opened 2023-02-28
Tests fail when run outside the CLIMADA working directory


aleeciu:
Tests fail when run from a directory other than the climada directory, because `CONFIG` cannot find the paths it's looking for. @peanutfun and I wondered whether this is expetected behaviour or not.

emanuel-schmid:
Yes this is the expected behavior - to some extent. The basic installation test as described in the [Simple (installation) instructions 6.](https://climada-python.readthedocs.io/en/latest/guide/install.html#simple-instructions) is supposed to pass no matter where but the others depend on config values that are not defined in the default config file `climada/conf/climada.conf`.
If you want them to pass everywhere you can just move or copy the "test-config" file `./climada.conf` file to your home directory or to your working directory and if the test still fails, it's a bug. 😀

peanutfun:
If I read the [docs on config files](https://climada-python.readthedocs.io/en/latest/guide/Guide_Configuration.html#Configuration-files) correctly, the `/climada.conf` is only read when working in this directory. Why are its contents not part of the "global" `/climada/conf/climada.conf`? I would expect the tests to always pass, independent from the working directory 🤔

emanuel-schmid:
Just to restrict the "global" config to its essentials, without stuff that is not strictly required for using climada out of the box.
Tests are primarily meant for developers, so they can see what has been broken by their progress. The ordinary user doesn't need to run them.

aleeciu:
Ok, I see it's intended behaviour and that there are good reasons for it. But would it make sense to throw a warning when tests are run from the "wrong" repo?

Currently, we get an error saying that CONFIG does not contain some of the sought paths, and it's not immediate at all to understand what is happening.

Maybe, when we try to load CONFIG and this fails, we can issue a warning like: "This likely happens because CONFIG cannot find the paths is looking for, use climada as wd or make sure CONFIG is properly modified".

PR #670, opened 2023-03-03
Accurate error message for missing configuration

emanuel-schmid:
Changes proposed in this PR:
- When ever a config value is missing in (beneath) `climada.CONFIG` an error message is provided that is clear about it and suggests a list of files that could be changed to solve the problem.

This PR fixes (to some extent) #665

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/util/config.py
+++ b/climada/util/config.py
@@ -50,6 +50,21 @@ class Config():
     Like this it's possible to refer to e.g. test files in the climada sources.
     """
 
+    def __getattribute__(self, __name):
+        # the only reason for overriding __getattribute__: better error messages in case a
+        # configuration value is asked for but never defined in any of the possible config files
+        try:
+            return super().__getattribute__(__name)
+        except AttributeError:
+            conf_files = [Path(_find_in_parents(conf_dir, CONFIG_NAME))
+                          if _find_in_parents(conf_dir, CONFIG_NAME)
+                          else conf_dir / CONFIG_NAME
+                          for conf_dir in CONFIG_DIRS[::-1]]
+            raise AttributeError(  # pylint: disable=raise-missing-from
+                f"there is no '{__name}' configured for '{super().__getattribute__('_name')}'."
+                f" check your config files: {conf_files}"
+            )
+
     def __str__(self):
         # pylint: disable=bare-except,multiple-statements,too-complex
         try: return self.str()
@@ -63,13 +78,13 @@ class Config():
         try: return str(self.list())
         except: pass
         return '{{{}}}'.format(", ".join([
-            f'{k}: {v}' for (k, v) in self.__dict__.items() if not k == '_root'
+            f'{k}: {v}' for (k, v) in self.__dict__.items() if not k in {'_name', '_root'}
         ]))
 
     def __repr__(self):
         return self.__str__()
 
-    def __init__(self, val=None, root=None):
+    def __init__(self, name, val=None, root=None):
         """
         Parameters
         ----------
@@ -81,6 +96,7 @@ class Config():
             the value of the Config in case it's basic, by default None,
             when a dictionary like object is created
         """
+        self._name = name
         if val is not None:
             self._val = val
         if root is None:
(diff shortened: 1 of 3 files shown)


climada/util/config.py
```diff
@@ -50,6 +50,21 @@ class Config():
...
+    def __getattribute__(self, __name):
+        # the only reason for overriding __getattribute__: better error messages in case a
+        # configuration value is asked for but never defined in any of the possible config files
+        try:
+            return super().__getattribute__(__name)
+        except AttributeError:
+            conf_files = [Path(_find_in_parents(conf_dir, CONFIG_NAME))
+                          if _find_in_parents(conf_dir, CONFIG_NAME)
+                          else conf_dir / CONFIG_NAME
+                          for conf_dir in CONFIG_DIRS[::-1]]
```
chahank, line 62: I do not fully understand where the variables `CONFIG_NAME` and `CONFIG_DIRS` are defined (or imported).

climada/util/config.py
```diff
@@ -50,6 +50,21 @@ class Config():
...
 
+    def __getattribute__(self, __name):
+        # the only reason for overriding __getattribute__: better error messages in case a
+        # configuration value is asked for but never defined in any of the possible config files
+        try:
+            return super().__getattribute__(__name)
+        except AttributeError:
+            conf_files = [Path(_find_in_parents(conf_dir, CONFIG_NAME))
+                          if _find_in_parents(conf_dir, CONFIG_NAME)
+                          else conf_dir / CONFIG_NAME
```
chahank, line 61:
```suggestion
                          else Path(conf_dir / CONFIG_NAME)
```
Would this not be more consistent?

climada/util/test/test_config.py
```diff
@@ -61,6 +61,12 @@ def test_substitute(self):
         self.assertEqual(conf.a._root, conf._root)
         self.assertEqual(conf.a.str(), 'https://host/page.domain')
 
+    def test_missing(self):
+        with self.assertRaises(AttributeError) as ve:
+            CONFIG.hazard.fire_fly.population.str()
+        self.assertIn("there is no 'fire_fly' configured for 'hazard'", str(ve.exception))
```
chahank, line 67: What does one get for a message for `CONFIG.random_module` and for `CONFIG.hazard.tropical_cyclone.some_file`? Should this also be tested maybe?

chahank:
Looks good to me, thanks for the update! A few optional minor comments.

climada/util/config.py
emanuel-schmid: :shrug: `conf_dir` is a path though and even with the proposed change it must be one.

climada/util/config.py
emanuel-schmid: Because `CONFIG` kind of _has_ to be at the bottom of the file and both `CONFIG_NAME` and `CONFIG_DIRS` have _one_ purpose: to define `CONFIG`. So I want them to be together, then I don't have to wonder or scroll.

climada/util/test/test_config.py
emanuel-schmid: This:
- "there is no 'random_module' configured for 'climada.CONFIG' ..."
- "there is no 'some_file' configured for 'tropical_cyclone' ..."
- 
The latter should not be tested as there is nothing such a test could contribute. The former possibly to some extent as the 'climada.CONFIG' is something special. I wouldn't but if you if you think it's worth it I can add it.

chahank: Well, the given that there is nothing currently, but it would be possible (and in particular a user can add it right?) I would suggest adding it since it is very little extra work.

emanuel-schmid: 👍 sure.
Just for understanding - what do you mean by 
> Well, the given that there is nothing currently, but it would be possible (and in particular a user can add it right?)

?

aleeciu:
Thanks @emanuel-schmid, it looks good. Fun fact: I tried to reproduce the error on #665 from `main` to compare the output with the one I'd get in this branch, but I couldn't 😵 .. tests worked in `main` even when run from different folders ...

emanuel-schmid:
What was the config value you were looking for? Perhaps it got established in the default config file (`climada/conf/climada.conf`) meanwhile?
I'll merge this nevertheless though...

PR #676, opened 2023-03-20
Change test data impf.id to test non-default params

chahank:
Changes proposed in this PR:
This PR does NOT provide a bugfix after all, as this was fixed by chance in PR #666 . However, the tests do not test for this case, and thus it was added to the tests. It is very minimalistic and could be made even more consistent. But at least it provides a minimal improvement.

Basically, the impact function id is by default set to 1 under the hood in CLIMADA if none is specified. Currently, the tests only work with impact function id 1. Thus, the use case when multiple impact functions with different ids are used is not tested. The default exposures for polygons was now set to use impact functions with id = 2.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ Removed:
 - Refactored `Exposure.assign_centroids` using a new util function `u_coord.match_centroids` [#602](https://github.com/CLIMADA-project/climada_python/pull/602)
 - Renamed `climada.util.coordinate.assign_grid_points` to `match_grid_points` and `climada.util.coordinates.assign_coordinates` to `match_coordinates`
 [#602](https://github.com/CLIMADA-project/climada_python/pull/602)
-- Modified the method to disaggregate lines in the lines_polygons_handler utility module in order to better conserve the total length of all lines on average [#679](https://github.com/CLIMADA-project/climada_python/pull/679).
+- Modified the method to disaggregate lines in the `lines_polys_handler` utility module in order to better conserve the total length of all lines on average [#679](https://github.com/CLIMADA-project/climada_python/pull/679).
+- Added test for non-default impact function id in the `lines_polys_handler` [#676](https://github.com/CLIMADA-project/climada_python/pull/676)
 
 ### Fixed
 


Evelyn-M:
I'm not sure how the tests really checks for this case now. The two impact functions in the currently tested impact function set are functionally the same (the welker ones), and i cannot see how it is really ensured that not the imp func with ID 1 is taken (as the impact values would be the same)

Evelyn-M:
aka, would it make sense to explicitly check that the interpolated exp_poly still has imp func id 2? and perhaps with a different impact function altogether?

chahank:
> aka, would it make sense to explicitly check that the interpolated exp_poly still has imp func id 2? and perhaps with a different impact function altogether?

This is exactly what is done in all the tests: 

```
def check_unchanged_geom_gdf(self, gdf_geom, gdf_pnt):
    """Test properties that should not change"""
    for n in gdf_pnt.index.levels[1]:
        sub_gdf_pnt = gdf_pnt.xs(n, level=1)
        rows_sel = sub_gdf_pnt.index.to_numpy()
        sub_gdf = gdf_geom.loc[rows_sel]
        self.assertTrue(np.alltrue(sub_gdf.geometry.geom_equals(sub_gdf_pnt.geometry_orig)))
    for col in gdf_pnt.columns:
        if col not in COL_CHANGING:
            np.testing.assert_allclose(gdf_pnt[col].unique(), gdf_geom[col].unique())
```

chahank:
> I'm not sure how the tests really checks for this case now. The two impact functions in the currently tested impact function set are functionally the same (the welker ones), and i cannot see how it is really ensured that not the imp func with ID 1 is taken (as the impact values would be the same)

That is correct.

Evelyn-M:
> > aka, would it make sense to explicitly check that the interpolated exp_poly still has imp func id 2? and perhaps with a different impact function altogether?
> 
> This is exactly what is done in all the tests:
> 
> ```
> def check_unchanged_geom_gdf(self, gdf_geom, gdf_pnt):
>     """Test properties that should not change"""
>     for n in gdf_pnt.index.levels[1]:
>         sub_gdf_pnt = gdf_pnt.xs(n, level=1)
>         rows_sel = sub_gdf_pnt.index.to_numpy()
>         sub_gdf = gdf_geom.loc[rows_sel]
>         self.assertTrue(np.alltrue(sub_gdf.geometry.geom_equals(sub_gdf_pnt.geometry_orig)))
>     for col in gdf_pnt.columns:
>         if col not in COL_CHANGING:
>             np.testing.assert_allclose(gdf_pnt[col].unique(), gdf_geom[col].unique())
> ```

ah yeah ok, that was a bit cryptic to spot.
looks good, then.

Issue #674, opened 2023-03-20
Read storm type ("nature") from IBTrACS data


peanutfun:
@ChrisFairless proposed to read the storm type data from IBTrACS data in PR #646:

> Changes proposed in this PR:
>
> * When creating a trackset from IBTrACS also read the 'Nature' of a storm. The nature gives information about whether a system is a disturbance, tropical storm, post-transition extratropical storm, or subtropical storm (or whether agencies disagree or don't report). This is potentially useful in visualisations and for generating synthetic tracks.

@tovogt https://github.com/CLIMADA-project/climada_python/pull/646#issuecomment-1420974181:

> I tend to prefer supporting a keyword argument for "other variables" to include instead of adding explicit support for the "nature" variable. Maybe we can start by supporting non-agency-specific variables, because the agency-specific need to be treated a bit differently?

bguillod:
I assume this can be closed following PR #728
If not please re-open

PR #688, opened 2023-03-30
Improve random walk synthetic track generation

ChrisFairless:
*** DRAFT: WORK IS STILL ONGOING! ***

Changes proposed in this PR:
- This is a full overhaul for how CLIMADA generates synthetic tracks from IBTrACS
- It's designed to fix the low bias in the existing track generation scheme
- More details once it's ready for review!

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Since this is a very large proposed change to a CLIMADA core module, it would be useful to have a report/publication showing the improvements provided by the changes. How good are the generate random tracks for risk assessment? What are the limitations? How is it an improvement over the previous version? 

Thanks for improving the tropical cyclone track generation!

chahank:
> The best way to do this was discussed in #646 and soon I'll implement it in the more generalised way we agreed on.

Issues #646 is closed. Do you mean  #674 ?

chahank:
Separate idea: will it be possible after this pull request to generate random tracks with a frequency and intensity shift compared to the historical tracks? Then it could also be used to generate tracks for future climate projections using the Knutson parameters. That would be very neat.

bguillod:
> Separate idea: will it be possible after this pull request to generate random tracks with a frequency and intensity shift compared to the historical tracks? Then it could also be used to generate tracks for future climate projections using the Knutson parameters. That would be very neat.

I do not understand what you mean. What can be done is altering frequency or intensity of each synthetic track of historical tracks according to Knutson parameters (after one fixes the code there, as I've already highlighted in an issue, since applying both intensity and frequency changes in fundamentally wrong; but this is a separate problem).

But this does not change between the original methodology and this new one, right? I might well misunderstand what your point was though.

chahank:
Sorry to have been unclear. Let me try again.

For generating a random track, the algorithm perturbs the direction at each step, and as I understood also the intensity (with this PR). As I understood, this intensity perturbation is also done randomly (not each random track has the same intensity perturbation, and the perturbation is not uniform over the whole track).

If this is correct, and only if, one could with the Knutson parameter not just shift entire tracks, but instead shift the distribution of intensity changes used in the random track generation. Then, one would have two different random track sets for the climate today and the climate projection.

ChrisFairless:
This is a thing that confuses me with the way CLIMADA implements Knutsen (raised by @bguillod in #453 ).

You can implement a change in event intensities by adjusting the event frequencies. The intensities change as a consequence. CLIMADA could serve the same events in every TC trackset, but with different frequencies, determined by climate change scenario and year.

Doing this has a very nice extra benefit: we don't have to scale pre-calculated windfields, which would make any TC modeller shudder.

chahank:
> This is a thing that confuses me with the way CLIMADA implements Knutsen (raised by @bguillod in #453 ).
> 
> You can implement a change in event intensities by adjusting the event frequencies. The intensities change as a consequence. CLIMADA could serve the same events in every TC trackset, but with different frequencies, determined by climate change scenario and year.
> 
> Doing this has a very nice extra benefit: we don't have to scale pre-calculated windfields, which would make any TC modeller shudder.

Sure. This is because the CLIMADA Knutson parameter implementation is, and was NEVER meant to be an actual tropical cyclone climate model. So, it is great to see that you might be able to elevate it to such as level.

climada/hazard/tc_tracks_synth.py
```diff
@@ -834,6 +1720,1497 @@ def _check_decay_values_plot(x_val, v_lf, p_lf, v_rel, p_rel):
...
+
+        if np.any(extended_cat >= 0):  
+            # Kill the extended track 12 hours after dropping below Cat 1
+            buffer_frames = int(np.ceil(12/time_step_h))
+            cutoff_idx = min(np.where(extended_cat >= 0)[0][-1] + buffer_frames, extended_cat.size-1)
+            track = track.isel(time=slice(None, cutoff_idx))
+
+        # If the parent storm finished as extratropical, kill any extension frames at times beyond the original track end
+        original_size = tracks_list[i].time.size
+        if hasattr(track, 'nature'):
```
bguillod, line 2652: @ChrisFairless Is my understanding correct that with this check, the model would work also with tracks without the 'nature' field but simply would not be apply to cut off the track in such a case?

climada/hazard/tc_tracks_synth.py
```diff
@@ -274,87 +647,255 @@ def _one_rnd_walk(track, nb_synth_tracks, max_shift_ini, max_dspeed_rel, max_ddi
...
-
-    ens_track.append(track)
-    cutoff_track_ids_ts = []
-    cutoff_track_ids_tc = []
+    ens_track.append(track.copy(True))
+
+    # calculate historical track values that are used for synthetic track modelling
+    if land_geom is not None:
+        # compute minimum pressure occurring on or after each timestep
+        if int(os.getenv('TRACKGEN_TEST_TARGET_PRESSURE_OVER_SEA')):
```
bguillod, line 665: Can you explain the rationale behind this @ChrisFairless ?
My understanding is: you only consider the minimum central pressure of all ocean points only after each timestep (including also the last central pressure value, which I guess is to make sure one value is available for each time step).
Why is that? In which case would you expect minimum future central pressure to be lower (i.e. stronger) if we include also land points? And if this case arises, why not accounting for it?

chahank:
Due to low activity, this draft pull request will be closed. Please feel free to reopen it at any time!

peanutfun:
@emanuel-schmid Can we expect some updates on this? If not, I would close again due to inactivity.

peanutfun:
Closing due to inactivity. Feel free to reopen at any time!

bguillod:
Yes fine to close the PR as long as the branch is kept, we will re-open it in time.

PR #692, opened 2023-04-06
Add module for calibrating impact functions

### LINKED ISSUE #680: Add utility module for calibrating impact functions
This issue collects ideas for an impact function calibration module

## Input information

For calibrating any impact function, we need to specify four main things:

* The impact data. This is the "real world" data we want to use to calibrate the impact function. The computed ("modeled") impact should be as close to it as possible.
* The impact function. We have to specify what its overall shape is and how it is parameterized.
* The cost function. It relates the modeled impact with the impact data and returns a scalar that indicates how close they are. The task for calibrating an impact function is reduced to minimizing the cost function.
* The impact estimation task. As the impact function is varied by the optimization algorithm, we "only" need to supply the usual Hazard and Exposure.

## Calibration with Bayesian Optimization

This idea is based on the Petals `calib_opt` module by @timschmi95, see https://github.com/CLIMADA-project/climada_petals/blob/feature/calib_opt/climada_petals/engine/calib_opt.py

The module wants us to instantiate a `BayesianOptimization` instance. Its main parameter is a function that has the search space parameters as input and returns the target function value. In this case, the target function value is *maximized*, which simply means that we have to invert the cost function value.

With regard to the impact data I would take a simplistic approach. To accomodate multiple events and multiple regions, the input data should be a `pandas.DataFrame` with the `event_id` and index and arbitrary (!) columns that can (!) represent the target impact values. The cost function then takes this data frame and the impact object of each iteration and the user can define how the cost should then be computed from both objects. In this case, we assume that the columns in the impact data are impacts per region. We compute the difference for each region and event, square all results, take the log for the sum of all regions and then sum the result of all events (as an overly complicated example for what we can do with this):

```python
def my_cost_function(data: pd.DataFrame, impact: climada.engine.Impact) -> float:
    return np.sum(np.log10(((data - impact.impact_at_reg())**2).sum(axis=1)))
```

The other important input it the impact function. Its parameters are what the optimizer calibrates in order to minimize the cost function. The user has to supply a "generator" function that takes the set of calibrated parameters and returns a new impact function set. In this example, we take the TC impact function by Emanuel 2011 and intend to vary two of the three parameters:

```python
def my_impf_set_gen(v_thresh: float, v_half: float) -> climada.entity.ImpactFuncSet:
    return climada.engine.ImpactFuncSet([
        climada.entity.ImpfTropCyclone.from_emanuel_usa(
            v_thresh=v_thresh, v_half=v_half
        )
    ])
```

Now we only need a utility function to plug everything together:

```python
from bayes_opt import BayesianOptimization

def calibrate(
    hazard: Hazard,
    exposure: Exposures,
    impact_data: pd.DataFrame,
    impact_func_gen: Callable[..., ImpactFuncSet],
    cost_func: Callable[[pd.DataFrame, Impact], float],
    param_bounds: Mapping[str, Tuple[float, float]],
    optimizer_kwds: Mapping[str, Any],
    impact_calc_kwds: Mapping[str, Any],
):
    """Calibrate parameters for an impact function"""
    # Match event IDs in hazard and impact_data
    hazard = hazard.select(event_id=impact_data.index)

    # Assign the centroids
    exposure.assign_centroids(hazard)

    # Create the calibration function
    def func(**kwargs):
        impf_set = impact_func_gen(**kwargs)
        impact = ImpactCalc(exposure, impf_set, hazard).impact(
            assign_centroids=False, **impact_calc_kwds
        )
        return 1 / cost_func(impact_data, impact)

    # Create the optimizer (with some default settings)
    optimizer_opts = dict(random_state=1, allow_duplicate_points=True)
    optimizer_opts.update(**optimizer_kwds)
    optimizer = BayesianOptimization(f=func, pbounds=param_bounds, **optimizer_opts)

    return optimizer
```

The user can then call the optimizer like advertised by the BayesianOptimization docs. We assume we have a hazard, an exposure, and the impact data defined somwhere. Then we only need to define the bounds for the parameters we are looking for. The BayesianOptimization package automatically deduces the parameter names and initial values from these automatically.

```python
optimizer = calibrate(
    hazard=hazard,
    exposure=exposure,
    impact_data=impact_data,
    impact_func_gen=my_impf_set_gen,
    cost_func=my_cost_function,
    # NOTE: Bad example. These parameters must be constrained because v_thresh < v_half is required!
    param_bounds={"v_half": (0, 100), "v_thresh": (0, 50)},  
    optimizer_kwds=dict(verbose=2),
    impact_calc_kwds=dict(save_mat=True),  # To be able to call impact.imact_at_reg()
)
optimizer.maximize(init_points=100, n_iter=200)
result = optimizer.max["params"]

# Use generator to receive the function we want
result_impf_set = my_impf_set_gen(result)
```

peanutfun:
Add a generalized (!) module for calibrating impact functions.

* Base module defining data structures to derive from for building your own impact function calibrator
* Calibration module based on `scipy.minimize`.
* Calibration module based on `bayesian-optimization` package.
* Helper structures to evaluate and visualize the calibration results.
* Calibration tutorial based on `BayesianOptimizer`.
* Unit and integration tests.

This PR fixes #680 

### PR Author Checklist

### PR Reviewer Checklist


climada/util/calibrate/impact_func.py
```diff
@@ -0,0 +1,111 @@
...
+    """Define the output of a calibration task"""
+
+    params: Mapping[str, Number]
+    target: Number
+    success: bool
+    result: Optional[OptimizeResult] = None
+
+
+@dataclass
+class Optimizer(ABC):
```
chahank, line 54: As I first thought of this class, I did not expect it to be a `dataclass`. Is there a strong reason to make it a dataclass instead of a normal class? The purpose of this class is to compute values right? Not to store data?

peanutfun: `dataclass` relieves us from writing trivial, yet error-prone `__init__` methods. This is particularly important because we want to derive from `Optimizer` and might want to add more attributes in the derived class. Compare:
```python
@dataclass
class Optimizer(ABC):
    input: Input

@dataclass
class Derived(Optimizer):
    other_input: OtherInput

class OptimizerAlt(ABC):
    def __init__(input):
        self.input = input

class DerivedAlt(OptimizerAlt):
    def __init__(input, other_input):
        self.other_input = other_input
        super().__init__(input)

# Called the same way
Derived(my_input, my_other_input)
DerivedAlt(my_input, my_other_input)
```

chahank: I see. I am a bit confused still. This is just how class inheritance is made in Python. Do you suggest using a dataclass just to handle the `__init__`, even though it is not a dataclass?

peanutfun: I can't quite follow. It is a dataclass when I decorate it as such. The term "dataclass" might be a bit misleading, but the [docs](https://docs.python.org/3/library/dataclasses.html) just state:

> [dataclasses](https://docs.python.org/3/library/dataclasses.html#module-dataclasses) — Data Classes
>
> This module provides a decorator and functions for automatically adding generated [special method](https://docs.python.org/3/glossary.html#term-special-method)s such as __init__() and __repr__() to user-defined classes.

This is my whole intention here.

chahank: I mean there are certainly different opinions on how to best make classes, but dataclasses are meant to be used to create classes that mostly simply store data. The proposed class `Optimizer` however is exactly the opposite, a class that does mostly process data. What I understand from your argument above is that all classes should always be with the decorator dataclass. 

If you could clarify when you would NOT use a dataclass, that might help. 

https://stackoverflow.com/questions/47955263/what-are-data-classes-and-how-are-they-different-from-common-classes

peanutfun: I see that this is the intended use but I still feel like you are focusing on the fact that the decorator is called "dataclass" and you percive the class as not being a data class. [attrs](https://www.attrs.org/en/stable/examples.html) does the exact same thing (and more) using the `@define` decorator. Would you also argue against that? I feel my use of `@dataclass` is justified here. I just want to remove boilerplate code. I would argue that the use case for `dataclass` is that a class stores objects that are passed through its constructor with little to no modifications (hence resulting in a lot of trivial boilerplate code). Using `dataclass` to simplify that is independent from any other functionality of the class.

I was trying things out and I find the resulting code clearer. But we can switch back to the usual `__init__` at any time.

climada/util/calibrate/impact_func.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Parameters
+    ----------
+    hazard : climada.Hazard
+        Hazard object to compute impacts from
+    exposure : climada.Exposures
+        Exposures object to compute impacts from
+    data : pandas.Dataframe
+        The data to compare computed impacts to. Index: Event IDs matching the IDs of
+        ``hazard``. Columns: Arbitrary columns.
```
chahank, line 56: Is this not too restrictive to have it event-matching? It might be sensible to optimize for the total over many events for instance. So, if this is not adding too much complexity, I would allow for it (since the only important factor, at first sight, is the cost function matching the data).

climada/util/calibrate/impact_func.py
```diff
@@ -0,0 +1,225 @@
...
+    cost_func : Callable
+        Function that takes an ``Impact`` object and a ``pandas.Dataframe`` as argument
+        and returns a single number. The optimization algorithm will try to minimize this
+        number. See this module for a suggestion of cost functions.
+    impact_func_gen : Callable
```
chahank, line 61: I would avoid the `_gen` as this can easily be confused for a `generator` in Python.

peanutfun: How would you call a function that returns a function?

chahank: I would try for a name that helps me understand what happens, not what it technically is. In this case, I think `impact_function` would be enough, as this is what it is. Or maybe `parametrized_impact_function`.

peanutfun: Careful, it's not am impact function, it is a function that *returns* an impact function. So `impact_function` would be severely misleading

chahank: Well, it is a function. It is true that it might be confused with the object `ImpactFunction`. But it is definitely simply a parametrized function.

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -0,0 +1,160 @@
...
+            pbounds=self.input.bounds,
+            verbose=verbose,
+            random_state=random_state,
+            allow_duplicate_points=allow_duplicate_points,
+            **bayes_opt_kwds,
+        )
+
+    def _target_func(self, impact: Impact, data: pd.DataFrame) -> Number:
+        """Invert the cost function because BayesianOptimization maximizes the target"""
+        return 1 / self.input.cost_func(impact, data)
```
tovogt, line 89: This is not mathematically equivalent to minimizing the cost function. You should rather switch signs:
```suggestion
        return -self.input.cost_func(impact, data)
```
First of all, it is undefined if the cost function takes the value 0, but it also has very different slopes for large values so that convexity is lost.

If the user engineers the cost function in a way that it is convex and has good optimization properties, this transformation will break this.

bguillod:
Hi,
Is there any plan to calibrate impact functions based on several hazards? To be more specific: I am here thinking that the sum of impacts from both TC wind and surge should be calibrated to reported damages for TC events, as opposed to calibrating wind only and assuming wind is a proxy for everything.

chahank:
This might come in the future. Please feel free to contribute if you would like to see more features.

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Absolutely right! I was afraid that something internal in the optimizer might break when using negative numbers, but it works just fine.

peanutfun:
@chahank, correct, it would have warranted even more input information, namely how to combine impacts and cost functions from different impacts. Calibrating on multiple instances of hazard and exposure will not be a feature of this PR.

climada/util/calibrate/func.py
```diff
@@ -0,0 +1,16 @@
...
+from typing import Sequence, Optional
+
+import pandas as pd
+import numpy as np
+
+from climada.engine import Impact
+
+def rmse(impact: pd.DataFrame, data: pd.DataFrame):
+    return np.sqrt(np.mean(((impact - data) ** 2).to_numpy()))
+
```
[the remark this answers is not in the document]
tovogt: I wouldn't use "+ 1" in such a general setting since we do not know the scaling of the impact variable. In some applications it might be an abstract indicator that has a very small value range. I would rather use something like ` + np.spacing(1)`.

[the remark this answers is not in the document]
chahank: I agree, but would make it a variable, as for certain cases 1 might be a bad choice.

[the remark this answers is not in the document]
tovogt: You can make it a variable, but `np.spacing(1)` should be general enough.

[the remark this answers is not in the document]
chahank: Aaa yes, I misunderstood what spacing does. I think the only 'problem' might arise if there are negative and positive values, but this is probably to much of an edge case.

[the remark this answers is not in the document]
peanutfun: @tovogt Very good idea with `spacing`! I was not even aware of that function. However, I found that `scikit-learn` is a dependency of Climada and they define error measures useful for regression problems. Maybe we just borrow them? Note that they also define the mean squared logarithmic error with "+1": https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-logarithmic-error

@chahank I am unsure about negative values. It is common to compute log(x) - log(y) instead of log(x/y) due to numerical accuracy and the former does not support any negative numbers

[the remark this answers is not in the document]
peanutfun: Any particular feelings towards using the "unrooted" versions of the squared errors? I really have no idea when to use what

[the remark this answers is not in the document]
tovogt: Taking the square root or not means having a "norm" or a "norm squared". And "squared norms" are mathematically much better behaved in optimization contexts than "norms" because they are strictly convex and the derivative at 0 is continuous. So yes, it's probably better to remove the sqrt.

However, I still think that the `+1` term is not save. Maybe it's numerically more robust if you take `log(x) - log(y)` instead of `log(x/y)` because you avoid subtracting very large numbers. But if you really have an application where your values are between 0 and 1, then this function behaves completely different than in cases where your values are orders of magnitude larger. Close to 0, the mapping `x -> log(1 + x)` is almost the same as `x -> x`...

`MSLE(100, 200)` is approximately equal to `MSLE(1000, 2000)`, but MSLE(0.1, 0.2) is completely different:
```python
>>> import numpy as np
>>> msle = lambda x, y: (np.log(1 + x) - np.log(1 + y))**2
>>> msle(100, 200)
0.4735977563158361
>>> msle(1000, 2000)
0.4797606358194477
>>> msle(10000, 20000)
0.48038370689796994
>>> msle(0.1, 0.2)
0.007570979725631454
>>> msle(0.01, 0.02)
9.706774520097977e-05
```

climada/util/calibrate/base.py
```diff
@@ -0,0 +1,399 @@
...
+        An impact object calculated using the optimal :py:attr:`impf_set`
+    """
+
+    input: Input
+    output: Output
+
+    def __post_init__(self):
+        """Compute the impact for the optimal parameters"""
+        self.impf_set = self.input.impact_func_creator(**self.output.params)
+        self.impact = ImpactCalc(
```
peanutfun, line 138: TODO: Calculate impact in separate method (for passing kwargs)

climada/util/calibrate/base.py
```diff
@@ -0,0 +1,399 @@
...
+            exposures=self.input.exposure,
+            impfset=self.impf_set,
+            hazard=self.input.hazard,
+        ).impact(assign_centroids=True, save_mat=True)
+        self._impact_label = f"Impact [{self.input.exposure.value_unit}]"
+
+    def plot_impf_set(self, **plot_kwargs):
```
peanutfun, line 145: TODO: This is trivial. Delete?

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -0,0 +1,259 @@
...
+        if x is not None:
+            params.remove(x)
+            iterable = zip(repeat(x), params)
+        elif y is not None:
+            params.remove(y)
+            iterable = zip(params, repeat(y))
+
+        # Iterate over parameter combinations
+        for p_first, p_second in iterable:
+            plot_single(p_first, p_second)
```
peanutfun, line 259: TODO: These plots are never returned!

climada/util/calibrate/base.py
```diff
@@ -0,0 +1,399 @@
...
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field, InitVar
+from typing import Callable, Mapping, Optional, Tuple, Union, Any, Dict
+from numbers import Number
+
+import pandas as pd
+import numpy as np
+from scipy.optimize import Bounds, LinearConstraint, NonlinearConstraint
+import seaborn as sns
```
peanutfun, line 11: TODO: Fix requirements

doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -0,0 +1,3101 @@
...
+   "metadata": {},
+   "source": [
+    "Each entry in the database refers to an economic impact for a specific country and TC event.\n",
+    "The TC events are identified by the ID assigned from the International Best Track Archive for Climate Stewardship (IBTrACS).\n",
+    "We now want to reshape this data so that impacts are grouped by event and country.\n",
+    "\n",
+    "To achieve this, we iterate over the unique track IDs, select all reported damages associated with this ID, and concatenate the results.\n",
+    "For missing entries, `pandas` will set the value to NaN.\n",
+    "We assume that missing entries means that no damages are reported (this is a strong assumption), and set all NaN values to zero.\n",
+    "Then, we transpose the dataframe so that each row represents and event and each column states the damage for a specific country.\n",
```
leonie-villiger, line 896: very small typo: "each row represents _and_ event" should be "each row represents _an_ event"

climada/util/calibrate/base.py
emanuel-schmid: @peanutfun which seaborn runs in your local environment? (addding seaborn from pypi just like that yields errors)

peanutfun: @emanuel-schmid Currently `seaborn=0.13.0`

climada/util/calibrate/base.py
peanutfun: I decided against that because all methods of this class require the impact data

climada/util/calibrate/base.py
peanutfun: Deleted

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Now they are returned

climada/util/calibrate/base.py
peanutfun: @emanuel-schmid That was wrong, it only works with `seaborn>=0.13`, see the new Jenkins config

climada/util/calibrate/impact_func.py
peanutfun: We decided that using `dataclass` is fine and justified, especially because it reduces the amount of boilerplace code.

climada/util/calibrate/impact_func.py
peanutfun: The attribute is now called `impact_func_creator`

doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -0,0 +1,3031 @@
...
+    "            for track in track_ids\n",
+    "        ],\n",
+    "        axis=1,\n",
+    "    )\n",
+    "    .fillna(0)\n",
+    "    .transpose()\n",
+    "    .set_index(track_ids)\n",
+    ")\n",
+    "data.index.rename(\"ibtracsID\", inplace=True)\n",
+    "data"
```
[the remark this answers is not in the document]
peanutfun: I used the very elegant `pivot_table` solution. Thanks for suggesting it!

climada/util/calibrate/func.py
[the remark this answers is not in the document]
peanutfun: @tovogt Thanks for digging into that! You are correct about the "distortion" effects of the MSLE with values between 0 and 1. However, I would argue that these values rarely occur in Climada impact calculations. Overall, it seems like this is treated as a "property" the user has to be aware of.

Instead of using our own cost functions I now suggest relying on those defined by `scikit.metrics`, see https://scikit-learn.org/stable/modules/model_evaluation.html#regression-metrics. The metrics there are typically squared and the MSLE is also defined with +1. Not supplying our own cost functions should signal sufficiently to the user that they have to decide for a metric themselves, and be more aware of these intricacies.

doc/tutorial/climada_util_calibrate.ipynb
peanutfun: Fixed in 472d0c56fc00891f24255ae58f7387ed2b2589d8

peanutfun:
I received a lot of feedback and tried to include it in this PR as much as possible. I think this is complete, useful, and quite elegant now. I would by happy for another round of reviews!

@chahank I added new methods and attributes to `Input`, which control aligning the data with the calculated impact. Not requiring some form of event-matching should give a lot of freedom to users, but we still have full control over the dataframe alignment in the end (which is strictly required). See https://github.com/CLIMADA-project/climada_python/pull/692#discussion_r1404382609. This is also mentioned in a new section in the tutorial.

@tovogt I decided to ditch all self-defined cost functions and use those from `sklearn.metrics`. They may have their own pitfalls but I would like to shift that responsibility to the user.

@timschmi95 I decided to move your plotting function into a new class `BayesianOptimizerOutputEvaluator`. This removes the clumsy test for `p_space_to_dataframe`. I did it also because overwriting the `output` attribute of the original dataclass base with the same attribute of another type actually works very elegantly, see https://github.com/CLIMADA-project/climada_python/blob/472d0c56fc00891f24255ae58f7387ed2b2589d8/climada/util/calibrate/bayesian_optimizer.py#L296

timschmi95:
@peanutfun Thanks, great work! The new Bayesion-optimizer-specific Outputevaluator which inherits from OutputEvaluator is indeed an elegant solution :) 
The usecase of having a non-bayesian optimizer but still a (user created) parameter space DataFrame with the same format (which is passed to the function as argument `p_space_df`) would now no longer be possible. But I think that's not a common use-case, so it's fine.

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -0,0 +1,448 @@
...
+        Parameters
+        ----------
+        init_points : int, optional
+            Number of initial samples taken from the parameter space. Defaults to 10^N,
+            where N is the number of parameters.
+        n_iter : int, optional
+            Number of iteration steps after initial sampling. Defaults to 10^N, where N
+            is the number of parameters.
+        opt_kwargs
+            Further keyword arguments passed to ``BayesianOptimization.maximize``.
```
tovogt, line 272: The `maximize` method now only accepts the three arguments `init_points`, `n_iter` and `acquisition_function`: https://github.com/bayesian-optimization/BayesianOptimization/pull/406

Therefore, we should probably drop the `opt_kwargs` here.

climada/test/test_util_calibrate.py
```diff
@@ -0,0 +1,244 @@
...
+            data={"a": [3, 1], "b": [0.2, 0.01]}, index=self.events
+        )
+        self.impact_to_dataframe = lambda impact: impact.impact_at_reg(["a", "b"])
+        self.impact_func_creator = lambda slope: ImpactFuncSet(
+            [
+                ImpactFunc(
+                    intensity=np.array([0, 10]),
+                    mdd=np.array([0, 10 * slope]),
+                    paa=np.ones(2),
+                    id=1,
```
tovogt, line 65: Specify a `haz_type` to prevent `ImpactFuncSet` from logging a warning. Can we maybe just give the output of `hazard()` an arbitrary `haz_type` like `"test"` or something similar? Just rename the column in `exposure()` to `"impf_test"` and pass `haz_type="test"` to all calls of `ImpactFunc` in this test file.

climada/test/test_util_calibrate.py
```diff
@@ -0,0 +1,244 @@
...
+            self.impact_func_creator,
+            self.impact_to_dataframe,
+            mean_squared_error,
+        )
+
+    def test_single(self):
+        """Test with single parameter"""
+        self.input.bounds = {"slope": (-1, 3)}
+        optimizer = BayesianOptimizer(self.input)
+        output = optimizer.run(init_points=10, n_iter=20, random_state=1)
```
tovogt, line 174:
```suggestion
        optimizer = BayesianOptimizer(self.input, random_state=1)
        output = optimizer.run(init_points=10, n_iter=20)
```
Passsing gaussian process parameters to `maximize` is no longer supported: https://github.com/bayesian-optimization/BayesianOptimization/pull/406

tovogt (commented):
I don't have time for a thorough review, but I ran the unit tests and quickly scanned through the code and made some remarks.

climada/util/calibrate/base.py
```diff
@@ -0,0 +1,461 @@
...
+
+    Attributes
+    ----------
+    input : Input
+        The input object for the optimization task. See :py:class:`Input`.
+    """
+
+    input: Input
+
+    def _target_func(self, true: pd.DataFrame, predicted: pd.DataFrame) -> Number:
```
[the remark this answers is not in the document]
peanutfun: Agreed. Two reasons why I did that:
1. `sklearn.metrics` calls the parameters `y_true` and `y_pred`.
2. The `sklearn.metrics` [user guide states](https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-logarithmic-error):

   > Note that this metric penalizes an under-predicted estimate greater than an over-predicted estimate.

    I thought this implied that the metric was not symmetrical in its parameters, but the mentioned property is only due to the logarithm.

I would then propose to simply rename "true" to "data", which would be in line with the nomenclature of the module so far ("data" refers to the reference or observed values).

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Not sure. I would like to keep the option to add further arguments. However, I also do not want to suggest users have to change the acquisition function. So I would rather keep it as is, which is also relatively safe in terms of forward- and backward-compatibility.

climada/test/test_util_calibrate.py
peanutfun: Indeed, not sure why I put it there. Setting `random_state` is suggested through the parameter list of the `BayesianOptimizer` initializer.

peanutfun: Fixed in d4d677770815dbc62d93c0be4f03049831be25fa

climada/test/test_util_calibrate.py
peanutfun: Set `haz_type=TEST` in all tests, see d4d677770815dbc62d93c0be4f03049831be25fa

climada/util/calibrate/base.py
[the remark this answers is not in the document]
peanutfun: See d4d677770815dbc62d93c0be4f03049831be25fa

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -0,0 +1,448 @@
...
+            random_state=random_state,
+            allow_duplicate_points=allow_duplicate_points,
+            **bayes_opt_kwds,
+        )
+
+    def _target_func(self, data: pd.DataFrame, predicted: pd.DataFrame) -> Number:
+        """Invert the cost function because BayesianOptimization maximizes the target"""
+        return -self.input.cost_func(data, predicted)
+
+    def run(self, **opt_kwargs) -> BayesianOptimizerOutput:
```
peanutfun, line 256: TODO: Can add explicit parameters here

peanutfun: We can, and it works!

ChrisFairless:
Hey! Some useful context, just in case you've not looked at this already.

I ran the tutorial and plotted the calibrated impact function for the NA1 region against the functions given by the out-of-the-box Emanuel and the regionally-calibrated Eberenz functions:

image

It's quite a change. Any idea why? Looking at the input data loaded at the start of the analysis, it hasn't been changed for at least three years, so I assume it dates back to the Eberenz calibrations. Given the uncertainty shown in those original calibrations (see below and Fig 5 of https://nhess.copernicus.org/articles/21/393/2021/ ) I assume it's entirely due to the cost function?

image

Code to add these curves:
```
from climada.util.calibrate import BayesianOptimizerOutputEvaluator, select_best
from climada.entity import ImpactFuncSet, ImpfTropCyclone, ImpfSetTropCyclone
from matplotlib.lines import Line2D

output_eval = BayesianOptimizerOutputEvaluator(input, bayes_output)

x = np.arange(0, 100, 5)
y1 = 100 * ImpfTropCyclone.from_emanuel_usa().calc_mdr(x)

region_vhalf = ImpfSetTropCyclone.calibrated_regional_vhalf()['NA1']
y2 = 100 * ImpfTropCyclone.from_emanuel_usa(v_thresh=25.7, v_half=region_vhalf, scale=1).calc_mdr(x)

# Plot the impact function variability
plt = output_eval.plot_impf_variability(select_best(p_space_df, 0.03), plot_haz=False)
plt.plot(x, y1, color='darkgreen')
plt.plot(x, y2, color='lightgreen')
plt.legend(
    [
        Line2D([0], [0], color='blue', lw=2),
        Line2D([0], [0], color='lightblue', lw=2),
        Line2D([0], [0], color='darkgreen', lw=2),
        Line2D([0], [0], color='lightgreen', lw=2)],
    [
        'Calibrated here',
        'Uncertainty',
        'emanuel_usa default',
        'Eberenz calibrated NA1'
    ]
)
```

doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -0,0 +1,3920 @@
...
+      "2024-04-29 14:09:55,673 - climada.util.calibrate.bayesian_optimizer - INFO - Optimization iteration: 5\n",
+      "2024-04-29 14:09:58,819 - climada.util.calibrate.bayesian_optimizer - INFO - Optimization iteration: 6\n",
+      "2024-04-29 14:10:02,072 - climada.util.calibrate.bayesian_optimizer - INFO - Optimization iteration: 7\n",
+      "2024-04-29 14:10:06,238 - climada.util.calibrate.bayesian_optimizer - INFO - No improvement. Stop optimization.\n"
+     ]
+    }
+   ],
+   "source": [
+    "from climada.util.calibrate import Input, BayesianOptimizer, BayesianOptimizerController\n",
+    "from sklearn.metrics import mean_squared_error, mean_squared_log_error\n",
```
sunt05, line 2130: `mean_squared_error` is imported but not used.

doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -0,0 +1,3920 @@
...
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Execute the Calibration\n",
+    "\n",
+    "We created a class `BayesianOptimizerController` to control and guide the calibration process.\n",
+    "It is intended to walk through several optimization iterations and stop the process if the buest guess cannot be improved.\n",
```
sunt05, line 2075: please fix typo: `buest` -> ` best`.

naingeet:
@editorialbot commands

PR #702, opened 2023-04-19
Feature/change impact total value

chahank:
Changes proposed in this PR:
- Proposal to update the affected value of the exposures to only those centroids with intensity above 0 (or threshold)

This PR would make the impact computation slower by roughly 10%. Do we want that? 

The code can be made prettier.

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


chahank:
@peanutfun : an alternative could be to have two methods: one that just computes the total sum of the exposures in the vicinity of a hazard centroid (so as it is currently) and which is used in the impact computation. Another method that would only be for the exposures object and which does what the here proposed method does (compute the value of the exposures affected by a given value of intensity).

climada/entity/exposures/base.py
```diff
@@ -1037,11 +1040,15 @@ def affected_total_value(self, hazard):
...
         """
+        assigned_centroids = self.gdf[hazard.centr_exp_col]
         nz_mask = (
             (self.gdf.value.values > 0)
-            & (self.gdf[hazard.centr_exp_col].values >= 0)
+            & (assigned_centroids.values >= 0)
         )
-        return np.sum(self.gdf.value.values[nz_mask])
+        cents = np.unique(assigned_centroids[nz_mask])
+        cent_with_inten_above_thres = (hazard.intensity[:,cents] > threshold_affected).sum(axis=0).nonzero()[1]
```
peanutfun, line 1049: Since it's not important which event surpassed the threshold, it might make sense to compute the row-wise maximum first, and then query `nonzero`:
```suggestion
        cent_with_inten_above_thres = (hazard.intensity[:,cents].max(axis=0) > threshold_affected).nonzero()[1]
```
(I did not benchmark this)

chahank: Good idea! I will benchmark it.

chahank:
> I like the new thing much better, however, the computation overhead could be a problem. Is `Impact.tot_value` really something useful? I would suggest keeping the current version of the method for backward-compatiblity, and deprecating `Impact.tot_value`. Users should instead use the new `affected_total_value` with a threshold of their choice and retrieve it from the exposure when they actually need it.

Good idea. Before doing so, it is probably to be discussed once with users from the insurance sector (which are at the origin of this feature).  Thanks!

climada/entity/exposures/base.py
peanutfun: I meant *column-wise*, of course (i.e., the maximum intensity at every centroid), but `axis=0` should mean exactly that 🙇

chahank: My benchmarking shows a slight improvement, so we keep the new version. Thanks!

chahank:
Ok, my suggestion is the following:

We keep the total value as it is currently implemented in the impact computation. We add a new method which does compute the total people subject to a given threshold. 

Thus, we do not slow down the impact computation and keep backwards compatibility, and give the possibility to users to obtain the more useful value of affected people directly from the hazard and exposures. Potentially we could also simply ditch completely the tot_value from the impact computation in order to simplify the code base.

peanutfun:
@chahank I approve!

> Potentially we could also simply ditch completely the tot_value from the impact computation in order to simplify the code base.

When? Do you want to create an issue or deprecate the value immediately? With the latter approach, I think we are more likely to get some feedback on that decision 😉

chahank:
I would go for the more direct approach -> deprecate.

climada/engine/impact.py
```diff
@@ -246,7 +244,7 @@ def from_eih(cls, exposures, impfset, hazard,
                                  axis=1),
             crs = exposures.crs,
             unit = exposures.value_unit,
-            tot_value = exposures.affected_total_value(hazard),
+            tot_value = exposures._centroids_total_value(hazard),
```
peanutfun, line 247: The linter complains about accessing a private method, and I agree. I think there is no harm in making this method public and adding a hint/warning to its documentation that it is there for legacy purposes and should not be used (by others)

chahank, line 247:
```suggestion
            tot_value = exposures.centroids_total_value(hazard),
```

climada/entity/exposures/base.py
```diff
@@ -1019,7 +1019,7 @@ def concat(exposures_list):
 
         return exp
 
-    def affected_total_value(self, hazard):
+    def _centroids_total_value(self, hazard):
```
chahank, line 1022:
```suggestion
    def centroids_total_value(self, hazard):
```

CHANGELOG.md
```diff
@@ -30,13 +30,16 @@ Removed:
 - Added test for non-default impact function id in the `lines_polys_handler` [#676](https://github.com/CLIMADA-project/climada_python/pull/676)
 - The sigmoid and step impact functions now require the user to define the hazard type. [#675](https://github.com/CLIMADA-project/climada_python/pull/675)
 - Improved error messages produced by `ImpactCalc.impact()` in case hazard type is not found in exposures/impf_set [#691](https://github.com/CLIMADA-project/climada_python/pull/691)
+- `Exposures.affected_total_value` now takes a hazard intensity threshold as argument.  [#702](https://github.com/CLIMADA-project/climada_python/pull/702)
```
chahank, line 33:
```suggestion
- `Exposures.affected_total_value` now takes a hazard intensity threshold as argument.  [#702](https://github.com/CLIMADA-project/climada_python/pull/702)
- Bugfix for the `Hazard.select` method: any iterable are now accepted for `event_id` and `event_names` [#719](https://github.com/CLIMADA-project/climada_python/pull/719) 
```

chahank:
I sneaked-in the missing changelog entry from the pull request #719. Hope this is fine for you @peanutfun

climada/entity/exposures/base.py
```diff
@@ -1043,6 +1048,38 @@ def affected_total_value(self, hazard):
...
+           Hazard affecting Exposures
+        threshold_affected : int or float
+            Hazard intensity threshold above which an exposures is
+            considere affected.
+
+        Returns
+        -------
+        float
+            Sum of value of all exposures points for which
+            a centroids is assigned
```
chahank, line 1069:
```suggestion
            Sum of value of all exposures points for which
            a centroids is assigned and that have at least one
            event intensity above threshold.
```

chahank:
@peanutfun ready to merge?

peanutfun:
> I sneaked-in the missing changelog entry from the pull request https://github.com/CLIMADA-project/climada_python/pull/719. Hope this is fine for you @peanutfun

I personally much prefer using a separate PR for that, but let's not overcomplicate things. You'll have to add this yourself, I cannot commit your suggestion because it is outdated.

> @peanutfun ready to merge?

Yes!

peanutfun (approved):

PR #737, opened 2023-06-16
cleaning up data api file cache

emanuel-schmid:
Changes proposed in this PR:
- introduce a purge method in the api_client that removes files from the api client file cache if they are outdated

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/test/test_api_client.py
+++ b/climada/test/test_api_client.py
@@ -19,8 +19,8 @@ with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
 Test save module.
 """
 from pathlib import Path
+import tempfile
 import unittest
-from shutil import rmtree
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


emanuel-schmid:
@chahank: let me know if this acceptably solves the ever increasing api data heap problem. If so I'll add documentation.

climada/util/api_client.py
```diff
@@ -595,13 +596,13 @@ def _download_file(self, local_path, fileinfo, check=checksize, retries=3):
             downloaded = self._tracked_download(remote_url=fileinfo.url, local_path=local_path)
             if not downloaded.enddownload:
                 raise Download.Failed("Download seems to be in progress, please try again later"
-                                      " or remove cache entry by calling"
-                                      f" `Client.purge_cache(Path('{local_path}'))`!")
+                                      " or remove cache entry from database by calling"
+                                      f" `Client.purge_cache_db(Path('{local_path}'))`!")
```
chahank, line 600: I think this is a bit to difficult to understand. When should a user purge the cache? When should a user wait for the download? Any way to help the user better here?

climada/util/api_client.py
```diff
@@ -1009,3 +1010,55 @@ def into_files_df(dataset_infos):
         """
         return Client.into_datasets_df(dataset_infos) \
             .merge(pd.DataFrame([dsfile for ds in dataset_infos for dsfile in ds.files]))
+
+    def purge_cache(self, target_dir=SYSTEM_DIR, keep_testfiles=True):
+        """Removes downloaded dataset files from the given directory if they have been downloaded
+        with the API client, if they are beneath the given directory and if one of the following
+        is the case:
+        - there status is neither 'active' nor 'test_dataset'
+        = their status is 'test_dataset' and keep_testfiles is set to False
```
chahank, line 1019:
```suggestion
        - their status is 'test_dataset' and keep_testfiles is set to False
```

climada/util/api_client.py
```diff
@@ -1009,3 +1010,55 @@ def into_files_df(dataset_infos):
...
+            files downloaded beneath this directory and empty subdirectories will be removed.
+            default: SYSTEM_DIR
+        keep_testfiles : bool, optional
+            if set to True, files from datasets with status 'test_dataset' will not be removed.
+            default: True
+        """
+
+        # collect urls from datasets that should not be removed
+        test_datasets = self.list_dataset_infos(status='test_dataset') if keep_testfiles else []
+        test_urls = set(filinf.url for dsinf in test_datasets for filinf in dsinf.files)
```
chahank, line 1035: I do not understand the short variables `dsinf` and `filinf`. Are these some standard abbreviations?

climada/util/api_client.py
```diff
@@ -1009,3 +1010,55 @@ def into_files_df(dataset_infos):
...
+                obsolete.delete_instance()
+
+        # clean up: remove all empty directories beneath target_dir
+        def rm_empty_dirs(directory: Path):
+            for subdir in directory.iterdir():
+                if subdir.is_dir():
+                    rm_empty_dirs(subdir)
+            try:
+                directory.rmdir()
+            except OSError:
```
chahank, line 1062: What error is ignored here? I do not understand.

climada/test/test_api_client.py
```diff
@@ -233,6 +233,46 @@ def test_multiplicity_split(self):
...
+
+            client.purge_cache(target_dir=temp_dir)
+            self.assertFalse(  # outdated data set removed
+                Path(temp_dir).joinpath('exposures/litpop/LitPop_150arcsec_ABW/v1').is_dir()
+            )
+            self.assertFalse(  # expired data set removed
+                Path(temp_dir).joinpath('hazard/tropical_cyclone/rename_files2/v1').is_dir()
+            )
+            self.assertEqual(  # test files are still there
+                3,
```
chahank, line 264: Why are there 3 test files?

chahank (changes requested):
That was a very quick fix, thanks! Just a few small comments, but otherwise it looks great to me (when the documentation will be added :)).

climada/test/test_api_client.py
emanuel-schmid: that's just the nature of that test case. datasets can have any number of files.

climada/util/api_client.py
emanuel-schmid: they're called `ds_info` and `file_info` now

climada/util/api_client.py
emanuel-schmid: I've added an inline comment `# raised when directory is not empty`

climada/util/api_client.py
emanuel-schmid: The exception is more verbose now.

climada/test/test_api_client.py
chahank: thanks. I meant, why does this particular test result in 3 files?

climada/util/api_client.py
chahank: Very clear now!

climada/test/test_api_client.py
emanuel-schmid: Beats me. I picked it for being small (in file size) and expired. I suspect it's an experimental dataset that was used to explore the data api itself.

chahank: Should we use a better-known test file then? The test looks rather mysterious like this.

emanuel-schmid: Oh. Wrong. Sorry. The one with 3 files is used in `TestStormEurope.test_icon_read`. Reading icon files takes a directory as input and collects data from there. Having more than one makes complete sense. And the test is fairly known. Apart from that I think it doesn't really matter which dataset we pick as long as size is acceptable and status and version make a difference.

chahank: All right, if it is clear to you, I am fine with it.

PR #743, opened 2023-07-04
Remove all uses of Tag from climada/engine

peanutfun:
Changes proposed in this PR:
- Remove all uses of `Tag` from `climada/engine/`

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -51,7 +51,6 @@ import climada.util.coordinates as u_coord
 import climada.util.dates_times as u_dt
 import climada.util.plot as u_plot
 from climada.util.select import get_attributes_with_matching_dimension
-from climada.util.tag import Tag
 
 LOGGER = logging.getLogger(__name__)
 
@@ -88,9 +87,6 @@ class Impact():
     imp_mat : sparse.csr_matrix
         matrix num_events x num_exp with impacts.
         only filled if save_mat is True in calc()
-    tag : dict
-        dictionary of tags of exposures, impact functions set and
-        hazard: {'exp': Tag(), 'impf_set': Tag(), 'haz': Tag()}
     haz_type : str
         the hazard type of the hazard
     """
@@ -109,7 +105,6 @@ class Impact():
                  aai_agg=0,
                  unit='',
                  imp_mat=None,
-                 tag=None,
                  haz_type=''):
         """
         Init Impact object
@@ -145,15 +140,11 @@ class Impact():
             value unit used (given by exposures unit)
         imp_mat : sparse.csr_matrix, optional
             matrix num_events x num_exp with impacts.
-        tag : dict, optional
-            dictionary of tags of exposures, impact functions set and
-            hazard: {'exp': Tag(), 'impf_set': Tag(), 'haz': Tag()}
         haz_type : str, optional
             the hazard type
         """
 
         self.haz_type = haz_type
-        self.tag = tag or {}
         self.event_id = np.array([], int) if event_id is None else event_id
         self.event_name = [] if event_name is None else event_name
         self.date = np.array([], int) if date is None else date
@@ -260,10 +251,6 @@ class Impact():
             at_event = at_event,
             aai_agg = aai_agg,
             imp_mat = imp_mat if imp_mat is not None else sparse.csr_matrix((0, 0)),
-            tag = {'exp': exposures.tag,
-                   'impf_set': impfset.tag,
-                   'haz': hazard.tag
-                   },
             haz_type = hazard.haz_type,
             )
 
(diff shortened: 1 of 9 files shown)


climada/engine/impact.py
```diff
@@ -1215,20 +1171,6 @@ def from_hdf5(cls, file_path: Union[str, Path]):
             ├─ event_name
             ├─ frequency
             ├─ imp_mat
```
chahank, line 1173: I think the `haz_type` is now missing from the description docstring. Should be in the `.attrs` ?

climada/engine/impact.py
```diff
@@ -1719,7 +1652,7 @@ def concat(cls, imp_list: Iterable, reset_event_ids: bool = False):
           ``frequency``, ``imp_mat``, ``at_event``,
         - sums up the values of attributes ``eai_exp``, ``aai_exp``
         - and takes the following attributes from the first impact object in the passed
-          impact list: ``coord_exp``, ``crs``, ``unit``, ``tot_value``, ``tag``,
+          impact list: ``coord_exp``, ``crs``, ``unit``, ``tot_value``,
           ``frequency_unit``
```
chahank, line 1656:
```suggestion
          impact list: ``coord_exp``, ``crs``, ``unit``, ``tot_value``,
          ``frequency_unit``, ``haz_type``
```

climada/util/yearsets.py
```diff
@@ -82,7 +82,6 @@ def impact_yearset(imp, sampled_years, lam=None, correction_fac=True, seed=None)
 
     #save calculations in yimp
     yimp.event_id = np.arange(1, n_sampled_years+1)
-    yimp.tag['yimp object'] = True
```
chahank, line 85: Does this potentially break a functionality? This is one of the rare instances of a possibly relevant variable to be stored in the tags object.

chahank (changes requested):
Great progress on the quest towards removing the `Tag` from CLIMADA. Thanks!

climada/engine/impact.py
peanutfun: True, looks like we are not writing it anymore. Will fix!

climada/util/yearsets.py
peanutfun: Not sure 😅 All tests pass on my end, so I don't think it does? 🤷

climada/engine/impact.py
peanutfun: It was written, but the test was not adapted properly. Everything works now!

peanutfun:
@chahank There is another issue: `Impact.from_eih` now has an unused argument `impfset`: https://github.com/CLIMADA-project/climada_python/pull/743/files#diff-1c276908f04d1a0ca5b1692295a586ce54b0937673711203582d3d72099ecdd0R210

The `impfset` was only used to populate the tag. I propose to deal with this the following way:
* Remove the `impfset` parameter from the method immediately
* Update all uses of `from_eih` within Climada (there are not many)
* Add a type check for the second argument. If it is an `ImpactFuncSet`, raise an error with useful error message
This way, we are *not* backward compatible but users know how to fix the issue.

chahank:
> @chahank There is another issue: `Impact.from_eih` now has an unused argument `impfset`: https://github.com/CLIMADA-project/climada_python/pull/743/files#diff-1c276908f04d1a0ca5b1692295a586ce54b0937673711203582d3d72099ecdd0R210
> 
> The `impfset` was only used to populate the tag. I propose to deal with this the following way:
> 
>     * Remove the `impfset` parameter from the method immediately
> 
>     * Update all uses of `from_eih` within Climada (there are not many)
> 
>     * Add a type check for the second argument. If it is an `ImpactFuncSet`, raise an error with useful error message
> 
> 
> This way, we are _not_ backward compatible but users know how to fix the issue.

Excellent idea! Although I think we do not need to make the error message. Most users will not directly use this method anyway imho.

chahank:
Oh, and probably you should go through the documentation and check whether the `tag` is explicitly mentioned. If yes, please remove it.

peanutfun:
@chahank I removed the `impfset` parameter from `Impact.from_eih` without adding an type check, as discussed. I removed the mention of Tag from the impact tutorial. It is still mentioned in other tutorials (e.g. exposures and hazard) though, but I would remove these instances once tags are removed from the respective classes.

climada/util/yearsets.py
carmensteinmann: I quickly checked and don't think it would break any functionality.

chahank:
Excellent, looks good to me!

peanutfun:
Thanks for the input, merging now!

PR #756, opened 2023-07-14
Remove Tag from Exposures

emanuel-schmid:
Changes proposed in this PR:
- removes tag attribute from Exposures and subclasses
- introduce new attribute of LitPop: `description`, which is used in `plot_hexbin` and `plot_scatter` to set the title

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


climada/entity/exposures/base.py
```diff
@@ -546,17 +543,19 @@ def plot_scatter(self, mask=None, ignore_zero=False, pop_name=True,
...
 
         Returns
         -------
         cartopy.mpl.geoaxes.GeoAxesSubplot
         """
         crs_epsg, _ = u_plot.get_transformation(self.crs)
-        title = "\n".join(self.tag.description)
-        cbar_label = f'Value ({self.value_unit})'
+        title = (self.description.strip() if isinstance(self.description, str) else ""
+                 ) if title is None else title
```
peanutfun, line 558: `title = something if title is None else title` seems redundant to me. Also: What other type than `str` would `description` support? Should we maybe cast it to `str` in the constructor to avoid testing its type here?
```suggestion
        if title is None:
            title = self.description.strip()
```

climada/entity/exposures/base.py
```diff
@@ -615,8 +621,8 @@ def plot_hexbin(self, mask=None, ignore_zero=False, pop_name=True,
...
```
peanutfun, line 625: Same as above. `title` is already assigned to `title` 😉

climada/engine/forecast.py
```diff
@@ -1079,18 +1079,6 @@ def _plot_warn(
...
-            & isinstance(decision_dict["area_aggregation"], float)
-        ):
-            ValueError(
-                " If decision_level is 'exposure_point',"
-                + "parameters probability_aggregation and "
-                + "area_aggregation of "
-                + "Forecast.plot_warn_map() must both be "
-                + "floats between [0..1]. Which each "
-                + "specify quantiles."
-            )
```
peanutfun, line 1093: How is this related to the Exposures tag?

peanutfun (changes requested):
Two minor issues. The rest looks well! 👌

climada/entity/exposures/base.py
emanuel-schmid: `None` is the default. So if you call it with a `title`-argument - fine, up to you, even if it's an empty string.
If you don't bother - also fine, then the title from the description is plotted.

emanuel-schmid: your suggestion doesn't work if the `description` is `None`. You would need another `if` block. From there on it's a question of taste I guess.

peanutfun: Then let's keep it as simple as can be: Users need to take care of formatting their descriptions. So I suggest we just do:
```suggestion
    if title is None:
        title = self.description
```
This should also work when `description` is `None`, `ax.set_title(None)` should have no (negative) effect.

climada/engine/forecast.py
peanutfun: Forecast will be overhauled anyway. I thinks it's best not to touch that class now. Can you revert the changes, even though the error clearly should be raised? 😅

climada/entity/exposures/base.py
emanuel-schmid: some util.plot function doesn't like `None`. Hence I did `tjitle = self.description or None`.

PR #773, opened 2023-08-18
Make `Impact.impact_at_reg` support all zero impacts

peanutfun:
Changes proposed in this PR:
- Detect if the shape product of the impact matrix is zero instead of the nonzero entries. The latter could also be the case for a correctly-shaped matrix whose values are all zero.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ Removed:
 - Problem with `pyproj.CRS` as `Impact` attribute, [#706](https://github.com/CLIMADA-project/climada_python/issues/706). Now CRS is always stored as `str` in WKT format.
 - Correctly handle assertion errors in `Centroids.values_from_vector_files` and fix the associated test [#768](https://github.com/CLIMADA-project/climada_python/pull/768/)
 - Text in `Forecast` class plots can now be adjusted [#769](https://github.com/CLIMADA-project/climada_python/issues/769)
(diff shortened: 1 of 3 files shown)


timschmi95 (commented):
The code looks good to me. I'm wondering if there is a use case to calculate imp_at_reg if all the values are zero? If not, one could also throw a different error (or warning) saying that all impacts are zero if `self.imp_mat.nnz == 0` .

peanutfun:
When calibrating impact functions (see https://github.com/CLIMADA-project/climada_python/pull/692), one usually aggregates impacts in space, and `impact_at_reg` is the "canonical" way of doing that. When the optimization algorithm explores possible parameter sets, it might happen that the computed impact is zero everywhere. However, the calibration cannot recover from the error thrown in this case and aborts.

One has to distinguish two cases.
* If `save_mat=False`, `impact_at_reg` cannot work, so I think throwing an error is fine. However, there is no way of knowing that for sure from the `Impact` object. By default, the impact matrix should be empty with shape (0, 0).
* If `save_mat=True`, it might very well be that all values are zero and computing the spatial aggregate will also return zero. However, this should not be an error (currently, it is), and I also don't see why this should issue a warning. It's simply the arithmetic result.

peanutfun:
Thanks for the review and discussion, @timschmi95!

PR #786, opened 2023-09-28
Remove python version pin from conda specs

peanutfun:
Changes proposed in this PR:
- Allow users to create environments for Python 3.9, 3.10, and 3.11, reverting part of the changes from https://github.com/CLIMADA-project/climada_python/commit/185ef967cdd6a37510421b7d1a69871ed7e4f20d

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/doc/guide/install.rst
+++ b/doc/guide/install.rst
@@ -147,7 +147,7 @@ For advanced Python users or developers of CLIMADA, we recommed cloning the CLIM
 
    .. code-block:: shell
 
-      conda env create -n climada_env python=3.9
+      conda create -n climada_env python=3.9
 
    .. note::
 
--- a/requirements/env_climada.yml
+++ b/requirements/env_climada.yml
@@ -26,7 +26,7 @@ dependencies:
   - pycountry>=22.3
   - pyepsg>=0.4
   - pytables>=3.7
-  - python=3.9
+  - python>=3.9,<3.12
   - pyxlsb>=1.0
   - rasterio>=1.3
   - requests>=2.31
--- a/script/jenkins/install_env.sh
+++ b/script/jenkins/install_env.sh
@@ -1,7 +1,8 @@
 #!/bin/bash -e
 
 mamba remove --name climada_env --all
-mamba env create -f requirements/env_climada.yml --name climada_env
+mamba create -n climada_env python=3.9
+mamba env update -n climada_env -f requirements/env_climada.yml
 
 source activate climada_env
 python -m pip install -e "./[test]"


emanuel-schmid:
With the current configuration this leads to Jenkins running all tests with Python 3.11 and I think for now it should stay at 3.9 there.
The installation scripts `script/jenkins/install_env.sh` of both petals and core need to be modified. I was thinking of something like

```bash
#!/bin/bash -e

mamba remove --name climada_env --all
mamba env create -f <(sed 's/ - python.*$/ - python=3.9/' requirements/env_climada.yml) --name climada_env

source activate climada_env
python -m pip install -e "./[test]"

make install_test

conda deactivate
```

but it doesn't look so nice, does it?

peanutfun:
Damn, sorry about that! 🙈  In the tests, I was relying on adding specs to the environment file definitions, see https://github.com/mamba-org/setup-micromamba/tree/main#create-environment-from-environment-specs

Apparently, this is an exclusive feature of micromamba. With Conda/Mamba, one can only define environments by either package specs or an environment file, but never both at the same time. So if you want to set a specific Python version, you have to create a clean environment first in which you pin the python version and then update it with the spec file:

```
mamba create -y -n climada_env python=3.9
mamba env update -n climada_env -f requirements/env_climada.yml
```

peanutfun:
Oh no, the solution above still upgrades Python in `mamba env update`. This is super annoying. I'll try to find a solution...

peanutfun:
@emanuel-schmid This works, but it's annoying because it first creates the environment with Python 3.11 and then re-installs all packages for Python 3.9:
```
mamba env create -y -n climada_env -f requirements/env_climada.yml
mamba install -n climada_env python=3.9
```

peanutfun:
Can we maybe just remove the Python version? We have a version requirement in the [setup.py](https://github.com/CLIMADA-project/climada_python/blob/main/setup.py#L60). The only issue is that the error of a wrong version will trigger only after creating an environment

emanuel-schmid:
Removing the Python version seems to work.

In any case we should update the  [Advanced Instrucions](https://github.com/CLIMADA-project/climada_python/blob/main/doc/guide/install.rst#advanced-instructions) as well, otherwise they result in an environment with the Python version from the base environment. We certainly don't want that for Python 3.12 or 3.8.

emanuel-schmid:
should be: "Removing the whole Python entry seems to work"

peanutfun:
Alright. So the resolution is
* Remove `python` from requirements in `env_climada.py`
* Update advanced installation instructions to do the following:
    ```shell
    mamba create -n climada_env python=X.Y  # Python version of choice
    mamba env update -n climada_env -f requirements/env_climada.yml
    mamba activate climada_env
    python -m pip install -e ./  # This will throw an error if the Python version is not supported
    ```
@chahank Is that okay for you?

chahank:
Hihi! I think the `#Python version of choice` line must be clearer. What versions are recommended? What versions are working?

Otherwise good for me.

peanutfun:
Yes, of course, we will list the supported versions (3.9, 3.10, 3.11 at the moment) ✌️

peanutfun:
@emanuel-schmid @chahank See https://climada-python--786.org.readthedocs.build/en/786/guide/install.html#advanced-instructions

chahank:
Thanks @peanutfun . So, we have to remember to ALWAYS update this line with new releases. I think the naming is a bit confusing though with `Supported` and `allowed` .

peanutfun:
I did not want to call the other Python versions "Supported", because that implies that we can support people with any kind of issues occurring with these versions. I think we are not ready for that, yet.

doc/guide/install.rst
```diff
@@ -143,13 +143,31 @@ For advanced Python users or developers of CLIMADA, we recommed cloning the CLIM
       cd climada_python
       git checkout develop
 
-#. Create an Anaconda environment called ``climada_env`` for installing CLIMADA.
-   Use the default environment specs in ``env_climada.yml`` to create it.
+#. Create an Anaconda environment called ``climada_env`` for installing CLIMADA:
+
+   .. code-block:: shell
+
+      conda env create -n climada_env python=3.9
```
emanuel-schmid, line 150:
```suggestion
      conda create -n climada_env python=3.9
```

peanutfun: Very good! Thanks for noticing 🙌

emanuel-schmid:
> So, we have to remember to ALWAYS update this line with new releases.

Er - how is that?

peanutfun:
@chahank Are you referring to Python releases? These occur once a year.

chahank:
I am referring to CLIMADA release. We have to update these installation instructions in case we support new versions of Python and if we do not support any more other versions.

PR #787, opened 2023-09-28
Feature/centroids as gdf

chahank:
Changes proposed in this PR:
- Make all the centroids data to be stored in a geodataframe attribute
- Removes and simplifies a lot of the methods. 

~~This is a DRAFT, and comments are welcome. Not all tests are already running through as many tests are using the centroids class even though they should not. But this will be fixed once we converge to a final result.~~

This pull request is getting close to review-ready, pending some open discussion points (see below).
### PR Author Checklist

### PR Reviewer Checklist


peanutfun:
How do we handle differing CRS from now on? I think the goal should be that centroids can be defined in any CRS. Some points for discussion:

* There are some instances in the new code where the CRS is checked and an error is thrown if it is not a specific one. However, for the function to work, we can simply transform into the required CRS and back (e.g. `set_region_id`).
* The naming of some attributes still relates to a geographic CRS, in particular `lat` and `lon`. If the CRS is projected, `lat` and `lon` (or `gdf.geometry.y` and `gdf.geometry.x`, respectively) are not lat/lon coordinates anymore. However, we could make it that `lat` and `lon` always report the respective lat/lon values, like so:
    ```python
    @property
    def lon(self):
        return self.gdf.geometry.to_crs(DEF_CRS).x
    ```
* When merging centroids, the CRS should be aligned instead of throwing an error:
   ```python
    def append(self, other):
        self.gdf = pd.concat([self.gdf, other.gdf.to_crs(self.gdf.crs)])
   ```

chahank:
I like the idea of having more consistent CRS handling! However, I am not sure whether this should be done under the hood as you suggest. CRS conversion can be computationally very expensive and can lead to inaccuracies. Thus, my approach so far was to keep the handling as is but allow for explicit handling of the CRS (sub-optimal, but at least now consistent and possible to use other CRS). 

I also like the idea of having a special handling of the epsg4326 lat/lon projections as it is quite pervasive in the CLIMADA code, and also widely used. However, I am not yet convinced that one needs to rename the variables, and I am not convinced by having a property that requires large amounts of computations (without clear benefit).

bguillod:
I like very much a refactoring on how centroids are being treated as a geospatial object since this is what they represent! This makes a lot of sense to me on a general level.

My main question would be how to distinguish **raster** versus **points** centroids? So far (i.e., before the proposed changes), the code mostly assumes that centroids are points and they are assigned to exposure based on their proximity. But centroids can alternatively be defined as raster using the `meta` attribute, in which case it is not the distance to the centroid that matters but whether the exposure lies within a raster grid cell (using `climada.util.coordinates.match_grid_points` under the hood within `climada.util.coordinates.match_centroids`, itself called from the exposure method`assign_centroids`). This is cleaner for all cases where the hazard data is raster-based, for instance based on climate models output (and it is also faster to compute than to calculate all the distances), but from my first glance at the changes in the code, you have removed that raster feature (e.g., in `climada.util.coordinates.match_centroids`).

To expand a bit on the topic: To me, strictly speaking I would envision three types of hazard data:

- Raster data, whereby exposures should be assigned to a raster cell based on whether it lies within the cell.
- Point data, whereby exposures are assigned to the nearest centroid point up to a maximum distance (which is hazard data dependent).
- Vector data, whereby there might at some point exist hazard data where each event has an extent defined by a polygon (e.g., each landslide could be a polygon of the affected area).

Admittedly, the last one is somewhat custom, so let's ignore it for now. But I would suggest we do not drop the raster data case, since it is a very common one in the field. Perhaps we should define two types of "centroids", one for points (with a geodataframe as you now have implemented), and one for raster using existing raster tools such as rasterio (or, alternatively, a geodataframe with raster cells defined as polygons; however the latter would imply using shapely's intersection to assign centroids to exposures, which is likely going to be less computationally-efficient - or to rely on some raster metadata as it was in the previous implementation, but defining locations and values as a point geodataframe instead of individual fields).

What do you think? Happy to brainstorm more if you want to pursue the idea. I really believe climada would benefit a lot from dealing cleanly with raster vs point-based hazard data.

peanutfun:
@bguillod, many thanks for bringing this up!

Indeed, we have thought about the issue of raster and pointwise data quite a bit. The main issue is that Climada does not actually use raster data structures. The GeoDataframes in Centroids and Exposures are lists with no inherent relation between the objects they contain. The `meta` attribute was envisaged as a way of storing and communicating a raster/grid relationship, but it is not used consistently throughout Climada.

We therefore decided to remove the `meta` attribute. As you mention, the only instance where grid/raster information might become handy is the assignment of exposure points to centroids. However, the algorithms for calculating distances between points already implement search trees, see https://github.com/CLIMADA-project/climada_python/blob/e41222210bfb99e92829c0a86f1db2f31355abf5/climada/util/coordinates.py#L1206
These trees optimize well for actually gridded data and therefore the cost of dropping the grid information is very low.

Further arguments why we want to drop `meta`:
* Any raster input is already transformed to points
   * NB: The OpenStreetMap module disaggregates polygons and lines to points for creating an exposure, and re-aggregates them after impact calculation
* Plot functions for raster data (like `scatter`) already do not rely on `meta`
* Climada already offers functions for recalculating the grid/raster from points, but these do not access the `meta` attribute, see
   https://github.com/CLIMADA-project/climada_python/blob/e41222210bfb99e92829c0a86f1db2f31355abf5/climada/util/coordinates.py#L2443

An interesting discussion might be the native support for raster data structures like xarray datasets. `Hazard` and `Exposures` could have two subclasses, one implementing a GeoDataFrame, and one implementing an xarray Dataset. However, this would mean a major rewrite of the Climada core engine. We will not be able to tackle that in the near future.

chahank:
Currently, assigning centroids works differently for points than for rasters. 

- Points: assign the nearest point within `Threshold` distance
- Raster: assign raster cell, and all points outside of the raster are NOT assigned

With the proposed changes here, all rasters are now converted to points at initialization. Thus, the assigning of centroids is now only based on the Points logic. This means that some points that previously were outside of the raster and assigned the values -1 (no match), will now be assigned to the closest point within Threshold distance.

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +82,175 @@ class Centroids():
...
-        area_pixel: Optional[np.ndarray] = None,
-        on_land: Optional[np.ndarray] = None,
+        longitude: np.ndarray,
+        latitude: np.ndarray,
+        crs: str = DEF_CRS,
         region_id: Optional[np.ndarray] = None,
-        elevation: Optional[np.ndarray] = None,
-        dist_coast: Optional[np.ndarray] = None
+        on_land: Optional[np.ndarray] = None,
+        **kwargs
```
peanutfun, line 113: I think the initializer should be the most basic form of defining the object. As such, I suggest to have an initializer that simply passes all its arguments directly to the `GeoDataFrame`. We can then define classmethods that initialize it from more "high-level" data, like so:
```python
class Centroids:

    def __init__(self, *args, **kwargs):
        self.gdf = GeoDataFrame(*args, **kwargs)

    @classmethod
    def from_lat_lon(cls, lat, lon, crs, *args, **kwargs):
        return cls(*args, geometry=gpd.points_from_xy(lon, lat, crs=crs), **kwargs)
```

chahank: While I agree it should be basic, I really really do not like when the initializer is not clear. Also, this would require some major rewriting as I understand it. The current version specifies what is necessary (we want to force any user to ALWAYS define the crs for instance), plus two optional ones `region_id` and `on_land`. We could remove these two from the initializer if you prefer. It is however then not clear that these can be set at all, and that they are useful for many other applications. So, a user would have to know somehow that these attributes are to be called this way.

peanutfun: Proposal after personal discussion: No arbitrary GeoDataFrame construction, but user can pass columns and *must* pass a geometry and crs.

Open questions:
* Is that feasible? Centroids are currently constructed from lat/lon coordinates. This changes would require refactoring of a lot of code. (`Centroids()` to `Centroids.from_lat_lon()`).
* Should we enforce the geometry type in the `__init__`? Not sure how costly this is for large frames. Maybe have an option to skip this if we are sure only points are passed?

```python
class Centroids:

    def __init__(self, geometry, crs, check_geometry=True, **kwargs):
        if check_geometry:
            if not (geometry.geom_type == shapely.Point).all():
                raise ValueError

        self.gdf = GeoDataFrame(geometry=geometry, crs=crs, data=kwargs)

    @classmethod
    def from_lat_lon(cls, lat, lon, crs, **kwargs):
        return cls(geometry=gpd.points_from_xy(lon, lat, crs=crs), check_geometry=False, **kwargs)
```

bguillod:
Thanks @chahank and @peanutfun for your answers.

> Currently, assigning centroids works differently for points than for rasters.
> 
> * Points: assign the nearest point within `Threshold` distance
> * Raster: assign raster cell, and all points outside of the raster are NOT assigned
> 
> With the proposed changes here, all rasters are now converted to points at initialization. Thus, the assigning of centroids is now only based on the Points logic. This means that some points that previously were outside of the raster and assigned the values -1 (no match), will now be assigned to the closest point within Threshold distance.

This is exactly the point that I am concerned about. I have no particular concern about performance by using points rather than raster (following @peanutfun comment on on search trees, fair enough), nor do I have concerns with using geopandas (definitely a good idea). But I am concerned that with the proposed changes, the user cannot prevent exposure datapoints that are outside of a raster cell to be assigned to it if it's the closest one and within the maximum allowed distance.

In particular, this also raises another issue when calculating impacts in chunks. If I have an event set with 20 million event and cannot load them all into memory at once, I might want to store the event set spread across several files, and read one file, calculate impact, store the result and move to the next file. Without clearly defined raster:

- Exposure locations that are close to the border of the area of my hazard set (but outside of the hazard spatial domain) will be assigned to centroids that actually are grid cells that do not contain that exposure. This exposure should actually not be assigned any value, now it will be assigned one.
- One exposure location might be assigned to different centroids depending on which centroids the hazard sub-data contain. This also implies that calculating the impact with the same hazard and exposure might lead to different results depending on whether all events are processed at once or whether subset of events (chunks) are processed and then concatenated.

I agree with @peanutfun that the cleanest solution would be to have two sub-classes for raster vs point centroids, but I also see that this is much more work (personally I think this is less needed for exposure, as the user can aggregate exposure to any raster resolution as a pre-processing so that's a user choice). Given this, my suggestion is to opt for one of the two following options:

1. Keep the current, simple raster possibility using the meta attribute, which does not prevent you from moving to a geopandas GeoDataFrame structure.
2. Add the possibility to state that the data are raster at a certain resolution. If so, create polygons from points as the geometries in the geodataframe, and add a field such as `is_raster`. If the latter is `True`, assign exposure to centroids using intersects with exposure; if `False`, use the nearest neighbour as implemented.

I think that throwing away the possibility to ensure raster data are treated appropriately would really be a pity - it would lead to expert climada users avoiding this pitfall by implementing their own raster trick, while the feature is currently there. Perhaps @gabrielaznar has alternative suggestions on possible ways forward (?), as I recall she once mentioned that's also from her perspective a super important but unfortunately thus far incomplete feature in CLIMADA.

chahank:
Dear @bguillod,
these are indeed two interesting edge cases. Fortunately, they can I think easily be solved by using a) padding and b) using assign_centroids only once.

bguillod:
> Dear @bguillod, these are indeed two interesting edge cases. Fortunately, they can I think easily be solved by using a) padding and b) using assign_centroids only once.

Great if there are solutions. Can you be more specific on what you mean by "padding" and where/how to use it?

emanuel-schmid:
We probably best cover these edge cases with according unit tests - right?

chahank:
I am not sure that a unit test would do the right thing here, it would probably have to be an integration test as it is about the impact computation.

chahank:
Side question: can we transform the comments of this pull request into a github conversation? I think the points raised by @bguillod might be useful in future development planning.

peanutfun:
> Exposure locations that are close to the border of the area of my hazard set (but outside of the hazard spatial domain) will be assigned to centroids that actually are grid cells that do not contain that exposure. This exposure should actually not be assigned any value, now it will be assigned one.

@bguillod This is indeed a case where raster information would be helpful, but I have to concur with @chahank that it is a corner case. I would argue that very few users are aware of these intricacies when reading in raster hazard data, and even fewer can make use of it. I also think there are several workarounds to this particular issue, and I would be happy to discuss them in a [Discussion](https://github.com/CLIMADA-project/climada_python/discussions) thread. Could you maybe post your last comment there?

I would also be happy about suggestions on handling raster data differently following up to this. For now, I argue that this "removal" of raster data is actually improving consistency within Climada and streamlines our internal data structures. Climada never used raster data internally, and the `meta` object was only a makeshift tool to transport original raster information. Its use was inconsistent throughout Climada, as I already pointed out. If anything, removing it actually paves the way for a future, consistent implementation of raster data formats.

aleeciu:
My two cents:

- 100 % in favour of using a geodataframe for centroids
- agree with @peanutfun of dropping the meta argument if this not used consistently in CLIMADA
- agree with @bguillod that dropping the `match_grid_points` is a bad idea because in many instances hazard are indeed rasters!

What about keeping `match_grid_points` but forcing the call of `points_to_raster` beforehand, such that `meta` is calculated only when needed and need not be an argument of the `Centroids` object?

peanutfun:
> agree with @bguillod that dropping the match_grid_points function is a bad idea because in many instances hazard are indeed rasters!

@aleeciu Thanks for commenting on this. I stand by my earlier argument that these intricacies of the current grid handling are lost to most Climada users. Using the `match_grid_points` method within `Exposures.assign_centroids` is currently triggered by the existence of `centroids.meta`, see 

https://github.com/CLIMADA-project/climada_python/blob/e41222210bfb99e92829c0a86f1db2f31355abf5/climada/util/coordinates.py#L1127

If we want to remove a permanent `meta`, we will need to make this trigger a parameter to the function, something like `assign_centroids(raster=True)`. If you have a concrete suggestion on how this could look like in the code, please share it. I think this is confusing to users because they themselves have to decide whether they want "legacy raster treatment" or not.

On the other hand, there are several ways of achieving the specific use case of @bguillod with other manners, e.g.
* restricting the exposure to the hazard area before assigning centroids
* manually assigning centroids with the help of [sjoin](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.sjoin.html) or similar methods

aleeciu:
> If we want to remove a permanent `meta`, we will need to make this trigger a parameter to the function, something like `assign_centroids(raster=True)`. If you have a concrete suggestion on how this could look like in the code, please share it. I think this is confusing to users because they themselves have to decide whether they want "legacy raster treatment" or not.

Yes exactly I would pass a kwarg like you suggest, which by default is set to False so nothing would change for those that want to work with points. I could share a code snippet of course, but I don't think this will convince anybody at this point :-)

To me it's just a very critical functionality that would be lost, and I think most users work with centroids because this is how CLIMADA works under the hood, not because it's their conscious choice (at least this is what happens with me), and dropping alternatives instead of strenghening them will create a lock-in where for large-scale applications that require raster data (e.g. handling GCM outputs) they would opt for alternatives.

peanutfun:
@bguillod @aleeciu Since this is obviously very important to you: Could you please provide a dataset (or a stub thereof) and a code example that demonstrates the issue? Then we can see if we would actually need to keep `match_grid_points` to preserve the old functionality. Another option would simply be to store the raster bounds and then use the pointwise-assignment we want to use exclusively, excluding all exposures outside the bounds. I claim, again, that we do not need `meta` for that.

I feel I have to reiterate that with "old functionality" I only mean the restriction of possibly assigned exposure points to the hazard/centroids raster bounds.

emanuel-schmid:
to be discussed:

- have one uniform label/name for lat/latitude and lon/longitude all over climada -> rename arguments in `__init__` to `lat`, `lon
- drop methods `set_on_land` and `set_region_id`, instead make the corresponding arguments accept `Union[str('AUTO'), None, ndarray, list, function]`
- drop the forwarded `**kwargs` arguments in the constructor. `Centroids` should not be a proxy to an arbitrary GeoDataFrame. It should have a clearly defined interface. And the internal GeoDataFrame should be essentially private.

chahank:
Good points!

> to be discussed:
> 
>     * have one uniform label/name for lat/latitude and lon/longitude all over climada -> rename arguments in `__init__` to `lat`, `lon`

Yes, what should we use? In `Exposures` it is `latitude` and `longitude`. Also, the order could be harmonized: lat/lon or x/y (lon/lat)?
> 
>     * drop methods `set_on_land` and `set_region_id`, instead make the corresponding arguments accept `Union[str('AUTO'), None, ndarray, list, function]`

I do not understand this one. This is important functionality to be able to set the on land and the region id properties...
> 
>     * drop the forwarded `**kwargs` arguments in the constructor. `Centroids` should not be a proxy to an arbitrary GeoDataFrame. It should have a clearly defined interface. And the internal GeoDataFrame should be essentially private.

Good idea. This restricts a bit the flexibility of the centroids class, but maybe this is exactly what we need.

emanuel-schmid:
> I do not understand this one. This is important functionality to be able to set the on land and the region id properties...
Sorry that was unclear, I've updated this in the orig. comment.

chahank:
> > I do not understand this one. This is important functionality to be able to set the on land and the region id properties...
> 
> Sorry that was unclear, I've updated this in the orig. comment.

I think this does not really resolve my question. The point is that users in general need the centroid class (or another function) to set these properties. Thus, if they are only an input value, then we need some other mechanism to set them...

climada/hazard/centroids/test/test_centr.py
```diff
@@ -134,20 +135,48 @@ def test_from_csv(self):
...
+            [40.68, -14.56],
+            [36.89, -17.88],
+            [33.58, -16.17],
+            [33.64, -25.04],
+            [36.98, -15.46],
+            [35.34, -23.87],
+            [35.24, -13.3],
+            [40.48, -12.96],
+            [34.73, -19.62],
+            [39.91, -16.23]
```
chahank, line 160: @sarah-hlsn : Why this set of data? Seems hard to read and to check whether it is correct, while also not covering well possible lat/lon values (-180 to 180, -90 to 90).

climada/hazard/centroids/test/test_centr.py
```diff
@@ -134,20 +135,48 @@ def test_from_csv(self):
...
+        n_centroids = 15
         self.assertEqual(centroids.coord.shape[0], n_centroids)
         self.assertEqual(centroids.coord.shape[1], 2)
         self.assertEqual(centroids.coord[0][0], -25.95)
         self.assertEqual(centroids.coord[0][1], 32.57)
-        self.assertEqual(centroids.coord[n_centroids - 1][0], -24.7)
-        self.assertEqual(centroids.coord[n_centroids - 1][1], 33.88)
+        self.assertEqual(centroids.coord[n_centroids - 1][0], -16.23)
+        self.assertEqual(centroids.coord[n_centroids - 1][1], 39.91)
+        self.assertEqual(centroids.crs, 'epsg:4326')
```
chahank: @sarah-hlsn : Better use a short list of clear values, and test for all of them. This random set of points is not very helpful. Also, no need to test for the lenght if you test for all values. Just test that the arrays are equal (`np.testing.assert_array_equal`).

climada/hazard/centroids/test/test_centr.py
```diff
@@ -134,20 +135,48 @@ def test_from_csv(self):
         self.assertEqual(centroids.coord[0][1], 32.57)
         self.assertEqual(centroids.coord[n_centroids - 1][0], -16.23)
         self.assertEqual(centroids.coord[n_centroids - 1][1], 39.91)
+        self.assertEqual(centroids.crs, 'epsg:4326')
 
         Path(tmpfile).unlink()
     
     def test_from_excel(self):
```
chahank, line 141: @sarah-hlsn : there should also be a test for when the data is written and read with climada. So, please write another test which does not use `pd.write_excel`, but `centroids.write_excel`. Then the `from_excel` should work with only using the filename.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -96,7 +98,46 @@ def test_to_crs(self):
 class TestCentroidsReader(unittest.TestCase):
     """Test read functions Centroids"""
 
-    def test_centroid_pass(self):
+    def test_from_csv(self):
```
chahank, line 101: @sarah-hlsn : idem comments as for the `from_excel` tests.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -96,7 +98,46 @@ def test_to_crs(self):
...
+        """Read a centroid csv file correctly"""
+        tmpfile = Path('test_write_csv.csv')
+        test_crs = CRS.from_user_input(4326).to_wkt()
```
chahank, line 104: @sarah-hlsn : for the default, please use the `DEF_CRS` variable. Also, add one test for a non-default crs.

climada/hazard/centroids/centr.py
```diff
@@ -44,20 +44,6 @@
...
-        'latitude': 'lat',
-        'longitude': 'lon',
-        'on_land': 'on_land',
-        'dist_coast': 'dist_coast',
-        'elevation': 'elevation',
-       ' area_pixel': 'area_pixel'
-    }
-}
-"""Excel variable names"""
-
```
chahank: @sarah-hlsn : this breaks the test of `storm_europe`.

climada/hazard/centroids/test/test_centr.py
sarah-hlsn: I used a subset of the centroids data from `hazard_template.xlsx`, but I can change this accordingly

climada/hazard/centroids/centr.py
```diff
@@ -709,17 +696,15 @@ def from_csv(cls, file_path):
...
         Parameters
         ----------
         file_name : str
             absolute or relative file path
-        crs : dict() or rasterio.crs.CRS, optional
-            CRS. Default: DEF_CRS
-        var_names : dict, default
-            name of the variables. Default: DEF_VAR_EXCEL
+        sheet_name : str
+            name of sheet in excel file containing centroid information
```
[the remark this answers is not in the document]
sarah-hlsn: which write_excel method exactly are you referring to? I left this optional because even some of the tests we have (e.g. for storm europe) don't stick to the default of having a "centroids" sheet, so I wasn't sure if there may be applications or users where another sheet name is used

climada/hazard/centroids/test/test_centr.py
sarah-hlsn: I may be mistaken but I don't think there is a centroids.write_excel method?

climada/hazard/centroids/centr.py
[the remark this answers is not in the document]
chahank: The important is that you define a clear default option. A user should not have to open the file in excel (or pandas) before being able to open it with CLIMADA.

ooo and I see there is no `write_excel` and `write_csv` methods now for centroids. I think these are then missing.. We should support read/write in general I would say.

climada/hazard/centroids/test/test_centr.py
chahank: Yes, that is why they are not well suited for a unit test. Unit tests should be obviously right or wrong, and test reasonable ranges for the values. 
A file instead belongs more to the integration tests where one checks that the methods also work with "real" data. However, when a test fails, one can then not easily know whether new  code is correct or wrong, only that it potentially produces a different result as the old code. Therefore the unit tests are needed that allow to really decide which code is correct.

climada/hazard/centroids/test/test_centr.py
chahank: Correct. This is missing :). I thought that these were already provided by you.

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-        on_land: Optional[np.ndarray] = None,
-        region_id: Optional[np.ndarray] = None,
-        elevation: Optional[np.ndarray] = None,
-        dist_coast: Optional[np.ndarray] = None
+        *,
+        lat: Union[np.ndarray, list],
+        lon: Union[np.ndarray, list],
+        crs: str = DEF_CRS,
+        region_id: Union[str('country'), None, np.ndarray, list] = None,
+        on_land: Union[str('natural_earth'), None, np.ndarray, list] = None,
```
peanutfun, line 75: Use `typing.Literal` for specific values. This requires `from typing import Literal`:
```suggestion
        region_id: Union[Literal["country"], None, np.ndarray, list[float]] = None,
        on_land: Union[Literal["natural_earth"], None, np.ndarray, list[float]] = None,
```

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-        geometry: Optional[gpd.GeoSeries] = None,
-        meta: Optional[Dict[Any, Any]] = None,
-        area_pixel: Optional[np.ndarray] = None,
-        on_land: Optional[np.ndarray] = None,
-        region_id: Optional[np.ndarray] = None,
-        elevation: Optional[np.ndarray] = None,
-        dist_coast: Optional[np.ndarray] = None
+        *,
+        lat: Union[np.ndarray, list],
+        lon: Union[np.ndarray, list],
```
peanutfun, line 72: List of what? Maybe add type hint here (float)
```suggestion
        lat: Union[np.ndarray, list[float]],
        lon: Union[np.ndarray, list[float]],
```

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
+    @property
+    def geometry(self):
+        """ Return the geometry """
+        return self.gdf['geometry']
+
+    @property
+    def on_land(self):
+        """ Get the on_land property """
+        if self.gdf.on_land.isna().all():
+            return None
```
peanutfun, line 126: I think checking all values for NaN can be very costly. We should instead check if the `on_land` column exists at all, and return `None` if it does not.

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
+        return self.gdf['on_land'].values
+
+    @property
+    def region_id(self):
+        """ Get the assigned region_id """
+        if self.gdf.region_id.isna().all():
+            return None
```
peanutfun, line 133: Same as for `on_land`

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
+        return self.gdf['region_id'].values
+
+    @property
+    def crs(self):
+        """ Get the crs"""
+        return self.gdf.crs
+
+    @property
+    def size(self):
+        """Get size (number of lat/lon paris)"""
```
peanutfun, line 143: Typo
```suggestion
        """Get size (number of lat/lon pairs)"""
```

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-        ----------
-        land : bool, optional
-            If True, restrict to grid points on land. Default: False.
-        res_as : int, optional
-            Base grid resolution in arc-seconds (one of 150, 360). Default: 360.
-        base_file : str, optional
-            If set, read this file instead of one provided with climada.
+    def to_default_crs(self):
+        """Project the current centroids to the default CRS (epsg4326)
+        Modifies the object in place.
```
peanutfun, line 177: Remove this method in favor of setting the default CRS as default parameter of `to_crs`?

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-        The geometry, lat, and lon attributes are set from the GeoDataFrame.geometry attribute,
-        while the columns are copied as attributes to the Centroids object in the form of
-        numpy.ndarrays using pandas.Series.to_numpy. The Series dtype will thus be respected.
-
-        Columns named lat or lon are ignored, as they would overwrite the coordinates extracted
-        from the point features. If the geometry attribute bears an alias, it can be dropped by
-        setting the geometry_alias parameter.
+    def to_crs(self, crs):
+        """ Project the current centroids to the default CRS (epsg4326)
+        Modifies the object in place.
```
peanutfun, line 183: No, it returns a new object

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-        >>> gdf.region_id = gdf.region_id.astype(int)  # type coercion
-        >>> centroids = Centroids.from_geodataframe(gdf)
+        Centroids
+            Centroids in the new crs
+        """
+        return Centroids.from_geodataframe(self.gdf.to_crs(crs))
+
+    @classmethod
+    def from_geodataframe(cls, gdf):
+        """Initialize centroids from a geodataframe
```
peanutfun, line 199: We decided that the class and not the user should handle the layout of the underlying geodataframe. So I think that this classmethod should not just take an arbitrary geodataframe and create a new object from it.

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
+            return cls.from_geodataframe(gdf)
+
+        if 'latitude' in exposures.gdf.columns and 'longitude' in exposures.gdf.columns:
+            gdf = exposures.gdf[col_names]
+            return cls(
+                lat = exposures.gdf['latitude'],
+                lon = exposures.gdf['longitude'],
+                crs = exposures.crs,
+                **dict(gdf.items())
+            )
```
peanutfun, line 263: No fallback `return` (this is the source of the `inconsistent-return-statements` linter issue)

climada/hazard/centroids/centr.py
```diff
@@ -1244,152 +813,103 @@ def _from_hdf5(cls, data):
...
+        -------
+        geo : gpd.GeoSeries
+        """
+        if u_coord.equal_crs(self.gdf.crs, u_coord.NE_CRS):
+            return self.gdf.geometry
+        return self.gdf.geometry.to_crs(u_coord.NE_CRS)
 
-    def set_geometry_points(self, scheduler=None):
-        """Set `geometry` attribute with Points from `lat`/`lon` attributes.
+    def _set_region_id(self, level='country', overwrite=False):
```
peanutfun, line 846: Why is this method private? Shouldn't users should be able to set the region ID after initialization?

climada/hazard/centroids/centr.py
```diff
@@ -1244,152 +813,103 @@ def _from_hdf5(cls, data):
...
-                                 .map_partitions(apply_point, meta=Point)
-                                 .compute(scheduler=scheduler))
-
-    def _ne_crs_geom(self, scheduler=None):
-        """Return `geometry` attribute in the CRS of Natural Earth.
+                raise NotImplementedError(
+                    'The region id can only be assigned for countries so far'
+                    )
+
+    def _set_on_land(self, source='natural_earth', overwrite=False):
```
peanutfun, line 870: Why is this private?

climada/hazard/centroids/centr.py
```diff
@@ -79,280 +52,215 @@ class Centroids():
...
-                raise ValueError('Affine transformations with positive y-orientation '
-                                 'or negative x-orientation are not supported.')
-
-    def equal(self, centr):
+        attr_dict = {
+            'geometry': gpd.points_from_xy(lon, lat, crs=crs),
+            'region_id': region_id,
+            'on_land': on_land
+        }
+        self.gdf = gpd.GeoDataFrame(data=attr_dict)
```
peanutfun, line 98: Maybe make this attribute private with `self._gdf`? This way we show users that we don't want them to modify it themselves.

peanutfun, line 98: This variable serves no purpose. Simply initialize the GeoDataFrame with the dict directly!
```suggestion
        self.gdf = gpd.GeoDataFrame(
            data={
                'geometry': gpd.points_from_xy(lon, lat, crs=crs),
                'region_id': region_id,
                'on_land': on_land,
            }
        )
```

peanutfun (changes requested):
There is quite a number of linter warnings. I think nearly all of them can and should still be fixed within this PR.

Not sure how far you got along with what we last discussed on this topic, but I think there are still some inconsistencies regarding the handling of the internal geodataframe. It should be clear that users should not modify it, only the class does when the users call specific methods.

climada/hazard/centroids/centr.py
chahank: I do not like typing much, so I am happy if someone makes it consistent.

climada/hazard/centroids/centr.py
chahank: I disagree. I think it makes more sense to have a consistent object with the same variables always defined in the geodataframe. The operation is not particularly costly. But if you have a better suggestion for handling this shoot.

climada/hazard/centroids/centr.py
chahank: idem as `on_land`

climada/hazard/centroids/centr.py
chahank: Currently `to_crs` is thought as a user method and returns a new centroids object. On the other hand, `to_default_crs` is mostly there to handle internal methods that require the default crs. Could also be merged.

climada/hazard/centroids/centr.py
chahank: Yes, the docstrings need to be adjusted (as per our offline discussions), thanks!
@idetring

climada/hazard/centroids/centr.py
chahank: No, we decided that data should NOT be modified in place after initialization. It is done in the init if desired by the user.

climada/hazard/centroids/centr.py
chahank: idem as above.

climada/hazard/centroids/centr.py
chahank: I am not sure this is necessary.

chahank: True, but really not particularly important either.

chahank:
Thanks @peanutfun for the review! From what I gather from your review is that there are mostly minor (linting and such) things to clarify.

climada/hazard/centroids/test/test_centr.py
sarah-hlsn: @chahank Following the reasoning of the back and forth compatibility between the read and write methods, I am thinking the output files of the write methods should probably not contain the geometry column of the centroids gdf, but only lat, lon and crs columns (plus potentially on_land, region_id). Correct?

sarah-hlsn: Or should we just keep the geometry columnn anyways since it won't be imported when reading the file?

chahank: hmmm what would be the use to save the geometry column? Can this be read by any means? If there is no use-case, I would not save it.

climada/hazard/centroids/centr.py
peanutfun: If the gdf is (intended to be) never modified once the class is built, [as you state](https://github.com/CLIMADA-project/climada_python/pull/787#discussion_r1440297148), then we can simply set our own logic. I think it's very wasteful to *always* check *all* values of `on_land` if this method is called, and it seems like a clear antipattern to me. Simply not setting the `on_land` column in the init should be safe, so I stand by my initial proposal.

climada/hazard/centroids/centr.py
peanutfun: Alright, then new proposal: Make this method a free function, which takes a geometry series as (additional) argument and returns a series of corresponding region IDs. The constructor of Centroids then may call this function (depending on the init arguments), and users are free to do so, too. Same for `_set_on_land`.

I see no reason why this function should be private to Centroids.

climada/hazard/centroids/centr.py
peanutfun: If you want the gdf to be never modified by a user, it should be marked as private. This will also help with tracking bugs, as the linter will raise issues when a private attribute is accessed from outside the class.

climada/hazard/centroids/centr.py
```diff
@@ -1244,152 +813,103 @@ def _from_hdf5(cls, data):
...
-            if left > right:
-                left, right = right, left
-            top = self.meta['transform'].yoff
-            bottom = top + self.meta['transform'][4] * self.meta['height']
-            if bottom > top:
-                bottom, top = top, bottom
-            return left, bottom, right, top
-        return self.lon.min(), self.lat.min(), self.lon.max(), self.lat.max()
+    def _ne_crs_geom(self):
+        """Return `geometry` attribute in the CRS of Natural Earth.
```
peanutfun, line 873: Is it necessary to have this method? Why not simply call `Centroids.to_crs`?

chahank: For legacy it is simpler yes.

climada/hazard/centroids/centr.py
chahank: I think it is still useful to have a coherent dataframe, in particular it makes saving and loading much simpler.

climada/hazard/centroids/centr.py
chahank: The question is: are we really sure that we do not want a user at all to be able to add columns to the geodataframe for convenience?

chahank: SO, I have tried it, but this creates several legacy issues (that can be fixed with enough effort). I am not yet convinced that making the gdf attribute fully private is the right way to go. Thus, before doing this, we need to be very sure that this is the best imho. At the moment I don't see it. The disadvantages (legacy, removal of some flexibility, extra work) seem not to be worth it for me.

climada/hazard/centroids/centr.py
chahank: Done in [1f6ec54](https://github.com/CLIMADA-project/climada_python/pull/787/commits/1f6ec54f6aa3891b67abc105d44c57ffd526b064)

climada/hazard/centroids/centr.py
chahank: Done in [e5dffd9](https://github.com/CLIMADA-project/climada_python/pull/787/commits/e5dffd97a6909e51173d3fceb20863bf7d740c4a)

climada/hazard/centroids/centr.py
chahank: Done in [f59278c](https://github.com/CLIMADA-project/climada_python/pull/787/commits/f59278cf542473ccbb398ff1168d1068dd0447fd)

climada/hazard/centroids/centr.py
chahank: This is already what the method does `u_coord.get_country_code` does (not on geometries, but on lat/lon coordinates). The `_set_region_id` is here just a convenience method to make the calls in the `_init_` simpler. It is private because it is intended to be called only privately.

climada/hazard/centroids/centr.py
```diff
@@ -694,14 +693,31 @@ def from_csv(cls, file_path):
...
+        file_path : str
+            absolute or relative file path and name to write to
+        """
+        LOGGER.info('Writing %s', file_path)
+        df = pd.DataFrame(self.gdf)
+        df['lon'] = self.gdf['geometry'].x
+        df['lat'] = self.gdf['geometry'].y
+        df = df.drop(['geometry'], axis=1)
+        crs = CRS.from_user_input(self.crs).to_wkt()
+        df['crs'] = crs
```
chahank, line 736: Since you use this code bit at least twice, you could consider making it a separate method/function that is then called. This avoids the duplication of code.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -130,11 +130,33 @@ def test_from_csv(self):
...
+        Path(tmpfile).unlink()  
+    
+    def test_write_read_csv(self):
+        """Write and read a Centroids CSV file correctly."""
+        # Create Centroids with latitude and longitude arrays
+        tmpfile = Path('test_write_csv.csv')
+        lat = np.array([10.0, 20.0, 30.0])
+        lon = np.array([-10.0, -20.0, -30.0])
+        centroids_out = Centroids(lat=lat, lon=lon)
+
```
chahank, line 148: @sarah-hlsn : You should also test the variables `on_land` and `region_id`.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -130,11 +130,33 @@ def test_from_csv(self):
...
+        # Write CSV file from Centroids using write_csv
+        centroids_out.write_csv(tmpfile)
+
+        # Read CSV file using read_csv
+        centroids_in = Centroids.from_csv(tmpfile)
+
+        # Test attributes
+        np.testing.assert_array_equal(centroids_in.lat, centroids_out.lat)
+        np.testing.assert_array_equal(centroids_in.lon, centroids_out.lon)
+        self.assertEqual(centroids_in.crs, centroids_out.crs)
```
chahank, line 158: This is a bit dangerous, as the `crs` can be written in different manners, and the strings might not be identical before/after loading. Better to use the `util.coordinates.equal_crs` method.

sarah-hlsn: I can of course change it but this is checking the crs of the centroids gdf, not the strings in the file, so it shouldn't be an issue no?

climada/hazard/centroids/centr.py
sarah-hlsn: And define that method/function in the test file? Because I don't think there is a general use case for it. Also, from an external point of view, isn't it easier for someone to understand what the test does if they see these lines instead of having to search for the function?

climada/hazard/centroids/test/test_centr.py
chahank: In theory, I agree, but I remember having some trouble with the equality of `crs`. So I would go for the more cautious approach here.

climada/hazard/centroids/centr.py
chahank: In general, one wants to avoid duplicating code. In general, if the function has a good name, there is no need to search for the function. One understands quicker and easier what is being done with a good function name. Also, think for instance that in the future one wants to write a `to_json` method. If the suggested method exists, this becomes easy. If not, then one has to again copy a lot of code.

What do you mean by 'And define that method/function in the test file? '? Do you mean writing a unit test for it?

sarah-hlsn: Okay I understand! What I meant is where should I define the function? E.g. in the first test I use it or just on top of the testing file?

chahank: Not in the test file, but in the `centr.py` file. This is where you use the code bits right?

sarah-hlsn: Ah yes sorry I misread and thought this was referring to some code I wrote in the tests, so that's where my confusion came from

climada/hazard/centroids/centr.py
peanutfun: I also think we are mixing two things here. We should first aim for a consistent use of the GeoDataFrame while maintaining the original interface and functionality as far as possible. Once this is done, we can talk about a new, more handy interface of the Centroids class.

chahank: That makes sense. But so, should we first make it private, or first leave it as is and consider this later? 

Note however that we won't have much capacity after this work to further improve the Centroid class since there are many other important things to do, so we should still have a very functional version.

chahank:
Ok, I implemented all suggested changes except the `.gdf` to `._gdf`. Now all tests are passing (test still fails because of some unrelated module), including for the `from_excel` and `from_csv` methods. 

I would now stop making changes unless they are very very critical. 

What needs to be done is
- to complete/amend the docstrings where needed: @idetring
- some more tests could be written for certain methods 
- clean-up where needed.

peanutfun:
> some more tests could be written for certain methods

@chahank Can you list which ones?

climada/hazard/centroids/centr.py
idetring: `centroids.to_default_crs(inplace=False)` currently results in a dead end - neither returning centroids nor setting the projection inplace.

chahank:
> > some more tests could be written for certain methods
> 
> @chahank Can you list which ones?

Mostly some conditionals are not tested and some methods have no direct unit tests:

- from_exposures
- _meta_to_lat_lon
- _centroids_to_df
- _from_dataframe

peanutfun:
From personal discussion (@emanuel-schmid): Check for side-effects! `from_geodataframe` should create a copy. Make sure that the new geodataframe is not affected by changes to the old.

climada/hazard/centroids/centr.py
```diff
@@ -820,386 +638,274 @@ def get_closest_point(self, x_lon, y_lat, scheduler=None):
...
-            self.set_geometry_points(scheduler)
-            close_idx = self.geometry.distance(Point(x_lon, y_lat)).values.argmin()
+        close_idx = self.geometry.distance(Point(x_lon, y_lat)).values.argmin()
         return self.lon[close_idx], self.lat[close_idx], close_idx
 
-    def set_region_id(self, scheduler=None):
-        """Set region_id as country ISO numeric code attribute for every pixel or point.
+    # NOT REALLY AN ELEVATION FUNCTION, JUST READ RASTER
+    def get_elevation(self, topo_path):
+        """Return elevation attribute for every pixel or point in meters.
```
tovogt, line 653: Maybe we can just remove it? It doesn't seem to be used anywhere, and it's really extremely simplistic, just calling `u_coord.read_raster_sample` with trivial arguments.

climada/hazard/centroids/centr.py
```diff
@@ -820,386 +638,274 @@ def get_closest_point(self, x_lon, y_lat, scheduler=None):
...
-        ne_geom = self._ne_crs_geom(scheduler)
-        LOGGER.debug('Setting region_id %s points.', str(self.lat.size))
-        self.region_id = u_coord.get_country_code(
-            ne_geom.geometry[:].y.values, ne_geom.geometry[:].x.values)
+        return u_coord.read_raster_sample(topo_path, self.lat, self.lon)
 
-    def set_area_pixel(self, min_resol=1.0e-8, scheduler=None):
-        """Set `area_pixel` attribute for every pixel or point (area in m*m).
+    def get_dist_coast(self, signed=False, precomputed=False):
+        """Get dist_coast attribute for every pixel or point in meters.
```
tovogt, line 668: We have to be careful when merging this because https://github.com/CLIMADA-project/climada_python/pull/840 is also making changes here.

tovogt (commented):
Thanks for all the work!

I pushed some changes regarding code linting and docstrings. See my comments in the code below for more thoughts.

emanuel-schmid (changes requested):
Looks already very good 😃. Many thanks!

My major remaining concerns are these:

- The integration test fails due to `rasterio.features.rasterize` behaving unexpectedly in `Hazard.write_raster`.
- The use of resolution seems inconsistent - sometimes it's a x/y pair, sometimes a single value. I'm not sure whether we always handle it correctly. Ideally it would always be a pair - but perhaps that's not so easily implemented.
- The removed methods cause climada_petals 4.0 to fail on compilation. I'd somehow prefer it to fail only when the methods are called and with an adequate error message.

climada/hazard/base.py
```diff
@@ -1810,35 +1698,68 @@ def size(self):
...
+                            (geom, value)
+                            for geom, value
+                            in zip(geometry, var_to_write[i_ev].toarray().flatten())
+                        ),
+                        out_shape=(meta['height'], meta['width']),
+                        transform=meta['transform'],
+                        fill=0,
+                        all_touched=True,
+                        dtype=meta['dtype'],
+                    )
```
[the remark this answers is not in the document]
tovogt: The Polygons are exactly at the pixel boundaries, and with `all_touched=True`, it's a question of floating point precision whether or not a value will be set.

However, I can't reproduce this with rasterio=1.3.6 and gdal=3.6.2. Which version of these packages are you using?

climada/util/test/test_coordinates.py
```diff
@@ -539,30 +539,33 @@ def test_match_centroids(self):
...
+        )
+        assigned = u_coord.match_centroids(gdf, centroids)
 
         expected_result = [
             # constant y-value, varying x-value
-            -1, 0, 0, 0, 0, 1,
+            0, 0, 0, 0, 0, 1,
             # constant x-value, varying y-value
-            -1, 0, 0, 20,
+            0, 0, 0, 20,
```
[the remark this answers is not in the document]
tovogt: The assignment logic of `match_centroids` used to be different for vector and raster centroids regarding "outside" points. With raster centroids there was a rectangular boundary to check if a point is within or outside of the raster. In this example, the point happened to be just outside of that rectangular boundary. Now that we only have vector centroids, the rectangular boundary has been replaced by a particular maximum distance (by default 100 km, or appx. 1 degree) to the closest point in `Centroids`. The resulting boundary is not rectangular, but it's the result of taking the union of circles with radius of approximately 1 degree. In this example, the point was outside of that rectangular raster boundary, but it's still closer than 1 degree to the next point in `Centroids`.

climada/hazard/centroids/test/test_vec_ras.py
```diff
@@ -472,69 +300,81 @@ def test_get_closest_point(self):
...
         """Test equal"""
         centr_ras = Centroids.from_raster_file(HAZ_DEMO_FL, window=Window(0, 0, 50, 60))
         centr_bis = Centroids.from_raster_file(HAZ_DEMO_FL, window=Window(51, 61, 10, 10))
-        self.assertFalse(centr_ras.equal(centr_bis))
-        self.assertFalse(centr_bis.equal(centr_ras))
-        self.assertTrue(centr_ras.equal(centr_ras))
-        self.assertTrue(centr_bis.equal(centr_bis))
+        self.assertFalse(centr_ras == centr_bis)
+        self.assertTrue(centr_ras == centr_ras)
+        self.assertTrue(centr_bis == centr_bis)
```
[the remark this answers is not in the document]
tovogt: Are you absolutely sure that this is checking the same thing, namely the `__eq__` dunder method of `Centroids`?

climada/hazard/centroids/centr.py
```diff
@@ -22,1184 +22,897 @@
...
 
-    def set_area_approx(self, min_resol=1.0e-8):
-        """Set `area_pixel` attribute for every pixel or point (approximate area in m*m).
+    def get_meta(self, resolution=None):
+        """Returns a meta raster based on the centroids bounds.
 
-        Values are differentiated per latitude. Faster than `set_area_pixel`.
+        Note that this function is not perfectly inverse with `from_meta` since `get_meta` enforces
+        a grid with equal resolution in x- and y-direction with coordinates increasing in
+        x-direction and decreasing in y-direction.
```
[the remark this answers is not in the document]
tovogt: Good question, I just described the current implementation. But I would be happy to change this.

climada/hazard/centroids/centr.py
```diff
@@ -22,1184 +22,897 @@
...
         Parameters
         ----------
-        min_resol : float, optional
-            if centroids are points, use this minimum resolution in lat and lon. Default: 1.0e-8
-        scheduler : str
-            used for dask map_partitions. “threads”, “synchronous” or “processes”
+        signed : bool, optional
+            If True, use signed distances (positive off shore and negative on land). Default: False.
+        precomputed : bool, optional
+            If True, use precomputed distances (from NASA). Works only for crs=epsg:4326
```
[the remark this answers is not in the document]
tovogt: Well spotted. I think it's just nonsense, I removed it.

climada/hazard/base.py
```diff
@@ -1810,35 +1698,68 @@ def size(self):
...
+                f"The variable {variable} is not valid. Please use 'intensity' or 'fraction'."
+            )
+
+        meta = self.centroids.get_meta(resolution=output_resolution)
+        meta.update(driver='GTiff', dtype=rasterio.float32, count=self.size)
+        res = meta["transform"][0]  # resolution from lon coordinates
+
+        if meta['height'] * meta['width'] == self.centroids.shape[0]:
+            # centroids already in raster format
+            u_coord.write_raster(file_name, var_to_write.toarray(), meta)
```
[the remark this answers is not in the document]
tovogt: Changed to `centroids.size`.

climada/util/test/test_coordinates.py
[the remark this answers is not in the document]
emanuel-schmid: thanks for the explanation!

climada/hazard/centroids/test/test_vec_ras.py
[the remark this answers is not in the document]
emanuel-schmid: Yes, I am. This is the code from `unittest.case` (with comments from me):

```python
    def _getAssertEqualityFunc(self, first, second):
        """Get a detailed comparison function for the types of the two args.

        Returns: A callable accepting (first, second, msg=None) that will
        raise a failure exception if first != second with a useful human
        readable error message for those types.
        """
        if type(first) is type(second):   # True, 1)
            asserter = self._type_equality_funcs.get(type(first))
            if asserter is not None:  # False, 2)
                if isinstance(asserter, str):
                    asserter = getattr(self, asserter)
                return asserter

        return self._baseAssertEqual

    def _baseAssertEqual(self, first, second, msg=None):
        """The default assertEqual implementation, not type specific."""
        if not first == second:
            standardMsg = '%s != %s' % _common_shorten_repr(first, second)
            msg = self._formatMessage(msg, standardMsg)
            raise self.failureException(msg)

    def assertEqual(self, first, second, msg=None):
        """Fail if the two objects are unequal as determined by the '=='
           operator.
        """
        assertion_func = self._getAssertEqualityFunc(first, second)
        assertion_func(first, second, msg=msg)
```

1) in our case both are `Centroids`
2) in our case, `Centroids` has never been registered with `TestCase.addTypeEqualityFunc()`, so we don't get a special Error message in case the test fails. See [https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual]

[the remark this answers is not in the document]
tovogt: I agree that it currently checks the same, but I'm not fully convinced that there is an advantage of using `assertEqual`. Instead, there is a very small risk that this implementation might silently change in the future - which is fine (because the risk is very small) if we at least have some kind of benefit for now. But if there is no benefit at all, then it's risky overall...

climada/hazard/centroids/centr.py
```diff
@@ -820,169 +351,90 @@ def get_closest_point(self, x_lon, y_lat, scheduler=None):
...
-            i_lon = np.clip(i_lon, 0, self.meta['width'] - 1)
-            close_idx = int(i_lat * self.meta['width'] + i_lon)
-        else:
-            self.set_geometry_points(scheduler)
-            close_idx = self.geometry.distance(Point(x_lon, y_lat)).values.argmin()
+        close_idx = self.geometry.distance(Point(x_lon, y_lat)).values.argmin()
         return self.lon[close_idx], self.lat[close_idx], close_idx
 
-    def set_region_id(self, scheduler=None):
+    def set_region_id(self, level='country', overwrite=False):
```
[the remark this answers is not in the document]
chahank: I think this is needed for user-friendliness. You do not want to have to recreate new centroids all the time when you change the region_id. Also, if I remember correctly this is needed for the way Litpop is implemented.

The question is whether the name `region_id` should be reserved for `country_id` or it should also allow other types of regional information such as admin1.

climada/hazard/centroids/centr.py
```diff
@@ -1055,151 +504,258 @@ def select_mask(self, reg_id=None, extent=None):
...
-                            " meta object", centr_name, centr_val.__class__.__name__)
-        hf_str = data.create_dataset('crs', (1,), dtype=str_dt)
-        hf_str[0] = CRS.from_user_input(self.crs).to_wkt()
+                centroids.gdf.set_crs(dst_crs, inplace=True)
+        return centroids
+
+    #TODO: Check whether other variables are necessary, e.g. dist to coast
+    @classmethod
+    def from_csv(cls, file_path, crs=DEF_CRS, var_names=None):
+        """
```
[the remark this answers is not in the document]
chahank: But where do we store the defaults? This is potentially dangerous if we add/change one of these names in the future since there are more than one method that use them, so they can be missed easily.

climada/hazard/centroids/centr.py
```diff
@@ -22,1184 +22,897 @@
...
-            longitude step (positive)
-        n_lat : int
-            number of latitude points
-        n_lon : int
-            number of longitude points
-        crs : dict() or rasterio.crs.CRS, optional
-            CRS. Default: DEF_CRS
+        crs : str
+            coordinate reference system
+        inplace: bool, default False
```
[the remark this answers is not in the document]
chahank: For me `to_crs` is something typically a user would do to change the `crs`. On the other hand, `to_default_crs` would be to handle internal methods that implicitly require the default crs.

climada/util/coordinates.py
```diff
@@ -1543,7 +1540,8 @@ def get_country_code(lat, lon, gridded=False):
         region_id = region_id.astype(int)
     else:
         (lon_min, lat_min, lon_max, lat_max) = latlon_bounds(lat, lon, 0.001)
-        countries = get_country_geometries(extent=(lon_min, lon_max, lat_min, lat_max))
+        countries = get_country_geometries(
+            extent=(lon_min, lon_max, lat_min, lat_max), center_crs=False)
```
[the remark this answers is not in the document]
chahank: good point, I guess so.

climada/hazard/trop_cyclone.py
```diff
@@ -262,20 +262,15 @@ def from_tracks(
...
             # Select centroids with lat <= max_latitude
             coastal_idx = (np.abs(centroids.lat) <= max_latitude).nonzero()[0]
         else:
             # Select centroids which are inside max_dist_inland_km and lat <= max_latitude
-            if not centroids.dist_coast.size:
-                centroids.set_dist_coast()
-            coastal_idx = ((centroids.dist_coast <= max_dist_inland_km * 1000)
+            if 'dist_coast' not in centroids.gdf.columns:
+                centroids.gdf['dist_coast'] = centroids.get_dist_coast()
+            coastal_idx = ((centroids.gdf['dist_coast'].values <= max_dist_inland_km * 1000)
```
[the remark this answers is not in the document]
chahank: good point!

climada/hazard/centroids/centr.py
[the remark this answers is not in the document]
sarah-hlsn: @chahank I think this is an outdated comment, the csv and excel methods work by using **kwargs and the private method _from_dataframe, which you wrote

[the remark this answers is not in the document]
chahank: aaaa excellent!

climada/hazard/centroids/centr.py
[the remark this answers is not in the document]
chahank: Because it is quite expensive both in memory and cpu if you have the whole world at 1x1km resolution for instance.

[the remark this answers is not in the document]
tovogt: That's mainly because of the `gpd.points_from_xy`, right? Because just copying an array that contains the whole world in 1x1km resolution several times shouldn't be an issue.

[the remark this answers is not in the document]
tovogt: I can't imagine a scenario where this is repeated more than two or three times without stuff being garbage collected.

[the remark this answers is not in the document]
chahank: Good point.

tovogt:
I just wanted to drop here for reference the following point:

Centroids coordinates used to be reliably in lat/lon coordinates (epsg4326). Or at least that's what I assumed. Now that they are in a GeoDataFrame with a "serious" crs attribute, we should probably make sure that none of the functions that use centroids just blindly assume that the `Centroids.coord` attribute is in lat/lon. That applies to trop_cyclone (and tc_rainfield in petals), but it might also apply to other modules.

PR #791, opened 2023-10-17
Add function to plot Impf variability of calibration

timschmi95:
Changes proposed in this PR:
- add a function which plots the impact function variability given a certain percentage difference in the cost function values. Additionaly has the option to plot the distribution of hazard intensity values at the selected exposure points

As discussed at the retreat. For me this function was quite useful to get a "feel" for the data. In some cases, it highlighted completely different impact functions with an almost equal cost-function value. Let me know if you think such a function is useful for users in general.

The two added plots and description are in the Section "Analyze the Calibration" of the tutorial Jupyter notebook.

### PR Author Checklist

### PR Reviewer Checklist


doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -2512,6 +2485,65 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#select all hazard values at Exposure points\n",
+    "haz_vals = output_eval.input.hazard.intensity[:,output_eval.input.exposure.gdf.centr_TC]"
```
leonie-villiger, line 2502: If I understood correctly, calling the method `.plot_impf_variability` with `plot_haz=True` would do this step automatically, i.e., no need to provide the `haz_vals`. 

Do you perform this step explicitly here to show the functionality? Consider removing this step to illustrate the most "straight-forward" usages of `.plot_impf_variability`. i.e., do not encourage users to perform unnecessary steps.

doc/tutorial/climada_util_calibrate.ipynb
```diff
@@ -2512,6 +2485,65 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "#Plot the impact function variability\n",
+    "output_eval.plot_impf_variability(cost_func_diff=0.1)\n",
+    "output_eval.plot_impf_variability(cost_func_diff=0.2,plot_haz=True,haz_vals=haz_vals)"
```
leonie-villiger, line 2544: See previous comment
```suggestion
    "output_eval.plot_impf_variability(cost_func_diff=0.2,plot_haz=True)"
```

leonie-villiger (commented):
Thanks for adding the method `.plot_impf_variability()` to the `OutputEvaluator`! I believe this nice and informative visualisation is very helpful when calibrating impact functions. My comments: 
- code comments could be a bit more specific (see suggestions in `base.py`)
- don't fully understand the reasoning to show the intermediate step of getting the `haz_vals` before plotting (see `climada_util_calibrate.ipynb`)
- several trailing whitespace Linter issues (VS can remove the automatically, https://linux.how2shout.com/remove-trailing-spaces-automatically-in-visual-code-studio/)
- complete the PR checklist (just now saw that the PR is only a draft...uuuups. I can give a final review once you're fully ready).

doc/tutorial/climada_util_calibrate.ipynb
timschmi95: The haz_vals agrument was added for faster repeated plotting. However, with the newest version this is no longer required as now the default is fast enough. Thus, 'haz_vals' will be removed

timschmi95:
Thanks @leonie-villiger for the review, especially the automatic whitespace removal tip! :) 
Since this is a pull request within a hanging pull request I didn't update the documentation or Changelog, but marked it as done. 

@peanutfun I managed to get the histogram much faster by using the .data attribute. Do you want to have a look over it before I merge it? The failing unit test are not related to this specific pull request, and I didn't add any tests for the new plotting function for now (as with the other plotting function in the calibration module). Is that all fine?

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,81 @@ def plot_impf_set(self, **plot_kwargs):
...
+        # and remove the dimension 'Cost Function'.
+        params = p_space_df.columns.tolist()
+        try:
+            params.remove('Cost Function')
+        except ValueError:
+            pass
+
+        # Retrieve parameters of impact functions with cost function values
+        # within 'cost_func_diff' % of the best estimate
+        max_cost_func_val = p_space_df['Cost Function'].min()*(1+cost_func_diff)
```
peanutfun, line 195: I am wondering: Should this be an absolute value (as it is right now) or rather a relative one, like
```suggestion
        max_cost_func_val = p_space_df['Cost Function'].min() + cost_func_diff * (p_space_df['Cost Function'].max() - p_space_df['Cost Function'].min())
```
?

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,81 @@ def plot_impf_set(self, **plot_kwargs):
...
+        params_within_range = p_space_df.loc[p_space_df['Cost Function'] <=
+                                             max_cost_func_val,params]
+
+        # Initialize figure
+        fig,ax = plt.subplots()
+
+        #Plot best-fit impact function
+        best_impf = self.impf_set.get_func(haz_type=haz_type)[0]
```
peanutfun, line 203: This only selects the first impact function for the hazard type. How would we deal with multiple hazard functions?

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,81 @@ def plot_impf_set(self, **plot_kwargs):
         """
         return self.impf_set.plot(**plot_kwargs)
 
+    def plot_impf_variability(
+        self,
+        cost_func_diff: float = 0.1,
+        p_space_df: Optional[pd.DataFrame] = None,
+        plot_haz: bool = True,
+        **plot_kwargs
```
peanutfun, line 167: `**plot_kwargs` are only passed to the first, but not the second plot. Maybe we should add to arguments, like `plot_impf_kws` and `plot_hist_kws`

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,81 @@ def plot_impf_set(self, **plot_kwargs):
...
+        params_within_range = p_space_df.loc[p_space_df['Cost Function'] <=
+                                             max_cost_func_val,params]
+
+        # Initialize figure
+        fig,ax = plt.subplots()
+
+        #Plot best-fit impact function
+        best_impf = self.impf_set.get_func(haz_type=haz_type)[0]
+        ax.plot(best_impf.intensity,best_impf.mdd*best_impf.paa*100,
+                color='tab:blue',lw=2,zorder=3,label='best fit',**plot_kwargs)
```
peanutfun, line 205: Using `**plot_kwargs` like this does not let you update the default parameters like `color` or `lw`. I suggest pulling the default arguments from `plot_kwargs` using `pop`, as I did in the other plots of the module: https://github.com/CLIMADA-project/climada_python/pull/692/files#diff-c11ede847f0d2a20c8d179c645028e1b00b8a81735677b29c3538edb7446ed02R208

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,81 @@ def plot_impf_set(self, **plot_kwargs):
...
+            pass
+
+        # Retrieve parameters of impact functions with cost function values
+        # within 'cost_func_diff' % of the best estimate
+        max_cost_func_val = p_space_df['Cost Function'].min()*(1+cost_func_diff)
+        params_within_range = p_space_df.loc[p_space_df['Cost Function'] <=
+                                             max_cost_func_val,params]
+
+        # Initialize figure
+        fig,ax = plt.subplots()
```
peanutfun, line 200: Fix unused variable issue
```suggestion
        _, ax = plt.subplots()
```

peanutfun (changes requested):
@timschmi95 Very helpful contribution, thanks a lot for that! My main concerns are how to deal with multiple impact functions within the impact function set and with the handling of plot parameters.

The failing tests are fine, we will fix this in the "main" PR.

climada/util/calibrate/base.py
timschmi95: I think it should be an absolute one. The range of cost function values may be largely dependent on outliers. By saying "within 10% of the best fit" I would refer to estimates within 10% of the minimum (absolute) cost function value. If the 'worst' guess it 100 or 1'000 times as bad shouldn't matter.
Of course an alternative metric could be considered too.

climada/util/calibrate/base.py
timschmi95: Sounds good. For now, I didn't add a third kwargs argument for the impact functions in grey, so that's the only thing that still cannot be changed, but it may be an overkill to add this too. 
With plot_impf_kws and plot_hist_kws the plot can be adjusted quite flexibly.

climada/util/calibrate/base.py
timschmi95: Nice!

climada/util/calibrate/base.py
timschmi95: Ah I assumed the calibration can only handle one impact function at a time. But I see that with the custom impact_func_creator also a impf_set with multiple functions could be created. 
I'll make it into a loop, plotting each function then

timschmi95:
@peanutfun Thanks for the review, let me know if you think the added changes are suitable. Currently, the function will only return the last axis object in case there are multiple impact functions, but I this shouldn't be an issue in my opinion.

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,97 @@ def plot_impf_set(self, **plot_kwargs):
...
+        label = plot_impf_kws.pop('label','best fit')
+
+        #get number of impact functions and create a plot for each
+        n_impf = len(self.impf_set.get_func(haz_type=haz_type))
+        for impf_idx in range(n_impf):
+
+            #Plot best-fit impact function
+            best_impf = self.impf_set.get_func(haz_type=haz_type)[impf_idx]
+            ax.plot(best_impf.intensity,best_impf.mdd*best_impf.paa*100,
+                    color=color,lw=lw,zorder=zorder,label=label,**plot_impf_kws)
```
peanutfun, line 216: This plots all impact functions into the same plot. I fear this might look very confusing since there is currently no way to distinguish the grey impact functions. I think we either have to color the ensemble of impact functions the same way as the best fit, or have a facetted plot with one subplot for each impact function

timschmi95: @peanutfun Let me know if you think it's ready to merge with this latest change

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
         """
         return self.impf_set.plot(**plot_kwargs)
 
+    def plot_impf_variability(
+        self,
+        cost_func_diff: float = 0.1,
+        p_space_df: Optional[pd.DataFrame] = None,
+        plot_haz: bool = True,
+        plot_impf_kws : dict = {},
+        plot_hist_kws : dict = {},
```
peanutfun, line 168: Careful! No mutable default arguments!
```suggestion
        plot_impf_kws: Optional[dict] = None,
        plot_hist_kws: Optional[dict] = None,
```
Needs some more adjustments in the code then.

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
         """
         return self.impf_set.plot(**plot_kwargs)
 
+    def plot_impf_variability(
+        self,
+        cost_func_diff: float = 0.1,
+        p_space_df: Optional[pd.DataFrame] = None,
```
peanutfun, line 165: Can we change this around a bit? I am thinking of two cases:
1. I want the default parameter space and sample it using `cost_func_diff`. Then these default arguments make sense
2. I insert a custom parameter space. In this case, I probably want to plot the entire function space, because I already did a selection myself. But then I would have to set `cost_func_diff = np.inf`, which is very unintuitive.

What would then be a sensible default value for `cost_func_diff`? Or could we maybe add the option to set it to `None`, meaning that the entire parameter space is plotted?

climada/util/calibrate/base.py
peanutfun: Alright! Then let's leave it as is until further notice 🙃

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
...
+                sel_params = params_within_range.iloc[row,:].to_dict()
+                temp_impf_set = self.input.impact_func_creator(**sel_params)
+                temp_impf = temp_impf_set.get_func(haz_type=haz_type)[impf_idx]
+
+                ax.plot(temp_impf.intensity,temp_impf.mdd*temp_impf.paa*100,
+                        color='grey',alpha=0.4,label=label_temp)
+
+            # Plot hazard intensity value distributions
+            if plot_haz:
+                haz_vals = self.input.hazard.intensity[:,self.input.exposure.gdf[f"centr_{haz_type}"]]
```
peanutfun, line 232: Line too long
```suggestion
                haz_vals = self.input.hazard.intensity[
                    :, self.input.exposure.gdf[f"centr_{haz_type}"]
                ]
```

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
...
+            _,ax = plt.subplots()
+
+            #Plot best-fit impact function
+            best_impf = self.impf_set.get_func(haz_type=haz_type)[impf_idx]
+            ax.plot(best_impf.intensity,best_impf.mdd*best_impf.paa*100,
+                    color=color,lw=lw,zorder=zorder,label=label,**plot_impf_kws)
+
+            #Plot all impact functions within 'cost_func_diff' % of best estimate
+            for row in range(params_within_range.shape[0]):
+                label_temp = f'within {int(cost_func_diff*100)} percent of best fit' if row==0 else None
```
peanutfun, line 221: Line too long
```suggestion
                label_temp = (
                    f"within {int(cost_func_diff*100)} percent of best fit"
                    if row == 0
                    else None
                )
```

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
...
+            cost_func_diff (float, optional): Max deviation from optimal cost
+            function value (as fraction). Defaults to 0.1 (i.e. 10%).
+            p_space_df (pd.DataFrame, optional): parameter space. Defaults to None.
+            plot_haz (bool, optional): Whether or not to plot hazard intensity
+            distibution. Defaults to False.
+        """
+
+        haz_type = self.input.hazard.haz_type
+        if p_space_df is None:
+            p_space_df = self.output.p_space_to_dataframe()
```
peanutfun, line 184: Big problem: The `OutputEvaluator` takes any `Output` object, but `p_space_to_dataframe()` is only available from `BayesianOptimizerOutput`. To avoid confusion, it might make sense the derive a `BayesianOutputEvaluator` from `OutputEvaluator` and move this function there, or have a type check in the beginng of the function. Sorry I didn't notice that before 🙈

climada/util/calibrate/base.py
```diff
@@ -158,6 +159,102 @@ def plot_impf_set(self, **plot_kwargs):
...
+            ax.set(xlabel=f"Intensity ({self.input.hazard.units})",
+                ylabel="Mean Damage Ratio (MDR) in %",
+                xlim=(min(best_impf.intensity),max(best_impf.intensity)))
+            ax.legend()
+            axes.append(ax)
+
+        if n_impf > 1:
+            return axes
+        else:
+            return ax
```
peanutfun, line 255: Unnecessary else:
```suggestion
        if n_impf > 1:
            return axes

        return ax
```

peanutfun (changes requested):

climada/util/calibrate/base.py
timschmi95: Good point. In my opinion a type check at the beginning of the function would be a good solution.

peanutfun: Alright, let's do it like this for now.

climada/util/calibrate/base.py
timschmi95: Hm, how about we keep 10% as default value and if someone adds cost_func_diff=None, we  add something like:
```
if cost_func_diff is None:
    # If no value is given, plot the complete parameter space
    max_cost_func_val = np.inf
else:
    max_cost_func_val = p_space_df['Cost Function'].min()*(1+cost_func_diff)
```
What do you think? Having 'None' as default may be confusing if someone has the complete (e.g. Bayesian) parameter space.

Furthermore, I realize that if the parameter space is given, the output doesn't have to be a BayesianOptimizerOutput anymore. So, I'll add the assertion within the "if p_space_df" code block.

peanutfun: > Having 'None' as default may be confusing if someone has the complete (e.g. Bayesian) parameter space.

True, sounds good. Then you could do:

```python
if p_space_df is 
    p_space_df = self.output.p_space_to_dataframe()  # Check if that exists first!

if cost_func_diff is None:
    params_within_range = p_space_df[params]
else:
    params_within_range = p_space_df.loc[
        p_space_df['Cost Function'] <= max_cost_func_val, params
    ]
```

climada/util/calibrate/base.py
```diff
@@ -158,6 +160,131 @@ def plot_impf_set(self, **plot_kwargs):
...
+            plot_impf_kws = {}
+        if plot_hist_kws is None:
+            plot_hist_kws = {}
+
+        # Retrieve hazard type and parameter space
+        haz_type = self.input.hazard.haz_type
+        if p_space_df is None:
+            # Assert that self.output has the p_space_to_dataframe() method,
+            # which is defined for the BayesianOptimizerOutput class
+            if not hasattr(self.output,"p_space_to_dataframe"):
```
timschmi95, line 198: @peanutfun if you're happy with this _implicit_ typecheck (which is only required if p_space_df is not given) I think we can merge?

peanutfun: Yes, I think we can merge. Thanks a lot for this very useful contribution!

PR #792, opened 2023-10-18
Fix dist_approx with "geosphere" method and close points

tovogt:
Changes proposed in this PR:
- Fix the `dist_approx` util function when used with `method="geosphere"` and `log=True`.
- This only affects very small input values, i.e. computing the distance between points that are very close. I added a unit test for that case. I also added a test case where points are identical because I think that this is a kind of sanity check that should be included.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/util/test/test_coordinates.py
+++ b/climada/util/test/test_coordinates.py
@@ -277,13 +277,20 @@ class TestFunc(unittest.TestCase):
     def test_dist_approx_pass(self):
         """Test approximate distance functions"""
         data = np.array([
-            # lat1, lon1, lat2, lon2, dist, dist_sph
+            # lat1, lon1, lat2, lon2, dist_equirect, dist_geosphere
             [45.5, -32.1, 14, 56, 7702.88906574, 8750.64119051],
             [45.5, 147.8, 14, -124, 7709.82781473, 8758.34146833],
             [45.5, 507.9, 14, -124, 7702.88906574, 8750.64119051],
             [45.5, -212.2, 14, -124, 7709.82781473, 8758.34146833],
             [-3, -130.1, 4, -30.5, 11079.7217421, 11087.0352544],
         ])
+        # conversion factors from reference data (in km, see above) to other units
+        factors_km_to_x = {
+            "m": 1e3,
+            "radian": np.radians(1.0) / u_coord.ONE_LAT_KM,
+            "degree": 1.0 / u_coord.ONE_LAT_KM,
+            "km": 1.0,
+        }
         compute_dist = np.stack([
             u_coord.dist_approx(data[:, None, 0], data[:, None, 1],
                                 data[:, None, 2], data[:, None, 3],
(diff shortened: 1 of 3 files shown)


climada/util/test/test_coordinates.py
```diff
@@ -309,21 +314,29 @@ def test_dist_approx_pass(self):
...
 
     def test_dist_approx_log_pass(self):
         """Test log-functionality of approximate distance functions"""
         data = np.array([
-            # lat1, lon1, lat2, lon2, dist, dist_sph
+            # lat1, lon1, lat2, lon2, dist_equirect, dist_geosphere
             [0, 0, 0, 1, 111.12, 111.12],
             [-13, 179, 5, -179, 2011.84774049, 2012.30698122],
+            [24., 85., 23.99999967, 85., 3.666960e-5, 3.666960e-5],
+            [24., 85., 24., 85., 0, 0],
```
emanuel-schmid, line 329: Just to be sure: this test is supposed to pass even with the former version of `dist_approx`, right?
I'm asking because it actually does and I was expecting it to fail due to the PR description stating

> This only affects very small input values, i.e. computing the distance between points that are very close _**or even identical**_. I added a unit test for that case.

tovogt: Yes, the test fails with the former version:
```bash
$ python -m unittest climada.util.test.test_coordinates.TestFunc.test_dist_approx_log_pass
F
======================================================================
FAIL: test_dist_approx_log_pass (climada.util.test.test_coordinates.TestFunc)
Test log-functionality of approximate distance functions
  File "/home/tovogt/code/climada_python/climada/util/test/test_coordinates.py", line 333, in test_dist_approx_log_pass
    np.testing.assert_allclose(np.linalg.norm(vec, axis=-1), dist)
  File "/home/tovogt/.local/share/miniconda3/envs/climada_env_3.9/lib/python3.9/site-packages/numpy/testing/_private/utils.py", line 1592, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/tovogt/.local/share/miniconda3/envs/climada_env_3.9/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/tovogt/.local/share/miniconda3/envs/climada_env_3.9/lib/python3.9/site-packages/numpy/testing/_private/utils.py", line 862, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 4 (25%)
Max absolute difference: 0.14939717
Max relative difference: 25938870.97724836
 x: array([0.017453, 0.316067, 0.149397, 0.      ])
 y: array([1.745329e-02, 3.160672e-01, 5.759587e-09, 0.000000e+00])

Ran 1 test in 0.002s

FAILED (failures=1)
```
Are you sure that you adapted the `test_dist_approx_log_pass` function and not the `test_dist_approx_pass`? They have a very similar name and very similar code.

tovogt: Ah, your question was only about the last case, right, where the points are identical... Wait a second..

tovogt: I think you are right. I am not 100% sure, but it's well possible that I originally added this test case because I thought that it is a "natural" aspect to test that was missing so far. And then, when I wrote the PR text, I forgot about that. Sorry for the confusion. There were more than 3 weeks between my commit and the PR ...

emanuel-schmid:
Looks good to me. Obviously fixes a bug.

PR #793, opened 2023-10-18
Fix exponent clipping in Holland 2010 implementation

tovogt:
Changes proposed in this PR:
- This PR is in response to a mail by a PhD student who had questions about the implementation of the Holland 2010 TC wind model in CLIMADA.
- The current implementation clips the `hol_x` exponent to the [0.0, 0.5] interval, even though this is not mentioned in the literature. I understand that negative values are undesirable because it would cause the wind speeds to increase outwards (which is unphysical). However, I don't think that clipping high values to 0.5 is necessary. That would just mean that wind speeds are decreasing very fast outwards, which seems okay to me.
- While nobody currently seems to use the Holland 2010 wind model (including me), it might still be interesting to ensure the correctness of the implementation.

### PR Author Checklist

### PR Reviewer Checklist


climada/hazard/test/test_trop_cyclone.py
```diff
@@ -306,20 +306,42 @@ def test_holland_2010_pass(self):
...
+        ], dtype=bool)
         hol_x = _x_holland_2010(si_track, d_centr, close_centr)
-        np.testing.assert_array_almost_equal(
-            hol_x, [[0.5, 0.5, 0.47273], [0.5, 0, 0.211602]])
+        np.testing.assert_array_almost_equal(hol_x, [
+            [0.5, 0.500000, 0.472730],
+            [0.5, 0.000000, 0.211602],
+            [0.5, 0.519094, 0.625478],
+            [0.5, 0.495423, 0.049174],
+            [0.5, 0.464344, 0.000000],
```
peanutfun, line 333: I am guessing the two zeros here are an effect of the clipping an something you wanted to explicitly test for?

peanutfun (commented):
The actual change looks easy enough, but it's really hard for me to understand the tests. Could we maybe reduce the number of values we are testing against? 🤔

climada/hazard/test/test_trop_cyclone.py
leonie-villiger: In my understanding, the first zero results from the large distance between track and centroids (`close_centr = False`), thus does not test the clipping. The last zero tests the clipping by creating the conflict v_max_s < v_n (v_max_s = 12 and v_n defaulting to 17). To me, it is also unclear what is tested in the other two, added cases.

tovogt:
Thanks for checking!

I modified the test cases to be more systematical, and also added explanations in the code.

leonie-villiger:
Thanks for adapting the test. Thanks to your explanations in the code, it is very clear what is tested. In my opinion ready to merge.

leonie-villiger:
Thanks for adapting the test. Your explanations make it is very clear what is tested. In my opinion, ready to merge.

peanutfun (approved):
This is much clearer now. Thank you, @tovogt! Merging.

PR #795, opened 2023-10-23
Change default Hazard.event_name to string of event_id

### LINKED ISSUE #789: `Hazard.from_raster_xarray` stores native values of dates in `Hazard.event_name`
The current implementation uses the exact date/event values as `Hazard.event_name` if no explicit event names can be found in the dataset. This causes issues when trying to store the hazard object or a related impact object as H5 file. We should somehow make sure that `Hazard.event_name` is sanitized. Suggestions:

* Using a different default value for `event_name` including none at all (empty list)
* Trying to cast any time information to a string, e.g. via `pandas.Timestamp(x).isoformat()`

sarah-hlsn:
Changes proposed in this PR:
- changed default Hazard.event_name to string representation of event_id
- changed Hazard test_base_xarray.py accordingly

This PR fixes #789 

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -518,7 +518,7 @@ class Hazard():
               :py:meth:`Hazard.__init__` for details.
             * ``hazard_type``: Empty string
             * ``frequency``: 1.0 for every event
-            * ``event_name``: String representation of the event time
+            * ``event_name``: String representation of the event id
             * ``event_id``: Consecutive integers starting at 1 and increasing with time
         crs : str, optional
             Identifier for the coordinate reference system of the coordinates. Defaults
@@ -840,7 +840,7 @@ class Hazard():
                     None,
                     np.ones(num_events),
                     np.array(range(num_events), dtype=int) + 1,
-                    list(data[coords["event"]].values),
+                    [str(x) for x in np.array(range(num_events), dtype=int) + 1],
                     np.array(u_dt.datetime64_to_ordinal(data[coords["event"]].values)),
                 ],
                 # The accessor for the data in the Dataset
--- a/climada/hazard/test/test_base_xarray.py
+++ b/climada/hazard/test/test_base_xarray.py
@@ -72,7 +72,7 @@ class TestReadDefaultNetCDF(unittest.TestCase):
         self.assertEqual(hazard.units, "")
         np.testing.assert_array_equal(hazard.event_id, [1, 2])
         np.testing.assert_array_equal(
-            hazard.event_name, [np.datetime64(val) for val in self.time]
+            hazard.event_name, [str(x) for x in hazard.event_id]
         )
         np.testing.assert_array_equal(
             hazard.date, [val.toordinal() for val in self.time]
@@ -342,9 +342,7 @@ class TestReadDefaultNetCDF(unittest.TestCase):
             ds = ds.isel(time=0).squeeze()
             hazard = Hazard.from_xarray_raster(ds, "", "")
             self._assert_default_types(hazard)
-            np.testing.assert_array_equal(
-                hazard.event_name, [np.datetime64(self.time[0])]
-            )
+
             np.testing.assert_array_equal(hazard.date, [self.time[0].toordinal()])
             np.testing.assert_array_equal(hazard.centroids.lat, [0, 0, 0, 1, 1, 1])
             np.testing.assert_array_equal(hazard.centroids.lon, [0, 1, 2, 0, 1, 2])
@@ -369,9 +367,6 @@ class TestReadDefaultNetCDF(unittest.TestCase):
             ds = ds.expand_dims(time=[np.datetime64("2022-01-01")])
             hazard = Hazard.from_xarray_raster(ds, "", "")
             self._assert_default_types(hazard)
-            np.testing.assert_array_equal(
-                hazard.event_name, [np.datetime64("2022-01-01")]
-            )
             np.testing.assert_array_equal(
                 hazard.date, [dt.datetime(2022, 1, 1).toordinal()]
             )


climada/hazard/base.py
```diff
@@ -840,7 +840,7 @@ def maybe_repeat(values: np.ndarray, times: int) -> np.ndarray:
                     None,
                     np.ones(num_events),
                     np.array(range(num_events), dtype=int) + 1,
-                    list(data[coords["event"]].values),
+                    [str(x) for x in np.array(range(num_events), dtype=int) + 1],
```
peanutfun, line 843: This does actually not behave as expected: If the dataset supplies custom event IDs (like 100...100+X), this will still set the event name to values 1...X. However, I don't see a way of inserting the event ID data as default value here. @chahank, what do you think?

climada/hazard/test/test_base_xarray.py
```diff
@@ -342,9 +342,7 @@ def test_missing_dims(self):
             ds = ds.isel(time=0).squeeze()
             hazard = Hazard.from_xarray_raster(ds, "", "")
             self._assert_default_types(hazard)
-            np.testing.assert_array_equal(
-                hazard.event_name, [np.datetime64(self.time[0])]
-            )
```
peanutfun, line 347: Why remove the test instead of adapting it to the new expected values? 🤔

climada/hazard/test/test_base_xarray.py
```diff
@@ -369,9 +367,6 @@ def test_missing_dims(self):
             ds = ds.expand_dims(time=[np.datetime64("2022-01-01")])
             hazard = Hazard.from_xarray_raster(ds, "", "")
             self._assert_default_types(hazard)
-            np.testing.assert_array_equal(
-                hazard.event_name, [np.datetime64("2022-01-01")]
-            )
```
peanutfun, line 374: Same as above, why not check the new expected value?

climada/hazard/test/test_base_xarray.py
sarah-hlsn: To my understanding this test (test_missing_dims) is supposed to check whether everything still works after removing time as a dimension. Since the default hazard.event_name used to depend on the time it was necessary to test whether this still returned the correct default event_name. My thinking was that if event_name is not connected to time anymore it would not be necessary to test the default name in this specific test

climada/hazard/test/test_base_xarray.py
sarah-hlsn: see above

climada/hazard/base.py
chahank: Good question, but I think it does not matter that much. I was actually expecting that this PR would change the `Hazard.init` to make a good default instead of just the read xarray datasets method. But, this might be a larger endeavor. 

Also, you do not need to cast the generator into an array (this is actually very sub-optimal, both memory-wise, speed-wise and style-wise). Keeping the original values, this would be it:

```suggestion
                    [str(x) for x in range(1, num_events+ 1), dtype=int)],
```

Another side-note to @peanutfun : why do you use a list here instead of a dictionary? How do you keep track of the correct ordering? Is there a particular reason to use a list?

peanutfun: @chahank We could indeed adjust the `Hazard.__init__` to handle that. I think this would be the cleanest solution. However, changing the default initialization of this class might have some repercussions elsewhere.

> why do you use a list here instead of a dictionary?

Not sure I understand. The whole thing is a dictionary which is inserted into a dataframe. This transforms the dictionary keys into column names and dictionary values into the column series. Order is preserved. I agree that it might be a bit less confusing if we used [`DataFrame.from_records`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.from_records.html), where the dataframe can be initialized row-wise. But I recall you insisted that I create a separate variable `DEF_DATA_VARS` instead of hardcoding the data variables here, so I thought this was the cleanest approach.

chahank: > @chahank We could indeed adjust the Hazard.__init__ to handle that. I think this would be the cleanest solution. However, changing the default initialization of this class might have some repercussions elsewhere.

Agreed. This will come later when we change the hazard attributes to a better data format I suppose.

>Not sure I understand.
Sorry that I was unclear! What I meant is why is the `default_var` variable a list (because then the ordering matters in that list, which is not directly obvious)?

peanutfun: > Agreed. This will come later when we change the hazard attributes to a better data format I suppose.

Then we need an alternative. I would not like to have `event_id` set to 100..100+X and `event_name` to 1..1+X. My proposal is then to simply leave the event name empty, and let the Hazard initializer handle these cases in the future.

> What I meant is why is the default_var variable a list (because then the ordering matters in that list, which is not directly obvious)?

Still unclear. Do you mean `default_value`? Yes, of course ordering matters, because the list relates to the `DEF_DATA_VARS`. Again, I can change initialization to row-wise, that would probably make it less confusing.

chahank: > Then we need an alternative. I would not like to have event_id set to 100..100+X and event_name to 1..1+X. My proposal is then to simply leave the event name empty, and let the Hazard initializer handle these cases in the future.

I think having an empty hazard event is not an option, this would break some functionalities. At the moment, the initialization does NOT handle empty event_names properly, so we would have to wait until we fix that (or do it now).

chahank: > Still unclear. Do you mean default_value? Yes, of course ordering matters, because the list relates to the DEF_DATA_VARS. Again, I can change initialization to row-wise, that would probably make it less confusing.

Exactly, the ordering of the values for `default_value` at that point in the code is not obvious. But it is also not a big deal. Maybe just a comment referring to `DEF_DATA_VARS` for the ordering/content?

peanutfun: > I think having an empty hazard event is not an option, this would break some functionalities. At the moment, the initialization does NOT handle empty event_names properly, so we would have to wait until we fix that (or do it now).

Okay, then I suggest the following: We create a follow-up issue to update the default initialization of Hazard. For now, we keep the functionality of `from_xarray_raster` as is, i.e. `event_name` stores the date. We just make sure that the `event_name` is indeed a string (because we should always be able to write those). We can do that by calling [`dt.strftime`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.dt.strftime.html#xarray-dataarray-dt-strftime) on the array and storing the result.

> Maybe just a comment referring to DEF_DATA_VARS for the ordering/content?

Will do.

chahank: > Okay, then I suggest the following: We create a follow-up issue to update the default initialization of Hazard. For now, we keep the functionality of from_xarray_raster as is, i.e. event_name stores the date. We just make sure that the event_name is indeed a string (because we should always be able to write those). We can do that by calling [dt.strftime](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.dt.strftime.html#xarray-dataarray-dt-strftime) on the array and storing the result.

Good idea! Let us do this then. @sarah-hlsn can you make the event_name to be the string version of the date as suggested by @peanutfun ?

sarah-hlsn: Yes, will do!

climada/hazard/test/test_base_xarray.py
peanutfun: Yes, I agree. Thanks for clarifying my own test 🙈

sarah-hlsn: Anytime :D Now I will keep it in and adapt it based on the discussion above though

peanutfun:
@sarah-hlsn This is coming along very nicely! Could you add an explicit check for `event_name` in [`test_2D_time`](https://github.com/CLIMADA-project/climada_python/blob/30464eedff808cddcc1520cb38e7d4527003d55a/climada/hazard/test/test_base_xarray.py#L516)? I am not sure if calling `list` on the DataArray will always result in a "flat" list. Maybe we have to create the list from `<arr>.values.flat`.

chahank:
@sarah-hlsn : could you then please update the changelog too? Then we can soon merge this bugfix.

sarah-hlsn:
@chahank Yes will do, I am currently working on the additional test suggested by @peanutfun but I may need some more time because it is not working as I expected

peanutfun:
@sarah-hlsn Let me know if you need any help

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- Change default event name to string of event date when importing Hazard data from xarray [#795](https://github.com/CLIMADA-project/climada_python/pull/795)
```
chahank, line 17: Could you please also add under `fixed` the issue that was fixed with this pull request please?

chahank:
Looks good to me. @peanutfun merge?

CHANGELOG.md
```diff
@@ -14,12 +14,14 @@ Code freeze date: YYYY-MM-DD
...
 - Add `shapes` argument to `geo_im_from_array` to allow flexible turning on/off of plotting coastline in `plot_intensity`. [#805](https://github.com/CLIMADA-project/climada_python/pull/805)
 - Update `CONTRIBUTING.md` to better explain types of contributions to this repository [#797](https://github.com/CLIMADA-project/climada_python/pull/797)
 - The default tile layer in Exposures maps is not Stamen Terrain anymore, but [CartoDB Positron](https://github.com/CartoDB/basemap-styles). Affected methods are `climada.engine.Impact.plot_basemap_eai_exposure`,`climada.engine.Impact.plot_basemap_impact_exposure` and `climada.entity.Exposures.plot_basemap`. [#798](https://github.com/CLIMADA-project/climada_python/pull/798)
 
 ### Fixed
 
+- Hazard.from_raster_xarray used to native values of dates in Hazard.event_name, now it is only a string representation of the event date [#795](https://github.com/CLIMADA-project/climada_python/pull/795)
```
peanutfun, line 24: We always only link the pull requests. The fixed issue is mentioned in the PR description.
```suggestion
- `Hazard.from_xarray_raster` now stores strings as default values for `Hazard.event_name` [#795](https://github.com/CLIMADA-project/climada_python/pull/795)
```

CHANGELOG.md
peanutfun: I think just mentioning this under `Fixed` is enough

CHANGELOG.md
```diff
@@ -14,12 +14,14 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- Change default event name to string of event date when importing Hazard data from xarray [#795](https://github.com/CLIMADA-project/climada_python/pull/795)
```
peanutfun, line 17:
```suggestion
```

climada/hazard/test/test_base_xarray.py
```diff
@@ -343,8 +343,8 @@ def test_missing_dims(self):
             hazard = Hazard.from_xarray_raster(ds, "", "")
             self._assert_default_types(hazard)
             np.testing.assert_array_equal(
-                hazard.event_name, [np.datetime64(self.time[0])]
-            )
+                hazard.event_name, [self.time[0].strftime("%Y-%m-%d")]
+                )
```
peanutfun, line 347: black would format like this
```suggestion
            )
```

climada/hazard/test/test_base_xarray.py
```diff
@@ -551,6 +551,7 @@ def test_2D_time(self):
         np.testing.assert_array_equal(
             hazard.date, [val.toordinal() for val in time.flat]
         )
+        np.testing.assert_array_equal(hazard.event_name, ["1999-01-01", "1999-02-01", "2000-01-01", "2000-02-01"])
```
peanutfun, line 554: black would format like this
```suggestion
        np.testing.assert_array_equal(
            hazard.event_name, ["1999-01-01", "1999-02-01", "2000-01-01", "2000-02-01"]
        )
```

peanutfun (changes requested):
Very good @sarah-hlsn! 🙌  I only have a few stylistic suggestions.

peanutfun:
@chahank, would you go trough my suggestions and merge afterwards?

CHANGELOG.md
chahank: FYI: I think it is in general a good idea to also mention which issue was fixed.

chahank:
Thanks @sarah-hlsn and @peanutfun !

PR #837, opened 2024-01-15
Allow reading Hazard events that are not dates from xarray

### LINKED ISSUE #829: `Hazard.from_xarray_raster` requires that "event" dimension can be interpreted as datetime
https://github.com/CLIMADA-project/climada_python/pull/795 has an unexpected repercussion (for me): The inserted hazard data now has to provide an "event" dimension that can be interpreted as datetime. I think this is unnecessarily strict. Consider the following code:

```python
import xarray as xr
import numpy as np

intensity = np.array([[[0, 1, 2], [3, 4, 5]], [[6, 7, 8], [9, 10, 11]]])
number = np.array([1, 2])
latitude = np.array([0, 1])
longitude = np.array([0, 1, 2])
dset = xr.Dataset(
    {
        "intensity": (["number", "latitude", "longitude"], intensity),
    },
    dict(number=number, latitude=latitude, longitude=longitude),
)

Hazard.from_xarray_raster(
    dset, "", "", intensity="intensity", coordinate_vars={"event": "number"}
)
```
This results in
```
TypeError: '.dt' accessor only available for DataArray with datetime64 timedelta64 dtype or for arrays containing cftime datetime objects.
```
Providing event names explicitly does **not** help, because the default values are always evaluated in the current implementation.

```python
dset["event_name"] = ["a", "b"]
Hazard.from_xarray_raster(
    dset, "", "", intensity="intensity", coordinate_vars={"event": "number"}
)  # Same error
```
Note that following line is *always* evaluated, even if the default value is not used: https://github.com/CLIMADA-project/climada_python/blob/0b06f3f8277b4c392a7e9cdd80a5c258c1ff1175/climada/hazard/base.py#L843

So, could we maybe just drop this default value altogehter? The default event name could just be a list of empty strings.

peanutfun:
Changes proposed in this PR:
* Try interpreting values of the event coordinate as dates or ordinals for default values of `Hazard.date`. If that fails, issue a warning and set default values to zeros.
* Try interpreting values of the event coordinate as dates for default values of `Hazard.event_name`. If that fails, issue a warning and set default values to empty strings.
* Update tests.

The method still tries to read the values as dates but issues a warning if that does not work. The fallback is zeros in case of `Hazard.date` and empty strings for `Hazard.event_name`.

This extends #795, which actually introduced tighter restrictions for the 'event' coordinate. This causes several issues in my new flood module, see https://github.com/CLIMADA-project/climada_petals/pull/64.

This fixes #829.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -513,12 +513,13 @@ class Hazard():
 
             Default values are:
 
-            * ``date``: The ``event`` coordinate interpreted as date
+            * ``date``: The ``event`` coordinate interpreted as date or ordinal, or
+              zeros if that fails (which will issue a warning).
             * ``fraction``: ``None``, which results in a value of 1.0 everywhere, see
               :py:meth:`Hazard.__init__` for details.
             * ``hazard_type``: Empty string
             * ``frequency``: 1.0 for every event
-            * ``event_name``: String representation of the event time
+            * ``event_name``: List of empty strings
             * ``event_id``: Consecutive integers starting at 1 and increasing with time
         crs : str, optional
             Identifier for the coordinate reference system of the coordinates. Defaults
@@ -553,13 +554,13 @@ class Hazard():
           and Examples) before loading the Dataset as Hazard.
         * Single-valued data for variables ``frequency``. ``event_name``, and
           ``event_date`` will be broadcast to every event.
+        * The ``event`` dimension need not be a time or date.
         * To avoid confusion in the call signature, several parameters are keyword-only
           arguments.
         * The attributes ``Hazard.haz_type`` and ``Hazard.unit`` currently cannot be
           read from the Dataset. Use the method parameters to set these attributes.
         * This method does not read coordinate system metadata. Use the ``crs`` parameter
           to set a custom coordinate system identifier.
-        * This method **does not** read lazily. Single data arrays must fit into memory.
 
         Examples
         --------
@@ -802,14 +803,28 @@ class Hazard():
                 raise ValueError(f"'{array.name}' data must be larger than zero")
             return array.values
 
-        def date_to_ordinal_accessor(array: xr.DataArray) -> np.ndarray:
+        def date_to_ordinal_accessor(
+            array: xr.DataArray, strict: bool = True
+        ) -> np.ndarray:
             """Take a DataArray and transform it into ordinals"""
-            if np.issubdtype(array.dtype, np.integer):
-                # Assume that data is ordinals
-                return strict_positive_int_accessor(array)
+            try:
+                if np.issubdtype(array.dtype, np.integer):
+                    # Assume that data is ordinals
+                    return strict_positive_int_accessor(array)
 
-            # Try transforming to ordinals
-            return np.array(u_dt.datetime64_to_ordinal(array.values))
+                # Try transforming to ordinals
+                return np.array(u_dt.datetime64_to_ordinal(array.values))
+
+            # Handle access errors
+            except (ValueError, TypeError) as err:
+                if strict:
+                    raise err
+
+                LOGGER.warning(
+                    "Failed to read values of '%s' as dates or ordinals. Hazard.date "
+                    "will be zeros only" % array.name
+                )
+                return np.zeros(array.shape)
 
         def maybe_repeat(values: np.ndarray, times: int) -> np.ndarray:
             """Return the array or repeat a single-valued array
@@ -840,8 +855,8 @@ class Hazard():
                     None,
                     np.ones(num_events),
                     np.array(range(num_events), dtype=int) + 1,
-                    data[coords["event"]].dt.strftime("%Y-%m-%d").values.flatten().tolist(),
-                    np.array(u_dt.datetime64_to_ordinal(data[coords["event"]].values)),
+                    [""] * num_events,
+                    date_to_ordinal_accessor(data[coords["event"]], strict=False),
                 ],
                 # The accessor for the data in the Dataset
                 accessor=[
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -826,6 +826,24 @@ class Hazard():
                 )
                 return np.zeros(array.shape)
 
+        def year_month_day_accessor(
+            array: xr.DataArray, strict: bool = True
+        ) -> np.ndarray:
+            """Take an array and return am array of YYYY-MM-DD strings"""
+            try:
+                return array.dt.strftime("%Y-%m-%d").values
+
+            # Handle access errors
+            except (ValueError, TypeError) as err:
+                if strict:
+                    raise err
+
+                LOGGER.warning(
+                    "Failed to read values of '%s' as dates. Hazard.event_name will be "
+                    "empty strings" % array.name
+                )
+                return np.full(array.shape, "")
+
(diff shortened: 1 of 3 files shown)


climada/hazard/base.py
```diff
@@ -513,12 +513,13 @@ def from_xarray_raster(
...
 
-            * ``date``: The ``event`` coordinate interpreted as date
+            * ``date``: The ``event`` coordinate interpreted as date or ordinal, or
+              zeros if that fails (which will issue a warning).
             * ``fraction``: ``None``, which results in a value of 1.0 everywhere, see
               :py:meth:`Hazard.__init__` for details.
             * ``hazard_type``: Empty string
             * ``frequency``: 1.0 for every event
-            * ``event_name``: String representation of the event time
+            * ``event_name``: List of empty strings
```
peanutfun, line 522:
```suggestion
            * ``event_name``: String representation of the event date or empty strings
              if that fails (which will issue a warning).
```

peanutfun:
Failing tests have nothing to do with this PR

climada/hazard/base.py
```diff
@@ -513,12 +513,14 @@ def from_xarray_raster(
 
             Default values are:
 
-            * ``date``: The ``event`` coordinate interpreted as date
+            * ``date``: The ``event`` coordinate interpreted as date or ordinal, or
+              zeros if that fails (which will issue a warning).
```
chahank, line 517: I am not sure about the value `0` since this is then not a valid ordinal. Thus, one would get a non-valid `event_date` which then creates easily problems down the line when a method expects an ordinal. This then might lead to hard-to-debug/understand error messages for users. Any other idea?

climada/hazard/base.py
```diff
@@ -553,13 +555,13 @@ def from_xarray_raster(
           and Examples) before loading the Dataset as Hazard.
         * Single-valued data for variables ``frequency``. ``event_name``, and
           ``event_date`` will be broadcast to every event.
+        * The ``event`` dimension need not be a time or date.
```
chahank, line 558: I do not understand what this is meaning here.

climada/hazard/base.py
```diff
@@ -553,13 +555,13 @@ def from_xarray_raster(
...
         * To avoid confusion in the call signature, several parameters are keyword-only
           arguments.
         * The attributes ``Hazard.haz_type`` and ``Hazard.unit`` currently cannot be
           read from the Dataset. Use the method parameters to set these attributes.
         * This method does not read coordinate system metadata. Use the ``crs`` parameter
           to set a custom coordinate system identifier.
-        * This method **does not** read lazily. Single data arrays must fit into memory.
```
chahank, line 562: Does it load lazily now? That would be great.

climada/hazard/base.py
```diff
@@ -513,12 +513,14 @@ def from_xarray_raster(
...
-            * ``date``: The ``event`` coordinate interpreted as date
+            * ``date``: The ``event`` coordinate interpreted as date or ordinal, or
+              zeros if that fails (which will issue a warning).
             * ``fraction``: ``None``, which results in a value of 1.0 everywhere, see
               :py:meth:`Hazard.__init__` for details.
             * ``hazard_type``: Empty string
             * ``frequency``: 1.0 for every event
-            * ``event_name``: String representation of the event time
+            * ``event_name``: String representation of the event date or empty strings
+              if that fails (which will issue a warning).
```
chahank, line 523: Just to be sure I understand: if the data contains a field 'name', it will be ignored by default?

climada/hazard/test/test_base_xarray.py
```diff
@@ -162,6 +159,29 @@ def test_type_and_unit(self):
...
 
+    def test_event_no_time(self):
+        """Test if an event coordinate that is not a time works"""
+        with xr.open_dataset(self.netcdf_path) as dataset:
+            size = dataset.sizes["time"]
+
+            # Integers
+            time = np.arange(size)
+            dataset["time"] = time
+            self.time = ["1970-01-01", "1970-01-02"]  # These will be 0, 1 as ordinals
```
chahank, line 170: I think ordinals must be larger equal to 1. Is the comment thus incorrect?

chahank:
Great, good improved fix. A few comments mostly for clarity and one main question: is it smart to use a non-valid ordinal as the default date?

climada/hazard/base.py
peanutfun: You are correct. I actually messed up the test. It is fixed now. I also did not realize that 0 does not work as ordinal. I now chose 1 as default value, the ordinal of date 01.01.0001.

climada/hazard/base.py
peanutfun: Yes, could be clearer. This way, maybe?
```suggestion
        * The values of the ``event`` coordinate need not be times or dates.
```

climada/hazard/base.py
peanutfun: It has been doing so for quite a while, I just missing deleting this line 🙃

climada/hazard/base.py
peanutfun: This list only specifies the *default* values. These are chosen if the dataset does *not*  contain a field `event_name` OR the user chose to ignore it via `read_xarray_raster(..., data_vars=dict(event_name=""))`

climada/hazard/test/test_base_xarray.py
peanutfun: Yes, the test was wrong.

climada/hazard/base.py
chahank, line 558:
```suggestion
        * The values of the ``event`` coordinates can be of any type (times, dates, strings, ...)
```

chahank:
Suggestion for the intro docstring wording:

```
This method reads data that can be interpreted using three coordinates: event,
        latitude, and longitude. The three coordinates to be read can be
        specified via the ``coordinate_vars`` parameter. The data and the coordinates themselves may be organized
        in arbitrary dimensions in the Dataset (e.g. four dimensions 'year', 'month',
        'day' and 'altitude' for the coordinate 'event').  See Notes and Examples if you
        want to load single-event data that does not contain an event dimension.
```

chahank:
Excellent work! Ready to merge upon consideration of the suggested docstring changes.

peanutfun:
@chahank Please have another look at https://github.com/CLIMADA-project/climada_python/pull/837/commits/eeda6b3719f47a748df64139dcab4acc8e2f9b16 and merge if you concur.

climada/hazard/base.py
```diff
@@ -553,13 +556,16 @@ def from_xarray_raster(
           and Examples) before loading the Dataset as Hazard.
         * Single-valued data for variables ``frequency``. ``event_name``, and
           ``event_date`` will be broadcast to every event.
+        * The ``event`` coordinate may take arbitrary values. In case these values
+          cannot be interpreted as dates or date ordinals, the default values for
+          ``Hazard.date`` and ``Hazard.event_name`` will not be useful, see the
+          ``data_vars``` parameter documentation above.
```
chahank, line 562:
```suggestion
        * The ``event`` coordinate may take arbitrary values. In case these values
          cannot be interpreted as dates or date ordinals, the default values for
          ``Hazard.date`` and ``Hazard.event_name`` are used, see the
          ``data_vars``` parameter documentation above.
```

PR #839, opened 2024-01-16
trop_cyclone: fix for tracks crossing antimeridian

### LINKED ISSUE #831: Persistent Wind Intensity Patterns in Historical Typhoon Tracks Beyond 180 Degrees Longitude
**Issue Description:**

I'm encountering consistent patterns while plotting wind intensity for historical typhoons that occurred in Fiji using CLIMADA v4.0.1. I'm using the TropCyclone class from climada.hazard and extracting tracks from specific Fiji historical typhoons via `TCTracks.from_ibtracs_netcdf()`.

**Problem Statement:**

The issue arises consistently when dealing with typhoon tracks where longitudes exceed 180 degrees. Despite the typhoons occurring in different years, the wind intensity patterns remain unchanged in these regions. Notably, this occurs when analyzing typhoons such as Tomas in 2010, Evan in 2012, and Yasa in 2020.

**Steps to Reproduce:**

- Use `TCTracks.from_ibtracs_netcdf(storm_id)` to extract tracks for historical Fiji typhoons. For instance, consider typhoons Yasa (id = 2020346S13168), Tomas (id = 2010069S12188), and Evan (id = 2012346S14180) as examples for track extraction.
- Employ the `TropCyclone.from_tracks(track, centroids, store_windfields=True)` function to derive windfields for each historical typhoon track. Specify centroids to locate the area of interest (Fiji). While optional, defining centroids allows focused analysis on the targeted region (Fiji) instead of the entire globe during windfield plotting.
- Generate wind intensity plots based on the extracted tracks using `TropCyclone.from_tracks().plot_intensity(storm_id)`.
- Observe and analyze wind intensity patterns specifically in regions where longitudes surpass 180 degrees.

Sample figures depicting wind intensity patterns for typhoons Tomas 2010, Evan 2012, and Yasa 2020 are attached below.
The issue persists across different typhoon events despite variations in their occurrence years.

**Attached Figures:**

evan_raw_windfield
tomas_raw_windfield
yasa_raw_windfield

Your assistance in addressing this issue would be greatly appreciated.

Thank you.

tovogt:
Changes proposed in this PR:
- Implement a new unit test for `TropCyclone` to cover the case where a TC track crosses the antimeridian. Note that this test fails in the current develop branch!
- Fix the code of `TropCyclone` to make the new test pass. This mostly affects the helper function `get_close_centroids` and surrounding code. Note that this function is also used in petals, and there is a compatibility PR here: https://github.com/CLIMADA-project/climada_petals/pull/105
- Change variable names in `trop_cyclone.py` to make binary numpy arrays start with `mask_` and arrays of indices with `idx_`. I found the previous nomenclature confusing while working on this fix.
- Implement a new unit test for `util.coordinates.latlon_bounds` to cover the case where the specified buffer is very large so that the bounds cover more than the full longitudinal range [-180, 180].

This PR does NOT fix #831!

### PR Author Checklist

### PR Reviewer Checklist


climada/hazard/trop_cyclone.py
```diff
@@ -270,13 +270,15 @@ def from_tracks(
 
         if ignore_distance_to_coast:
             # Select centroids with lat <= max_latitude
-            coastal_idx = (np.abs(centroids.lat) <= max_latitude).nonzero()[0]
+            [idx_centr_filter] = (np.abs(centroids.lat) <= max_latitude).nonzero()
```
tovogt, line 273: The variable name used to be misleading because it was describing a filtered list of centroids, and filtering was not only done based on the property of being a "coastal" centroid or not. The "coastal" criterion was only one of several (optional) criteria.

climada/hazard/trop_cyclone.py
```diff
@@ -740,50 +743,31 @@ def _compute_windfields_sparse(
...
-    max_dist_eye_lon = max_dist_eye_km / (
-        u_const.ONE_LAT_KM * np.cos(np.radians(np.abs(coastal_centr[:, 0]) + max_dist_eye_lat))
-    )
-    coastal_idx = coastal_idx[
-        (t_lat.min() - coastal_centr[:, 0] <= max_dist_eye_lat)
-        & (coastal_centr[:, 0] - t_lat.max() <= max_dist_eye_lat)
-        & (t_lon.min() - coastal_centr[:, 1] <= max_dist_eye_lon)
-        & (coastal_centr[:, 1] - t_lon.max() <= max_dist_eye_lon)
-    ]
-    coastal_centr = centroids.coord[coastal_idx]
```
tovogt, line 760: The lines above are now moved to the `get_close_centroids` helper function. This has the advantage that this functionality does not need to be re-implemented in `TCRain` (in CLIMADA petals) and we can cover this part of the functionality with our unit tests more easily.

climada/hazard/trop_cyclone.py
```diff
@@ -792,35 +776,35 @@ def _compute_windfields_sparse(
...
+        indptr = np.arange(npositions + 1) * n_centr_filter * 2
         windfields_sparse = sparse.csr_matrix((windfields.ravel(), indices, indptr),
                                               shape=windfields_shape)
         windfields_sparse.eliminate_zeros()
 
     return intensity_sparse, windfields_sparse
 
 def _compute_windfields_sparse_chunked(
-    track_centr_msk: np.ndarray,
+    mask_close_alongtrack: np.ndarray,
```
tovogt, line 807: I know that this renaming is not clarifying everything perfectly. But at least the naming is more consistent within `trop_cyclone.py` regarding the aspect of using the `mask_` prefix (instead of a `_msk` suffix) to indicate that this is a bool array.

climada/hazard/trop_cyclone.py
```diff
@@ -901,8 +885,9 @@ def _compute_windfields(
         Output of `tctrack_to_si`. Which data variables are used in the computation of the wind
         speeds depends on the selected model.
     centroids : np.ndarray with two dimensions
-        Each row is a centroid [lat, lon].
-        Centroids that are not within reach of the track are ignored.
+        Each row is a centroid [lat, lon]. Centroids that are not within reach of the track are
+        ignored. Longitudinal coordinates are assumed to be normalized consistently with the
+        longitudinal coordinates in `si_track`.
```
tovogt, line 890: This was the reason for the issue that we are fixing here: This function assumes that the lon coordinates are normalized consistently with the lon coordinates in `si_track` - but they were not, in some cases!

climada/hazard/trop_cyclone.py
```diff
@@ -1128,29 +1112,27 @@ def compute_angular_windspeeds(si_track, d_centr, close_centr_msk, model, cyclos
     return result
 
 def get_close_centroids(
-    t_lat: np.ndarray,
-    t_lon: np.ndarray,
+    si_track: xr.Dataset,
```
tovogt, line 1115: The advantage of passing the whole track dataset instead of only the lat/lon coordinates is that the track dataset comes with an attribute  `"mid_lon"` that specifies the normalization of the longitudinal coordinates.

climada/hazard/trop_cyclone.py
```diff
@@ -1373,7 +1397,7 @@ def _B_holland_1980(si_track: xr.Dataset):  # pylint: disable=invalid-name
 def _x_holland_2010(
     si_track: xr.Dataset,
     d_centr: np.ndarray,
-    close_centr: np.ndarray,
+    mask_centr_close: np.ndarray,
```
tovogt, line 1400: The variable name `close_centr` is used in many places in `trop_cyclone.py`, and I found it confusing because the name sounds like the data is describing centroids (i.e. coordinates of points), but instead, it's just a mask (a bool array) over the set of centroids. The renaming should clarify this.

climada/util/coordinates.py
```diff
@@ -195,9 +198,11 @@ def lon_bounds(lon, buffer=0.0):
     gap_max = np.argmax(lon_diff)
     lon_diff_max = lon_diff[gap_max]
     if lon_diff_max < 2:
-        # looks like the data covers the whole range [-180, 180] rather evenly
-        lon_min = max(lon_uniq[0] - buffer, -180)
-        lon_max = min(lon_uniq[-2] + buffer, 180)
+        # since the largest gap is comparably small, enforce the [-180, 180] value range
+        gap_max = lon_diff.size - 1
+    if lon_diff_max <= 2 * buffer:
```
tovogt, line 203: We are now explicitly handling exceedingly large `buffer` sizes. This fixes a bug in this helper function (see unit test).

climada/util/test/test_coordinates.py
```diff
@@ -248,6 +248,11 @@ def test_latlon_bounds(self):
         bounds = u_coord.latlon_bounds(lat, lon, buffer=1)
         self.assertEqual(bounds, (179.5, -3.1, 186, 6))
 
+        # exceedingly large buffers are truncated to [-180, 180]
+        lat, lon = np.array([0, -2.1, 5]), np.array([-179.5, -175, -178])
+        bounds = u_coord.latlon_bounds(lat, lon, buffer=234)
+        self.assertEqual(bounds, (-180, -90, 180, 90))
```
tovogt, line 254: Before this PR, this would fail because the buffer is so large that adding it both to the right and to the left of the bounds would actually cover more than a full circle.

tovogt (commented):
See below for comments that explain some of the changes directly in the code.

peanutfun:
GitHub tests pass, but Jenkins tests fail with
```
climada.hazard.test.test_trop_cyclone.TestReader.test_cross_antimeridian

ValueError: The following given IDs are not in IBTrACS: 2020346S13168.
```

emanuel-schmid:
The file is updated, and hence the test climada_branches passes, but the petals_compatibility test still fails with
`TypeError: get_close_centroids() got multiple values for argument 'metric'`

tovogt:
> The file is updated, and hence the test climada_branches passes, but the petals_compatibility test still fails with `TypeError: get_close_centroids() got multiple values for argument 'metric'`

Good point, that means that the PR https://github.com/CLIMADA-project/climada_petals/pull/105 is actually a compatibility PR and needs to be merged at the same time with this one.

climada/hazard/trop_cyclone.py
```diff
@@ -1128,29 +1112,27 @@ def compute_angular_windspeeds(si_track, d_centr, close_centr_msk, model, cyclos
...
+    impossible, but might happen with synthetical or test data.
 
     Parameters
     ----------
-    t_lat : np.ndarray of shape (npositions,)
-        Latitudinal coordinates of track positions.
-    t_lon : np.ndarray of shape (npositions,)
-        Longitudinal coordinates of track positions, normalized around a central longitude.
+    si_track : xr.Dataset
+        Track information as returned by `tctrack_to_si`. Hence, longitudinal coordinates are
```
peanutfun, line 1129: docstring formatting
```suggestion
        Track information as returned by ``tctrack_to_si``. Hence, longitudinal coordinates are
```

climada/hazard/trop_cyclone.py
```diff
@@ -1128,29 +1112,27 @@ def compute_angular_windspeeds(si_track, d_centr, close_centr_msk, model, cyclos
...
+        normalized around a central longitude. This makes sure that the buffered bounding box
+        around the track doesn't cross the antimeridian. The data variables used by this function
+        are "lat", and "lon".
```
peanutfun, line 1132:
```suggestion
        are "lat" and "lon" (with length "npositions", see return values below),
        "time", and the "mid_lon" attribute indicating the longitude normalization.
```

climada/hazard/trop_cyclone.py
```diff
@@ -740,50 +743,31 @@ def _compute_windfields_sparse(
...
 
-    # After the previous filtering step, finding and storing the reachable centroids is not a
-    # memory bottle neck and can be done before chunking.
-    track_centr_msk = get_close_centroids(
-        t_lat, t_lon, coastal_centr, max_dist_eye_km, metric=metric,
+    # when done properly, finding and storing the close centroids is not a memory bottle neck and
+    # can be done before chunking:
+    centroids_close, mask_close, mask_close_alongtrack = get_close_centroids(
+        si_track, centroids.coord[idx_centr_filter], max_dist_eye_km, metric=metric,
     )
```
peanutfun, line 751: Why would this function return both the close centroids and the mask for the close centroids? I see no reason why the "closeness" information should be carried around twice. What am I missing?

climada/hazard/trop_cyclone.py
peanutfun: Maybe make clear that this is a mask for the centroids?
```suggestion
    mask_centr_along_track: np.ndarray,
```
Also, why does this take the centroids and a mask separately, and not just the masked centroids?

peanutfun (changes requested):
Great to see that this now works, thanks for adding the necessary unit tests! 🙌 

I am a bit confused about functions returning both centroids and masks for centroids. I think they should return only one.

I agree with most of the naming changes, and will simply resolve the respective conversations 😇

climada/hazard/trop_cyclone.py
tovogt: The `_alongtrack` part means that it's not just a single mask, but it's an "along-track mask", meaning that it is actually one mask for each track position. That's why the array is not one-dimensional, but two-dimensional. This is already made clear from the docstrings:
https://github.com/CLIMADA-project/climada_python/blob/83cff963a9d1f2311cfd16122a9e0b0dd54334d3/climada/hazard/trop_cyclone.py#L817-L818

climada/hazard/trop_cyclone.py
tovogt: The lon-coordinates of the returned centroids are normalized. We could of course apply the mask, and then normalize outside of the function (again). But that was the approach that caused the original bug that is fixed with this PR: We normalized in several places, and then in one place we forgot to normalize and everything silently "failed". 

On the other hand, we cannot just return the centroids without the mask, because other parts of the code need to know which centroids from the original array are the ones that we mask out.

tovogt:
Thanks for the review! Yes, understanding masks can be challenging in this module. For performance reasons, we have a mask for each track position, but at some point we apply a mask to the centroids to filter out centroids that are not affected by any of the track positions. But we need the masks to stick around in order to inflate the result (that is only defined on the masked centroids) to the original set of centroids in the end.

climada/hazard/trop_cyclone.py
peanutfun: Indeed, I missed the fact that centroids are 1D, but the masks are 2D. Very good! Thanks for the clarification both here and in the code.

peanutfun (approved):

peanutfun:
@tovogt The double-backticks are very annoying and easy to miss 😅

From my point of view, this is ready! Can I merge?

tovogt:
Wait a second, CHANGELOG...

And let's also merge https://github.com/CLIMADA-project/climada_petals/pull/105 soonish due to compatibility.

tovogt:
Okay, ready to be merged!

tovogt:
@emanuel-schmid Thanks for fixing the CHANGELOG! I should have merged develop before adding the new points...

peanutfun:
@tovogt @emanuel-schmid Nice work, thank you!

PR #851, opened 2024-02-20
update to xarray 2024.2

### LINKED ISSUE #852: xarray 2024.2 breaks unit test
climada.hazard.test.test_base_xarray.TestReadDefaultNetCDF.test_event_no_time fails after update from 2024.1

emanuel-schmid:
Changes proposed in this PR:

- xarray 2024.2 introduced an AttributeError in `year_month_day_accessor` of `Hazard.from_xarray_raster` where: the `array` argument is not reallly a date-time array. This PR fixes the problem by simply adding the Exception type to the list of caught Exceptions.
- However, the update to xarray 2024.2 also seems to slow down the procedure significantly. On a personal computer the test `test_base_xarray.TestReadDefaultNetCDF.test_event_no_time` took between 6 and 8 seconds where it finished in 1 to 2 seconds with version 2024.1

This PR fixes #852

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -840,7 +840,7 @@ class Hazard():
                 return array.dt.strftime("%Y-%m-%d").values
 
             # Handle access errors
-            except (ValueError, TypeError) as err:
+            except (ValueError, TypeError, AttributeError) as err:
                 if strict:
                     raise err
 
--- a/requirements/env_climada.yml
+++ b/requirements/env_climada.yml
@@ -38,6 +38,6 @@ dependencies:
   - tabulate>=0.9
   - tqdm>=4.66
   - unittest-xml-reporting>=3.2
-  - xarray>=2024.1,<2024.2  # 2024.2 introduced an unhandled AttributeError in Hazard.from_xarray_raster
+  - xarray>=2024.2
   - xlrd>=2.0
   - xlsxwriter>=3.1


requirements/env_climada.yml
```diff
@@ -38,6 +38,6 @@ dependencies:
   - tabulate>=0.9
   - tqdm>=4.66
   - unittest-xml-reporting>=3.2
-  - xarray>=2024.1,<2024.2  # 2024.2 introduced an unhandled AttributeError in Hazard.from_xarray_raster
+  - xarray>=2024.2
```
peanutfun, line 41: Why raise to 2024.2? We are now compatible to it but we should still be compatible to 2024.1 🤔 
```suggestion
  - xarray>=2024.1
```

climada/hazard/base.py
```diff
@@ -840,7 +840,7 @@ def year_month_day_accessor(
                 return array.dt.strftime("%Y-%m-%d").values
 
             # Handle access errors
-            except (ValueError, TypeError) as err:
+            except (ValueError, TypeError, AttributeError) as err:
```
peanutfun, line 843: This is correct and expected. From the v2024.2 changelog:
> DataArray.dt now raises an AttributeError rather than a TypeError when the data isn’t datetime-like.
Thanks for the fix!

requirements/env_climada.yml
emanuel-schmid: true but sooner or later the version will be updated to whatever is current
and this would make it easier to adapt to the new xarray. i suspect it to slow down "something" and i'm not sure whether we want that. 🤔

peanutfun: Sure, we have to adapt to later versions, and this is what we do here. Also, the GitHub pipeline always tries to install the latest possible version. I am just a bit hesitant to immediately block any older versions. But go ahead! If something breaks, we can revert.

peanutfun (approved):

tovogt (approved):
The PR looks fine to me, in general. However, the fact that version 2024.2 slows down the unit test by a factor 4 is quite suspicious. I would suggest to merge this, but to create a separate issue that makes note of the slowdown. I would suspect that it's possible to avoid the slowdown somehow, but I don't have time to investigate this any further.

PR #855, opened 2024-02-23
Make GitHub CI execute unit tests in Petals via reusable workflow

peanutfun:
Changes proposed in this PR:
- Call the resuable workflow introduced in https://github.com/CLIMADA-project/climada_petals/pull/114 to run compatibility tests for Petals from Core

### PR Author Checklist

### PR Reviewer Checklist


.github/workflows/ci.yml
```diff
@@ -59,12 +57,22 @@ jobs:
...
+          name: coverage-report-core-unittests-py${{ matrix.python-version }}
           path: coverage/
+
+  test-petals:
+    name: Petals Compatibility
+    uses: CLIMADA-project/climada_petals/.github/workflows/testing.yml@github-actions
+    needs: build-and-test
+    with:
+      core_branch: ${{ github.ref }}
+      petals_branch: github-actions  # TODO: Change this to 'develop' once Petals PR is merged!
```
peanutfun, line 76:
```suggestion
      petals_branch: develop
```

.github/workflows/ci.yml
```diff
@@ -59,12 +57,22 @@ jobs:
         if: always()
         with:
           junit_files: tests_xml/tests.xml
-          check_name: "Unit Test Results Python ${{ matrix.python-version }}"
+          check_name: "Core / Unit Test Results (${{ matrix.python-version }})"
```
chahank, line 60: I am not sure to understand now what is the difference between the test

`GitHub CI / Core / Unit Tests`

and 

`GitHub CI / Core / Unit Test Results`

Other reviewers might also be confused to see these separate tests.

peanutfun: I named them similar on purpose because one is simply the result report of the other. I agree that it is a bit confusing that the report appears as additional "Check". I can disable that. However, the report and check also include Annotations of the failing tests, see https://github.com/CLIMADA-project/climada_python/actions/runs/8817066788 for an example. These allow to quickly hop to the failing code and also appear in the PR "Changes" overview. As far as I can see, Annotations are only available if we keep the Check. See [the docs](https://github.com/marketplace/actions/publish-test-results#commit-and-pull-request-checks) of the associated GitHub action.

If we disable the additional check, one will have to look into the [full test report](https://github.com/CLIMADA-project/climada_python/actions/runs/8817066788/job/24202578024) to find out which code is failing. That is likely more confusing.

If it's just an issue with naming, I welcome suggestions.

peanutfun: I don't really see how "Unit Test Summary Report" and "Unit Test Detailed Report" are more clear than "Unit Tests" and "Unit Test Results" 😅 How about "Unit Testing" and "Unit Test Report"?

chahank: Well, the point is that both cases are reports of test results, just in different formats. So, neither the words "report", "results" or "test" help differentiate both cases. 

Anyway, most people will probably be scared by the test reports now that we have 18 of them. So naming will not change much.

peanutfun: > Well, the point is that both cases are reports of test results, just in different formats. 

But the thing you wanted to call "Detailed Report" is actually the output of the entire testing pipeline. It contains much more than a test result or report or whatever you want to call it 😅

peanutfun: > How about "Unit Testing" and "Unit Test Report"?

Do we go with this or do we leave it as is?

chahank: As you prefer. I think `Unit Testing` and `Unit Test Report` are not super clear, but it probably does not matter. @emanuel-schmid may have a better idea, otherwise, let's keep it as you propose.

peanutfun: We now have `Unit Test Pipeline` and `Unit Test Report`

peanutfun:
Petals Compatibility tests are failing because we are testing against Petal's `develop`. Should be fixed by switching to the centroids branch or by waiting until https://github.com/CLIMADA-project/climada_petals/pull/122 is merged.

PR #869, opened 2024-04-10
Catch case where new Hazard fraction is unintentionally zero

### LINKED ISSUE #866: Handle the case where `Hazard.select` creates a selection where the entire fraction matrix is zero
Throw an error or a warning in this case. This will not actually handle the underlying issue: If subselecting a fraction matrix to values that are only zero, the fraction will suddenly be considered a fraction matrix of only ones (implicitly).

```python
import numpy as np
from scipy.sparse import csr_matrix
from climada.hazard import Hazard, Centroids

centroids = Centroids(lat=np.zeros(2), lon=np.zeros(2), region_id=np.array([0, 1]))

hazard = Hazard(
    centroids=centroids,
    event_id=np.array([1, 2]),
    intensity=csr_matrix([[1, 1], [1, 1]]),
    fraction=csr_matrix([[0, 0], [1, 1]]),
)

hazard._get_fraction().toarray()
# array([[0, 0],
#        [1, 1]])

hazard.select(reg_id=[0])._get_fraction().toarray()
# array([[0],
#        [1]])

hazard.select(event_id=[1])._get_fraction()
# Returns None, equivalent to fraction of 1 everywhere, but should be
# array([[0, 0]])
```

First reported in #838

peanutfun:
Changes proposed in this PR:
- Throw an error when selecting a part of the hazard where the fraction is zero everywhere, but only if it has had non-zero entries before.

This PR fixes #866

### PR Author Checklist

### PR Reviewer Checklist


climada/hazard/base.py
```diff
@@ -1457,6 +1457,18 @@ def select(self, event_names=None, event_id=None, date=None, orig=None,
...
+        # Check if new fraction is zero everywhere
+        if self.fraction.nnz != 0 and haz.fraction.nnz == 0:
+            raise RuntimeError(
+                "Your selection created a Hazard object where the fraction matrix is "
+                "zero everywhere. This hazard will have zero impact everywhere. "
+                "We are catching this condition because of an implementation detail: "
+                "A fraction matrix without nonzero-valued entries will be completely "
+                "ignored. This is surely not what you intended. If you really want to, "
+                "you can circumvent this error by setting your original fraction "
+                "matrix to zero everywhere, but there probably is no point in doing so."
```
chahank, line 1469: Very pedagogical, I like it!

climada/hazard/base.py
```diff
@@ -1457,6 +1457,18 @@ def select(self, event_names=None, event_id=None, date=None, orig=None,
                                    dt.datetime.fromordinal(haz.date.min()).year) + 1
             haz.frequency = haz.frequency * year_span_old / year_span_new
 
+        # Check if new fraction is zero everywhere
+        if self.fraction.nnz != 0 and haz.fraction.nnz == 0:
```
chahank, line 1461: Just a small efficiency comment: `crs_matrix.nnz` is equivalent to `np.count_nonzero(a.toarray())`. Thus, it first projects to an array and then counts. This is rather costly for large matrices, and here is also not needed. It is much faster to check `self.fraction.size != 0` and `haz.fraction.size == 0`. The only downside is IF the fraction matrix is not well handled (i.e., some explicit 0s are present).

peanutfun: > crs_matrix.nnz is equivalent to np.count_nonzero(a.toarray())

This is incorrect. `nnz` returns the number of explicitly stored values, including explicitly stored zeros. `np.count_nonzero(a.toarray())` would count all zeros in the assembled dense matrix.

> Thus, it first projects to an array and then counts.

As far as I can see in the scipy source code, this is also incorrect. Where do you take this information from?

> The only downside is IF the fraction matrix is not well handled (i.e., some explicit 0s are present).

This is indeed correct. We can call `csr_matrix.eliminate_zeros` during `select` to make sure this does not happen.

chahank: ooo you are right, sorry, I misread one element of the documentation, sorry! I was so confused now trying to find where `.nnz` is defined exactly. In the base class of sparse, `.size` and `.nnz` are identical. Anyway, it seems I was mistaken. I thought `.nzz` is an inefficient way of checking whether a matrix is empty or not. But that was wrong. Sorry for the confusion!


> This is indeed correct. We can call csr_matrix.eliminate_zeros during select to make sure this does not happen.

Not sure if we need to do that since a proper climada hazard should not contain explicit 0s... but probably it does not hurt. Then we have to do it for intensity and for fraction right?

chahank, line 1461: Another more sensible possible alternative that is potentially easier to read:

```suggestion
        if self.get_fraction() is not None and haz.fraction is None:
```

peanutfun: We technically only need it for `fraction` here, because the values of `intensity` do not matter during selection 🤷

peanutfun: Agreed! One last issue with this: `_get_fraction()` is marked private for some reason, and this will cause an additional linter warning.

peanutfun: After personal discussion: Go with `_get_fraction` although the linter will complain 😵

peanutfun: After personal discussion: Go with `_get_fraction` although the linter will complain 😵

peanutfun: When implementing this for `fraction`, I noticed that the csr matrix automatically eliminates zeros when you do "regular" value assignment. Only when manipulating the attributes directly, the matrix might end up in an inconsistent state. So, I think there is no action needed for `intensity`, as few users will do such a thing.

Example:
```python
fraction[1, 2] = 0  # This zero will *not* be stored explicitly

fraction.data[0] = 0  # This zero *will* be stored explicitly, therefore...
fraction.eliminate_zeros()  # ...is needed.
```

chahank: Good to know!

chahank:
There remains an error with the compatibility with Petals, but I think this is unrelated. Good to merge!

peanutfun:
Yes, the error is related to the centroids PRs. Will merge.

emanuel-schmid:
Very well. Many thanks! 🙌

PR #878, opened 2024-04-30
Add polynomial s shape impact function

chahank:
Changes proposed in this PR:
- 

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/entity/impact_funcs/base.py
+++ b/climada/entity/impact_funcs/base.py
@@ -273,3 +273,70 @@ class ImpactFunc():
         LOGGER.warning("The use of ImpactFunc.set_sigmoid_impf is deprecated."
                        "Use ImpactFunc.from_sigmoid_impf instead.")
         self.__dict__ = ImpactFunc.from_sigmoid_impf(*args, **kwargs).__dict__
+
+    @classmethod
+    def from_poly_s_shape(
+        cls,
+        intensity: tuple[float, float, float],
+        threshold: float,
+        half_point: float,
+        upper_limit: float,
+        exponent: float,
+        haz_type: str,
+        impf_id: int = 1,
+        **kwargs):
+        """S-shape polynomial impact function hinging on four parameter.
+
+        f(I) = D(I)**exponent / (1 + D(I)**exponent) * upper_limit
+
+        D(I) = max[I - threshold, 0] / (half_point - threshold)
+
+        This function is inspired by Emanuel et al. (2011)
+        https://doi.org/10.1175/WCAS-D-11-00007.1
+
+        Parameters
+        ----------
+        intensity: tuple(float, float, float)
+            tuple of 3 intensity numbers along np.arange(min, max, step)
+        threshold : float
+            Intensity threshold below which there is no impact.
+            Should in general be larger than 0 for computational efficiency
+            of impacts.
+        half_point : float
+            Intensity at which 50% of maxixmum impact is expected.
+            Must be larger than thres.
+        upper_limit : float
+            Maximum impact value. Must be larger than 0.
+        exponent: float
+            Exponent of the polynomial. Must be larger than 0.
+        haz_type: str
+            Reference string for the hazard (e.g., 'TC', 'RF', 'WS', ...)
+        impf_id : int, optional, default=1
+            Impact function id
+        kwargs :
+            keyword arguments passed to ImpactFunc()
+
+        Returns
+        -------
+        impf : climada.entity.impact_funcs.ImpactFunc
+            s-shapep polynomial impact function
+        """
+        intensity = np.arange(*intensity)
+
+        if threshold > half_point:
+            mdd = np.zeros_like(intensity)
+        else:
+            D = (intensity - threshold) / (half_point - threshold)
+            D[D < 0] = 0
+            mdd = D**exponent / (1 + D**exponent) * upper_limit
+        paa = np.ones_like(intensity)
+
+        impf = cls(
+            haz_type=haz_type,
+            id=impf_id,
+            intensity=intensity,
+            paa=paa,
+            mdd=mdd,
+            **kwargs
+        )
+        return impf
--- a/climada/entity/impact_funcs/test/test_base.py
+++ b/climada/entity/impact_funcs/test/test_base.py
@@ -47,7 +47,6 @@ class TestInterpolation(unittest.TestCase):
         self.assertEqual(imp_fun.haz_type, 'TC')
         self.assertEqual(imp_fun.id, 2)
 
-
     def test_from_sigmoid(self):
         """Check default impact function: sigmoid function"""
         inten = (0, 100, 5)
@@ -59,6 +58,34 @@ class TestInterpolation(unittest.TestCase):
         self.assertTrue(np.array_equal(imp_fun.intensity, np.arange(0, 100, 5)))
         self.assertEqual(imp_fun.haz_type, 'RF')
         self.assertEqual(imp_fun.id, 2)
+
+    def test_from_poly_s_shape(self):
+        """Check default impact function: polynomial s-shape"""
+        inten = (0, 5, 1)
+        impf = ImpactFunc.from_poly_s_shape(
+            inten, threshold=0.2, half_point=1, upper_limit=0.8, exponent=4,
+            haz_type='RF', impf_id=2, intensity_unit='m'
+            )
+        correct_mdd = np.array([0, 0.4, 0.76995746, 0.79470418, 0.79843158])
+        self.assertTrue(np.array_equal(impf.paa, np.ones(5)))
+        np.testing.assert_array_almost_equal(impf.mdd, correct_mdd)
+        self.assertTrue(np.array_equal(impf.intensity, np.arange(0, 5, 1)))
+        self.assertEqual(impf.haz_type, 'RF')
+        self.assertEqual(impf.id, 2)
+        self.assertEqual(impf.intensity_unit, 'm')
+
+        # If threshold > half_point, mdd should all be 0
+        impf = ImpactFunc.from_poly_s_shape(
+            inten, threshold=2, half_point=1, upper_limit=0.8, exponent=4,
+            haz_type='RF', impf_id=2, intensity_unit='m'
+            )
+        correct_mdd = np.array([0, 0.4, 0.76995746, 0.79470418, 0.79843158])
+        self.assertTrue(np.array_equal(impf.paa, np.ones(5)))
+        np.testing.assert_array_almost_equal(impf.mdd, np.zeros(5))
+        self.assertTrue(np.array_equal(impf.intensity, np.arange(0, 5, 1)))
+        self.assertEqual(impf.haz_type, 'RF')
+        self.assertEqual(impf.id, 2)
+        self.assertEqual(impf.intensity_unit, 'm')
 
 # Execute Tests
 if __name__ == "__main__":


climada/entity/impact_funcs/test/test_base.py
```diff
@@ -60,6 +59,34 @@ def test_from_sigmoid(self):
...
         self.assertEqual(imp_fun.id, 2)
 
+    def test_from_poly_s_shape(self):
+        """Check default impact function: polynomial s-shape"""
+        inten = (0, 5, 1)
+        impf = ImpactFunc.from_poly_s_shape(
+            inten, threshold=0.2, half_point=1, upper_limit=0.8, exponent=4,
+            haz_type='RF', impf_id=2, intensity_unit='m'
+            )
+        correct_mdd = np.array([0, 0.4, 0.76995746, 0.79470418, 0.79843158])
```
peanutfun, line 69: It's not clear why these are the correct values. You should check that the function is
* zero below the threshold
* zero at the threshold
* 0.5x the upper limit at the half_point *(this is covered)*
* (very) close to the upper limit somewhere 10x half_point
* behaving as expected when threshold < 0, half_point < threshold, upper_limit < 0, exponent < 0.

climada/entity/impact_funcs/test/test_base.py
```diff
@@ -60,6 +59,34 @@ def test_from_sigmoid(self):
...
+            inten, threshold=2, half_point=1, upper_limit=0.8, exponent=4,
+            haz_type='RF', impf_id=2, intensity_unit='m'
+            )
+        correct_mdd = np.array([0, 0.4, 0.76995746, 0.79470418, 0.79843158])
+        self.assertTrue(np.array_equal(impf.paa, np.ones(5)))
+        np.testing.assert_array_almost_equal(impf.mdd, np.zeros(5))
+        self.assertTrue(np.array_equal(impf.intensity, np.arange(0, 5, 1)))
+        self.assertEqual(impf.haz_type, 'RF')
+        self.assertEqual(impf.id, 2)
+        self.assertEqual(impf.intensity_unit, 'm')
```
peanutfun, line 88: No need to test this again. If you really want to, please write a function instead of copy-pasting.

peanutfun (changes requested):
Thanks for the addition! I think the function will be very useful.

Please improve the documentation and the tests. One linter issue can also easily be fixed.

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        upper_limit: float,
+        exponent: float,
+        haz_type: str,
+        impf_id: int = 1,
+        **kwargs):
+        """S-shape polynomial impact function hinging on four parameter.
+
+        f(I) = D(I)**exponent / (1 + D(I)**exponent) * upper_limit
+
+        D(I) = max[I - threshold, 0] / (half_point - threshold)
```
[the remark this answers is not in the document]
chahank: Thanks!

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+
+        This function is inspired by Emanuel et al. (2011)
+        https://doi.org/10.1175/WCAS-D-11-00007.1
+
+        Parameters
+        ----------
+        intensity: tuple(float, float, float)
+            tuple of 3 intensity numbers along np.arange(min, max, step)
```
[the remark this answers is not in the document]
chahank: Ok, good to know! It does not matter in the case of the impact function in most cases, but good to improve, thanks!

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        threshold : float
+            Intensity threshold below which there is no impact.
+            Should in general be larger than 0 for computational efficiency
+            of impacts.
+        half_point : float
+            Intensity at which 50% of maxixmum impact is expected.
+            Must be larger than thres.
```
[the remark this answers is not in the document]
chahank: Aaa yes, I had it first throw an error. Good catch.

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        upper_limit : float
+            Maximum impact value. Must be larger than 0.
```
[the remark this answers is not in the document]
chahank: Good point. But then it should maybe also be called `scale` parameter or something as it would become the lower limit ^^.

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        exponent: float
+            Exponent of the polynomial. Must be larger than 0.
```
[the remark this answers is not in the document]
chahank: Indeed, must is again too strong.

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,70 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        -------
+        impf : climada.entity.impact_funcs.ImpactFunc
+            s-shapep polynomial impact function
+        """
+        intensity = np.arange(*intensity)
+
+        if threshold > half_point:
+            mdd = np.zeros_like(intensity)
+        else:
+            D = (intensity - threshold) / (half_point - threshold)
```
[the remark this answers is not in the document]
chahank: For math formula I think single-letter variables are much clearer. Bad linter! But can do.

climada/entity/impact_funcs/test/test_base.py
chahank: In this case, I opted for a direct check of the values as one can easily recompute by hand what the correct values are. I think the checks you propose are less good as they do not directly test the mathematical formula. I can add some tests for the edge cases, although I think this is a clear overkill here.

climada/entity/impact_funcs/test/test_base.py
chahank: I think it is good to test that the object remains consistent with the edge case.

climada/entity/impact_funcs/base.py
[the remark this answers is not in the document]
chahank: I made the exponent to be positive. The function is senseless otherwise. Negative exponents now raises an error.

climada/entity/impact_funcs/test/test_base.py
chahank: Done.

climada/entity/impact_funcs/base.py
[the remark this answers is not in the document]
chahank: The mathmode seems not to compile correctly. This was also already the case for the existing method `from_sigmoid_impf`. I will see if I can find the reason.

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +273,74 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        scale: float,
+        exponent: float,
+        haz_type: str,
+        impf_id: int = 1,
+        **kwargs):
+        """S-shape polynomial impact function hinging on four parameter.
+
+        .. math:: f(I) = \\frac{luk(I)**exponent}{1 + luk(I)**exponent}}\\cdot scale
+
+        .. math:: luk(I) = \\frac{max[I - threshold, 0]}{half_point - threshold}
```
peanutfun, line 292: Please write valid LaTeX. You don't have to excape characters here. You can use a single math directive for both formulae. Please use `\max` and either use `\mathrm{threshold}` or a single symbol, as in any math formula.

climada/entity/impact_funcs/base.py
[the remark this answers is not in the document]
peanutfun: But an exponent of exactly zero would work, right? The function would just be constant then (thinking about calibration: It might be better to have as few restrictions as possible to avoid setting strange parameter bounds)

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +272,83 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+        exponent: float,
+        haz_type: str,
+        impf_id: int = 1,
+        **kwargs):
+        r"""S-shape polynomial impact function hinging on four parameter.
+
+        .. math::
+
+            f(I) = \frac{\textrm{luk}(I)^{\textrm{exponent}}}{1 + \textrm{luk}(I)^{\textrm{exponent}}}\cdot \textrm{scale} \\
+            \textrm{luk}(I) = \frac{\max[I - \textrm{threshold}, 0]}{\textrm{half_point} - \textrm{threshold}}
```
peanutfun, line 292: Fix line-too-long
```suggestion
            f(I) = \frac{\textrm{luk}(I)^{\textrm{exponent}}}{
                1 + \textrm{luk}(I)^{\textrm{exponent}}
            }
                \cdot \textrm{scale} \\
            \textrm{luk}(I) = \frac{\max[I - \textrm{threshold}, 0]}{
                \textrm{half_point} - \textrm{threshold}
            }
```

climada/entity/impact_funcs/base.py
```diff
@@ -273,3 +272,83 @@ def set_sigmoid_impf(self, *args, **kwargs):
...
+            raise ValueError('Exponent value must larger than 0')
+
+        inten = np.linspace(*intensity)
+
+        if threshold > half_point:
+            mdd = np.zeros_like(inten)
+        else:
+            luk = (inten - threshold) / (half_point - threshold)
+            luk[luk < 0] = 0
+            mdd = luk**exponent / (1 + luk**exponent) * scale
```
peanutfun, line 343: Clarify that `scale` is not in the denominator
```suggestion
            mdd = scale * luk**exponent / (1 + luk**exponent)
```

climada/entity/impact_funcs/test/test_base.py
```diff
@@ -60,6 +59,50 @@ def test_from_sigmoid(self):
...
+
+        # If threshold > half_point, mdd should all be 0
+        impf = ImpactFunc.from_poly_s_shape(
+            intensity=intensity, threshold=half_point*2, half_point=half_point, scale=scale,
+            exponent=exponent, haz_type=haz_type, impf_id=impf_id, intensity_unit=unit
+            )
+        np.testing.assert_array_almost_equal(impf.mdd, np.zeros(5))
+        test_aux_vars(impf)
+
+        with self.assertRaises(ValueError):
```
peanutfun, line 98: Check actual error
```suggestion
        with self.assertRaisesRegex(ValueError, "Exponent value"):
```

chahank: Thanks for the correction! Why should we priorize using `assertRaisesRegex` ?

climada/entity/impact_funcs/base.py
chahank: Also for math mode? Ok, why not.

climada/entity/impact_funcs/test/test_base.py
peanutfun: Please don't patronize the readers of your code ("can easily be computed"). Can't you find one or multiple sets of parameters where the true values can be calculated *without* a calculator?

> I can construct a function that fulfils all your tests but that does not represent the correct formula.

That is true for most tests. I agree that some tests have to be specific, but the corner cases have to be checked nonetheless.

chahank:
Ok, I fixed all that you proposed, thanks! I also fixed one test and made sure that for `threshold = half_point` the method returns `mdd=0` (else there was a divide by 0).

climada/entity/impact_funcs/test/test_base.py
peanutfun: Because it tests the actual error message. `assertRaises` would succeed with *any* `ValueError` thrown, but you want to test for a specific one here.

Note that this is basically a shortcut for the idiom we were using so far:
```python
with self.assertRaises(Exception) as cm:
    do_stuff()
self.assertIn("Text", str(cm.exception))
```

peanutfun (approved):

peanutfun:
Great work! Will merge.

climada/entity/impact_funcs/test/test_base.py
chahank: Makes total sense, thanks!

PR #886, opened 2024-05-30
Update unsequa to fully support the latest version of SALib

### LINKED ISSUE #828: Update unsequa to fully support the latest version of SALib
The unsequa module directly refers to some of the methods from the SALib package for sampling and for sensitivity analysis. In the latest version of SALib, some methods exist that are not yet documented in unsequa. 

- To resolve this, the sampling and sensitivity methods' docstrings should be updated to list all the existing methods in the SALib package. 
- Also, the compatibility dictionary `SALIB_COMPATIBILITY` in the `unc_output` module which tells the user which sampling method can be used in combination with which sensitivity method should be accordingly updated. 
- Ideally, a few tests are added to check whether the "new" methods from SALib work as in intended in unsequa.

luseverin:
Changes proposed in this PR:
- Update SALib sensitivity and sampling methods from newest version (SALib 1.4.7) [#828](https://github.com/CLIMADA-project/climada_python/issues/828)
- Allow for computation of relative and absolute delta impacts in `CalcDeltaClimate`

This PR fixes #828

### PR Author Checklist

### PR Reviewer Checklist


luseverin:
A few more comments on this PR:
- Three new sensitivity methods required a specific treatment to be included into unsequa: `ff`, `hdmr`, and `delta`
- The output of the `ff` and `hdmr` methods from SALib required refactoring in order to fit unsequa's `_sens_df` output format. I had to remove a few variables from the output of `hdmr` as I could not find a way to refactor them to the `_sens_df`. These include notably all the details of the surrogate model used, included in the `Em` (Emulator) variable.  Additionally, all output variables from `hdmr` are refactored to 2D arrays, even the first-order sensitivity indices.
- The `delta` method systematically failed with a matrix inversion error: ```
LinAlgError: Singular matrix detected
This may be due to the sample size (30000) being too small. If this is not the case, check Y values or raise an issue with the
SALib team.
``` I could not find why that was so I removed `delta` from the compatible sensitivity methods.
- I also have some doubts on the stability of the `hdmr` method as it seem to yield different results between identical model runs. I decreased the sensitivity of the `AssertEqual` tests in the unit test in consequence so that it should not be failing.

File: CHANGELOG.md [hunk 1]
```diff
@@ -152,6 +154,7 @@ Changed:
 
 - `geopandas` >=0.13 &rarr; >=0.14
 - `pandas` >=1.5,<2.0 &rarr; >=2.1
+- `salib` >=1.3.0 &rarr; >=1.4.7
```
chahank, line 157: @emanuel-schmid : why would we indicate two different larger than versions (as for geopandas)?

climada/engine/unsequa/calc_base.py
```diff
@@ -231,12 +242,20 @@ def make_sample(self, N, sampling_method='saltelli',
             'names' : param_labels,
             'bounds' : [[0, 1]]*len(param_labels)
             }
-
+        #not sure if we want to still force users to enter a value for N in the
+        #case of the 'ff' sampler. In the meantime, trigger a warning.
```
chahank, line 246: I think I would leave it as is at the moment and just have the warning.
```suggestion
        #for the ff sampler, no value of N is needed. For API consistency the user
        #must input a value that is ignored and a warning is given. 
```

climada/engine/unsequa/calc_base.py
```diff
@@ -231,12 +242,20 @@ def make_sample(self, N, sampling_method='saltelli',
...
+        if sampling_method == 'ff':
+            LOGGER.warning("""You are using the 'ff' sampler which does not require
+                           a value for 'N. The entered N value will be ignored
```
chahank, line 249:
```suggestion
                           a value for 'N'. The entered N value will be ignored
```

climada/engine/unsequa/calc_base.py
```diff
@@ -231,12 +242,20 @@ def make_sample(self, N, sampling_method='saltelli',
...
+                           in the sampling process.""")
         uniform_base_sample = self._make_uniform_base_sample(N, problem_sa,
                                                              sampling_method,
                                                              sampling_kwargs)
         df_samples = pd.DataFrame(uniform_base_sample, columns=param_labels)
         for param in list(df_samples):
+            if param not in self.distr_dict: #dummy_0 param added to uniform_base_sample
+                                             #when using ff method, need to ignore it?
+                continue
```
chahank, line 258: I think we should rather explicitly catch the case for `dummy_` variables and only ignore these instead of all possible params.

climada/engine/unsequa/calc_base.py
```diff
@@ -271,7 +290,7 @@ def _make_uniform_base_sample(self, N, problem_sa, sampling_method,
             SALib sampling method.
         sampling_method: string
             The sampling method as defined in SALib. Possible choices:
-            'saltelli', 'fast_sampler', 'latin', 'morris', 'dgsm', 'ff'
+            'dgsm', 'fast_sampler', 'ff', 'finite_diff', 'latin', 'morris', 'saltelli',
```
chahank, line 293: Idem.

climada/engine/unsequa/calc_base.py
```diff
@@ -292,8 +311,21 @@ def _make_uniform_base_sample(self, N, problem_sa, sampling_method,
         #c.f. https://stackoverflow.com/questions/2724260/why-does-pythons-import-require-fromlist
         import importlib # pylint: disable=import-outside-toplevel
         salib_sampling_method = importlib.import_module(f'SALib.sample.{sampling_method}')
-        sample_uniform = salib_sampling_method.sample(
-            problem = problem_sa, N = N, **sampling_kwargs)
+
+        if sampling_method == 'ff': #the ff sampling has a fixed sample size and
+                                    #does not require the N parameter
+            if (problem_sa['num_vars'] & (problem_sa['num_vars'] - 1) != 0):
+                raise ValueError("The number of parameters must be a power of 2 "
```
chahank, line 318:
```suggestion
                raise ValueError("The number of parameters must be a power of 2. "
```

climada/engine/unsequa/calc_base.py
```diff
@@ -323,17 +355,21 @@ def sensitivity(self, unc_output, sensitivity_method = 'sobol',
...
             Uncertainty data object in which to store the sensitivity indices
         sensitivity_method : str, optional
-            sensitivity analysis method from SALib.analyse
-            Possible choices:
-                'fast', 'rbd_fact', 'morris', 'sobol', 'delta', 'ff'
-            The default is 'sobol'.
-            Note that in Salib, sampling methods and sensitivity analysis
-            methods should be used in specific pairs.
+            Sensitivity analysis method from SALib.analyse. Possible choices: 'fast', 'rbd_fast',
+            'morris', 'sobol', 'dgsm', 'ff', 'pawn', 'rhdm', 'rsa', 'discrepancy', 'hdmr'.
```
chahank, line 359: Idem.

climada/engine/unsequa/calc_base.py
```diff
@@ -500,10 +536,44 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
...
+                #parse interaction terms of sens_indices to a square matrix
+                #to ensure consistency with unsequa
+                interaction_names = sens_indices.pop('interaction_names')
+                interactions = np.full((nparams, nparams), np.nan)
+                for i,ie in enumerate(interaction_names):
+                    interactions[param_labels.index(ie[0]),
+                                 param_labels.index(ie[1])] = sens_indices['IE'][i]
+                sens_indices['IE'] = interactions
+
+        if method.__name__ == 'SALib.analyze.hdmr':
```
chahank, line 552: idem.

climada/engine/unsequa/calc_base.py
```diff
@@ -500,10 +536,44 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
...
                                                     **sensitivity_kwargs)
+        #refactor incoherent SALib output
+        nparams = len(param_labels)
+        if method.__name__ == 'SALib.analyze.ff':
+            if sensitivity_kwargs['second_order']:
+                #parse interaction terms of sens_indices to a square matrix
+                #to ensure consistency with unsequa
+                interaction_names = sens_indices.pop('interaction_names')
+                interactions = np.full((nparams, nparams), np.nan)
+                for i,ie in enumerate(interaction_names):
```
chahank, line 547: I think you could use clearer variable names. What does ie mean? What does 'IE' mean?

climada/engine/unsequa/calc_base.py
```diff
@@ -500,10 +536,44 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
...
+                    interactions[param_labels.index(ie[0]),
+                                 param_labels.index(ie[1])] = sens_indices['IE'][i]
+                sens_indices['IE'] = interactions
+
+        if method.__name__ == 'SALib.analyze.hdmr':
+            keys_to_remove = ['Em','Term','X','Y'] #need to delete emulator as is a dict
+                                                #and other useless variables
+            [keys_to_remove.append(si)  for si, si_val_array in sens_indices.items()
+             if np.array(si_val_array).size > nparams**2]
+            sens_indices = {k: v for k, v in sens_indices.items()
```
chahank, line 557: There seems to be a space too much.

climada/engine/unsequa/calc_base.py
```diff
@@ -500,10 +536,44 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
...
+                interactions = np.full((nparams, nparams), np.nan)
+                for i,ie in enumerate(interaction_names):
+                    interactions[param_labels.index(ie[0]),
+                                 param_labels.index(ie[1])] = sens_indices['IE'][i]
+                sens_indices['IE'] = interactions
+
+        if method.__name__ == 'SALib.analyze.hdmr':
+            keys_to_remove = ['Em','Term','X','Y'] #need to delete emulator as is a dict
+                                                #and other useless variables
+            [keys_to_remove.append(si)  for si, si_val_array in sens_indices.items()
```
chahank, line 555: What is this line? It is never used.

climada/engine/unsequa/calc_base.py
```diff
@@ -515,6 +585,9 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
         sens_second_order_dict[submetric_name] = sens_second_order
 
     sens_first_order_df = pd.DataFrame(sens_first_order_dict, dtype=np.number)
+    # Assume sens_first_order_dict is a dictionary where values are lists/arrays of varying lengths
+    #sens_first_order_df = pd.DataFrame({k: pd.Series(v, dtype=object) for k, v in sens_first_order_dict.items()}) #!for some reason this make the plotting methods fail
```
chahank, line 589: Too long lines. Please break.
```suggestion
    #Assume sens_first_order_dict is a dictionary where values are lists/arrays of varying lengths
    #sens_first_order_df = pd.DataFrame({k: pd.Series(v, dtype=object) for k, v in sens_first_order_dict.items()}) #!for some reason this make the plotting methods fail
```

chahank: Why is this line commented?

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -614,6 +715,75 @@ def test_calc_sensitivity_morris_pass(self):
                                      len(unc_data.param_labels) * 4
                                      )
 
+    def test_calc_sensitivity_ff_pass(self):
+        """Test compute sensitivity using the 'ff' method"""
+
+        exp_unc, impf_unc, haz_unc = make_input_vars()
+        unc_calc = CalcImpact(exp_unc, impf_unc, haz_unc)
+        unc_data = unc_calc.make_sample(N=4, sampling_method='ff')  # Generate samples
```
chahank, line 723: You should test for the warning messages.

climada/engine/unsequa/unc_output.py
```diff
@@ -56,13 +56,15 @@
 #Table of recommended pairing between salib sampling and sensitivity methods
 # NEEDS TO BE UPDATED REGULARLY!! https://salib.readthedocs.io/en/latest/api.html
 SALIB_COMPATIBILITY = {
+    #'delta': ['fast_sampler', 'ff', 'finite_diff', 'latin', 'morris', 'saltelli'],
```
chahank, line 59: Add comment why it is commented.

chahank (changes requested):
Great, thanks! 

I think there are a few small things to update, and maybe one general idea for the tests. Since the tests for the different method repeat themselves, it would be better to define a function and then call it in a loop for all methods instead of copying large parts of the code.

CHANGELOG.md
emanuel-schmid: This is about dependency version constraint changes. They start from the given minimal version of the last release (left hand side) and end at the given minimal version of this release (right hand side).
However: the whole version change section is eventually done by a script I wrote. Don't bother, it will be overwritten!

climada/engine/unsequa/calc_base.py
luseverin: I think it is used. It is there to remove every variable which has a size greater than nparams^2, the rationale being that variables that have size > nparams^2 cannot be sensitivity indices of degree 2 or smaller. It catches the variables ['select', 'RT', 'Y_em', 'idx', 'X', 'Y']. Alternatively I can put all of these directly in the keys_to_remove list without doing this loop, it might be clearer.

climada/engine/unsequa/calc_base.py
luseverin: I think sens_first_order_df = pd.DataFrame({k: pd.Series(v, dtype=object) for k, v in sens_first_order_dict.items()}) was added in a previous commit by @simonameiler. I commented it back because it seemed to create issues with the plot_sensitivity method (empty Axes not displaying anything). I think this commented line can be removed. @simonameiler what do you think?

climada/engine/unsequa/calc_base.py
luseverin: Do you mean in front of the if? It was added automatically by VScode..

climada/engine/unsequa/calc_base.py
```diff
@@ -500,10 +536,44 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
         else:
             sens_indices = method.analyze(problem_sa, Y,
                                                     **sensitivity_kwargs)
+        #refactor incoherent SALib output
+        nparams = len(param_labels)
+        if method.__name__ == 'SALib.analyze.ff':
```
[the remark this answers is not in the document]
luseverin: Because it is not available from within _calc_sens_df, I think? But we can pass it as an argument of _calc_sens_df if you think it is worth it.

climada/engine/unsequa/calc_base.py
luseverin: I changed 'ie' to 'interaction_name'. 'IE' actually stands for 'Interaction Effect'. As it is the actual name of the indice for the 'ff' method from SALib I would keep it this way, what do you think?

climada/engine/unsequa/calc_base.py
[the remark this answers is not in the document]
chahank: But it is already... the argument `method` is a string.

climada/engine/unsequa/calc_base.py
luseverin: Do you suggest simply adding a . after the 2 or also removing everything that comes after it?

chahank: Just add the '.' Otherwise the message will look incomplete.

climada/engine/unsequa/calc_base.py
luseverin: Do we still need this line of code? As we now raise a `ValueError` when the number of parameters is not a power of 2 and require the user to input their own dummy parameter beforehand. Then if the user inputs their own dummy parameter, the dummy parameters should be covered by the ```df_samples[param] = df_samples[param].apply(self.distr_dict[param].ppf)```, correct?

climada/engine/unsequa/calc_base.py
luseverin: I simplified the approach, and just directly set keys_to_remove as ['Em', 'Term', 'select', 'RT', 'Y_em', 'idx', 'X', 'Y'].

climada/engine/unsequa/calc_base.py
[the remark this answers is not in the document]
luseverin: Ok then I changed it to only check if the corresponding last characters of the method name string correspond to the target method name instead of the entire `Salib.analyse.ff`.

chahank:
Great job, we are almost there! What is missing:

- Address all the new linter issues (and any older ones that you solve are bonus points!)
- The method for the test is in the right direction. I would maybe propose to simplify the whole thing a bit. More on that in the code directly.

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -471,6 +473,33 @@ def test_make_sample_pass(self):
...
+            # Warning ff sampling
+            unc_calc = CalcImpact(exp_unc, impf_unc, haz_unc)
+            warning_msg = "You are using the 'ff' sampler which does not require "
+            "a value for N. The entered N value will be ignored"
+            "in the sampling process."
+
+            with self.assertLogs(ILOG, level='WARNING') as logs:
+                unc_data = unc_calc.make_sample(N=4, sampling_method='ff')
+            self.assertEqual(len(logs.output), 1)
+            self.assertIn(warning_msg, logs.output[0])
```
chahank, line 491: I think this should be in the indention of the with. (so add one tab)
```suggestion
            self.assertEqual(len(logs.output), 1)
            self.assertIn(warning_msg, logs.output[0])
```

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -614,6 +800,76 @@ def test_calc_sensitivity_morris_pass(self):
...
+        self.assertTrue(unc_data.eai_exp_unc_df.empty)
+        self.assertTrue(unc_data.at_event_unc_df.empty)
+
+    def test_calc_sensitivity_hdmr_pass(self):
+            """Test compute sensitivity using the 'hdmr' method"""
+
+            exp_unc, impf_unc, haz_unc = make_input_vars()
+            haz = haz_dem()
+            unc_calc = CalcImpact(exp_unc, impf_unc, haz_unc)
+            unc_data = unc_calc.make_sample(N=100)  # Generate samples
```
chahank, line 843: Should you not specify the sampling method here?

luseverin:
Ok so I tried to simplify the tests of the sensitivity methods as much as possible in the generic test functions, and removed the previous method-specific test functions. I wasn't sure about a few lines that only appeared in the test for the morris method though:
```
for name, attr in unc_data.__dict__.items():
            if 'sens_df' in name:
                np.testing.assert_array_equal(
                    attr.param.unique(),
                    np.array(['x_exp', 'x_paa', 'x_mdd'])
                    )
                np.testing.assert_array_equal(
                    attr.si.unique(),
                    np.array(['mu', 'mu_star', 'sigma', 'mu_star_conf'])
                    )
                if 'eai' in name:
                    self.assertEqual(
                        attr.size,
                        len(unc_data.param_labels)*4*(len(exp_unc.evaluate().gdf) + 3)
                        )
                elif 'at_event' in name:
                    self.assertEqual(
                        attr.size,
                        len(unc_data.param_labels) * 4 * (haz.size + 3)
                        )
                else:
                    self.assertEqual(len(attr),
                                     len(unc_data.param_labels) * 4
                                     )
```
I did not include these lines in the generic test function as it did not seem absolutely necessary to me, but if you think those should be included just let me know.

Additionally, I need to fix the few new linter issues that popped up but I think it shouldn't be an issue. Otherwise I think I adressed the rest of your comments.

climada/engine/unsequa/calc_base.py
```diff
@@ -515,6 +585,11 @@ def _calc_sens_df(method, problem_sa, sensitivity_kwargs, param_labels, X, unc_d
         sens_second_order_dict[submetric_name] = sens_second_order
 
     sens_first_order_df = pd.DataFrame(sens_first_order_dict, dtype=np.number)
+    # Assume sens_first_order_dict is a dictionary where values are lists/arrays of varying lengths
+    # !for some reason this make the plotting methods fail
+    #sens_first_order_df = pd.DataFrame({k: pd.Series(v, dtype=object)
+    #                                    for k, v in sens_first_order_dict.items()})
+
```
chahank, line 592: Is this something that still needs to be addressed or is it a left-over?

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -36,6 +36,8 @@
 from climada.hazard import Hazard
 from climada.engine import ImpactCalc
 from climada.engine.unsequa import InputVar, CalcImpact, UncOutput, CalcCostBenefit, CalcDeltaImpact
+from climada.engine.unsequa.calc_base import LOGGER as ILOG
```
chahank, line 39: Any particular reason for renaming the logger? Why not just use `LOGGER` ?

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -527,92 +571,134 @@ def test_calc_uncertainty_pool_pass(self):
...
+                'sensitivity_kwargs' : {},
+                'test_param_name' : ['x_exp', 0],
+                'test_si_name' : ['mu', 1],
+                'test_si_value' : [5066460029.63911, 8]
+            },
+        }
+
+        def test_sensitivity_method(sensitivity_method, param_dict):
+            """Function to test each seaprate sensitivity method"""
+            unc_calc = CalcImpact(exp_unc, impf_unc, haz_unc)
```
chahank, line 666: It would be cleaner if the exp_unc, impf_unc and haz_unc are also passed as arguments.

climada/engine/unsequa/test/test_unsequa.py
```diff
@@ -527,92 +571,134 @@ def test_calc_uncertainty_pool_pass(self):
...
+            self.assertEqual(
+                unc_data.freq_curve_unc_df.size,
+                unc_data.n_samples * len(unc_calc.rp)
+                )
+            self.assertTrue(unc_data.eai_exp_unc_df.empty)
+            self.assertTrue(unc_data.at_event_unc_df.empty)
+
+        # loop over each method and do test
+        for sensitivity_method in test_dict:
+            test_sensitivity_method(sensitivity_method, test_dict[sensitivity_method])
```
chahank, line 700:
```suggestion
        for sensitivity_method, method_params in test_dict.items():
            test_sensitivity_method(sensitivity_method, method_params)
```

climada/engine/unsequa/test/test_unsequa.py
luseverin: No, no reason, it was just blind copying of the code I saw on stackoverflow, I'll change that!

climada/engine/unsequa/calc_base.py
luseverin: It's a left-over, I cleaned it up.

luseverin:
Ok so I incorporated your last suggestions. If there is no further comments from your side I suggest we merge this PR?

chahank:
I think it looks good!

Just FYI: you managed to remove 39 pylint warnings, and add 16. Maybe you can take a last look whether the 16 new ones are easy to avoid or not. 

Else, this is ready to merge from my point of view.

PR #889, opened 2024-06-07
set_geometry

### LINKED ISSUE #888: util.coordinates functions add 'geometry' column to GeoDataFrame
**Describe the bug**
Whenever `climada.util.coordinates.points_to_raster` or `climada.util.coordinates.set_df_geometry_points` are executed a warning is raised: 
```
FutureWarning: You are adding a column named 'geometry' to a GeoDataFrame constructed without an active
geometry column. Currently, this automatically sets the active geometry column to 'geometry' but in the
future that will no longer happen.
Instead, either provide geometry to the GeoDataFrame constructor (GeoDataFrame(... geometry=GeoSeries())
or use `set_geometry('geometry')` to explicitly set the active geometry column.
```

**To Reproduce**
1. Create an `Exposures` object without a proper geometry column.
2. Run `set_geometry_points` on it.
3. Run `plot_raster` on it.

Code example:
```python
import pandas as pd
from shapely.geometry import Point
from climada.entity.exposures import Exposures


exp = Exposures(data=pd.DataFrame(
    dict(
        latitude = [1,2,3]*3,
        longitude = [1]*3+[2]*3+[3]*3,
        value = range(9)
    )
), crs='EPSG:4326')
exp.set_geometry_points()
exp.plot_raster()
```

**Expected behavior**
Silent plotting.

**Climada Version:**
 - climada_python >= 4.0

 - Operating system and version: Windows 10
 - Python version:3.9

**Additional context**
 - geopandas >= 0.14

emanuel-schmid:
Changes proposed in this PR:
- use set_geometry in util.coordinates instead of direct column assignment

This PR fixes #888

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Quick note: while the fix works, I think this is not a very forward-looking fix.

Currently, the method assumes that the inputs are geodataframes without a geometry column, but instead latitude and longitude columns. This is what we are trying to remove from the codebase (done in centroids, ongoing in exposures).

I suggest that we at least flag this method as one that should be refactored as soon as exposures have proper geodataframes.

peanutfun:
You don't need a temporary column. `set_geometry` also accepts an array of values, and sets these as the geometry column, see https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.set_geometry.html

emanuel-schmid:
Er - are you sure? When running with a scheduler, dask.dataframe complains if there is no column to assign the values to. Isn't it?

emanuel-schmid:
> I suggest that we at least flag this method as one that should be refactored as soon as exposures have proper geodataframes.

No need to flag this. Any occurrence of geodataframes with latitude columns will eventually show up in the list of failed tests. 😁

climada/util/coordinates.py
```diff
@@ -2655,26 +2658,28 @@ def set_df_geometry_points(df_val, scheduler=None, crs=None):
         try:
             crs = df_val.geometry.crs
         except AttributeError:
-            crs = None
+            pass
```
peanutfun, line 2661: Any GeoDataFrame has the attribute [`.crs`](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.crs.html#geopandas.GeoDataFrame.crs). If it is not set, the attribute already returns None. So this can simply be
```python
crs = df_val.crs if crs is None else crs  # crs might now still be None
```

peanutfun (changes requested):
Thanks for updating this! I think the CRS handling can be improved a bit.

However, I am wondering if the `set_df_geometry_points` needs the bit about the scheduler at all. `geopandas.points_from_xy` has become very efficient since Shapely v2.0, because the latter now uses multithreading. As the final geometry is anyway stored in a regular GeoDataFrame, and not in a distributed dask dataframe, I think we can safely remove the whole scheduler thing from `set_df_geometry_points` and make the function much simpler.

The linter complains anyway that the scheduler part is not tested 😅

emanuel-schmid:
> I think we can safely remove the whole scheduler thing from set_df_geometry_points and make the function much simpler.

@peanutfun : I'm inclined to believe you but would like to separate that from this PR, as it is not related to #888

climada/util/coordinates.py
```diff
@@ -2655,26 +2658,28 @@ def set_df_geometry_points(df_val, scheduler=None, crs=None):
         try:
             crs = df_val.geometry.crs
         except AttributeError:
```
emanuel-schmid, line 2660:
```suggestion
        crs = df_val.crs if crs is None else crs  # crs might now still be None
```

climada/util/coordinates.py
emanuel-schmid: 👍

Issue #891, opened 2024-06-13
Wildfire data on API has non-canonical format, resulting in erroneous hazard


peanutfun:
**Describe the bug**
We had noticed some strange patterns before when plotting the wildfire data from the API. However, the statistics on the `Hazard.intensity.data` looked fine, so we assumed a plotting issue. Turns out, the `Hazard.intensity` is not in canonical format, meaning that multiple entries in `intensity.data` map to the same entry in the `intensity` matrix. This causes erroneously high intensities for about 6% of all centroids.

**To Reproduce**
Steps to reproduce the behavior/error:
1. Download a wildfire dataset (here: Mexico) and plot it. Observe weird rectangular patterns with high intensity
2. Notice that the matrix is not in canonical format, and that its maximum is way larger than the `data` maximum.
   * To add to the confusion, calling `max` will bring it into canonical format.

```python
from climada.util.api_client import Client

client = Client()
wf_mexico = client.get_hazard("wildfire", properties={"country_iso3alpha": "MEX"})

print(f"Canonical format: {wf_mexico.intensity.has_canonical_format}")
print(f"Min data: {wf_mexico.intensity.data.min()}")
print(f"Max data: {wf_mexico.intensity.data.max()}")
print(f"Min value: {wf_mexico.intensity.min()}"). # This implicitly ensures canonical format!
print(f"Max value: {wf_mexico.intensity.max()}")
print(f"Canonical format: {wf_mexico.intensity.has_canonical_format}")
print(f"Min data: {wf_mexico.intensity.data.min()}")
print(f"Max data: {wf_mexico.intensity.data.max()}")

wf_mexico.plot_intensity(0)
```
Output:
```
Canonical format: False
Min data: 300.2
Max data: 506.2
Min value: 0.0
Max value: 1695.1
Canonical format: True
Min data: 300.2
Max data: 1695.1
```
![240613-wf-issue-01](https://github.com/CLIMADA-project/climada_python/assets/34276446/3b31ce1e-16be-44bb-9787-a02c92a0d8be)

**Background**

The reason for this is that multiple entries in `intensity.data` point to the same matrix entries of `intensity`. `csr_matrix` supports this, and sums up these values. 

```python
wf_mexico = client.get_hazard("wildfire", properties={"country_iso3alpha": "MEX"})
indices_event_0 = wf_mexico.intensity.indices[
    wf_mexico.intensity.indptr[0] : wf_mexico.intensity.indptr[1]
]
[...]

PR #893, opened 2024-06-14
Ensure csr matrices are in "canonical format" in Hazard objects

peanutfun:
Changes proposed in this PR:
- Use getter/setter methods and private attributes to ensure that any csr matrix passed into a hazard is in "canonical format". This ensures that the `csr_matrix.data` attribute reflects the exact values present in the matrix.
- Update tests

This PR fixes inconsistencies in the csr matrix reported in #891, but does not fix the problem of unintentionally summed values (it just ensures this problem is visible when looking at the matrix data).

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -90,8 +90,8 @@ class Hazard(HazardIO, HazardPlot):
                   'centroids',
                   'event_id',
                   'frequency',
-                  'intensity',
-                  'fraction'
+                  '_intensity',
+                  '_fraction'
                   }
     """Name of the variables needed to compute the impact. Types: scalar, str,
     list, 1dim np.array of size num_events, scipy.sparse matrix of shape
@@ -181,15 +181,47 @@ class Hazard(HazardIO, HazardPlot):
         self.event_name = event_name if event_name is not None else list()
         self.date = date if date is not None else np.array([], int)
         self.orig = orig if orig is not None else np.array([], bool)
-        # following values are defined for each event and centroid
-        self.intensity = intensity if intensity is not None else sparse.csr_matrix(
-            np.empty((0, 0)))  # events x centroids
-        self.fraction = fraction if fraction is not None else sparse.csr_matrix(
-            self.intensity.shape)  # events x centroids
+
+        # The following values are defined for each event and centroid
+        # Shape: ( events x centroids )
+        self._intensity = intensity
+        self._fraction = fraction
 
         self.pool = pool
         if self.pool:
             LOGGER.info('Using %s CPUs.', self.pool.ncpus)
+
+    @property
+    def intensity(self) -> sparse.csr_matrix:
+        """Hazard intensity matrix"""
+        if self._intensity is None:
+            return sparse.csr_matrix(np.empty((0, 0)))
+
+        return self._intensity
+
+    @intensity.setter
+    def intensity(self, value: sparse.csr_matrix):
+        """Set intensity matrix to new value"""
+        self._intensity = value
+        self._intensity.check_format()
+        self._intensity.sort_indices()
+        self._intensity.sum_duplicates()
+
+    @property
+    def fraction(self):
+        """Hazard fraction matrix"""
+        if self._fraction is None:
+            return sparse.csr_matrix(self.intensity.shape)
+
+        return self._fraction
+
+    @fraction.setter
+    def fraction(self, value: sparse.csr_matrix):
+        """Set fraction matrix to new value"""
+        self._fraction = value
+        self._fraction.check_format()
+        self._fraction.sort_indices()
+        self._fraction.sum_duplicates()
 
     @classmethod
     def get_default(cls, attribute):
--- a/climada/hazard/test/test_base.py
+++ b/climada/hazard/test/test_base.py
@@ -124,18 +124,18 @@ class TestLoader(unittest.TestCase):
     def test_check_wrongInten_fail(self):
         """Wrong hazard definition"""
         self.hazard.intensity = sparse.csr_matrix([[1, 2], [1, 2]])
-
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaisesRegex(
+            ValueError, "Invalid Hazard._intensity row size: 3 != 2."
+        ):
             self.hazard.check()
-        self.assertIn('Invalid Hazard.intensity row size: 3 != 2.', str(cm.exception))
 
     def test_check_wrongFrac_fail(self):
         """Wrong hazard definition"""
         self.hazard.fraction = sparse.csr_matrix([[1], [1], [1]])
-
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaisesRegex(
+            ValueError, "Invalid Hazard._fraction column size: 2 != 1."
+        ):
             self.hazard.check()
-        self.assertIn('Invalid Hazard.fraction column size: 2 != 1.', str(cm.exception))
 
     def test_check_wrongEvName_fail(self):
         """Wrong hazard definition"""
@@ -211,6 +211,25 @@ class TestLoader(unittest.TestCase):
         self.assertEqual(len(haz.get_event_date()), haz.date.size)
         self.assertEqual(haz.get_event_date()[560],
                          u_dt.date_to_str(haz.date[560]))
+
+    def test_matrix_consistency(self):
+        """Check that the csr_matrix is brought in canonical format"""
+        # Non-canonical: All data points will be summed onto the first matrix entry
+        data = [0, 1, 2]
+        indices = [0, 0, 0]
+        indptr = [0, 3, 3, 3]
+        matrix = sparse.csr_matrix((data, indices, indptr), shape=(3, 2))
+        self.assertEqual(matrix.data.max(), 2)
+        self.assertEqual(matrix[0, 0], 3)
+        self.assertFalse(matrix.has_canonical_format)
+
+        # Check conversion to canonical format when assigning
+        for attr in ("intensity", "fraction"):
+            setattr(self.hazard, attr, matrix.copy())
+            hazard_matrix = getattr(self.hazard, attr)
+            self.assertTrue(hazard_matrix.has_canonical_format)
+            self.assertEqual(hazard_matrix[0, 0], 3)
+            np.testing.assert_array_equal(hazard_matrix.data, [3])
 
 class TestRemoveDupl(unittest.TestCase):
     """Test remove_duplicates method."""
--- a/climada/hazard/io.py
+++ b/climada/hazard/io.py
@@ -1004,12 +1004,18 @@ class HazardIO():
         LOGGER.info('Writing %s', file_name)
         with h5py.File(file_name, 'w') as hf_data:
             str_dt = h5py.special_dtype(vlen=str)
-            for (var_name, var_val) in self.__dict__.items():
+            for var_name in self.__dict__:
                 if var_name == 'centroids':
                     # Centroids have their own write_hdf5 method,
                     # which is invoked at the end of this method (s.b.)
-                    pass
-                elif isinstance(var_val, sparse.csr_matrix):
+                    continue
+                # Prune private attributes
+                if var_name in self.vars_oblig:
+                    var_name = var_name.lstrip("_")
+
+                var_val = getattr(self, var_name)  # Also works with properties
+
+                if isinstance(var_val, sparse.csr_matrix):
                     if todense:
                         hf_data.create_dataset(var_name, data=var_val.toarray())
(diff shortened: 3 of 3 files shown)


peanutfun:
Petals compatibility is addressed here: CLIMADA-project/climada_petals#129

climada/hazard/base.py
```diff
@@ -191,6 +191,32 @@ def __init__(self,
...
+        """Hazard intensity matrix"""
+        return self._intensity
+
+    @intensity.setter
+    def intensity(self, value: sparse.csr_matrix):
+        """Set intensity matrix to new value"""
+        self._intensity = value
+        self._intensity.check_format()
+        self._intensity.eliminate_zeros()
+        self._intensity.sum_duplicates()
```
chahank, line 205: I am not fully convinced that we should do that silently. Can we not instead raise an error if the input `crs_matrix` has duplicates? Maybe the property `crs_matrix.has_canonical_format` would be the thing to check?

chahank (commented):
Good idea to add more consistency to the hazard data! One suggestion / comment.

climada/hazard/base.py
peanutfun: Two reasons against this:
1. Several tests fail if we introduce an explicit check for `has_canonical_format`. I did not dig deeper to check why exactly. But it is not uncommon throughout the Climada code base.
2. A non-canonical matrix is completely fine from the implementation perspective. It just has two disadvantages I see:
   * `.data` does not reflect the actual matrix values. I think this is irrelevant for 90% of all users, they will never touch that attribute.
   * Accessing matrix values might be slower if indices are not ordered and there are duplicates. Not sure how much, though.

The overhead for eliminating zeros and summing duplicates should be very low, especially since it will speed up the impact calculation in turn. Because no "visible" data of the matrix is changed, I think silently calling these methods is fine

chahank: hmmm ok. Then maybe we should add it as information in the docstring of the class?

peanutfun: Yes, can do.

Come to think of it, we should probably also call these checks when the Hazard is plugged into `ImpactCalc`, because direct modifications of the csr_matrix might again produce duplicates or explicit zeros.

peanutfun:
@chahank What do you think about this updated version? I still have to adapt the test.

climada/hazard/base.py
```diff
@@ -49,6 +49,22 @@ class Hazard(HazardIO, HazardPlot):
...
+    `scipy.sparse.csr_matrix
+    <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html>`_
+    to store :py:attr:`intensity` and :py:attr:`fraction`. This data types comes with
+    its particular pitfalls. Depending on how the objects are instantiated and modified,
+    a matrix might end up in a "non-canonical" state. In this state, its ``.data``
+    attribute does not necessarily represent the values apparent in the final matrix.
+    In particular, a "non-canonical" matrix may store "duplicates", i.e. multiple values
+    that map to the same matrix position. This is supported, and the default behavior is
+    to sum up these values. To avoid any inconsistencies, call :py:meth:`check_matrices`
+    once you inserted all data.
```
chahank, line 64:
```suggestion
    once you inserted all data. This will explicitly sum all values at the same matrix position and eliminate explicit zeros.
```

climada/hazard/test/test_base.py
```diff
@@ -212,6 +212,47 @@ def test_get_date_strings_pass(self):
...
+        # Non-canonical: First three data points will be summed onto the first matrix
+        # entry, forth will be an explicit zero entry
+        data = [0, 1, 2, 0]
+        indices = [0, 0, 0, 1]
+        indptr = [0, 4, 4, 4]
+        matrix = sparse.csr_matrix((data, indices, indptr), shape=(3, 2))
+        np.testing.assert_array_equal(matrix.data, data)
+        np.testing.assert_array_equal(matrix[0, [0, 1]].toarray(), [[3, 0]])
+        self.assertEqual(matrix.nnz, 4)
+        self.assertFalse(matrix.has_canonical_format)
```
chahank, line 237: I am not sure we want to test the `has_canonical_format method` from scipy. I would just remove this line.

peanutfun: I left it in the test for `prune_matrix`, just to be sure that we are using the `csr_matrix` as intended, and things are indeed working as expected.

climada/engine/test/test_impact_calc.py
```diff
@@ -70,6 +70,16 @@ def test_init(self):
         np.testing.assert_array_equal(HAZ.event_id, icalc.hazard.event_id)
         np.testing.assert_array_equal(HAZ.event_name, icalc.hazard.event_name)
 
+        # Test check matrices
```
chahank, line 73:
```suggestion
        # Test check matrices
        # Check fraction and intensity have the same shape
        # Check that explicit zeroes are pruned
```

I am a bit confused why the pruning is tested here...

climada/hazard/base.py
```diff
@@ -49,6 +49,23 @@ class Hazard(HazardIO, HazardPlot):
...
+    `scipy.sparse.csr_matrix
+    <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html>`_
+    to store :py:attr:`intensity` and :py:attr:`fraction`. This data types comes with
+    its particular pitfalls. Depending on how the objects are instantiated and modified,
+    a matrix might end up in a "non-canonical" state. In this state, its ``.data``
+    attribute does not necessarily represent the values apparent in the final matrix.
+    In particular, a "non-canonical" matrix may store "duplicates", i.e. multiple values
+    that map to the same matrix position. This is supported, and the default behavior is
+    to sum up these values. To avoid any inconsistencies, call :py:meth:`check_matrices`
+    once you inserted all data. This will explicitly sum all values at the same matrix
```
chahank, line 64: Do we want to recommend checking that if the class does check it anyway? I am a bit sceptical with these `check_object` methods that we have in climada. Imho these should be handled in the `init` and that is it.

climada/hazard/base.py
```diff
@@ -186,11 +203,58 @@ def __init__(self,
...
+    def check_matrices(self):
+        """Ensure that matrices are consistently shaped and stored
+
+        See Also
+        --------
+        :py:func:`climada.util.checker.prune_csr_matrix`
+
+        Todo
+        -----
+        * Check consistency with centroids
```
chahank, line 243: Can't we do that in this PR directly? I think we only need to check that the number of lines of the hazard matrices is equal to the number of centroids right?

climada/util/checker.py
```diff
@@ -162,3 +163,36 @@ def array_default(exp_len, var, var_name, def_val):
...
+    See Also
+    --------
+    `csr_matrix.has_canonical_format
+    <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.has_canonical_format.html#scipy.sparse.csr_matrix.has_canonical_format>`_
+
+    Parameters
+    ----------
+    matrix : csr_matrix
+        The matrix to check. It will be modified *inplace*. No apparent matrix values
+        will change.
```
chahank, line 186:
```suggestion
        The matrix to check. It will be modified *inplace*. The .data attribute will change, but
        no apparent matrix values will change.
```

climada/util/test/test_checker.py
```diff
@@ -87,6 +87,28 @@ def test_check_optionals_fail(self):
             u_check.check_optionals(dummy.__dict__, dummy.vars_opt, "DummyClass.", dummy.id.size)
         self.assertIn('Invalid DummyClass.list size: 25 != 3.', str(cm.exception))
 
+    def test_prune_csr_matrix(self):
+        """Check that csr matrices are brought into canonical format"""
+        # Non-canonical: First three data points will be summed onto the first matrix
+        # entry, forth will be an explicit zero entry
```
chahank, line 93:
```suggestion
        # entry, fourth will be an explicit zero entry
```

climada/engine/test/test_impact_calc.py
peanutfun: I can also check via a mock if `check_matrices` is called, but I recall strong sentiments against such tests 😄

climada/hazard/base.py
peanutfun: Yes, but the new Hazard never did this check automatically, did it? I just wanted to keep things simple. If we add this, we also need to add another test, and so on. I would like to postpone it.

climada/util/checker.py
peanutfun: `indices` and `indptr` attributes might also change. Do we need to explain this as well?

peanutfun:
@chahank Waiting for your responses to move forward 😇 👉 👈

climada/util/checker.py
chahank: No, I think this if fine like that.

climada/hazard/base.py
chahank: All well, as long as we do not forget it ;)

climada/hazard/base.py
chahank: hmmm... not yet fully convinced. If the user messes up an object, which the user always can in Python, then it is their problem. I would rather have fewer checks at random places in the code that hide bad habits and save people from doing things they should not do, and instead force good habits by having checks in the inits.

climada/engine/test/test_impact_calc.py
chahank: I think my confusion comes from the fact that I am uneasy with having the impact class changing the hazard objects. See comment below.

climada/hazard/base.py
peanutfun: I generally agree that we should only check if necessary. However, the MDR calculation operates on the `data` attribute. It is therefore essential that canonical format is ensured before starting the impact calculation.

If you really want to boil it down to a minimum, I would suggest to remove the pruning from the Hazard setters and instead only call `check_matrices` from `ImpactCalc`. But this might increase confusion again, because consistency is never ensured *before* starting `ImpactCalc`.

Another thing to keep in mind is that `sum_duplicates` is O(N^2) for a non-canonical matrix, and a no-op O(1) for a canonical matrix, and `eliminate_zeros` is O(N), where N is the number of stored non-zeros entries.

climada/util/checker.py
peanutfun:
```suggestion
        The matrix to check. It will be modified *inplace*. Its ``.data`` attribute
        might change, but apparent matrix values will stay the same.
```

climada/hazard/base.py
chahank: Good points, and thanks for looking up the time cost of the methods.

Ok, let's try it like this. Maybe we can keep an eye on this change in the separate efforts to optimize computation times for ImpactCalc.

peanutfun: In this case, I'll revert the properties and make `intensity` and `fraction` proper attributes again.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ CLIMADA tutorials. [#872](https://github.com/CLIMADA-project/climada_python/pull
 - Centroids complete overhaul. Most function should be backward compatible. Internal data is stored in a geodataframe attribute. Raster are now stored as points, and the meta attribute is removed. Several methds were deprecated or removed. [#787](https://github.com/CLIMADA-project/climada_python/pull/787)
 - Improved error messages produced by `ImpactCalc.impact()` in case impact function in the exposures is not found in impf_set [#863](https://github.com/CLIMADA-project/climada_python/pull/863)
 - Changed module structure: `climada.hazard.Hazard` has been split into the modules `base`, `io` and `plot` [#871](https://github.com/CLIMADA-project/climada_python/pull/871)
+- Ensure `csr_matrix` stored in `climada.hazard.Hazard` have consistent data format and store no explicit zeros [#893](https://github.com/CLIMADA-project/climada_python/pull/893)
```
chahank, line 27:
```suggestion
- Ensure `csr_matrix` stored in `climada.hazard.Hazard` have consistent data format and store no explicit zeros [#893](https://github.com/CLIMADA-project/climada_python/pull/893)
- Before an impact calculation with `climada.engine.impact_calc.ImpactCal.calc()` is executed the hazard intensity and fraction matrices are checked and transformed to a consistent data format without explicit zeroes [#893](https://github.com/CLIMADA-project/climada_python/pull/893)
```

chahank:
Great fix! Ready to merge. Just one possible addition to the changelog.

CHANGELOG.md
peanutfun: Ah yes I forgot to update this

PR #894, opened 2024-06-17
Explicitly convert `event_name` to strings in `Impact.from_hdf5`

### LINKED ISSUE #807: Impact read hdf5 fails if hazard event name is not a string
Similarly to issue #789 , saving an impact to hdf5 if the hazard name is not a list of strings causes issues when the data is read again. 

Precisely, the attribute `impact.event_name` must be a list of strings at the moment of saving to hdf5 `impact.to_hdf5`, or the reading method `impact.from_hdf5` should be able to handle non-numerical values.

peanutfun:
Changes proposed in this PR:
- Call `str` on values if decoding strings from `event_name` fails, issue a warning in this case.
- Throw error during writing if `event_name` is not strings.
- Fix an issue where only the first value of `event_name` was checked.

This PR fixes  #807

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1240,10 +1240,18 @@ class Impact():
             ).intersection(file.keys())
             kwargs.update({attr: file[attr][:] for attr in array_attrs})
 
-            # Special handling for 'event_name' because it's a list of strings
+            # Special handling for 'event_name' because it should be a list of strings
             if "event_name" in file:
                 # pylint: disable=no-member
-                kwargs["event_name"] = list(file["event_name"].asstr()[:])
+                try:
+                    event_name = file["event_name"].asstr()[:]
+                except TypeError:
+                    LOGGER.warn(
+                        "'event_name' could not be decoded to a string. Reading raw "
+                        "values instead."
+                    )
+                    event_name = file["event_name"][:]
+                kwargs["event_name"] = list(event_name)
 
         # Create the impact object
         return cls(**kwargs)
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -937,11 +937,6 @@ class Impact():
 
         The impact matrix can be stored in a sparse or dense format.
 
-        Notes
-        -----
-        This writer does not support attributes with variable types. Please make sure
-        that ``event_name`` is a list of equally-typed values, e.g., all ``str``.
-
         Parameters
         ----------
         file_path : str or Path
@@ -950,6 +945,11 @@ class Impact():
             If ``True``, write the impact matrix as dense matrix that can be more easily
             interpreted by common H5 file readers but takes up (vastly) more space.
             Defaults to ``False``.
+
+        Raises
+        ------
+        TypeError
+            If :py:attr:`event_name` does not contain strings exclusively.
         """
         # Define writers for all types (will be filled later)
         type_writers = dict()
@@ -983,7 +983,7 @@ class Impact():
 
         def _str_type_helper(values: Collection):
             """Return string datatype if we assume 'values' contains strings"""
-            if isinstance(next(iter(values)), str):
+            if all((isinstance(val, str) for val in values)):
                 return h5py.string_dtype()
             return None
 
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1246,7 +1246,7 @@ class Impact():
                 try:
                     event_name = file["event_name"].asstr()[:]
                 except TypeError:
-                    LOGGER.warn(
+                    LOGGER.warning(
                         "'event_name' could not be decoded to a string. Reading raw "
                         "values instead."
                     )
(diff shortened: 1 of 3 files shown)


emanuel-schmid:
why don't we just force event names to be strings? If `asstr()` fails there must be other methods, e.g., 

```python
event_name = [str(x) for x in file["event_name"][:]]
```

peanutfun:
@emanuel-schmid There was no decision on what to do in #807, so I came up with the least intrusive solution (from my perspective).

I am fine with your proposal, but it would imply that an `Impact` object with non-string `event_name` attribute is not cleanly "cycled": After writing to H5 and reading it again, the `event_name` will be a list of strings.

emanuel-schmid:
Yeah, right. Sorry for coming late to the discussion - but...
Imho 1) a name ought to be a string 2) ambiguous data types are a constant source of sorrow.
Sooner or later someone will run into an issue because one of their hazards has event_name = 1, 2, 3 and another one '1', '2', '3'
If it's only about clean cycles (for not _so_ clean data): I estimate the cost of losing "cyclability" lower than having to deal with this for ever.

peanutfun:
@emanuel-schmid Thanks a lot for that perspective! This is exactly the kind of perspective I was hoping for in the discussion of #807. I would be fine with going the "string-only" route. However, there are some inconsistencies then we have to fix. In the tests for Impact, there are two instances where Impacts are instantiated with non-string event names:
* https://github.com/CLIMADA-project/climada_python/blob/9faf54bc219cf7f53b7000c87a4ec427e643aa52/climada/engine/test/test_impact.py#L54
* https://github.com/CLIMADA-project/climada_python/blob/9faf54bc219cf7f53b7000c87a4ec427e643aa52/climada/engine/test/test_impact.py#L603
There might be more such instances throughout the code base.

peanutfun:
@chahank What do you think?

chahank:
Thanks for the discussion @peanutfun and @emanuel-schmid : I here would also support your suggestion of the `string only` route. I think this makes things much clearer.

emanuel-schmid:
@peanutfun Yes, Sorry again. 😳 
thanks for pointing out the inconsistent tests. With regard to fixing #807 they have most likely no impact, 
but they are sending contradictory signals and we should eventually put every instance of an event_name within quotes.
However - I think we can follow this up by fixing these when we come across them.

peanutfun:
@emanuel-schmid @chahank When we go string-only, we still have to clarify: Do we (try to) convert any event name data to strings when writing, when reading, or both times?

emanuel-schmid:
Suggestion: When writing we simply _assume_ they're strings, when reading we convert if necessary.

peanutfun:
> (When writing it's kind of too late to do anything about it.)

Why? We can also call `str` on the elements before/while writing.

emanuel-schmid:
Sure we can. But what would be the point? By then we supposedly have already been working daredevil on an "irregular object". We should have converted the non string event names to strings long ago. If we didn't, the conversion to strings just for the writing cannot save us as it doesn't make a difference anymore. Or does it?

emanuel-schmid:
I'd agree if it really was dangerous. Is it though? I was under the impression that `str` can be called on literally every thing in Python. - ?

emanuel-schmid:
Right. This will also prevent storing pickled `event_name`s!

emanuel-schmid:
Hold on - the proper thing to do is fail if it is not a string, isn't it?

peanutfun:
@emanuel-schmid Great catch! Then we
1. Throw an error when writing non-strings.
2. Read strings or try to convert (in the latter case, issue a warning)

climada/engine/impact.py
```diff
@@ -983,7 +983,7 @@ def write(group: h5py.Group, name: str, value: Any):
 
         def _str_type_helper(values: Collection):
```
chahank, line 984: Is `Collection` the correct type? Does this not also allow for dictionaries and such? Just curious to learn the correct type hint.

peanutfun: `Collection` is any sized, iterable container, see https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes. The type might actually be too restrictive, `Iterable` would probably suffice here.

Since `dict` implements `__contains__`, `__iter__`, and `__len__`, it counts as `Collection`. Note the comment in the `type_writers` dict below: We try to cover the special types like `str`, `dict` and `csr_matrix` (which all count as `Collection`) before falling back to the collection writer.

peanutfun:
> Maybe we need an Issue to make sure that the event name is always a string in all hazard methods?

We would even need it for `Impact`, because this PR only fixes the HDF5 read/write methods. Elsewhere, the tests explicitly check if mixed-type event names work, see https://github.com/CLIMADA-project/climada_python/pull/894#issuecomment-2173307313

PR #857, opened 2024-03-07
Add functions to compute, plot, store the local hazard exceedence intensity and RP maps

simonameiler:
Changes proposed in this PR:
- addition of new functions to store the hazard maps that are generated in the Hazard.plot_rp_intensity() function as netCDF `write_netcdf_local_exceedance_inten` or raster files `write_raster_local_exceedance_inten.`
- addition of new functions to compute local return period maps `local_return_period` and `_loc_return_period` for given hazard intensities, incl. plotting function `plot_local_rp`.
- addition of new functions to store these local return period maps for given hazard intensities as NetCDF `write_netcdf_local_return_periods` or raster files `write_raster_local_return_periods`.

This PR fixes #854 

### PR Author Checklist

- [x] Read the [Contribution Guide][contrib]
- [x] Correct target branch selected (if unsure, select `develop`)
- [x] Descriptive pull request title added
- [ ] Source branch up-to-date with target branch
- [ ] [Documentation](https://climada-python.readthedocs.io/en/latest/guide/Guide_PythonDos-n-Donts.html#2.--Commenting-&-Documenting) updated
- [ ] [Tests][testing] updated
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]
- [ ] [Changelog](https://github.com/CLIMADA-project/climada_python/blob/main/CHANGELOG.md) updated

### PR Reviewer Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/guide/Guide_Reviewer_Checklist.html) passed
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]

[contrib]: https://github.com/CLIMADA-project/climada_python/blob/main/CONTRIBUTING.md
[testing]: https://climada-python.readthedocs.io/en/latest/guide/Guide_Continuous_Integration_and_Testing.html
[linter]: https://climada-python.readthedocs.io/en/stable/guide/Guide_Continuous_Integration_and_Testing.html#3.C.--Static-Code-Analysis

PR #898, opened 2024-06-20
Add functions to compute, plot, store the local hazard exceedence intensity and RP maps (new branch)

### LINKED ISSUE #854: Create a function to store the local exceedence intensity maps as netcdf file
It would be nice to have a new function to store the hazard maps that are generated in the `Hazard.plot_rp_intensity()` function as netCDF files. 

I wrote the following two functions below which allow me to store the local exceedence intensity maps. I also created a [branch](https://github.com/CLIMADA-project/climada_python/blob/feature/write_haz_rp_maps/climada/hazard/base.py#L1916), trying to incorporate these functions in the hazard base class but this does not work 100% yet and could also be improved. Particulary, the staticmethod I introduced to interpolate intensity data onto a regular grid may not be needed or may be better placed in `util.coordinates`.

Who can help?

`def interpolate_to_grid(latitudes, longitudes, coords, inten_stats, method='linear'):
    """
    Interpolates intensity data onto a regular grid defined by latitudes and longitudes.
    
    Parameters:
    - latitudes: 1D array of latitude values for the target grid.
    - longitudes: 1D array of longitude values for the target grid.
    - coords: (N, 2) array of original (latitude, longitude) pairs.
    - inten_stats: Original flattened array of intensity values.
    - method: Method of interpolation ('linear', 'nearest', 'cubic').
    
    Returns:
    - grid_inten_stats: 2D array of intensity values interpolated onto the regular grid.
    """
    grid_lat, grid_lon = np.meshgrid(latitudes, longitudes, indexing='ij')
    grid_inten_stats = griddata(coords, inten_stats, (grid_lat, grid_lon), method=method)
    
    return grid_inten_stats

def save_exceedance_intensity_as_netcdf(hazard, return_periods, filename):
    """
    Generates exceedance intensity data, interpolates it onto a regular grid,
    and saves the interpolated data into a NetCDF file.
    """
    inten_stats = hazard.local_exceedance_inten(return_periods=return_periods)
    coords = hazard.centroids.coord
    
    # Define the spatial extent and create regular grid arrays for latitude and longitude
    min_lat, max_lat = np.min(coords[:, 0]), np.max(coords[:, 0])
    min_lon, max_lon = np.min(coords[:, 1]), np.max(coords[:, 1])
    lat_interval = (max_lat - min_lat) / (len(np.unique(coords[:, 0])) - 1)
    lon_interval = (max_lon - min_lon) / (len(np.unique(coords[:, 1])) - 1)
    latitudes = np.arange(min_lat, max_lat + lat_interval*0.1, lat_interval)  # Add a small buffer to include max value
    longitudes = np.arange(min_lon, max_lon + lon_interval*0.1, lon_interval)  # Add a small buffer to include max value

    with nc.Dataset(filename, 'w', format='NETCDF4') as dataset:
        dataset.createDimension('latitude', len(latitudes))
        dataset.createDimension('longitude', len(longitudes))
        
        lat_var = dataset.createVariable('latitude', 'f4', ('latitude',))
        lon_var = dataset.createVariable('longitude', 'f4', ('longitude',))
        lat_var[:] = latitudes
        lon_var[:] = longitudes
        
        for i, period in enumerate(return_periods):
            grid_inten_stats = interpolate_to_grid(latitudes, longitudes, coords, inten_stats[i, :], method='linear')
            intensity_var_name = f'intensity_RP{period}'
            intensity_var = dataset.createVariable(intensity_var_name, 'f4', ('latitude', 'longitude'))
            intensity_var[:, :] = grid_inten_stats
            intensity_var.units = hazard.units
            intensity_var.description = f'Exceedance intensity map for {period}-year return period'
            
        dataset.description = 'Exceedance intensity data for various return periods'
        lat_var.units = 'degrees_north'
        lon_var.units = 'degrees_east'
`

ValentinGebhart:
Changes proposed in this PR (new version of old PR #857):
- addition of new function to compute local return period maps `local_return_period` and `_loc_return_period` for given hazard intensities, giving out a geodataframe
- addition of function `subplots_from_gdf` in util.plot that plots different subplots from the columns of a gdf

Earlier parts removed (commented out in code) until agreement:
- addition of new functions to store the hazard maps that are generated in the Hazard.plot_rp_intensity() function as raster files `write_raster_local_exceedance_inten.` (do we also need as netCDF files `write_netcdf_local_exceedance_inten`?)
- addition of new function to store these local return period maps for given hazard intensities as raster files `write_raster_local_return_periods` or as NetCDF files `write_netcdf_local_return_periods`.

This PR fixes #854
(new version of old PR #857)
### PR Author Checklist

### PR Reviewer Checklist


climada/hazard/plot.py
```diff
@@ -65,6 +67,45 @@ def plot_rp_intensity(self, return_periods=(25, 50, 100, 250),
...
+            Figure size for plt.subplots.
+        kwargs: optional
+            Arguments for pcolormesh matplotlib function used in event plots.
+    
+        Returns
+        -------
+        axis: matplotlib.axes._subplots.AxesSubplot
+            Matplotlib axis with the plot.
+        """
+        ### code to replace self._set_coords_centroids()
```
ValentinGebhart, line 92: Do we need the following six lines? (the ones replacing self._set_coords_centroids())
I mean, can it happen that the meta data exists but the centroids (or their coordinates) are not set? Also, I think one cannot manually set the centroids.geometry attribute anymore, right?

chahank: In the latest version of Centroids (is part of develop, will be in the next release), there is no more meta attribute. Thus, I would say that you can remove these lines.

ValentinGebhart: Ok perfect.
The write_raster_... functions in io.py however use the get_meta() method. That should be fine, right?

climada/hazard/io.py
```diff
@@ -1038,6 +1039,128 @@ def write_hdf5(self, file_name, todense=False):
...
+            Array or list of return periods (in years) for which to calculate 
+            and store exceedance intensities.
+        filename : str
+            Path and name of the file to write in tif format.
+        """
+        inten_stats = self.local_exceedance_inten(return_periods=return_periods)
+        num_bands = inten_stats.shape[0]
+        
+        if not self.centroids.get_meta():
+            raise ValueError("centroids.get_meta() is required but not set.")
```
chahank, line 1059: What is this line for? `centroids.get_meta()` always returns the meta. Please see the updated centroids methods on develop.

climada/hazard/test/test_base.py
```diff
@@ -1022,6 +1022,25 @@ def test_ref_all_pass(self):
...
+        threshold_intensities = np.array([1., 2., 4.])
+        return_stats = haz.local_return_period(threshold_intensities)
+        np.testing.assert_allclose(
+            return_stats,
+            np.array([
+                [0.66666667, 0.5, 0.66666667],
+                [1., 0.66666667, 0.66666667],
+                [2., np.nan, np.nan]
+                ])
+        )
```
chahank, line 1043: Can you make a simpler test? Why use numbers that are hard to compute instead of a simple set of numbers? Ideally, the test should be immediately clear. Here one needs to take a piece of paper and compute before knowing what is going on.

Also, you should test the method `_local_return_period`. 

Also, you should test that the chunking is working as intended.

climada/hazard/test/test_io.py
```diff
@@ -682,6 +686,75 @@ class CustomID:
         self.assertTrue(np.array_equal(hazard.date, hazard_read.date))
         self.assertTrue(np.array_equal(hazard_read.event_id, np.array([])))  # Empty array
 
+class TestWriteExceedAndRP(unittest.TestCase):
+    """Test writing raster and netCDF files from exceedance intensitiy and return period maps""" 
+
+    def test_write_raster_exceed_inten(self):
+        """Test write TIFF file from local exceedance intensity"""
+        self.temp_dir = TemporaryDirectory()
+        self.test_file_path = os.path.join(self.temp_dir.name, 'test_file.tif')
```
chahank, line 695: This works, but I think using `Pathlib` instead of `os.path` is cleaner.

climada/hazard/io.py
```diff
@@ -1038,6 +1039,128 @@ def write_hdf5(self, file_name, todense=False):
...
+
+        ### this code is to replace pixel_geom = self.centroids.calc_pixels_polygons()
+        if abs(abs(self.centroids.get_meta()['transform'].a) -
+               abs(self.centroids.get_meta()['transform'].e)) > 1.0e-5:
+            raise ValueError('Area can not be computed for not squared pixels.')
```
chahank, line 1064: Do you know why this is needed? Can't you save the raster directly with the values and the meta attribute? Why do you need to compute the area?

climada/hazard/io.py
```diff
@@ -1038,6 +1039,128 @@ def write_hdf5(self, file_name, todense=False):
...
+        pixel_geom = self.centroids.geometry.buffer(self.centroids.get_meta()['transform'].a / 2).envelope
```
chahank, line 1065: Why not use the `self.centroids.get_pixel_shapes` ?

chahank (changes requested):
Thanks for the updated clean version of this PR! I think this will be a quite useful set of methods.

I have 2 general points that itch me at the moment.
- I am not sure it makes sense that the plotting and saving methods always call and recalculate the values. Maybe we should rather store the values as a new attribute. Do you know how much computation this costs for large datasets? @emanuel-schmid @peanutfun : what do you think, would it be ok if the method `local_return_period` adds an attribute to the hazard object? 
- I find the method `local_return_period` and the method `_loc_return_period` quite hard to read at the moment. I cannot say exactly why, but probably a mix of not always clear variable names, hidden changing of variables (like the return_periods), and quite some loops.

I thus suggest to a) wait for @emanuel-schmid and @peanutfun opinion about the attribute b) rewrite the code to avoid updating the return periods in the back.

climada/hazard/base.py
```diff
@@ -613,6 +653,44 @@ def _loc_return_inten(self, return_periods, inten, exc_inten):
             exc_inten[:, cen_idx] = self._cen_return_inten(
                 inten_sort[:, cen_idx], freq_sort[:, cen_idx],
                 self.intensity_thres, return_periods)
+            
+    def _loc_return_period(self, threshold_intensities, inten, return_periods):
+        """Compute local return periods for given hazard intensities for a specific chunk of data.
```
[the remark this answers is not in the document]
peanutfun: I agree that it's much better to return a new array rather than modifying input parameters

peanutfun:
> I am not sure it makes sense that the plotting and saving methods always call and recalculate the values. Maybe we should rather store the values as a new attribute. Do you know how much computation this costs for large datasets? @emanuel-schmid @peanutfun : what do you think, would it be ok if the method local_return_period adds an attribute to the hazard object?

I agree that the values should *not* be calculated in the plotting method. Instead, the user should insert the data to plot into the appropriate function.

I very much dislike adding new attributes after the class initialization. I would return the data to the user, who should take care of "bookkeeping" themselves. They can then plug the values into a plotting function, store them separately, etc.

ValentinGebhart:
> > I very much dislike adding new attributes after the class initialization. I would return the data to the user, who should take care of "bookkeeping" themselves. They can then plug the values into a plotting function, store them separately, etc.
> 
> I also agree with you that adding new attributes after the class initialization is a bad habit. Here the point is that plotting the intensity return period on the map requires the centroids, and thus having it in the hazard object is convenient... not sure what the best solution is.

One way of returning data to the user without losing the centroids could be to return a geodataframe?

climada/hazard/base.py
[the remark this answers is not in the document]
ValentinGebhart: I agree with restructuring these functions as you say. I believe much of the structure of `_loc_return_period` and `local_return_period` was carried over from the quite old functions `_loc_return_inten` and `local_exceedance_inten`. 
If I find a better structure, would it make sense also to adapt the older functions?

peanutfun:
@ValentinGebhart Great idea! Returning a GeoDataFrame with the return period information and the centroids geometry should contain all data necessary for the plot.

emanuel-schmid:
@ValentinGebhart I agree the idea is not bad - but returning a GeoDataFrame is still somewhat fuzzy. A cleaner solution would be to return a tuple (intensity return period, centroids geometry) or a class with two attributes, intensity_return_period and geometry.
This seems cleaner. Ideally any requirement should be explicit in a method's signature and not hidden away in an object that could contain anything at all, like a dict or a dataframe.

chahank:
@emanuel-schmid I strongly disagree with this take as is. The proposed outputs (tuples or custom class) maybe would be cleaner from a programming point of view, but it would make the output mostly useless from a user perspective. I do not see the problem with giving back a geodataframe. Can you maybe explain why this would be a bad idea? 

Besides, we already have a precedent with the method `impact_at_reg` that returns a DataFrame :D

emanuel-schmid:
Yeah. Agreed. Although I must stress that the principle holds - in this particular case a dataframe seems alrightish. 😁

chahank:
> Yeah. Agreed. Although I must stress that the principle holds - in this particular case a dataframe seems alrightish. 😁

Indeed. I think there is a general question on how climada objects should return extra information. We had the questions about Centroids and things like pixel sizes, Impacts and regional values, or here Hazards and return period.

ValentinGebhart:
In the new commit `local_return_period` outputs a gdf and I add a function `subplots_from_gdf` to util_plot that plots subplots from the different columns of a gdf. I thought it would be nice to have this plot function as general as possible (so it could plot return periods and exceedance frequencies and more), but if this is too messy, we can also write a more specific "plot_return_period" function. 

I was not sure how to best include the meta data about what is written in the gdf and the corresponding units in the gdf. We decided to try the gdf.columns.name attribute, but I feel that how I did it now might not be the best way. To be discussed.

I also removed the write functions (only commented out for now) because we might want to generalize them to general gdf.

climada/hazard/base.py
```diff
@@ -450,6 +450,46 @@ def sanitize_event_ids(self):
...
+            Array containing computed local return periods for given hazard intensities.
+        """
+        # Ensure threshold_intensities is a numpy array
+        threshold_intensities = np.array(threshold_intensities)
+        
+        num_cen = self.intensity.shape[1]
+        return_periods = np.zeros((len(threshold_intensities), num_cen))  # Adjusted for 2D structure
+        
+        # Process each centroid in chunks as in local_exceedance_inten
+        cen_step = CONFIG.max_matrix_size.int() // self.intensity.shape[0]
```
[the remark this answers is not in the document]
ValentinGebhart: I think it should be fine? (max_matrix size) // (number of events) gives the number of centroids that can be handled at the same time?

climada/hazard/base.py
```diff
@@ -450,6 +450,46 @@ def sanitize_event_ids(self):
...
+            self._loc_return_period(
+                threshold_intensities,
+                self.intensity[:, chk * cen_step:(chk + 1) * cen_step].toarray(),
+                return_periods[:, chk * cen_step:(chk + 1) * cen_step])
+        
+        if (chk + 1) * cen_step < num_cen:  # Check if there's a remainder
+            self._loc_return_period(
+                threshold_intensities,
+                self.intensity[:, (chk + 1) * cen_step:].toarray(),
+                return_periods[:, (chk + 1) * cen_step:])
```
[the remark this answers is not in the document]
ValentinGebhart: I should be a bit more elegant now.

climada/hazard/base.py
```diff
@@ -450,6 +450,46 @@ def sanitize_event_ids(self):
...
+        Returns
+        -------
+        return_periods : np.array
+            Array containing computed local return periods for given hazard intensities.
+        """
+        # Ensure threshold_intensities is a numpy array
+        threshold_intensities = np.array(threshold_intensities)
+        
+        num_cen = self.intensity.shape[1]
+        return_periods = np.zeros((len(threshold_intensities), num_cen))  # Adjusted for 2D structure
```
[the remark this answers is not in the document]
ValentinGebhart: I don't know. The code should be fine, I have deleted the comment

climada/hazard/test/test_base.py
ValentinGebhart: I simplified the numbers in the test. 
I guess the test is a test of both local_return_period and _loc_return_period? Also, I don't know how to test the chunking with a "simple" test? Then I would need to use a hazard large enough that the chunking is activated? Feedback welcome :)

ValentinGebhart:
After the changes (see above comment) a new review is requested.

climada/hazard/plot.py
```diff
@@ -65,6 +67,37 @@ def plot_rp_intensity(self, return_periods=(25, 50, 100, 250),
...
+        threshold_intensities: np.array
+            Hazard intensities to consider for calculating return periods.
+        smooth: bool, optional
+            Smooth plot to plot.RESOLUTION x plot.RESOLUTION.
+        axis: matplotlib.axes._subplots.AxesSubplot, optional
+            Axis to use.
+        figsize: tuple, optional
+            Figure size for plt.subplots.
+        kwargs: optional
+            Arguments for pcolormesh matplotlib function used in event plots.
```
chahank, line 85: Some parameters are missing.

climada/hazard/plot.py
```diff
@@ -65,6 +67,37 @@ def plot_rp_intensity(self, return_periods=(25, 50, 100, 250),
                                         colbar_name, title, smooth=smooth, axes=axis,
                                         figsize=figsize, adapt_fontsize=adapt_fontsize, **kwargs)
         return axis, inten_stats
+    
+    def plot_local_rp(self, threshold_intensities, smooth=True, axis=None, figsize=(9, 13), adapt_fontsize=True, cmap = 'viridis_r', **kwargs):
```
chahank, line 71: Formatting.

climada/hazard/plot.py
```diff
@@ -65,6 +67,37 @@ def plot_rp_intensity(self, return_periods=(25, 50, 100, 250),
...
+            Axis to use.
+        figsize: tuple, optional
+            Figure size for plt.subplots.
+        kwargs: optional
+            Arguments for pcolormesh matplotlib function used in event plots.
+    
+        Returns
+        -------
+        axis: matplotlib.axes._subplots.AxesSubplot
+            Matplotlib axis with the plot.
```
chahank, line 90: Incomplete.

climada/hazard/plot.py
chahank, line 71: So, if a user wants to adapt the say color map, they must recompute the return periods every time? This seems rather inefficient. Would it not make sense to be able to pass the return periods to this function? Something like

```
plot_local_rp(self, threshold_intensities, return_periods=None, smooth=True, axis=None, figsize=(9, 13), adapt_fontsize=True, cmap = 'viridis_r', **kwargs):
```

Then, if return_periods is none, compute them. Else, use the return periods to make the graph. Maybe throw a warning if the threshold_itnensities are not in the gdf.

climada/hazard/test/test_base.py
```diff
@@ -1022,6 +1022,27 @@ def test_ref_all_pass(self):
...
+            ])
+        haz.frequency = np.full(4, 1.)
+        threshold_intensities = np.array([1., 2., 4.])
+        return_stats, _ = haz.local_return_period(threshold_intensities)
+        np.testing.assert_allclose(
+            return_stats[return_stats.columns[1:]].values.T,
+            np.array([
+                [0.25, 0.25, 0.33333333],
+                [0.33333333, 0.25, np.nan],
+                [np.nan, 1., np.nan]
```
chahank, line 1043: Are these the correct values? The frequencies are all `1/year`. So, an event with at least intensity 1 happens every year for centroids 1 and 2, and 3/4 years for centroid 3. Am I missing something?

climada/hazard/test/test_io.py
```diff
@@ -21,8 +21,10 @@
 import unittest
 from unittest.mock import patch
 import datetime as dt
+import os
```
chahank, line 24: Why? I do not see it used in this PR.

climada/hazard/test/test_io.py
```diff
@@ -21,8 +21,10 @@
...
 from pathlib import Path
 from tempfile import TemporaryDirectory
+import rasterio
```
chahank, line 27: Why? I do not see it used in this PR.

climada/test/test_plot.py
```diff
@@ -105,6 +105,14 @@ def test_hazard_rp_intensity(self):
...
         self.assertEqual('Return period: 25 years', axis1.get_title())
         self.assertEqual('Return period: 50 years', axis2.get_title())
+    
+    # def test_plot_local_rp(self):
+    #     """"Plot local return period maps for different hazard intensities"""
+    #     hazard = Hazard.from_hdf5(HAZ_TEST_TC)
+    #     (axis1, axis2), return_stats = hazard.plot_local_rp([10., 20.])
+    #     self.assertEqual('Intensity: 10.0 m/s', axis1.get_title())
+    #     self.assertEqual('Intensity: 20.0 m/s', axis2.get_title())
+    #     np.testing.assert_array_equal(return_stats.shape, (2, 100))
```
chahank, line 115: ?

climada/util/test/test_plot.py
```diff
@@ -25,6 +25,8 @@
 import matplotlib.pyplot as plt
 from matplotlib import colormaps as cm
 import cartopy.crs as ccrs
+from geopandas import GeoDataFrame
```
chahank, line 28: It think it is better to import geopandas instead.

climada/util/plot.py
```diff
@@ -874,3 +876,73 @@ def multibar_plot(ax, data, colors=None, total_width=0.8, single_width=1,
     # Draw legend if we need
     if legend:
         ax.legend(bars, data.keys())
+
+# Create GdfMeta class (as a named tuple) for GeoDataFrame meta data
+GdfMeta = namedtuple('GdfMeta', ['name', 'unit', 'col_name', 'col_unit'])
```
chahank, line 881: hmmm I am not sure whether this is a good idea. Maybe it is, maybe not. 

@emanuel-schmid @peanutfun What do you think? Should we add some nametuple in the utility plot for a specific case of geopandas plotting?

chahank (changes requested):
Thanks for this simplified version! It makes it much clearer. 

There are a few main points to discuss:
~a) is it computing what it should? The test looks incorrect to me, but I probably have mixed up something. Can you please enlighten me?~ Did it myself.
b) I am not sure whether the GdfMeta is a good idea.
c) I am not sure whether the chunking is working as intented (and whether it is needed at all.)

climada/util/plot.py
chahank: To add more details: this is called very generically `GdfMeta`, but refers to only one very specific data set of geodataframes.

chahank:
> In the new commit, `local_return_period` outputs a GeoDataFrame and gdf_meta (named tuple with plotting metadata information) and I add a function `subplots_from_gdf` to util.plot that plots subplots from the different columns of a gdf (using the gdf_meta information). I thought it would be nice to have this plot function as general as possible (so it could plot return periods and exceedance frequencies and more), but if this is too messy, we can also write a more specific "plot_return_period" function.

I like the generality! I only fear that the GdfMeta might make it not generic. 

> After discussion with Lukas, I create the meta data of the gdf as a named tuple, such that it is clear which attributes gdf_meta must have to correctly plot.
> 
> I removed the write functions (and their tests) because we might want to generalize them to general gdf. I made a backup if we change our minds.

Sounds good.

climada/util/plot.py
peanutfun: I suggested this to pass information to the plot that is not explicitly contained in the GeoDataFrame. `GdfMeta` can be renamed, of course. But I think a `namedtuple` makes more sense than a `dict`, because you can be sure which keys/attributes are defined. `col_name` specifies the name of the column to plot. But I agree that the names can be adjusted and made clearer.

chahank: I like the idea of passing some meta data for the plotting. I am not sure that this is the perfect way this way :D At first I thought that `GdfMeta` some new property of `GeoDataFrames`. Also, is the current proposal general enough for a generic geopandas plotting method in CLIMADA? Or should it then be made more specific.

And, `namedtuple` makes sense indeed.

climada/hazard/base.py
```diff
@@ -450,6 +452,60 @@ def sanitize_event_ids(self):
             LOGGER.debug('Resetting event_id.')
             self.event_id = np.arange(1, self.event_id.size + 1)
 
+    def local_return_period(self, threshold_intensities):
+        """Compute local return periods for given hazard intensities.
+    
+        Parameters
+        ----------
+        threshold_intensities : np.array
+            Hazard intensities to consider.
```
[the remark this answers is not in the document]
chahank: hmmm I think the word `test` would be quite confusing actually. Maybe just `intensities`? Or even avoid the plural and use `intensity`? Or keep `threshold_intensities` but with a clearer docstring.

climada/test/test_plot.py
ValentinGebhart: sorry forgot to delete this test

climada/util/plot.py
ValentinGebhart: I agree that the naming is not optimal. The included information should be: what quantity represent the values (+ units) in the gdf, what quantity represent the values (+ units) in the column labels. The rows always correspond to centroids. Maybe instead of `name` we could use `quantity` or `value_type`?

Otherwise, the plotting function should be "generic", in the sense that it produces one subplot per column of a gdf, using the values to color the different points, and using the gdf_meta varibale for figure and axis label. Of course, it is still using the geo_im_from_array util function, and I'm sure there are other ways to plot a gdf with several columns. 
Let me know what you think

climada/hazard/base.py
```diff
@@ -450,6 +452,60 @@ def sanitize_event_ids(self):
...
+        Returns
+        -------
+        gdf : gpd.GeoDataFrame
+            GeoDataFrame containing return periods for given threshold intensity
+            description of units of values and threshold intensities in gdf.columns.name
+        """
+        #check frequency unit
+        if self.frequency_unit not in ['1/year', 'annual', '1/y', '1/a']:
+            LOGGER.warning("The Hazard's frequency unit is %s and not %s which "
+                "most likely leads to incorrect results",
```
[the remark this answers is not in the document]
ValentinGebhart: I agree, I could just use the inverse of the frequency attribute. I now added using a if elif statement that can read years, months, and weeks. I don't know how to make this more generic and easier because the frequency attribute is just a string and not a "physical" frequency unit that can be converted eg to years.

climada/util/plot.py
ValentinGebhart: @peanutfun sorry I didn't see your suggestion in time. We can definitely add the labels and units as parameter(s) to the plot function. Then the user has to know about what their gdf represents without us helping by providing the information, which I guess is ok but is a bit more work to the user. If we put the as defaults to the plot function, the plot function is not generic anymore.

climada/hazard/base.py
```diff
@@ -450,6 +452,60 @@ def sanitize_event_ids(self):
...
+        if not block_size:
+            raise ValueError('Increase max_matrix_size configuration parameter to >'
+                             f'{self.intensity.shape[0]}')
+        
+        for start_col in range(0, num_cen, block_size):
+            end_col = min(start_col + block_size, num_cen)
+            return_periods[:, start_col:end_col] = self._loc_return_period(
+                threshold_intensities,
+                self.intensity[:, start_col:end_col].toarray()
+                )
```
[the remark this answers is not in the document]
ValentinGebhart: sorry, maybe I didn't understand what max_matrix_size represents. Is it the number of columns, the number of rows, or the (number cols) x (number rows). I thought it was the last one. 

I also think that the chunking is not needed, especially because I wanted to rewrite these functions using the general fit method discussed in issue #904. I though we'd just clean this up there, but if you say it's ok I can remove the chunking already here

climada/hazard/plot.py
ValentinGebhart: sorry, I forgot to remove this function. The functionality is now in `subplots_from_gdf`

climada/util/plot.py
emanuel-schmid: Basically along the same lines as @peanutfun 
I'd suggest to have two parameters instead of `gdf_meta: GdfMeta`, 
`colbar_name: str` and `title_subplots: lambda`.

To get the same result as implemented now you would instead of

```python
subplots_from_gdf(
    gdf=gdf, 
    gdf_meta=GdfMeta(name='Return Period', unit='Years', col_name='Threshold Intensity', col_unit='m/s')
)
```
you would

```python
subplots_from_gdf(
    gdf=gdf, 
    colbar_name="Return Period (Years)", 
    title_subplots=lambda column: f"Threshold Intensity: {column} m/s"),
)
```

Seems better readable and has more flexibilty

emanuel-schmid: consequently, I'd let the `local_return_period` return a triple
```python
Returns
-------
gdf : GeoDataFrame
label : str
    GeoDataFrame label
column_label : lambda
    Column label
```

climada/hazard/test/test_io.py
```diff
@@ -32,7 +32,9 @@
 from climada.hazard.base import Hazard
 from climada.util.constants import DEF_FREQ_UNIT, HAZ_TEMPLATE_XLS, HAZ_DEMO_FL, DEF_CRS
 from climada.hazard.test.test_base import DATA_DIR, dummy_hazard
+from climada.test import get_test_file
 
+HAZ_TEST_TC :Path = get_test_file('test_tc_florida')
```
chahank, line 37: I think this is not part of this PR.

chahank (changes requested):
Good work! Please finalize the cosmetics: the linter complains about trailing white spaces to be removed and some indention that is incorrect. There is one file that is modified by mistake I think.

 Also, don't forget to add your name to the list of authors: https://github.com/CLIMADA-project/climada_python/blob/feature/write_haz_rp_maps2/AUTHORS.md

ValentinGebhart:
> Good work! Please finalize the cosmetics: the linter complains about trailing white spaces to be removed and some indention that is incorrect. There is one file that is modified by mistake I think.
> 
> Also, don't forget to add your name to the list of authors: https://github.com/CLIMADA-project/climada_python/blob/feature/write_haz_rp_maps2/AUTHORS.md

Thanks for the review(s)! I now understood what the trailing white space means :) For me this looks ready to merge, let me know if there is something else to do.

chahank:
This is good, feel free to merge!

PR #929, opened 2024-07-30
plot NaNs in gray color and use log color bar

ValentinGebhart:
Changes proposed in this PR:
- In the plot function `geo_im_from_array`, NaN values in the data will be plotted in gray. Before, NaN value were not plotted (i.e. transparent), making them indistinguishable from plot regions for which there is no data (no centroids).
- In the plot function `plot_from_gdf`, the colorbar with will be shown on a logarithmic scale if a) the gdf is about return periods or impacts, b) there are no zeros in the data, c) the span of the data's values are at least two orders of magnitude

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- In `climada.util.plot.geom_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted, and renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+
 ### Fixed
 
 ### Deprecated
@@ -89,6 +91,8 @@ CLIMADA tutorials. [#872](https://github.com/CLIMADA-project/climada_python/pull
 - climada.hazard.centroids.centr.Centroids.to_default_crs
 - climada.hazard.centroids.centr.Centroids.write_csv
 - climada.hazard.centroids.centr.Centroids.write_excel
+- climada.hazard.local_return_period [#898](https://github.com/CLIMADA-project/climada_python/pull/898)
+- climada.util.plot.subplots_from_gdf [#898](https://github.com/CLIMADA-project/climada_python/pull/898)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
-- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted, and renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+ - Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
-- In `climada.util.plot.geom_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted, and renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted, and renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Code freeze date: YYYY-MM-DD
 ### Changed
 
 - In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
- - Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+- Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
 
 ### Fixed
 
--- a/climada/util/plot.py
+++ b/climada/util/plot.py
@@ -342,6 +342,9 @@ def geo_im_from_array(array_sub, coord, var_name, title,
                 extent[2]: extent[3]: complex(0, RESOLUTION)]
             grid_im = griddata((coord[:, 1], coord[:, 0]), array_im,
                                (grid_x, grid_y))
+            # Create mask for original NaN values in the data
+            nan_mask = griddata((coord[:, 1], coord[:, 0]), np.isnan(array_im.astype(float)),
+                                (grid_x, grid_y), fill_value=0).astype(bool)
         else:
             grid_x = coord[:, 1].reshape((width, height)).transpose()
             grid_y = coord[:, 0].reshape((width, height)).transpose()
@@ -350,6 +353,7 @@ def geo_im_from_array(array_sub, coord, var_name, title,
                 grid_y = np.flip(grid_y)
                 grid_im = np.flip(grid_im, 1)
             grid_im = np.resize(grid_im, (height, width, 1))
+            nan_mask = np.isnan(grid_im)
         axis.set_extent((extent[0] - mid_lon, extent[1] - mid_lon,
                          extent[2], extent[3]), crs=proj)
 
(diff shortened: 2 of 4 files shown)


ValentinGebhart:
> @ValentinGebhart Thank you for that contribution. Can you share an example and compare the resulting plots before and after your changes?

This is an example of plotting the return periods of a hazard object where there are some NaNs (because the centroid had never seen the given threshold intensity, so the return period is given as NaN), and some centroids are removed (left bottom corner). This is the code: 

import numpy as np
from climada.util import HAZ_DEMO_H5 # CLIMADA's Python file
# Hazard needs to know the acronym of the hazard type to be constructed!!! Use 'NA' if not known.
haz_tc_fl = Hazard.from_hdf5(HAZ_DEMO_H5)  # Historic tropical cyclones in Florida from 1990 to 2004
haz_tc_fl.check() # Use always the check() method to see if the hazard has been loaded correctly

# remove some centroids
centroids_mask = np.array(
    [ (i + j > 10) for j in range(50) for i in range(50)]
)
haz_tc_fl.centroids = haz_tc_fl.centroids.select(sel_cen=centroids_mask)
haz_tc_fl.intensity = haz_tc_fl.intensity[:, -2434:]

return_periods, label, column_label = haz_tc_fl.local_return_period([30, 40])

from climada.util.plot import plot_from_gdf
plot_from_gdf(return_periods, colorbar_name=label, title_subplots=column_label)

ValentinGebhart:
old plots
old

ValentinGebhart:
new plots
new

ValentinGebhart:
Note that if the value range of the hazard return periods would have extended two orders of magnitude (without having zeros), the color scale would also be logarithmic in the new plots

peanutfun (changes requested):
I very much like the overall contribution, but I have to say I dislike the approach. Calling `griddata` again on basically the same data, casting to bool, plotting with a weird colormap...

I think the same thing can be achieved much easier, using all the tools of Matplotlib. You can set "bad" and "over/under" colors for a colormap. Choosing the right `vmin` should then give you the expected outcome with a single call to `pcolormesh`

```python
# ...
if "norm" in kwargs:
    min_value = kwargs["norm"].vmin
    vmin = None  # We will pass norm
else:
    min_value = np.nanmin(array_im)
    vmin = kwargs.pop("vmin", min_value)

grid_im = griddata((coord[:, 1], coord[:, 0]), array_im, (grid_x, grid_y), fill_value = min_value-1)

# ...
cmap = plt.get_cmap(kwargs.pop("cmap", "viridis"))
cmap.set_bad("gray")  # For NaNs and infs
cmap.set_under("white", alpha=0)  # For values below vmin

axis.pcolormesh(
    grid_x - mid_lon,
    grid_y,
    np.squeeze(grid_im),
    transform=proj,
    cmap=cmap,
    vmin=vmin,
    **kwargs
)
```

climada/util/plot.py
```diff
@@ -927,20 +935,29 @@ def subplots_from_gdf(
...
+        colorbar_name.strip().startswith(('Return Period', 'Impact')) and
+        'norm' not in kwargs.keys() and
+        # check if there are no zeros values in gdf
+        np.sum(gdf_values == 0) == 0 and
+        # check if value range too small for logarithmic colorscale
+        (np.log10(np.nanmax(gdf_values)) - np.log10(np.nanmin(gdf_values))) > 2
+    ):
+        kwargs.update(
+            {'norm': mpl.colors.LogNorm(
+                vmin=gdf.values[:,1:].min(), vmax=gdf.values[:,1:].max()
```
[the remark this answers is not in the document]
ValentinGebhart: Thanks for the advice! I agree that the way you describe is easier. I implemented and tested it (example plots from above didn't change), with a small modification for the case of the log colorscale.

ValentinGebhart:
Thanks for the advice! I agree that the way you describe is easier. I implemented and tested it (example plots from above didn't change), with a small modification for the case of the log colorscale (min_value - 1 did not seem to work, so I used min_value/2).

climada/util/plot.py
```diff
@@ -328,9 +328,21 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    # prepare colormap
+    cmap = plt.get_cmap(kwargs.pop("cmap", CMAP_RASTER))
+    cmap.set_bad("gainsboro")  # For NaNs and infs
+    cmap.set_under("white", alpha=0)  # For values below vmin
+
+    # prepare boundary values
+    if "norm" in kwargs:
+        min_value = kwargs["norm"].vmin
+        vmin = None  # We will pass norm
+        del kwargs["vmin"], kwargs["vmax"]
```
peanutfun, line 340: Matplotlib will throw an error if vmin/vmax *and* norm are passed. We should not handle that silently.
```suggestion
        del kwargs["vmin"], kwargs["vmax"]
```

climada/util/plot.py
```diff
@@ -328,9 +328,21 @@ def geo_im_from_array(array_sub, coord, var_name, title,
...
+    else:
+        min_value = np.nanmin(array_sub)
+        vmin = kwargs.pop("vmin", min_value)
+    below_value = min_value/2 if min_value > 0 else min_value-1
```
peanutfun, line 344: `min_value` is not needed anymore afterwards, so we do not need to define a new variable
```suggestion
    min_value = min_value/2 if min_value > 0 else min_value-1
```

climada/util/plot.py
```diff
@@ -340,8 +352,11 @@ def geo_im_from_array(array_sub, coord, var_name, title,
             grid_x, grid_y = np.mgrid[
                 extent[0]: extent[1]: complex(0, RESOLUTION),
                 extent[2]: extent[3]: complex(0, RESOLUTION)]
-            grid_im = griddata((coord[:, 1], coord[:, 0]), array_im,
-                               (grid_x, grid_y))
+            grid_im = griddata(
+                (coord[:, 1], coord[:, 0]),
+                array_im,
+                (grid_x, grid_y),
+                fill_value=below_value)
```
peanutfun, line 359: see above
```suggestion
                fill_value=min_value,
            )
```

climada/util/plot.py
```diff
@@ -927,23 +949,32 @@ def subplots_from_gdf(
...
-                    vmin=gdf.values[:,1:].min(), vmax=gdf.values[:,1:].max()
-                    ),
-                'vmin': None, 'vmax': None}
-            )
+    # use log colorbar for return periods and impact
+    if (
+        colorbar_name.strip().startswith(('Return Period', 'Impact')) and
+        'norm' not in kwargs.keys() and
+        # check if there are no zeros values in gdf
+        np.sum(gdf_values == 0) == 0 and
```
peanutfun, line 957: More explicit and less error-prone than summing booleans
```suggestion
        not np.any(gdf_values == 0)
```

peanutfun (changes requested):
Okay, this is looking much better now, thanks for the update! I have a few nitpicky suggestions still 🙈 We can merge once these are resolved!

climada/util/plot.py
ValentinGebhart: This is why I am deleting these entries here. Earlier in the function (lines 312-315) we add the vmin/vmax kwargs, but as you say if we also input norm this will result in an error. How do you think we should handle this?

peanutfun: Woops, I was not aware of this 😅 I have to say these four lines confuse me a bit, because it should be the default behavior of any matplotlib norm to use `nanmin` and `nanmax` of the inserted values as `vmin` and `vmax`, respectively. So this looks like a no-op to me. Can you check if we can simply remove them?

ValentinGebhart: I believe the reason for these lines is that we want to have that same colorbar for all the subplots, not a colorbar that changes its min/max bounds depending on the subplot? At least this is what happens if I comment about the lines. I assume there could be a more elegant way to code this, do you have a suggestion?

CHANGELOG.md
```diff
@@ -89,6 +91,8 @@ CLIMADA tutorials. [#872](https://github.com/CLIMADA-project/climada_python/pull
 - climada.hazard.centroids.centr.Centroids.to_default_crs
 - climada.hazard.centroids.centr.Centroids.write_csv
 - climada.hazard.centroids.centr.Centroids.write_excel
+- climada.hazard.local_return_period [#898](https://github.com/CLIMADA-project/climada_python/pull/898)
+- climada.util.plot.subplots_from_gdf [#898](https://github.com/CLIMADA-project/climada_python/pull/898)
```
peanutfun, line 95:
```suggestion
- climada.util.plot.plot_from_gdf [#898](https://github.com/CLIMADA-project/climada_python/pull/898) [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted, and renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
```
peanutfun, line 17:
```suggestion
- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
 - Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
```

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
+ - Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
```
peanutfun, line 18:
```suggestion
- Renamed `climada.util.plot.subplots_from_gdf` to `climada.util.plot.plot_from_gdf` [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
```

peanutfun (approved):

PR #948, opened 2024-09-17
Bug fix in Hazard.get_mdr and impact calculation in case of a single exposure point and MDR values of 0

### LINKED ISSUE #933: Possible bug in impact calculation
**Describe the bug**
I noticed a weird behavior in the impact calculation. Inputting a valid exposure, hazard and impact function set leads to:

> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.

the error occurs at [this](https://github.com/CLIMADA-project/climada_python/blob/main/climada/engine/impact_calc.py#L412) line, i.e. when looping the impact matrix generator. After spending few hours, I still couldn't figure out what conditions lead to the error.

The code below reproduces it using one exposure point, one centroid and two hazard events. Interestingly enough, using an exposure with twice the same point makes it run just fine.

**To Reproduce**
Run the code below.

Code example:
```python
import numpy as np
import pandas as pd
from scipy import sparse
import geopandas as gpd
from shapely.geometry import Point

from climada.entity import Exposures
from climada.entity.impact_funcs import ImpactFuncSet, ImpfTropCyclone
from climada.engine import ImpactCalc
from climada.hazard import Centroids, Hazard

intensity = np.array([[31.5], [19.0]])
centroids = Centroids.from_lat_lon([-26.16], [28.20])

tc = Hazard(
    intensity=sparse.csr_matrix(intensity),
    event_id=np.arange(2),
    event_name=[0,1],
    frequency=np.ones(2) / 2,
    fraction=sparse.csr_matrix(np.zeros((2,1))),
    date=np.array([736018, 736019]),
    centroids=centroids,
    haz_type='TC'
)

impf_tc = ImpfTropCyclone.from_emanuel_usa()
impf_set = ImpactFuncSet([impf_tc])
impf_set.check()

exposure = gpd.GeoDataFrame(
    {'value': 1., 'geometry': [Point(28.22, -26.17)]}, crs="EPSG:4326"
)
exposure['latitude'] = exposure.geometry.y
exposure['longitude'] = exposure.geometry.x
exposure['impf_TC'] = 1

exp = Exposures(exposure)
#exp = Exposures(pd.concat([exposure, exposure]).reset_index()) # Use this and it works

impact = ImpactCalc(exp, impf_set, tc).impact(save_mat=True)
```

**Expected behavior**
The expected behaviour would be to get an impact object, the actual result is: 
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.

**Climada Version:** 4.1.2.dev0

 - Operating system and version: macOS Sonoma 14.6
 - Python version: 3.11.6

Any idea/clue?

bguillod:
Changes proposed in this PR:
- Fixes a bug in the impact calculation for cases with a single exposure point and a zero MDR value.

This PR fixes #933

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Code freeze date: YYYY-MM-DD
 
 ### Fixed
 
+- Avoids a ValueError in the impact calculation for cases with a single exposure point and MDR values of 0, by explicitly removing zeros in `climada.hazard.Hazard.get_mdr` [#933](https://github.com/CLIMADA-project/climada_python/pull/948)
+
 ### Deprecated
 
 ### Removed
--- a/climada/engine/test/test_impact_calc.py
+++ b/climada/engine/test/test_impact_calc.py
@@ -28,8 +28,8 @@ from pathlib import Path
 
 from climada import CONFIG
 from climada.entity.entity_def import Entity
-from climada.entity import Exposures, ImpactFuncSet, ImpactFunc
-from climada.hazard.base import Hazard
+from climada.entity import Exposures, ImpactFuncSet, ImpactFunc, ImpfTropCyclone
+from climada.hazard.base import Hazard, Centroids
 from climada.engine import ImpactCalc, Impact
 from climada.engine.impact_calc import LOGGER as ILOG
 from climada.util.constants import ENT_DEMO_TODAY, DEMO_DIR
@@ -471,6 +471,34 @@ class TestImpactCalc(unittest.TestCase):
         np.testing.assert_array_equal(eai_exp, [2.25, 1.25, 4.5])
         self.assertEqual(aai_agg, 8.0)  # Sum of eai_exp
 
+    def test_single_exp_zero_mdr(self):
+        """Test for case where exposure has a single value and MDR or fraction contains zeros"""
+        centroids = Centroids.from_lat_lon([-26.16], [28.20])
+        haz = Hazard(
+            intensity=sparse.csr_matrix(np.array([[31.5], [19.0]])),
+            event_id=np.arange(2),
+            event_name=[0,1],
+            frequency=np.ones(2) / 2,
+            fraction=sparse.csr_matrix(np.zeros((2,1))),
+            date=np.array([0, 1]),
+            centroids=centroids,
+            haz_type='TC'
+        )
+        exp = Exposures({'value': [1.],
+                            'longitude': 28.22,
+                            'latitude': -26.17,
+                            'impf_TC': 1},
+                            crs="EPSG:4326")
+        imp_evt = 0.00250988804927603
+        aai_agg = imp_evt/2
+        eai_exp = np.array([aai_agg])
+        at_event = np.array([imp_evt, 0])
+        exp.set_geometry_points()
+        impf_tc = ImpfTropCyclone.from_emanuel_usa()
+        impf_set = ImpactFuncSet([impf_tc])
+        impf_set.check()
+        imp = ImpactCalc(exp, impf_set, haz).impact(save_mat=True)
+        check_impact(self, imp, haz, exp, aai_agg, eai_exp, at_event, at_event)
 
 class TestImpactMatrixCalc(unittest.TestCase):
     """Verify the computation of the impact matrix"""
--- a/climada/hazard/base.py
+++ b/climada/hazard/base.py
@@ -1103,7 +1103,9 @@ class Hazard(HazardIO, HazardPlot):
             impf.id)
             mdr_array = impf.calc_mdr(mdr.toarray().ravel()).reshape(mdr.shape)
             mdr = sparse.csr_matrix(mdr_array)
-        return mdr[:, indices]
+        mdr_out = mdr[:, indices]
+        mdr_out.eliminate_zeros()
+        return mdr_out
 
     def get_paa(self, cent_idx, impf):
         """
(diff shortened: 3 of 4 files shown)


bguillod:
@emanuel-schmid The Test Pipelines fail due to no internet connection - can I safely assume this is an issues on the Jenkins side and unrelated to changes in this PR?

emanuel-schmid:
> @emanuel-schmid The Test Pipelines fail due to no internet connection - can I safely assume this is an issues on the github configuration side and unrelated to changes in this PR?

@bguillod I do assume it's unrelated to changes in this PR, but I couldn't blame the github configuration either. Atm, I suspect the climada data server had a glitch.

bguillod:
Ah, it works now and tests pass, so yes it must have been something.
I've assigned @peanutfun to this PR, but actually I guess he's off for a while now, correct? Who should we best assign to this PR @emanuel-schmid (beside @aleeciu who should check it indeed solves all the related issues for his cases)?

chahank:
Thanks for the fix! 

Can you please make a quick check that your modification does not lead to a significant loss of performance for large datasets?

bguillod:
According to @peanutfun it should rather lead to a performance improvement, however we can check this if you want. Do you have a suggested case which uses large datasets which I could run? It also depends what you mean by "large" as I don't have an Euler cluster at hand...

chahank:
Yes, as @peanutfun said, it might lead to an improvement too. I think it is enough if you run it for a case of a few thousand TC events and make a small uncertainty analysis (you can just execute the tutorial with one of your large datasets). There is no need to make extended tests, I would just like to have a quick check to be sure that we are not missing something obvious. Thanks!

bguillod:
Alright, so based on the uncertainty analysis tutorial I created the following test:
```
#Define the input variable functions
import numpy as np

from climada.entity import ImpactFunc, ImpactFuncSet, Exposures
from climada.util.constants import EXP_DEMO_H5, HAZ_DEMO_H5
from functools import partial
import scipy as sp
from climada.engine.unsequa import InputVar
from climada.engine.unsequa import CalcImpact
import time

def impf_func(G=1, v_half=84.7, vmin=25.7, k=3, _id=1):

    def xhi(v, v_half, vmin):
        return max([(v - vmin), 0]) / (v_half - vmin)

    def sigmoid_func(v, G, v_half, vmin, k):
        return G * xhi(v, v_half, vmin)**k / (1 + xhi(v, v_half, vmin)**k)

    #In-function imports needed only for parallel computing on Windows
    import numpy as np
    from climada.entity import ImpactFunc, ImpactFuncSet
    intensity_unit = 'm/s'
    intensity = np.linspace(0, 150, num=100)
    mdd = np.repeat(1, len(intensity))
    paa = np.array([sigmoid_func(v, G, v_half, vmin, k) for v in intensity])
    imp_fun = ImpactFunc("TC", _id, intensity, mdd, paa, intensity_unit)
    imp_fun.check()
    impf_set = ImpactFuncSet([imp_fun])
    return impf_set

haz = Hazard.from_hdf5(HAZ_DEMO_H5)
exp_base = Exposures.from_hdf5(EXP_DEMO_H5)

def prep_full_calc(haz, exp_base, impf_func):
    #It is a good idea to assign the centroids to the base exposures in order to avoid repeating this
    # potentially costly operation for each sample.
    exp_base.assign_centroids(haz)
    def exp_base_func(x_exp, exp_base):
        exp = exp_base.copy()
        exp.gdf['value'] *= x_exp
        return exp
    exp_func = partial(exp_base_func, exp_base=exp_base)
    #Define the InputVars


    exp_distr = {"x_exp": sp.stats.beta(10, 1.1)} #This is not really a reasonable distribution but is used
                                                #here to show that you can use any scipy distribution.

    exp_iv = InputVar(exp_func, exp_distr)

    impf_distr = {
        "G": sp.stats.truncnorm(0.5, 1.5),
        "v_half": sp.stats.uniform(35, 65),
        "vmin": sp.stats.uniform(0, 15),
        "k": sp.stats.uniform(1, 4)
        }
    impf_iv = InputVar(impf_func, impf_distr)
    calc_imp = CalcImpact(exp_iv, impf_iv, haz)
    output_imp = calc_imp.make_sample(N=2**7, sampling_kwargs={'skip_values': 2**8})
    return calc_imp, output_imp

def run_unc_calc(calc_imp, output_imp):
    calc_imp.uncertainty(output_imp, rp = [50, 100, 250])
```

I then did a few tests on the `develop` branch and on the branch used in this PR by first running:
```
calc_imp, output_imp = prep_full_calc(haz, exp_base, impf_func)
```
and then by profiling using this (everything here was run in a jupyter notebook):
```
%timeit -r 20 run_unc_calc(calc_imp, output_imp)
```

The results are not very stable (i.e. they change a little bit each time you run it), so sometimes there is a small difference and overall it tends to be slightly longer in this branch compared to `develop`. For instance, in the final run I did the outcome was:

1. In the develop branch: `3.19 s ± 13.7 ms per loop (mean ± std. dev. of 20 runs, 1 loop each)`.
2. In this branch: `3.2 s ± 21.6 ms per loop (mean ± std. dev. of 20 runs, 1 loop each)`.

Since it already takes several seconds to run this calculation I did not go on with a larger event set. This seems quite acceptable to me as it fixes a bug. Let me know what you think @chahank.

chahank:
Ok, thanks for checking, this looks good then.

aleeciu:
I checked the cases that allowed me to find the bug and this fix indeed solves the issue.. thanks @bguillod and @peanutfun

bguillod:
Great, thanks @aleeciu for checking it resolves your case and @chahank for reviewing! All good from my perspective, so feel free to merge @chahank :-)

PR #980, opened 2024-11-27
added get_bounds functions for EPSG 4326

ValentinGebhart:
- Introduced three new utility functions `bounding_box_global`, `bounding_box_from_countries`, and `bounding_box_from_cardinal_bounds` in the `climada.util.coordinates` module.
- Added unit tests.

**Notes**
- Methods only treat EPSG 4326 coordinate system. For more flexible methods, a bounds dataclass or class could be implemented in a future PR
- These methods were implemented for the Seasonal Forecast module but moved here for general uses

### PR Author Checklist

### PR Reviewer Checklist


climada/util/coordinates.py
```diff
@@ -783,12 +783,22 @@ def get_country_geometries(
...
+
         country_mask = np.isin(
             nat_earth[["ISO_A3", "WB_A3", "ADM0_A3"]].values,
             country_names,
         ).any(axis=1)
         out = out[country_mask]
 
+    # exit with Value error if no country code was recognized
+    if out.size == 0:
+        raise ValueError(f"None of the given country codes were regocnized.")
```
spjuhel, line 800:
```suggestion
        raise ValueError(f"None of the given country codes were recognized.")
```

climada/util/coordinates.py
```diff
@@ -1687,6 +1697,86 @@ def _ensure_utf8(val):
...
+        )
+
+    if not (360 >= eastern >= -180) or not (360 >= western >= -180):
+        raise ValueError("Given eastern/western bounds are out of range (-180, 360).")
+    # order eastern and western coordinates
+    if western > eastern:
+        eastern += 360
+    if eastern > 360 and western > 180:
+        eastern -= 360
+        western -= 360
```
spjuhel, line 1775:
```suggestion
    eastern = (eastern + 180) % 360 - 180
    western = (western + 180) % 360 - 180

    # Ensure eastern > western
    if western > eastern:
        eastern += 360
```

spjuhel: I am pretty sure this works as well and avoid some condition checking.

Edit: I realise this includes bounding box outside of the (-180,360). You can keep the initial `if`, if you don't want that.

climada/util/test/test_coordinates.py
```diff
@@ -2294,6 +2294,69 @@ def test_mask_raster_with_geometry(self):
...
+            ),
+            (170, -20, 190, 90),
+        )
+        np.testing.assert_array_almost_equal(
+            u_coord.bounds_from_cardinal_bounds(
+                northern=90, southern=-20, eastern=170, western=175
+            ),
+            (175, -20, 530, 90),
+        )
+
```
spjuhel, line 2353: You should add tests for invalid bounding box.

spjuhel (commented):
This looks good overall,
I made some minor suggestions, you may choose to ignore (except maybe the additional tests).

climada/util/coordinates.py
```diff
@@ -783,12 +783,22 @@ def get_country_geometries(
     if country_names:
         if isinstance(country_names, str):
             country_names = [country_names]
+
+        # print warning if ISO code not recognized
+        for country_name in country_names:
+            if not country_name in nat_earth[["ISO_A3", "WB_A3", "ADM0_A3"]].values:
+                LOGGER.warning(f"ISO code {country_name} not recognized.")
```
emanuel-schmid, line 790: wouldn't it be better to simply throw an exception in case the country is not recognized?
I can't see the point of filtering out the unrecognizable countries in here. If a user runs this with a typo in one of the country names would they be upset if it fails?

ValentinGebhart: Yes I agree, this would be better. I changed it to raise a ValueError now.

My only doubt is that the function util.coordinates.get_country_geometries() existed before and if you input a list of ISO codes with one invalid code, it did NOT raise an error before (this is why I "only" added a warning). Wouldn't this be a potential problem for existing code to break down due to the error, or is it better to make aware that the input was not correct?

ValentinGebhart: @emanuel-schmid What do you think?

emanuel-schmid: Good question. Imo, it is better to make them finally aware that their input has been wrong all the time.
We just need to add a note in the change log.

ValentinGebhart: Perfect, agreed. I'm adapting the change log accordingly.

PR #989, opened 2024-12-19
optimize concatenation of centroids

### LINKED ISSUE #988: TropCyclone.from_tracks() takes considerably longer in climada 5.0.0
TropCyclone.from_tracks() takes orders of magnitude longer in climada 5.0.0 compared to 3.3.0 (all else being equal in terms of numbers of tracks, centroid resolution etc.)

Code example:
```python
from climada.hazard import TCTracks, Centroids, TropCyclone, tc_tracks_synth
import numpy as np

# get tracks
tracks_SI_proba = TCTracks.from_ibtracs_netcdf(year_range=(1990, 2023), basin='SI')
tracks_SI_proba.equal_timestep(time_step_h=1)
tracks_SI_proba.calc_perturbed_trajectories(nb_synth_tracks=9, max_shift_ini=0.75, max_dspeed_rel=0.3)

# create centroids
min_lat, max_lat, min_lon, max_lon = -26, -12, 40, 54
resol = 10
grid = (np.mgrid[min_lat : max_lat : complex(0, resol),
min_lon : max_lon : complex(0, resol)]). \
reshape(2, resol*resol).transpose()
cent = Centroids.from_lat_lon(grid[:,0], grid[:,1])
cent.id = np.arange(cent.lat.size)

# Calculate windfields
tc_SI_proba = TropCyclone.from_tracks(tracks_SI_proba, centroids=cent) ## this is the line that takes so long
tc_SI_proba.check()
tc_SI_proba.plot_intensity(event=0, smooth=False);
```

**Screenshots**
Screenshot 2024-12-18 at 16 18 34
Screenshot 2024-12-18 at 16 18 22

**Climada Version:** 5.0.0

Run on IAC hub

NicolasColombi:
Changes proposed in this PR:

Refer to https://github.com/CLIMADA-project/climada_python/issues/988 for detailed information on the profiling results.

Modify `hazard.append()` function in `climada/hazard/centroids/centr.py` so that concatenation of centroids is made only once, when all centroids has already been appended, instead of concatenating every time a centroid is appended (extremely time consuming)

- Create centroid attribute `batch_gdf` in hazard.append()
- Append centroids to `batch_gdf`
- Define concatenation function: `finalize_append()`
- Call `finalize_append()` to concatenate at the end of the appending process
- Clear `batch_gdf` attribute

This PR fixes #988 

Optimize the function `TropCyclone.from_tracks` from 53 minutes to 4.03 minutes.

### PR Author Checklist

### PR Reviewer Checklist

- [x] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/guide/Guide_Review.html#reviewer-checklist) passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Code freeze date: YYYY-MM-DD
 
 ### Fixed
 
+- Resolved an issue where windspeed computation was much slower than in Climada v3 [#989](https://github.com/CLIMADA-project/climada_python/pull/989)
 - File handles are being closed after reading netcdf files with `climada.hazard` modules [#953](https://github.com/CLIMADA-project/climada_python/pull/953)
 - Avoids a ValueError in the impact calculation for cases with a single exposure point and MDR values of 0, by explicitly removing zeros in `climada.hazard.Hazard.get_mdr` [#933](https://github.com/CLIMADA-project/climada_python/pull/948)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,9 +31,7 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
-- `climada.hazard.centroids.centr.union` and `climada.hazard.centroids.centr.append`[#989](https://github.com/CLIMADA-project/climada_python/pull/989). Results: computing windfields is 10 times faster.
-  - `union` does not iterate anymore over `append` but calls directly `append` with the entire list of centroids objects.
-  - `append` concatenate centroids only once, when all centroids have been appended, instead of concatenating after every centroid is appended.
+- `Centroids.append` now takes multiple arguments and provides a performance boost when doing so [#989](https://github.com/CLIMADA-project/climada_python/pull/989)
 - `climada.util.coordinates.get_country_geometries` function: Now throwing a ValueError if unregognized ISO country code is given (before, the invalid ISO code was ignored) [#980](https://github.com/CLIMADA-project/climada_python/pull/980)
 - Improved scaling factors implemented in `climada.hazard.trop_cyclone.apply_climate_scenario_knu` to model the impact of climate changes to tropical cyclones [#734](https://github.com/CLIMADA-project/climada_python/pull/734)
 - In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- `climada.hazard.centroids.centr.union` and `climada.hazard.centroids.centr.append`[#989](https://github.com/CLIMADA-project/climada_python/pull/989). Results: computing windfields is 10 times faster.
+  - `union` does not iterate anymore over `append` but calls directly `append` with the entire list of centroids objects.
+  - `append` concatenate centroids only once, when all centroids have been appended, instead of concatenating after every centroid is appended.
 - `climada.util.coordinates.get_country_geometries` function: Now throwing a ValueError if unregognized ISO country code is given (before, the invalid ISO code was ignored) [#980](https://github.com/CLIMADA-project/climada_python/pull/980)
 - Improved scaling factors implemented in `climada.hazard.trop_cyclone.apply_climate_scenario_knu` to model the impact of climate changes to tropical cyclones [#734](https://github.com/CLIMADA-project/climada_python/pull/734)
 - In `climada.util.plot.geo_im_from_array`, NaNs are plotted in gray while cells with no centroid are not plotted [#929](https://github.com/CLIMADA-project/climada_python/pull/929)
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -344,8 +344,8 @@ class Centroids:
 
         Parameters
         ----------
-        centr : list
-            List of Centroids to append. The centroids need to have the same CRS.
+        centr : Centroids
+            Centroids to append. The centroids need to have the same CRS.
 
         Raises
         ------
@@ -374,8 +374,8 @@ class Centroids:
 
         Parameters
         ----------
-        others : list
-            List of Centroids contributing to the union.
+        others : Centroids
+            Centroids contributing to the union.
 
         Returns
         -------
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -331,16 +331,16 @@ class Centroids:
             }
         )
 
-    def append(self, centr):
-        """Append Centroids to the current object for batch concatenation.
+    def append(self, *centr):
+        """Append Centroids to the current centroid object for concatenation.
 
-        This method adds the `centr.gdf` to the list of DataFrames to be concatenated
-        later with `finalize_append()`. Instead of concatenating immediately, it accumulates
-        the centroids in `_batch_gdf` to perform the concatenation all at once, which is more
-        efficient for multiple appends.
+        This method check that all centroids use the same CRS, append the list of centroids to
+        the initial Centroid object and eventually concatenate them to create a single centroid
+        object with the union of all centroids.
 
         Note that the result might contain duplicate points if the object to append has an overlap
-        with the current object.
+        with the current object. Duplicates points will be removed in `union`
+        by calling `remove_duplicate_points`.
 
         Parameters
         ----------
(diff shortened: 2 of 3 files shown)


peanutfun:
@NicolasColombi Thanks for coming up with this bugfix! If I understand correctly, the problem is that `union` iterates over multiple `others`, and for each of these calls `append`, which in turn calls the costly `pd.concat` each time. Indeed, `concat` creates copies, so you should call it as rarely as possible with as many arguments as possible, see https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#concat.

You solve the issue by introducing a new attribute and function, which I think is error-prone. I would suggest that `append` should be updated to support a list of other centroids. `union` can then call `append` right away with the list.

```python
   def append(self, *centr):
        for cc in centr:
            if not u_coord.equal_crs(self.crs, cc.crs):
                raise ValueError(
                    f"The given centroids use different CRS: {self.crs}, {cc.crs}. "
                    "The centroids are incompatible and cannot be concatenated."
                )
        self.gdf = pd.concat([self.gdf] + [cc.gdf for cc in centr])
```

NicolasColombi:
> @NicolasColombi Thanks for coming up with this bugfix! If I understand correctly, the problem is that `union` iterates over multiple `others`, and for each of these calls `append`, which in turn calls the costly `pd.concat` each time. Indeed, `concat` creates copies, so you should call it as rarely as possible with as many arguments as possible, see https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#concat.
> 
> You solve the issue by introducing a new attribute and function, which I think is error-prone. I would suggest that `append` should be updated to support a list of other centroids. `union` can then call `append` right away with the list.
> 
> ```python
>    def append(self, *centr):
>         for cc in centr:
>             if not u_coord.equal_crs(self.crs, cc.crs):
>                 raise ValueError(
>                     f"The given centroids use different CRS: {self.crs}, {cc.crs}. "
>                     "The centroids are incompatible and cannot be concatenated."
>                 )
>         self.gdf = pd.concat([self.gdf] + [cc.gdf for cc in centr])
> ```
> 
> I think this is a better solution because calling `append` (and hence `concat`) multiple times seems to be bad practice, and with the new signature we suggest doing otherwise.

Hi Lukas, fine by me! With the way you propose we could even get rid completely of `union()` correct ? As if I understand you right, you would make `append()` do the job of `union()` with the constraints of concatenating the already appended list (hence, doing it only once and thus being fast). Or as you said, we make union call append with directly the list, but then, `union()` wouldn't do anything else except calling `append()`, or am I missing something ?

NicolasColombi:
> > @NicolasColombi Thanks for coming up with this bugfix! If I understand correctly, the problem is that `union` iterates over multiple `others`, and for each of these calls `append`, which in turn calls the costly `pd.concat` each time. Indeed, `concat` creates copies, so you should call it as rarely as possible with as many arguments as possible, see https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#concat.
> > You solve the issue by introducing a new attribute and function, which I think is error-prone. I would suggest that `append` should be updated to support a list of other centroids. `union` can then call `append` right away with the list.
> > ```python
> >    def append(self, *centr):
> >         for cc in centr:
> >             if not u_coord.equal_crs(self.crs, cc.crs):
> >                 raise ValueError(
> >                     f"The given centroids use different CRS: {self.crs}, {cc.crs}. "
> >                     "The centroids are incompatible and cannot be concatenated."
> >                 )
> >         self.gdf = pd.concat([self.gdf] + [cc.gdf for cc in centr])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is a better solution because calling `append` (and hence `concat`) multiple times seems to be bad practice, and with the new signature we suggest doing otherwise.
> 
> Hi @peanutfun, fine by me! With the way you propose we could even get rid completely of `union()` correct ? As if I understand you right, you would make `append()` do the job of `union()` with the constraints of concatenating the already appended list (hence, doing it only once and thus being fast). Or as you said, we make union call append with directly the list, but then, `union()` wouldn't do anything else except calling `append()`, or am I missing something ?

peanutfun:
> With the way you propose we could even get rid completely of union() correct ?

Not necessarily. It's a bit unclear, but `append` and `union` do different things:

* `append` adds more centroids to the current object
* `union` creates a copy of the current object, adds more centroids, removes duplicate points, and returns the new object

Arguably, `append` is a bit superfluous. But I would not want to remove an established function as a reaction to a bug report.

NicolasColombi:
> Not necessarily. It's a bit unclear, but `append` and `union` do different things:

Yes, that is why I thought that a single function could do both, but as you mentioned it is probably better to keep them separate, one of the reason being that all tests pass and that it shouldn't compromise the rest of the code this way.

climada/hazard/centroids/centr.py
```diff
@@ -331,11 +331,16 @@ def from_pnt_bounds(cls, points_bounds, res, crs=DEF_CRS):
...
+        """Append Centroids to the current centroid object for concatenation.
+
+        This method check that all centroids use the same CRS, append the list of centroids to
+        the initial Centroid object and eventually concatenate them to create a single centroid
+        object with the union of all centroids.
 
         Note that the result might contain duplicate points if the object to append has an overlap
-        with the current object.
+        with the current object. Duplicates points will be removed in `union`
+        by calling `remove_duplicate_points`.
```
peanutfun, line 343:
```suggestion
        with the current object. Remove duplicates by either using :py:meth:`union`
        or calling :py:meth:`remove_duplicate_points` after appending.
```
This should render correctly in the docs, but one needs to verify.

climada/hazard/centroids/centr.py
```diff
@@ -351,15 +356,18 @@ def append(self, centr):
         union : Union of Centroid objects.
         remove_duplicate_points : Remove duplicate points in a Centroids object.
         """
-        if not u_coord.equal_crs(self.crs, centr.crs):
-            raise ValueError(
-                f"The given centroids use different CRS: {self.crs}, {centr.crs}. "
-                "The centroids are incompatible and cannot be concatenated."
-            )
-        self.gdf = pd.concat([self.gdf, centr.gdf])
+        for cc in centr:
```
peanutfun, line 359: Please fix the linter issue. `cc` is too short, maybe use `other`?

climada/hazard/centroids/centr.py
```diff
@@ -331,11 +331,16 @@ def from_pnt_bounds(cls, points_bounds, res, crs=DEF_CRS):
             }
         )
 
-    def append(self, centr):
-        """Append Centroids
+    def append(self, *centr):
```
peanutfun, line 334: Please add a dedicated test for `append` with multiple arguments

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ Code freeze date: YYYY-MM-DD
 
 ### Changed
 
+- `climada.hazard.centroids.centr.union` and `climada.hazard.centroids.centr.append`[#989](https://github.com/CLIMADA-project/climada_python/pull/989). Results: computing windfields is 10 times faster.
+  - `union` does not iterate anymore over `append` but calls directly `append` with the entire list of centroids objects.
+  - `append` concatenate centroids only once, when all centroids have been appended, instead of concatenating after every centroid is appended.
```
peanutfun, line 36: Please merge these into a single change like
```suggestion
- `Centroids.append` now takes multiple arguments and provides a performance boost when doing so [#989](https://github.com/CLIMADA-project/climada_python/pull/989)
```
You can then add another point to "Fixed" with:
```
- Resolved an issue where windspeed computation was much slower than in Climada v3 [#989](https://github.com/CLIMADA-project/climada_python/pull/989)
```

peanutfun (changes requested):
Thanks for updating the PR! It looks much cleaner now. Please fix the remaining small issues and update the PR description.

NicolasColombi:
@peanutfun I added a test to verify that `append()` can accept multiple arguments, the docs looks good with `:py:meth:`, the changelog is also updated now as you suggested.

climada/hazard/centroids/test/test_centr.py
```diff
@@ -816,6 +816,26 @@ def test_append_dif_crs(self):
...
+        centr.append(*centroids_list)
+
+        self.assertEqual(centr.lat[0], 1)
+        self.assertEqual(centr.lat[1], 2)
+        self.assertEqual(centr.lat[2], 3)
+        self.assertEqual(centr.lat[3], 4)
+        self.assertEqual(centr.lon[0], 1)
+        self.assertEqual(centr.lon[1], 2)
+        self.assertEqual(centr.lon[2], 3)
+        self.assertEqual(centr.lon[3], 4)
```
peanutfun, line 837: For arrays, it's more straightforward to use the numpy testing functions:
```suggestion
        np.testing.assert_array_equal(centr.lat, [1, 2, 3, 4])
        np.testing.assert_array_equal(centr.lon, [1, 2, 3, 4])
```
They will also explicitly test if the arrays have the same size etc. See https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

climada/hazard/centroids/centr.py
```diff
@@ -331,16 +331,21 @@ def from_pnt_bounds(cls, points_bounds, res, crs=DEF_CRS):
...
-        with the current object.
+        with the current object. Remove duplicates by either using :py:meth:`union`
+        or calling :py:meth:`remove_duplicate_points` after appending.
 
         Parameters
         ----------
-        centr : Centroids
-            Centroids to append. The centroids need to have the same CRS.
+        centr : list
+            List of Centroids to append. The centroids need to have the same CRS.
```
bguillod, line 348: since `centr` is given as a `*args`, it does not have to be a list but can be several centroids, no? Formulated differently, can this method be called these two ways with equivalent results or does only one way work?
```
centr.append(new_centrs1, new_centrs2)
```
```
centr.append([new_centrs1, new_centrs2])
```

climada/hazard/centroids/centr.py
```diff
@@ -351,32 +356,35 @@ def append(self, centr):
...
 
         All centroids must have the same CRS. Points that are contained in more than one of the
         Centroids objects will only be contained once (i.e. duplicates are removed).
 
         Parameters
         ----------
-        others : list of Centroids
-            Centroids contributing to the union.
+        others : list
+            List of Centroids contributing to the union.
```
bguillod, line 378: Same comments as above: if it's an `*args` then it's not a list, no?

bguillod (commented):
Thanks a lot for this improvements, it makes a lot of sense and the performance gain is significant indeed! A few small comments but looks good to me overall.

sarah-hlsn (changes requested):
Looks really good to me, thanks a lot! 
Just fixed some minor typos in the docstrings. 

As @bguillod highlighted, there could be some inconsistency associated with how the `centr` argument is passed to the method/how it is defined in the docstrings. I think we should either add a fix to support different ways of passing this argument or clarify the docstrings (if my understanding is correct here, happy to defer to your expertise @NicolasColombi & @peanutfun).

climada/hazard/centroids/centr.py
NicolasColombi: Hi Benoit, I think you are right, I will update the docs strings, it shouldn't be a list.

NicolasColombi:
@sarah-hlsn exactly! I personally wouldn't modify it so that it can accept a list.

sarah-hlsn:
> @sarah-hlsn exactly! I personally wouldn't modify it so that it can accept a list.

Makes perfect sense, we can just adapt the docstrings as you say :)

sarah-hlsn (approved):

NicolasColombi:
May I merge ? 👍  @peanutfun

peanutfun (approved):
Sorry @NicolasColombi, I lost track of this. Excellent work! I'll merge...

PR #1003, opened 2025-02-06
Feature/density tracks

NicolasColombi:
Changes proposed in this PR:

This PR adds a function to compute track density. Tracks are considered lines but are in reality a succession of points. The function ensure equal time step for all tracks by calling `equal_timestep()`, then proceed to compute the number of points per grid cell with the desired resolution. `compute_track_density()` computes the data for `plot_track_density()`.

Functions added:

- `compute_track_density()`
- `test_compute_track_density()`
- `compute_grid_area()`
- `plot_track_density()`

tropical_cyclone_track_density

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/tc_tracks.py
+++ b/climada/hazard/tc_tracks.py
@@ -28,7 +28,6 @@ import logging
 import re
 import shutil
 import warnings
-
 from collections import defaultdict
 from enum import Enum
 from pathlib import Path


climada/hazard/tc_tracks.py
```diff
@@ -2867,3 +2867,48 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
     if np.issubdtype(data_var.dtype, float) or np.issubdtype(data_var.dtype, int):
         return True
     return False
+
+
+def compute_track_density(
+    tc_track: TCTracks, res: int = 5, time_step: float = 0.5, mode: str = "normalized"
+) -> tuple[np.ndarray, tuple]:
+    """Compute absolute and normalized tropical cyclone track density as the number of points per
+    grid cell.
```
chahank, line 2876: Interesting addition. I think though that even with the explanation in the pull request, I do not understand what this method does (and what is the use of it). Can you expand on the docstring to make it clearer?

NicolasColombi: I was exactly doing it in this moment ;)

climada/hazard/tc_tracks.py
```diff
@@ -2867,3 +2867,48 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    # Concatenate datasets along a new "track" dimension
+    all_tracks_ds = xr.concat(tc_track.data, dim="track")
+
+    # Extract flattened latitude and longitude arrays (all values)
+    latitudes = all_tracks_ds["lat"].values.flatten()
+    longitudes = all_tracks_ds["lon"].values.flatten()
+
+    # Define grid resolution and bounds for density computation
+    lat_bins = np.arange(-90, 91, res)  # 91 and not 90 for the bins (90 included)
+    lon_bins = np.arange(-180, 181, res)
```
chahank, line 2905: That would lead to values larger than 90 if say res=0.3. I think this is not what you want. Also, note that it probably is better to use np.linspace than np.arange for non-integer values (see https://numpy.org/doc/stable/reference/generated/numpy.arange.html)

climada/hazard/tc_tracks.py
```diff
@@ -2867,3 +2867,48 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+
+    # Compute 2D density
+    hist, lat_edges, lon_edges = np.histogram2d(
+        latitudes, longitudes, bins=[lat_bins, lon_bins]
+    )
+
+    hist = hist / hist.sum() if mode == "normalized" else hist
+
+    return hist, lat_bins, lon_bins
```
chahank, line 2914: Would it not be better to work with sparse matrices and rasters? Because in general, the output will be very very sparse.

NicolasColombi: Good point!

climada/hazard/tc_tracks.py
NicolasColombi: I updated the docstring which should be clearer now. The role of this function is to create the data for an other plotting function (which will follow soon) and should give something this (first approximation):

image

A plot like this allows you to compare TC activities between different climate scenarios, or different time periods, or models. It's a visualization for the hazard.

climada/hazard/tc_tracks.py
NicolasColombi: True, thanks for catching that!

climada/hazard/tc_tracks.py
chahank: What is the exact goal? Because the method now counts the number of track points per pixel. Thus, if a track is just staying a long time in the same place, it counts as high density. But, it is still just a single track in this pixel. Conversely, if a track is moving fast, it gets low density. I am not sure what this would indicate to be honest.

NicolasColombi: That is true, but, the goal is to display TC activity, and if a track stays longer in a grid cell, this rightfully represent  higher TC activity in my opinion. The question then becomes, how do you define TC activity? Is it how many **different** tracks cross a grid cell or how many tracks cross a grid, **and** how long they stay in it (that is the current version)? 

If you consider two different cyclones with two different translational speeds but with the same angular wind speed, the slower one will have more impact because it stays longer and has the same intensity as the other. So in an impact context, track density (a measure of TC activity) should reflect this. 

But I see your point, and it's worth discussing it with Kerry and Simona next week, frankly both methods would be valid, they just represent something slightly different.

For clarification, the image I sent represent the track density (defined as in this message) for 300 tracks in 2025 with CMIP5 data.

chahank: I agree with the analysis. But then I would ask: why not look at wind intensity density? Because if the argument is that a track staying longer in a position creates more impact, then you would also have to consider that higher winds create more impact. Hence, I would say:
1- if you are interested in impacts, you should consider the windfield distributions
2- if you are interested in track density, then you should consider different track densities.
But maybe I am missing the point and the approach for the track point density is the way to go?

NicolasColombi: Fair, I agree. The point approach was just the practical way I thought to program it. I need to rethink it a bit then :nerd_face:. I'll change it so that a maximum of one point per track per grid cell will be allowed to be counted in that cell.
And at a later stage, we can think of how to compute wind field densities.

chahank: Maybe it can be an argument in the method (if it is easy to implement)?

NicolasColombi: yes it's easy

bguillod: I like the overall idea a lot! We actually had implemented something quite similar to validate the new (still to be incorporated) TC stochastic tracks model. However we had done it a bit differently:

- First, there was a "minimum intensity" threshold, such that you can plot the density of tracks exceeding a certain wind speed (or Saffir-Simpson category). This enabled us to find out where there where too many or too few tracks in general (when doing for example the difference between probabilistic and historical tracks) versus how more intense events were over-/under-estimated. I'd suggest add this feature here as well.
- Ideally there is a valid unit to the "track density". To do so, we counted not the number of points (which is kind of the total duration of tracks being there) but instead counted the number of tracks (after identifying points within each grid cell, we counted the unique `cyc_id` values of these points). You might want to consider doing the same, or at least enabling the user to choose between both approaches.
- We had implemented that using lines, but it took ages to calculate (on a 10'000 years event set), so using points might be good - and I am already relieved to see the validation of the new synthetic tracks model will be made easier thanks to this method). In fact, there should then be a reasonable default for the temporal resolution of the tracks as a function of the raster cells size (i.e., if temporal resolution is too low there might be no point in a grid cell because it jumped from east to west of the cell but is not counted).
- Finally, just to chime on the "track versus windfield" discussion above: I think both are very valuable and complementary. One probably want to first have the tracks being right, and validate wind fields in a second step. This is the approach we had taken back then and it worked well.

climada/hazard/tc_tracks.py
```diff
@@ -2890,6 +2896,10 @@ def compute_track_density(
     density: bool (optional) default: False
         If False it returns the number of samples in each bin. If True, returns the
         probability density function at each bin computed as count_bin / tot_count.
+    filter_tracks: bool (optional) default: True
```
bguillod, line 2899: perhaps calling this argument something like `count_tracks` would be more explicit?

climada/hazard/tc_tracks.py
```diff
@@ -2867,3 +2868,64 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+        grid cell. Hence slower tracks increase the density if the parameter is set to False.
+
+    Returns:
+    -------
+    hist: 2D np.ndarray
+        2D matrix containing the track density
+    """
+
+    # ensure equal time step
+    # tc_track.equal_timestep(time_step_h=time_step)
```
bguillod, line 2911: I think there should be a reasonable check on the track's temporal resolution versus `res` to make sure all tracks will be counted (see my comment in the other thread).

climada/hazard/tc_tracks.py
```diff
@@ -2867,3 +2868,64 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    # Compute 2D density
+    hist_count = csr_matrix((len(lat_bins) - 1, len(lon_bins) - 1))
+    for track in tc_track.data:
+
+        # Compute 2D density
+        hist_new, _, _ = np.histogram2d(
+            track.lat.values, track.lon.values, bins=[lat_bins, lon_bins], density=False
+        )
+        hist_new = csr_matrix(hist_new)
+        hist_new[hist_new > 1] = 1 if filter_tracks else hist_new
```
bguillod, line 2926: why not something like this (seems easier to read):
```
if filter_tracks:
    np.clip(hist_new, a_max=1, out=hist_new)
```

bguillod: You might need to do this before converting to a csr_matrix

bguillod (commented):
I like the overall idea a lot! We actually had implemented something quite similar to validate the new (still to be incorporated) TC stochastic tracks model. However we had done it a bit differently:

- First, there was a parameter for the "minimum intensity" threshold to consider (something like `min_ max_sustained_wind`), such that you can plot the density of tracks exceeding a certain wind speed (or Saffir-Simpson category). This enabled us to find out where there where too many or too few tracks in general (when doing for example the difference between probabilistic and historical tracks) versus how more intense events were over-/under-estimated. I'd suggest add this feature here as well.
- Ideally there is a valid unit to the "track density". To do so, we counted not the number of points (which is kind of the total duration of tracks being there) but instead counted the number of tracks (after identifying points within each grid cell, we counted the unique cyc_id values of these points), which leads to units or "number of tracks per year" if you then scale by the number of years in your dataset. You might want to consider doing the same, or at least enabling the user to choose between both approaches.
- We had implemented that using lines, but it took ages to calculate (well, it was on a 10'000 years event set...), so using points might be good - and I am already relieved to see the validation of the new synthetic tracks model will be made easier thanks to this method 😄). In fact, there should then be a reasonable default for the temporal resolution of the tracks as a function of the raster cells size (i.e., if temporal resolution is too low there might be no point in a grid cell because it jumped from east to west of the cell but is not counted).
- Finally, just to chime on the "track versus windfield" discussion in earlier comments: I think both are very valuable and complementary. One probably want to be able to validate the tracks first, and validate wind fields in a second step. This is the approach we had taken back then and it enabled tuning tracks algorithm without doing the more expensive windfield computation for each test.

climada/hazard/tc_tracks.py
NicolasColombi: yes I like it, then I guess it doesn't really make sense to use sparse matrixes anymore. Since the core function  `np.histogram2d` doesn't exist with sparse, and the later operations are better off with numpy.

NicolasColombi: mmh actually np.clip creates some issues later on with the testing, I'll keep it as it is.

climada/hazard/tc_tracks.py
NicolasColombi: I agree `filter_track` it's a bit cryptic...

climada/hazard/tc_tracks.py
NicolasColombi: True, there should be a limit ratio between temporal resolution and grid cell resolution. Considering that the fastest recorded translational speed of a TC is around 110 km/h, a temporal resolution of 1h and grid res of 1° (~100km) should be the limit. In this case, I would make time_step a function of res and not the opposite.

climada/hazard/tc_tracks.py
chahank: If you keep it dense, then you should justmake it clear in the docstring that this code does not work well at high resolution. It would be good to do some simple testing (res=1, 0.1, 0.01, 0.001) and see how much memory / time it takes.

bguillod: Or use np.where(x>1,1,x)

NicolasColombi:
> I like the overall idea a lot! We actually had implemented something quite similar to validate the new (still to be incorporated) TC stochastic tracks model. However we had done it a bit differently:

Good to hear! Is your version somewhere on a climada branch ?

climada/hazard/tc_tracks.py
NicolasColombi: > If you keep it dense, then you should just make it clear in the docstring that this code does not work well at high resolution. It would be good to do some simple testing (res=1, 0.1, 0.01, 0.001) and see how much memory / time it takes.

At the moment, using sparse (and numpy), with 300 tracks at global scale it takes:

1°: 141ms
0.1°: 9.1s
0.01°: 17 minutes
0.001°: ...

**IF** it makes sense to use this method at really high resolutions (e.g > 0.1°) I will try to optimize it.

chahank: I do not know whether higher resolution is needed, but it would be good to mention in the docstring what kind of resolution is reasonable.

bguillod: From my experience anything below 0.1 degrees is far too noisy, typically one uses something like 1, 5 or even 10 degrees.

bguillod:
> Good to hear! Is your version somewhere on a climada branch ? 

No, it was in our code base at CelsiusPro so I don't have it anymore... @ChrisFairless probably has access to it but I think your approach is easier so I'm not sure you'd gain much (?).

The main thing missing so far in your code is being able to filter points by minimum wind speed.

NicolasColombi:
> The main main thing missing so far in your code is being able to filter points by minimum wind speed.
 
 I see, I have just added the feature :+1: now we can select above, below and in between wind speeds.

NicolasColombi:
@chahank @bguillod I added two functions to compute the grid cell area: One accepts different projections and one uses a spherical approximation, which is a more refined version of an already existing climada function (that doesn't account for the curvature of a tile vs rectangular tile). The two functions have a 0.27% difference on a spherical projection (which mutually validates them, or not). I also added the plotting function. Let me know your thoughts on the current state of the PR.

ChrisFairless:
Hey! Late to the party here! I really like this, thanks!

A few responses for the ongoing discussions:

## Lines vs points
this is complicated. When a track passes through the edge of a cell, a line-based approach would count this as an intersection (depending on the algorithm – see below) whereas with a point-based approach it would depend on your chosen time step, the speed of the system, the latitude of the cell, and the length of the intersection. I think this is fine, actually, but it needs to be documented. I would agree that the timestep should be set algorithmically with the option to pass it as a manual parameter.

Aside: for a really reliable approach, future work could even copy the TCTracks object and change the timestamps so that the systems all travel at a constant speed. Then the choice of time step can be made so that any track intersection that is more than e.g. 1/4 of a cell width is guaranteed to have a point in it (at least, at the equator – I don't know how to standardise this part without reprojection shenanigans).

Line-based intersection methods are in theory a better solution but implementation isn't simple. You also have to make some quite subjective decisions about what counts as an intersection (e.g. counting all intersections, or using something like https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm). We also found a very specific bug in the version of GDAL that CLIMADA uses which gave incorrect results when rasterizing lines, IIRC it did some double-counting of intersections. I thiiiink it's been fixed in more recent versions but I couldn't find the Issue with a quick Google.

But in short: I like the approach here as a pragmatic solution.

## Wind frequency vs track frequency
Both of these are nice. From an insurance industry perspective (and from most academic studies), track frequency is far more common, and maybe even expected. It's also cheaper to calculate and doesn't depend on your choice of wind algorithm. 

## Other validation scripts
@bguillod is right – we had an internal suite of tools at CelsiusPro for this kind of analysis. I don't think it's something that could be shared easily, since it got very bloated over the years and was also quite tied to the databases that we used

NicolasColombi:
> Hey! Late to the party here! I really like this, thanks!

Thanks Chris for the feedback! 

Please let me all know if code wise I shall change something, I made some changes from the first revision. Aside the weird pre-commit check failing here, the PR shall be basically ready.

climada/hazard/tc_tracks.py
```diff
@@ -2980,3 +2980,176 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+            tc_track=tc_track, lat_bins=lat_bins, lon_bins=lon_bins
+        )
+    else:
+        hist_count = np.zeros((len(lat_bins) - 1, len(lon_bins) - 1))
+        for track in tqdm(tc_track.data, desc="Processing Tracks"):
+
+            # select according to wind speed
+            wind_speed = track.max_sustained_wind.values
+            if wind_min and wind_max:
+                index = np.where((wind_speed >= wind_min) & (wind_speed <= wind_max))[0]
```
bguillod, line 3065: Personally I'd use `>= wind_min` but `< wind_max` (here and below), such that if you call the function once with `wind_min=10, wind_max=20` and then with `wind_min=20` you don't get track points with a wind speed of exactly 20 counted in both. In other words I'd make `wind_min` inclusive but `wind_max` exclusive (i.e., `[wind_min, wind_max)`).

climada/hazard/tc_tracks.py
```diff
@@ -2980,3 +2980,176 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+        )
+    elif res < 0.1:
+        warnings.warn(
+            "The resolution is too high. The computation might take several minutes \n"
+            "to hours. Consider using a resolution below 0.1°."
+        )
+
+    # define grid resolution and bounds for density computation
+    lat_bins: np.ndarray = np.linspace(-90, 90, int(180 / res))
+    lon_bins: np.ndarray = np.linspace(-180, 180, int(360 / res))
```
bguillod, line 3052: Actually, I strongly suggest to enable the user to define a range of `lon` and `lat` (new input parameters `lon_range` and `lat_range`, default to None which leads to the current behavior of covering the Earth). If a user only has the North Atlantic covered by their track they will be happy to define a sub-domain. This also applies to the plotting function (`plot_track_density`) of course.

climada/hazard/tc_tracks.py
```diff
@@ -2980,3 +2980,176 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+
+            hist_count += hist_new
+
+    if norm:
+        hist_count = normalize_hist(res=res, hist_count=hist_count, norm=norm)
+
+    return hist_count, lat_bins, lon_bins
+
+
+def compute_genesis_density(
```
bguillod, line 3091: Nice addition

climada/hazard/tc_tracks.py
```diff
@@ -2980,3 +2980,176 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    ----------
+    tc_track: TCT track object
+        track object containing a list of all tracks
+    res: int (optional), default: 5°
+        resolution in degrees of the grid bins in which the density will be computed
+    genesis: bool, Default = False
+        If true the function computes the track density of only the genesis location of tracks
+    norm: bool (optional), default = False
+        If False it returns the number of samples in each bin. If True, returns the
+        specified normalization function at each bin computed as count_bin / grid_area.
```
bguillod, line 3012: This docstring should be updated, since the `norm` parameter now is a string.

climada/hazard/tc_tracks.py
```diff
@@ -2980,3 +2980,176 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    Returns:
+    ---------
+
+    """
+
+    if norm == "area":
+        grid_area, _ = u_coord.compute_grid_cell_area(res=res)
+        norm_hist: np.ndarray = hist_count / grid_area
+    elif norm == "sum":
+        norm_hist: np.ndarray = hist_count / hist_count.sum()
```
bguillod, line 3153: Perhaps add
```
else:
    raise ValueError("Invalid value for input parameter 'norm', it should be one of 'area' or 'sum'")
```
or something like this? It would also remove the pylint warning below ("Possibly using variable 'norm_hist' before assignment").

climada/hazard/tc_tracks.py
bguillod: Also, you might want to specify the possible values that `norm` can take somewhere in the code (e.g., as a constant) and to check in this function whether the value passed by the user is allowed.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -1248,6 +1248,111 @@ def test_track_land_params(self):
...
+                "category": "1",
+            },
+        )
+
+        tc_tracks = tc.TCTracks([track])
+
+        hist_abs, *_ = tc.compute_track_density(
+            tc_tracks,
+            res=10,
+            norm=None,
```
bguillod, line 1285: Doesn't this fail because `norm=None` is not handled?

bguillod (commented):
Thanks for the additions, I think it is starting to looking quite good! It will definitely be useful, as stated in my previous review already 🙂 
I have found a few inconsistencies in how `norm` is handled, and I have a few suggestions too in my comments. I hope this helps to finalize the PR!

climada/hazard/tc_tracks.py
NicolasColombi: Hi Benoit, sorry for my late reply and thanks for the feedback! Good point, I just finished implementing that

climada/hazard/tc_tracks.py
NicolasColombi: Yes thanks!

climada/hazard/tc_tracks.py
NicolasColombi: Good point!

NicolasColombi:
@bguillod  Can I merge ?

climada/hazard/tc_tracks.py
```diff
@@ -2988,3 +2988,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    res: int (optional), default: 5°
+        resolution in degrees of the grid bins in which the density will be computed
+    bounds: tuple, (optional) dafault: None
+        (lat_min,lat_max,lon_min,lon_max) latitude and longitude bounds.
+    genesis: bool, (optional) default = False
+        If true the function computes the track density of only the genesis location of tracks
+    norm: str (optional), default: None
+        If None the function returns the number of samples in each bin. If True, it normalize the
+        bin count as specified: if norm = area -> normalize by gird cell area. If norm = sum ->
+        normalize by the total sum of each bin.
```
sarah-hlsn, line 3024:
```suggestion
        normalize by the total sum across all bins.
```

sarah-hlsn (approved):
Thanks for your work Nico! Fixed a couple of things in the docstrings. As discussed let's use point bound tuples consistently across functions. Otherwise ready to merge!

climada/hazard/tc_tracks.py
```diff
@@ -2988,3 +2988,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    To plot the output of this function use :py:meth:`plot_track_density`.
+
+    Parameters:
+    ----------
+    tc_track: TCT track object
+        track object containing a list of all tracks
+    res: int (optional), default: 5°
+        resolution in degrees of the grid bins in which the density will be computed
+    bounds: tuple, (optional) dafault: None
+        (lat_min,lat_max,lon_min,lon_max) latitude and longitude bounds.
```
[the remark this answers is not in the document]
sarah-hlsn: Double check tests after making this change :)

NicolasColombi:
I think this PR is ready to be merged, do I have the green light ? @chahank

climada/hazard/plot.py
```diff
@@ -158,6 +161,136 @@ def plot_intensity(
 
         raise ValueError("Provide one event id or one centroid id.")
 
+    @staticmethod
+    def plot_track_density(
```
chahank, line 165: I am a bit confused now about this method. Why is it a Hazard method? This should be a `TCTracks` method right? It has nothing to do with Hazards in general. At most maybe it could be a `TropCyclone` method. 

In general, we should not add methods that have only a very specific narrow application to a general basis class. This will lead back to the clutter of methods that we now spend years to reduce. 

Where do you think this would fit best? `TCTracks` or `TropCyclones`? Since it is only about tracks and not about intensities, I would move this to `TCTracks`.

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    lat_bins: np.ndarray
+        latitude bins in which the point were counted
+    lon_bins: np.ndarray
+        longitude bins in which the point were counted
+
+    Example:
+    --------
+    >>> tc_tracks = TCTrack.from_ibtracs_netcdf("path_to_file")
+    >>> tc_tracks.equal_timestep(time_step_h = 1)
+    >>> hist_count, *_ = compute_track_density(tc_track = tc_tracks, res = 1)
```
chahank, line 3154: I think this is cleaner..

```suggestion
    >>> hist_count, _, _ = compute_track_density(tc_track = tc_tracks, res = 1)
```

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+    >>> ax = plot_track_density(hist_count)
+
+    """
+
+    limit_ratio: float = 1.12 * 1.1  # record tc speed 112km/h -> 1.12°/h + 10% margin
```
chahank, line 3159: Would that become a problem if there is a future TC set with tracks with windspeeds above 112?

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+
+    return hist_count, lat_bins, lon_bins
+
+
+def compute_genesis_density(
+    tc_track: TCTracks, lat_bins: np.ndarray, lon_bins: np.ndarray
+) -> np.ndarray:
+    """Compute the density of track genesis locations. This function works under the hood
+    of :py:meth:`compute_track_density`. If it is called with the parameter genesis = True,
+    the function return the number of genesis points per grid cell.
```
chahank, line 3225: If this method is not intended to be accessible to the user, then please define as such `_compute_genesis_density`.

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+
+    Parameters:
+    -----------
+
+    tc_track: TCT track object
+        TC track object containing a list of all tracks.
```
chahank, line 3231: Why is this a function instead of a class method since the first attribute seems to be self?

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
     if np.issubdtype(data_var.dtype, float) or np.issubdtype(data_var.dtype, int):
         return True
     return False
+
+
+def compute_track_density(
```
chahank, line 3099: Why is this not a class method since the first argument seems to be the object itself?

climada/hazard/tc_tracks.py
```diff
@@ -3094,3 +3094,196 @@ def _zlib_from_dataarray(data_var: xr.DataArray) -> bool:
...
+        first_lats,
+        first_lons,
+        bins=[lat_bins, lon_bins],
+        density=False,
+    )
+
+    return hist_count
+
+
+def normalize_hist(
```
chahank, line 3257: Is this supposed to be a method called by the user? If yes, leave it as is. If not, please make it private with the `_`.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -1359,6 +1359,111 @@ def test_track_land_params(self):
...
+        )
+        # hist_norm, *_ = tc.compute_track_density(tc_tracks, res=10, density=True)
```
chahank, line 1398: ?

climada/hazard/test/test_tc_tracks.py
```diff
@@ -1359,6 +1359,111 @@ def test_track_land_params(self):
...
+        )
+        hist_wind_max, *_ = tc.compute_track_density(
+            tc_tracks, res=10, norm=None, wind_min=None, wind_max=30
+        )
+        hist_wind_max, *_ = tc.compute_track_density(
+            tc_tracks, res=10, norm=None, wind_min=None, wind_max=30
+        )
+        hist_wind_both, *_ = tc.compute_track_density(
+            tc_tracks, res=10, norm=None, wind_min=11, wind_max=29
+        )
```
chahank, line 1410: Please explain in comments what are the values that should come out so that , in case the test fails in the future, a reader can immediately understand whether the new values are to be expected or are wrong.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -1359,6 +1359,111 @@ def test_track_land_params(self):
...
+        self.assertEqual(hist_wind_both.sum(), 2)
+        # the track defined above occupy positions 0 to 4
+        np.testing.assert_array_equal(hist_abs[0, 0:4], [1, 1, 1, 1])
+
+    def test_normalize_hist(self):
+        """test the correct normalization of a 2D matrix by grid cell area and sum of
+        values of the matrix."""
+
+        M = np.ones((10, 10))
+        M_norm = tc.normalize_hist(res=10, hist_count=M, norm="sum")
```
chahank, line 1424: What is tc? This variable is not defined here... Also, what about the other possible values of the attributes of `nomarlize_hist`?

climada/hazard/test/test_tc_tracks.py
```diff
@@ -1359,6 +1359,111 @@ def test_track_land_params(self):
...
+        res = 10
+        tc_tracks = tc.TCTracks([track])
+        lat_bins = np.linspace(-90, 90, int(180 / res))
+        lon_bins = np.linspace(-180, 180, int(360 / res))
+        hist = tc.compute_genesis_density(
+            tc_track=tc_tracks, lat_bins=lat_bins, lon_bins=lon_bins
+        )
+        self.assertEqual(hist.shape, (17, 35))
+        self.assertEqual(hist.sum(), 1)
+        self.assertEqual(hist[0, 0], 1)
```
chahank, line 1465: Please add a comment to explain why these are the expected values.

climada/util/coordinates.py
```diff
@@ -459,6 +460,136 @@ def get_gridcellarea(lat, resolution=0.5, unit="ha"):
     return area
 
 
+def compute_grid_cell_area_validation(res: float) -> tuple[np.ndarray]:
```
chahank, line 463: I am confused by the name of this function. It says that it computes the area of grid cells. So why is called `validation`?

climada/util/coordinates.py
```diff
@@ -459,6 +460,136 @@ def get_gridcellarea(lat, resolution=0.5, unit="ha"):
...
+    res: int
+        The resolution of the grid in degrees. The grid will have cells of size `res x res` in
+        terms of latitude and longitude.
+
+    Returns:
+    -------
+    grid_area: np.ndarray
+        A 2D array of shape `(len(lat_bins)-1, len(lon_bins)-1)` containing the area of each grid
+        cell, in square kilometers. Each entry in the array corresponds to the area of the
+        corresponding grid cell on Earth.
```
chahank, line 498: Incomplete.

climada/util/coordinates.py
```diff
@@ -459,6 +460,136 @@ def get_gridcellarea(lat, resolution=0.5, unit="ha"):
...
+    for i, (lat1, lat2) in enumerate(zip(lat_edges[:-1], lat_edges[1:])):
+        for j, (lon1, lon2) in enumerate(zip(lon_edges[:-1], lon_edges[1:])):
+
+            # 5th point to close the loop
+            poly_lons = [lon1, lon2, lon2, lon1, lon1]
+            poly_lats = [lat1, lat1, lat2, lat2, lat1]
+
+            # Compute the area of the grid cell
+            poly_area, _ = geod.polygon_area_perimeter(poly_lons, poly_lats)
+            area[i, j] = abs(poly_area)  # Convert from m² to km²
```
chahank, line 585: This looks quite inefficient. In general, nested loops in Python are not a good sign as these are often not vectorized.

climada/util/test/test_coordinates.py
```diff
@@ -565,6 +565,31 @@ def test_get_gridcellarea(self):
...
         self.assertTrue(area2[0] <= 2500)
 
+    def test_compute_grid_area(self):
+        """Test that the two twin functions calculate the area of a gridcell correctly. Using
+        an absolute reference and mutual validation by comparison."""
+        res = 1
+        area = u_coord.compute_grid_cell_area(
+            res=res, projection="sphere", units="km^2"
+        )
+        area_test, *_ = u_coord.compute_grid_cell_area_validation(res=res)
```
chahank, line 575: In tests you cannot simply ignore the outputs. Please test the functions fully.

climada/util/test/test_coordinates.py
```diff
@@ -565,6 +565,31 @@ def test_get_gridcellarea(self):
...
+
+        self.assertEqual(area.shape, (179, 359))
+        self.assertEqual(area_test.shape, (179, 359))
```
chahank, line 578: Unclear why this tests the methods.

climada/util/coordinates.py
```diff
@@ -459,6 +460,136 @@ def get_gridcellarea(lat, resolution=0.5, unit="ha"):
...
+        areas[:, np.newaxis], (1, len(lon_bins) - 1)
+    )  # each row as same area
+
+    return grid_area, lat_bins, lon_bins
+
+
+def compute_grid_cell_area(
+    res: float = None,
+    bounds: tuple = None,
+    projection: str = "WGS84",
```
chahank, line 533: Does this also accept projections for instance in meters? Please make clear if not, and catch the error accordingly.

climada/util/coordinates.py
```diff
@@ -459,6 +460,136 @@ def get_gridcellarea(lat, resolution=0.5, unit="ha"):
...
+    )
+    # Broadcast to create a full 2D grid of areas
+    grid_area = np.tile(
+        areas[:, np.newaxis], (1, len(lon_bins) - 1)
+    )  # each row as same area
+
+    return grid_area, lat_bins, lon_bins
+
+
+def compute_grid_cell_area(
```
chahank, line 530: Please compare this method to the one implemented in the Centroids class `Centroids.get_pixel_area` https://github.com/CLIMADA-project/climada_python/blob/main/climada/hazard/centroids/centr.py#L571. If they do the same, please test and keep the most efficient one and consolidate in the utils.

chahank (changes requested):
Thanks for the addition! I think this is getting closer to merging. There are still a few fundamental questions to solve before proceeding such as why the methods are where they are, and the quality of the tests. 

Thanks for all the efforts. I know this is a longer process, but we really have to be careful to not fall in the feature bloating trap again as this has caused us a very large amount of refactoring work.

NicolasColombi:
Before I start to do anything, after a second thought I think this PR should be only about the track density function. Plotting the density is pretty straight forward and having a single plotting function just for that might not be ideal, so we might want to remove it. Normalizing by the area does not make sense anymore, as you should normalize the density by time, not space (ex: n°storms per grid cell per year), even though the function `compute_grid_cell_area` might be useful for something else, I think it does not belongs here. I have written for myself an additional function: `compute_windspeed_density`. I propose the following: remove the plotting function and area function keeping only the `track_density`, and **maybe** add the `compute_windspeed_density`. 
What do you think @chahank  ?

NicolasColombi:
> Before I start to do anything, after a second thought I think this PR should be only about the track density function. Plotting the density is pretty straight forward and having a single plotting function just for that might not be ideal, so we might want to remove it. Normalizing by the area does not make sense anymore, as you should normalize the density by time, not space (ex: n°storms per grid cell per year), even though the function `compute_grid_cell_area` might be useful for something else, I think it does not belongs here. I have written for myself an additional function: `compute_windspeed_density`. I propose the following: remove the plotting function and area function keeping only the `track_density`, and **maybe** add the `compute_windspeed_density`. What do you think @chahank ?

any thoughts ? @chahank

chahank:
> Before I start to do anything, after a second thought I think this PR should be only about the track density function. Plotting the density is pretty straight forward and having a single plotting function just for that might not be ideal, so we might want to remove it. Normalizing by the area does not make sense anymore, as you should normalize the density by time, not space (ex: n°storms per grid cell per year), even though the function `compute_grid_cell_area` might be useful for something else, I think it does not belongs here. I have written for myself an additional function: `compute_windspeed_density`. I propose the following: remove the plotting function and area function keeping only the `track_density`, and **maybe** add the `compute_windspeed_density`. What do you think @chahank ?

I agree. There is already a grid cell area method in the util.coordinates module anyway. Why would you remove the plotting function now? I think you can keep it.

climada/hazard/tc_tracks.py
NicolasColombi: Not really, I added a 10% safety buffer. If a TC will be faster than 123.2km (112 + 10%) worst case scenario is, one or a few grid cell will be skipped. We can increase the buffer to 20% or so if you wish.

NicolasColombi:
Thanks @chahank for the inputs! :raised_hands: 

summary:
- removed grid area function
- removed normalization
- moved plotting function into TCTtracks
- updated tests
- added test for plotting function (just calling the plotting function, probably to simple for a test)
- corrected the code with the suggestions you proposed

let me know if something is still missing.

@emanuel-schmid how does that look to you for the release ?

climada/hazard/tc_tracks.py
```diff
@@ -2089,6 +2088,292 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+    ----------
+    hist: np.ndarray
+        2D histogram of track density.
+    axis: GeoAxes, optional
+        Existing Cartopy axis.
+    projection: cartopy.crs, optional
+        Projection for the map.
+    add_features: dict
+        Dictionary of map features to add. Keys can be 'land', 'coastline', 'borders', and
+        'lakes'. Values are Booleans indicating whether to include each feature.
```
emanuel-schmid: why not simply add flags for the `add_features`? If this is about pleasing the linter (too many arguments), then flags are the lesser flaw. The problem with too many arguments is usually that they point to a design flaw. Plotting functions are often the exception.

climada/hazard/tc_tracks.py
```diff
@@ -2089,6 +2088,292 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+        >>> tc_tracks = TCTrack.from_ibtracs_netcdf("path_to_file")
+        >>> tc_tracks.equal_timestep(time_step_h = 1)
+        >>> hist_count, _, _ = compute_track_density(tc_track = tc_tracks, res = 1)
+        >>> ax = plot_track_density(hist_count)
+
+        """
+
+        limit_ratio: float = (
+            1.12 * 1.1
+        )  # record tc speed 112km/h -> 1.12°/h + 10% margin
```
emanuel-schmid: I guess having this limit hard coded is fine here. But out of curiosity: why 112km/h / 1.12 degree? is this about the expected value in tropical cyclones?

NicolasColombi: The fastest recorded TC to my knowledge was traveling at 112km/h. If the track time step is 1h we want the grid cell to be at least 112km x 112km, so that the points in a track do not skip a grid cell. Which means there is a ratio between max speed, time step and resolution that should not be crossed (I added 10% to the max speed as margin). Here I assume no TC will ever be faster than 112km/h + 10% margin.

Around the equator 112km equates to roughly 1°, so 112km/h becomes 1°/h .

(I should have written 1° instead of 1.12°)

climada/hazard/plot.py
```diff
@@ -21,6 +21,9 @@
 
 import logging
 
+import cartopy.crs as ccrs
```
chahank: I think all the changes to `hazard.plot` are not necessary. Are they maybe leftovers from something else?

climada/hazard/tc_tracks.py
```diff
@@ -2089,6 +2088,292 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+            grid cell. Hence slower tracks increase the density if the parameter is set to False.
+        wind_min: float (optional), default: None
+            Minimum wind speed above which to select tracks (inclusive).
+        wind_max: float (optional), default: None
+            Maximal wind speed below which to select tracks (exclusive if wind_min is also provided,
+            otherwise inclusive).
+        Returns:
+        -------
+        hist_count: np.ndarray
+            2D matrix containing the the absolute count per grid cell of track point or the
```
chahank:
```suggestion
            2D matrix containing the absolute count per grid cell of track point or the
```

climada/hazard/tc_tracks.py
```diff
@@ -2089,6 +2088,292 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+            normalized number of track points, depending on the norm parameter.
```
chahank: I do not see any norm parameter.

climada/hazard/tc_tracks.py
```diff
@@ -2089,6 +2088,292 @@ def _one_interp_data(track, time_step_h, land_geom=None):
...
+        self,
+        res: int = 5,
+        bounds: tuple = None,
+        genesis: bool = False,
+        filter_tracks: bool = True,
+        wind_min: float = None,
+        wind_max: float = None,
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+        """Compute tropical cyclone track density. Before using this function,
+        apply the same temporal resolution to all tracks by calling :py:meth:`equal_timestep` on
```
chahank, line 2285: Actually, you require more right? You require that the timestep is equal to one hour?

climada/hazard/tc_tracks.py
chahank: But this requires the equal time step to be set to 1 hour right?

NicolasColombi: Only if the chosen resolution is 1° the time step should be 1h or less, the code check if the ratio `resolution / time_step` is equal or greater than the `limit ratio`. If it is smaller, it throws an error: as either the resolution is to high or the time step to big. I'll try to write it more clearly.

climada/hazard/plot.py
NicolasColombi: Yes, I forgot to cancel them... (expect `plot_fraction` what was a merge conflict)

climada/hazard/tc_tracks.py
NicolasColombi: (See the other comment) You are probably right in pointing out that I should mention in the doc string the ratio between resolution and time step. Otherwise the only way to find out about it is getting an error.

climada/hazard/tc_tracks.py
NicolasColombi: Good point :grin:

PR #1012, opened 2025-02-14
Improve efficiency of local exceedance intensitry/impact and local return periods functions

### LINKED ISSUE #1010: Missing deprecation warning of `HazardPlot.plot_rp_intensity()`
Missing deprecation warning of method `HazardPlot.plot_rp_intensity()`. The new way of computing local exceedance intensities differs from the old one.

Generally, do we want to keep this (direct) function or just use `Hazard.local_exceedance_intensity` and then `util.plot.plot_from_gdf` instead, which does the same thing?

ValentinGebhart:
Changes proposed in this PR:
Changes address the methods `Hazard.local_exceedance_intensity`, `Hazard.local_return_period`, `Impact.local_exceedance_impact`, and `Impact.local_return_period`.
- In the computation, intensity/impact values are now only binned (and their frequency added in each bin) if bin_decimals is an integer (default: None). Then, the values are binned according to their bin_decimals decimal places. In all other cases, no binning occurs. In the notes, some possible values for bin_decimals are given.
- The interpolation util function is now only called for centroids with nonzero intensity/impact.
- docstring was extented to explain how data are extrapolated.
- The plotting functions `Hazard.plot_rp_intensity` and `Impact.plot_rp_imp` now directly call the corresponding functions above, and a warning is added about how to recover values that are computed with CLIMADA 5.0.0 or earlier.

Comments for future improvements:
- It would be nice to have a fit method, also for extrapolation.
- The impact method `calc_freq_curve` and the class `ImpactFreqCurve` could be adapted to also include the different methods of util.interpolation.

This PR fixes #1010 and #1015

### PR Author Checklist

- [ ] [Changelog](https://github.com/CLIMADA-project/climada_python/blob/develop/CHANGELOG.md) updated

### PR Reviewer Checklist


climada/engine/impact.py
```diff
@@ -1146,42 +1164,32 @@ def plot_rp_imp(
         imp_stats : np.array
             return_periods.size x num_centroids
         """
-        imp_stats = (
-            self.local_exceedance_impact(np.array(return_periods))[0].values[:, 1:].T
+
+        LOGGER.info(
+            "Some errors in the previous calculation of local exceedance impacts have been corrected,"
+            " see Impact.local_exceedance_impact. To reproduce data with the "
+            "previous calculation, use CLIMADA v5.0.0 or less."
```
chahank, line 1171: Cool idea, I am not sure though that this should be an info logging. @emanuel-schmid : what would be the most elegant way to inform the users that something changed, without having logs all the time?

climada/engine/impact.py
```diff
@@ -1146,42 +1164,32 @@ def plot_rp_imp(
...
-            imp_stats_log,
-            self.coord_exp,
-            colbar_name,
-            title,
-            smooth=smooth,
-            axes=axis,
-            **kwargs,
+
+        impacts_stats, title, column_labels = self.local_exceedance_impact(
+            return_periods
```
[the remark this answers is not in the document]
ValentinGebhart: I think the best way is to do `gdf, title, columns = local_exceedance_intensity()` and `plot_from_gdf(gdf, title, columns)`. Then you can just change the plot options by repeating the second step, without recalculating the gdf. The `plot_rp_intensity` is a wrapper around these two functions, which is why I asked if we want to keep it in the first place. 

One way of not repeating the calculation here is one calls the function twice would be to save the local exceedance gdf as an attribute of the Hazard object, but I don't think we want to do that?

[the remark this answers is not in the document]
ValentinGebhart: What about writing the info to better use the other two functions in the docstring? Is this enough?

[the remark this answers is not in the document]
chahank: Yeah, that might be a solution. For small hazards I guess the method is rather fast, and for lager ones there would the solution in the docstring.

ValentinGebhart:
This PR is now ready for review. Note that I did not add a code example to any docstring as I think the only understandable place would be util.interpolation._group_frequency which is a private function that will not be seen by the user.

ValentinGebhart:
just realized that the fix of the rounding method is much slower than the method before that had floating point problems. I will try to solve this, I'll say as soon as I'm ready

ValentinGebhart:
Quick update:
I removed the method of rounding to significant digits. Now the default is to not bin any values. Binning can be activated for extrapolation: by method="extrapolate" (default "interpolate") and setting bin_decimals to an integer (default None). Then values are binned according to bin_decimals decimal places.

chahank:
Does this mean that one cannot bin for the "interpolate" method? And does that mean that the user always has to come up with a number for the rounding?

ValentinGebhart:
Yes, one cannot bin anymore in interpolation, as binning does have little effect here (it just "smoothens" the data). 

In extrapolation, default is not binning, and the user has to provide the decimal places to bin if they want to bin. Binning in decimals and not n_sig_sigits because it is simpler.

In this way, binning has to be consciously chosen by the user, as it is also not clear what a good default value would be and if one would want to bin by default.

ValentinGebhart:
Update: 
Binning is now possible for all methods (default: no binning). Some example choices for bin_decimals (the decimals about how to bin) are added to the docstrings.

climada/engine/test/test_impact.py
```diff
@@ -594,7 +595,7 @@ def test_local_exceedance_impact_methods(self):
 
         # test log log extrapolation
         impact_stats, _, _ = impact.local_exceedance_impact(
-            return_periods=(1000, 30, 0.1), method="extrapolate"
+            return_periods=(1000, 30, 0.1), method="extrapolate", bin_decimals=2
```
chahank, line 598: Since it is a bit 'out of the ordinary', can you also try with `bin_decimals = -1` or any negative value?

climada/engine/impact.py
```diff
@@ -1145,43 +1204,40 @@ def plot_rp_imp(
         axis : matplotlib.axes.Axes
         imp_stats : np.array
             return_periods.size x num_centroids
+
```
chahank, line 1207:
```suggestion
        See Also
        --------
        engine.impact.local_exceedance_impact :
            inter- and extrapolation method
```

climada/engine/impact.py
```diff
@@ -1109,24 +1175,17 @@ def plot_basemap_impact_exposure(
...
-    @deprecated(
-        details="The use of Impact.plot_rp_imp is deprecated."
-        "Use Impact.local_exceedance_impact and util.plot.plot_from_gdf instead."
-    )
     def plot_rp_imp(
         self,
         return_periods=(25, 50, 100, 250),
         log10_scale=True,
-        smooth=True,
         axis=None,
```
chahank, line 1182:
```suggestion
        axis=None,
        kwargs_local_exceedance_impact,
```

chahank: Please add this to the docstring, I cannot suggest it here.

climada/engine/impact.py
```diff
@@ -1145,43 +1204,40 @@ def plot_rp_imp(
...
-            self.coord_exp,
-            colbar_name,
-            title,
-            smooth=smooth,
-            axes=axis,
-            **kwargs,
+
+        impacts_stats, title, column_labels = self.local_exceedance_impact(
+            return_periods
         )
```
chahank, line 1224:
```suggestion
        **kwargs_local_exceedance_impact
        )
```

climada/hazard/plot.py
```diff
@@ -37,62 +37,52 @@ class HazardPlot:
...
 
-    @deprecated(
-        details="The use of Hazard.plot_rp_intensity is deprecated."
-        "Use Hazard.local_exceedance_intensity and util.plot.plot_from_gdf instead."
-    )
     def plot_rp_intensity(
         self,
         return_periods=(25, 50, 100, 250),
-        smooth=True,
         axis=None,
```
chahank, line 43:
```suggestion
        axis=None,
        kwargs_local_exceedance_intensity={},
```

climada/hazard/plot.py
```diff
@@ -37,62 +37,52 @@ class HazardPlot:
...
-            figure size for plt.subplots
         kwargs: optional
             arguments for pcolormesh matplotlib function used in event plots
 
         Returns
         -------
         axis, inten_stats:  matplotlib.axes._subplots.AxesSubplot, np.ndarray
             intenstats is return_periods.size x num_centroids
+
+        Notes
```
chahank, line 65:
```suggestion
       See Also
       ---------
       hazard.local_exceedance_intensity: method to calculate local exceedance frequencies.

        Notes
```

climada/engine/impact.py
chahank: CAn you please test that this works? Idem for the case of the hazard.

ValentinGebhart:
Thanks for the review! I included the suggestions and checked that they work. Also, the tutorial now uses a hazard object with the correct frequencies, without redefining frequencies (Emanuel put it on the API). Let me know if we can merge :)

chahank:
Great, this makes the methods much clearer, faster and better tested! I think it is ready to merge. Before, please do not forget to update the changelog file!

climada/hazard/plot.py
```diff
@@ -37,62 +37,61 @@ class HazardPlot:
...
         axis=None,
-        figsize=(9, 13),
-        adapt_fontsize=True,
+        kwargs_local_exceedance_intensity=None,
         **kwargs,
     ):
         """
-        This function is deprecated,
-        use Impact.local_exceedance_impact and util.plot.plot_from_gdf instead.
+         Compute and plot hazard exceedance intensity maps for different
```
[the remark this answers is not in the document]
ValentinGebhart: nice catch, how did you find that? :D it doesn't show in my pylint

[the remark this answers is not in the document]
chahank: Haha, it just showed weirdly on the screen in the github file-change view. Laser-eyes :D

PR #1020, opened 2025-03-06
Remove implicit use of CRS in code base

luseverin:
Changes proposed in this PR:
- Add a cheap test for functions that must be in epsg:4326 but do take only value as input to catch when input coordinates are not geographic (e.g. in meters)
- Update functions that assume epsg:4326 without reason to be able to use other crs. (work in progress / requires more information)

This PR treats scrum user story number 129 Consistent use of CRS in CLIMADA

Further information:
- I added a quick test so function that require geographic coordinates (e.g. `get_country_geometries` or `dist_to_coast` that are based on naturalearth data) at least test that input data seems geographic. Currently allowed values are -91 < lat < 90 and -181 < lon < 541 and maximum extension of 181 (latitudinal) and 361 (longitudinal) but it is not exactly clear to me where to do the cut-off. For instance, do we want to allow longitudinal extension greater than 360, or longitudinal values smaller than -180 (some tests related to that e.g. `test_track_land_params` are still failing due to longitudinal extension > 360)?
- It was specifically mentionned to treat issues in the `trop_cyclone` module but I was not exactly sure of what to do there. For instance, the `get_close_centroids` does not check that input centroids and lat and lon from the TC track are geographic; I was not sure if this should be explicitly check or if it can be left as is.
- It was also supposed to adapt functions to allow use of other crs when possible but I did not find any instances of where this would be worth doing it or could be done without too much trouble. Some inputs on this are welcome!

### PR Author Checklist

### PR Reviewer Checklist


ValentinGebhart (changes requested):
Thank you for the lat lon check function! I think it is very valuable. I am not really sure in which other functions we could add it. What about: 
latlon_to_geosph_vector(lat, lon, rad=False, basis=False)
lon_bounds(lon, buffer=0.0)
latlon_bounds(lat, lon, buffer=0.0)
dist_approx()
dist_to_coast_nasa()

These were all I could find in climada.util.coordinates that looked like they would require geographical coordinates. But maybe you thought about it and it does not make sense here?

I only have two suggestions:
- describe the error message a bit better so the user has a better idea what could be wrong and what to do about it. 
- change the range to -540 to 540 (apparently there is no convention in CLIMADA, and this seems soft enough as a error range)

climada/util/coordinates.py
```diff
@@ -85,6 +85,24 @@
 are not considered."""
 
 
+def check_if_geo_coords(lat, lon):
+    """
+    Check if latitude and longitude arrays are likely in geographic coordinates,
+    testing if min/max values are within -90 to 90 for latitude and -180 to 540
+    for longitude.
+    Returns True if lat/lon ranges are in the geographic coordinates range, otherwise False.
```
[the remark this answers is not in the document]
luseverin: Not sure the explanation for the 540<> makes sense, let me know if this is not the case.

climada/util/coordinates.py
```diff
@@ -1596,6 +1644,12 @@ def get_country_code(lat, lon, gridded=False):
         return np.empty((0,), dtype=int)
     LOGGER.info("Setting region_id %s points.", str(lat.size))
     if gridded:
+        # first check that input lat lon are geographic
```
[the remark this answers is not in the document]
luseverin: Good question.. I think it should also test when griodded=False as it then calls get_country_geometries which requires WGS84.

DahyannAraya:
I tested the `Hazard.from_xarray_raster` function using daily humidity data (`hurs`) from DWD EPISODES, which uses the Lambert Conformal Conic projection (EPSG:3034). The data was **successfully loaded as a Hazard object** with no issues related to the CRS format.

However, by default, the coordinate reference system (CRS) recorded in the object was **EPSG:4326**, as this is coded in the function’s properties.

Like :
 haz.centroids.crs 
<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: World.
- bounds: (-180.0, -90.0, 180.0, 90.0)
Datum: World Geodetic System 1984 ensemble
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich_

However, my data properties are: ETRS_1989_LCC, Datum: D_ETRS_1989, Spheroid: GRS_1980, Authority: EPSG:3034, Prime Meridian: Greenwich (0.0), Projection: Lambert Conformal Conic.

Perhaps the function could trigger a warning indicating that the CRS should be manually defined, or (more ambitiously ) try to extract it automatically from the data. However, in my case, this is not straightforward. 

For now, I did: 

ds = xr.open_dataset("/Users/daraya/Downloads/hurs_day_GCFS21--DWD-EPISODES2022--DE-5_sfc20240601_r1i1p1_20240601-20241130.nc")
intensity_variable = "hurs"  
hazard_type = "HU"           
intensity_unit = "%"         

# Create hazard from xarray
haz = Hazard.from_xarray_raster(
    data=ds,
    hazard_type=hazard_type,
    intensity_unit=intensity_unit,
    intensity=intensity_variable,
    **crs="EPSG:3034",**
    coordinate_vars={
        "event": "time",          
        "longitude": "lon",      
        "latitude": "lat",
                
    }
) Obviously, with my data I can't use functions like dist_to_coast, which are based on EPSG:4326. I think it's a great idea to have the flexibility to incorporate data with different coordinate systems.

climada/util/coordinates.py
```diff
@@ -85,6 +85,66 @@
 are not considered."""
 
 
+def check_if_geo_coords(lat, lon):
+    """Check if latitude and longitude arrays are likely in geographic coordinates,
+    testing if min/max values are within -90 to 90 for latitude and -540 to 540
+    for longitude. Lon coordinates of <-360 or >360 are allowed to cover cases
+    of objects being defined close to the 180 meridian.
```
ValentinGebhart: I like this explanation more. You could even be more explicit and rephrase the final sentence to 
"Lon coordinates of <-360 or >360 are allowed to cover cases of objects that extend over the 180 meridian, for which longitude intervals such as [179, 538] or [-538, -179] might be used."

climada/util/coordinates.py
```diff
@@ -85,6 +85,66 @@
...
+    ValueError
+        If the coordinate system is neither geodetic nor projected, or if the
+        unit cannot be inferred.
+    """
+
+    if coords.crs.is_geographic:
+        unit = "degree"
+    elif coords.crs.is_projected:
+        unit = "m"
+    else:
```
ValentinGebhart, line 149: Did you check that this is always the correct assignment? For instance, are all objects with a projected crs (so crs.is_projected is true) in the unit "m"? Or could they also sometimes be in, e.g., "km" or "feet" or so?

climada/util/coordinates.py
```diff
@@ -1113,6 +1258,9 @@ def match_centroids(
         Distance to use in case of vector centroids.
         Possible values are "euclidean", "haversine" and "approx".
         Default: "euclidean"
+    unit : str, optional
+        Unit to use for non-exact matching. Possible values are "degree", "m", "km".
+        Default: None
```
ValentinGebhart: Add explanation what is done by the method if unit is None. EG: 
"If unit is None, the coordinate unit is retrieved from coord_gdf when possible."

climada/util/coordinates.py
```diff
@@ -1198,14 +1353,22 @@ def _nearest_neighbor_approx(
...
+        raise ValueError(
+            "Only degree unit is supported for nearest neighbor approximation."
+            "Please use euclidean distance for non-degree units."
+        )
     # Compute only for the unique coordinates. Copy the results for the
     # not unique coordinates
     _, idx, inv = np.unique(coordinates, axis=0, return_index=True, return_inverse=True)
     # Compute cos(lat) for all centroids
-    centr_cos_lat = np.cos(np.radians(centroids[:, 0]))
+    centr_cos_lat = np.cos(centroids[:, 0])
```
ValentinGebhart: why don't you transform to radians anymore? are the centroids already in radian? 

Maybe now I understand that before calling the function you transformed coords and coord_to_assign to radians? If that is the case, you could make clear in the docstrings of the underlying functions that coordinates should be in radian. Because as it is now, it looks like coordinates must be in radian but unit can only be "degree", which is a bit confusing.

climada/util/coordinates.py
```diff
@@ -1198,14 +1353,22 @@ def _nearest_neighbor_approx(
...
     assigned = np.zeros(coordinates.shape[0], int)
     num_warn = 0
+    # need to convert back to degrees for the approx distance and nearest neighbor
+    coordinates = np.rad2deg(coordinates)
+    centroids = np.rad2deg(centroids)
```
ValentinGebhart: same as above: I thought centroids and coordinates would be in "degree", not in "radian"?

climada/util/coordinates.py
```diff
@@ -1259,14 +1424,20 @@ def _nearest_neighbor_haversine(centroids, coordinates, threshold):
...
     """
+    # first check that unit is in degree
+    if unit != "degree":
+        raise ValueError(
+            "Only degree unit is supported for nearest neighbor approximation."
+            "Please use euclidean distance for non-degree units."
+        )
     # Construct tree from centroids
-    tree = BallTree(np.radians(centroids), metric="haversine")
+    tree = BallTree(centroids, metric="haversine")
```
ValentinGebhart: see above about radians

climada/util/coordinates.py
```diff
@@ -1259,14 +1424,20 @@ def _nearest_neighbor_haversine(centroids, coordinates, threshold):
...
     # Select unique exposures coordinates
     _, idx, inv = np.unique(coordinates, axis=0, return_index=True, return_inverse=True)
 
     # query the k closest points of the n_points using dual tree
     dist, assigned = tree.query(
-        np.radians(coordinates[idx]),
+        coordinates[idx],
```
ValentinGebhart: see above about radians

climada/util/coordinates.py
```diff
@@ -1320,27 +1493,30 @@ def _nearest_neighbor_euclidean(
         with as many rows as coordinates containing the centroids indexes
     """
     # Construct tree from centroids
-    tree = scipy.spatial.KDTree(np.radians(centroids))
+    tree = scipy.spatial.KDTree(centroids)
```
ValentinGebhart, line 1469: same as above: radians

climada/util/coordinates.py
```diff
@@ -1320,27 +1493,30 @@ def _nearest_neighbor_euclidean(
...
     # Select unique exposures coordinates
     _, idx, inv = np.unique(coordinates, axis=0, return_index=True, return_inverse=True)
 
     # query the k closest points of the n_points using dual tree
-    dist, assigned = tree.query(np.radians(coordinates[idx]), k=1, p=2, workers=-1)
+    dist, assigned = tree.query(coordinates[idx], k=1, p=2, workers=-1)
```
ValentinGebhart, line 1474: same as above: radians

climada/util/coordinates.py
```diff
@@ -1320,27 +1493,30 @@ def _nearest_neighbor_euclidean(
...
-            num_warn,
+            f"Distance to closest centroid is greater than {threshold} km for {num_warn} coordinates."
         )
-        assigned[dist * EARTH_RADIUS_KM > threshold] = -1
+        assigned[dist > threshold] = -1
 
     if check_antimeridian:
         assigned = _nearest_neighbor_antimeridian(
-            centroids, coordinates[idx], threshold, assigned
+            np.rad2deg(centroids), np.rad2deg(coordinates[idx]), threshold, assigned
```
ValentinGebhart: same as above: radians

climada/util/coordinates.py
```diff
@@ -1389,7 +1565,7 @@ def _nearest_neighbor_antimeridian(centroids, coordinates, threshold, assigned):
         if np.any(cent_strip_bool):
             cent_strip = centroids[cent_strip_bool]
             strip_assigned = _nearest_neighbor_haversine(
-                cent_strip, coord_strip, threshold
+                np.deg2rad(cent_strip), np.deg2rad(coord_strip), "degree", threshold
```
ValentinGebhart: same as above: radians

climada/util/coordinates.py
luseverin: You are right, this is not optimal. I think I have found a way to properly infer coordinates from a pyproj.crs. Now, there is no need to guestimate any more. I think the unit can also be in feet or arcseconds or seconds, but I will only allow for degree, metre, and kilometre and ask the user to reproject.

climada/util/coordinates.py
luseverin: I will now remove this optional keyword and directly obtain the unit of the coordinate system from the crs object.

climada/util/coordinates.py
luseverin: Now we also get coordinates that are not in degree, so we do not always convert to radians. I have tried to do the conversion outside of the `_nearest_neighbor` functions as I thought this was cleaner but did not realize it would be confusing (also as I did not update the docstrings accordingly). Now I have reverted the changes so that the `_nearest_neighbor` functions still receive coordinates in degree and do the conversion to radian internally when necessary. I hope that is clearer now!

luseverin:
Thanks a lot @ValentinGebhart! I think I successfully adressed your comments. I now wait for approval from any of the code owners.

climada/util/coordinates.py
```diff
@@ -85,6 +85,65 @@
...
+    Parameters
+    ----------
+    coords : GeoDataFrame
+        An object with a coordinate reference system (CRS) attribute.
+
+    Returns
+    -------
+    unit : str
+        The unit of measurement for the coordinate system, as specified in the
+        CRS axis information. Assumes that both axes have the same unit.
```
chahank: Can a geopandas attribute CRS have different units on the axes? If not, I would remove as this is slightly confusing extra information.

climada/util/coordinates.py
```diff
@@ -85,6 +85,65 @@
...
+    if unit == "degree":
+        pass
+    elif unit == "metre":
+        unit = "m"
+    elif unit == "kilometre":
+        unit = "km"
+    else:
+        raise ValueError(
+            f"Unknown unit: {unit}. Please provide a crs that has a unit "
+            "of 'degree', 'metre' or 'kilometre'."
```
chahank, line 152: Is there a reason to forbid other units? I am not sure to understand why at this stage.

climada/util/coordinates.py
```diff
@@ -1089,15 +1197,48 @@ def match_coordinates(
...
+                        " to geographic coordinates in degrees. This can be because"
+                        " total extents are > 180 for lat or > 360 for lon, lat coordinates"
+                        " are outside of -90<lat<90, or lon coordinates are outside of "
+                        "-540<lon<540. If you use degree values outside of these ranges,"
+                        " please shift the coordinates to the valid ranges."
+                    )
+
+            elif unit == "m":
+                coords *= 1e-3
+                coords_to_assign *= 1e-3
```
chahank: I am not sure to understand why a conversion to kilometres is necessary.

climada/util/coordinates.py
```diff
@@ -1146,15 +1287,25 @@ def match_centroids(
...
             raise ValueError("Set hazard and GeoDataFrame to same CRS first!")
-    except AttributeError:
-        # If the coord_gdf has no crs defined (or no valid geometry column),
-        # no error is raised and it is assumed that the user set the crs correctly
-        pass
+        if coord_gdf.crs is None or centroids.crs is None:
+            raise ValueError(
+                "Please provide coordinate GeoDataFrame and Hazard "
+                "object with a valid crs attribute."
+            )
```
chahank:
```suggestion
        if coord_gdf.crs is None:
            raise ValueError(
                "Please provide a coord_gdf (coordinate GeoDataFrame)"
                "object with a valid crs attribute."
            )
        if centroids.crs is None :
            raise ValueError(
                "Please provide centroids "
                "object with a valid crs attribute."
            )
```

climada/util/coordinates.py
```diff
@@ -1146,15 +1287,25 @@ def match_centroids(
...
+    except AttributeError as exc:
+        # a crs attribute is needed for unit inference
+        raise ValueError(
+            "Please provide coordinate GeoDataFrame and Hazard object with a valid crs attribute."
+        ) from exc
+
+    # get unit of coordinate systems from axis of crs
+    unit = get_crs_unit(coord_gdf)
```
chahank: I am confused. Why would the above exception be needed, since the function `get_crs_unit(coord_gdf)` is used after? Why would one first check if the CRS are equal, and afterwards check if they exist?

climada/util/coordinates.py
```diff
@@ -1198,7 +1351,12 @@ def _nearest_neighbor_approx(
     np.array
         with as many rows as coordinates containing the centroids indexes
     """
-
+    # first check that unit is in degree
+    if unit != "degree":
+        raise ValueError(
+            "Only degree unit is supported for nearest neighbor approximation."
+            "Please use euclidean distance for non-degree units."
```
chahank: The docstring says `Compute the nearest centroid for each coordinate using the
    euclidean distance`. But the error says that one should change to Euclidian distance. I think there is a mix-up here.

climada/util/coordinates.py
```diff
@@ -1259,6 +1419,12 @@ def _nearest_neighbor_haversine(centroids, coordinates, threshold):
     np.array
         with as many rows as coordinates containing the centroids indexes
     """
+    # first check that unit is in degree
+    if unit != "degree":
+        raise ValueError(
+            "Only degree unit is supported for nearest neighbor approximation."
+            "Please use euclidean distance for non-degree units."
```
chahank: I think this error message is not useful. The method cannot use Euclidian distance. What should the user do to fix this error?

climada/util/coordinates.py
```diff
@@ -1305,38 +1472,55 @@ def _nearest_neighbor_euclidean(
...
+    dist, assigned = tree.query(coordinates[idx], k=1, p=2, workers=-1)
+
+    if unit == "degree":
+        # convert back to degree for check antimeridian and convert distance
+        centroids = np.rad2deg(centroids)
+        coordinates = np.rad2deg(coordinates)
+        dist = dist * EARTH_RADIUS_KM
+    else:
+        # if unit is not in degree, check_antimeridian is forced to False
+        check_antimeridian = False
```
chahank: What happens if the antimeridian is crossed in a crs without degrees? Is this possible?

climada/util/coordinates.py
```diff
@@ -1305,38 +1472,55 @@ def _nearest_neighbor_euclidean(
     coordinates : 2d array
         First column contains latitude, second column contains longitude. Each
         row is a geographic point
+    unit : str
+        Unit to use for non-exact matching. Possible values are "degree", "m", "km".
```
chahank, line 1447: I think this is the point where the restriction to meters of kilometers would come into play, but I fail to see how. Could you please help me understand?

chahank (changes requested):
Thanks, it looks good! I would propose a few improvements and clarifications.

The error message is repeated quite often in the code. This should thus rather be handled by a function. Please either write a wrapper, or include into `check_if_geo_coords`.

>     if not check_if_geo_coords(lat, lon):
        raise ValueError(
            "Input lat and lon coordinates do not seem to correspond"
            " to geographic coordinates in degrees. This can be because"
            " total extents are > 180 for lat or > 360 for lon, lat coordinates"
            " are outside of -90<lat<90, or lon coordinates are outside of -540<lon<540."
            " If you use degree values outside of these ranges,"
            " please shift the coordinates to the valid ranges."
        )

climada/util/coordinates.py
luseverin: You are right, it probably does not exist. I will remove it.

climada/util/coordinates.py
luseverin: I forbade it essentially because of the `threshold` parameter that we use to assign or not the coordinates. If we allow for other units, we would need to ensure that the user overseeds the default threshold value (which is currently set to 100 km), or define default thresholds for other coordinate units. I figured it was simpler to force the input coordinates to be in either SI units (m, km) or degrees, but I guess we could be more flexible if you think that it adds value. Let me know!

climada/util/coordinates.py
luseverin: This is to be consistent with the default threshold defined in km. See my previous comment.

chahank: I think I would rather convert the threshold no?

climada/util/coordinates.py
chahank: I see, it makes sense. A well-spotted subtlety, thanks.

It is a bit unfortunate to introduce a limitation just to avoid incoherent results with some default value, which anyway, is quite sub-optimal at the moment. I think 100km for the threshold is often not very well chosen. I am not sure what to do here. 

I would tend not to introduce limitations in this PR, and rather open an issue to better handle the threshold value. So, concretely, I would continue to handle explicitly meters and kilometres, but only for the threshold in the assignement. What do you think?

luseverin: > It is a bit unfortunate to introduce a limitation just to avoid incoherent results with some default value, which anyway, is quite sub-optimal at the moment. I think 100km for the threshold is often not very well chosen. I am not sure what to do here.

Yes, I agree. 100 km is probably far from being optimal in many situations. But on the other hand I would not know what other value would be suited.. Maybe using a relative threshold i.e. that is based on the resolution of the input data would be better suited? E.g. `res_coords = np.roll(coord_gdf, 1, axis=(0,1)).mean()`,  `res_centr = np.roll(centroids, 1, axis=(0,1)).mean()` and then set a default `threshold` in function of that and alternatively warn if the difference of resolution between the two input data is too large? Not sure if that would make sense though. And also it would only work with gridded data... But yes maybe it is better to open a new issue for that.

> So, concretely, I would continue to handle explicitly meters and kilometres, but only for the threshold in the assignement. What do you think?

Not sure I understood right. Do you suggest to not raise an exception if `unit` is not one of (degree, km, m), and then just use a default value of 100 <unit> as a threshold?

climada/util/coordinates.py
luseverin: That would also work but require a specific `if` statement to treat the case of inputs in degrees. Unless we want to warn the users giving a distance in degrees (which might be less optimal).

climada/util/coordinates.py
luseverin: To be able to correctly handle different units, we need to have a crs attribute defined for both `coord_gdf` and `centroids`. They also need to be equal for consistency. This is essentially what those `try-except-raise` blocks do, I think:

`try:`
1. Check if crs of `coord_gdf` and centroids are equal
2. Check that none of the crs of both `coord_gdf` and `centroids` is `None`

`except AttributeError:`
- If `AttributeError` is raised, then it means that one of the input (e.g. `centroids`) does not have a `crs` attribute, hence we also need to raise an error as crs are required.

I don't know if that helps.. Also I am not sure why you mention the `get_crs_unit(coord_gdf)`; do you mean we should not raise before  the call to `get_crs_unit(coord_gdf)` and just let an exception be raised from it when there is no crs defined?

climada/util/coordinates.py
chahank: How about you raise a warning when the unit is not among the supported ones in the assign centroids part and explain that the threshold might be unsuitable? 

I like the suggestion for improving the default value of the threshold, but maybe this should be done in a separate PR?

climada/util/coordinates.py
chahank: Ok, I see.

My worry is that this part of the code is now modifying the input arrays directly, which can quickly lead to un-controlled results.

climada/util/coordinates.py
chahank: Sorry for being unclear. 

If you check that the centroids are equal, then there is no point in checking whether either is None, as this is only possible if they are unequal. The only question is what to do if both are None, and I do not know what equal_crs returns in this case.

For the AttributeError you are right we can leave it as is. The error message could be improved a bit, since the method does neither have GeoDataFrame nor Hazard as input.

climada/util/coordinates.py
luseverin: I modified the message to the following:
`raise ValueError(
            "Only degree unit is supported for nearest neighbor matching using"
            "'haversine' distance. Please use euclidean distance by setting "
            "'distance=euclidean' in the match_coordinates or match_centroids "
            "function when using units that are not degrees."
        )`
Let me know if that is clearer.

climada/util/coordinates.py
luseverin: Good question. I see two possible cases:
1. The user uses a local projected coordinates system which is locally defined around the antimeridian region. Hence, there is no discontinuity at the antimeridian and everything is fine.
2. The user uses a global projected coordinates system which is discontinuous at the antimeridian (e.g. EPSG:3857). In which case, I am not sure if crossing the antimeridian should be allowed, as it might be unphysical, and I am even less sure on how to handle this, given our current implementation of handling antimeridian crossing that relies on the haversine distance.

Maybe we could just raise a warning that "handling of antimeridian crossing is not implemented for non-degree coordinates. Check_antimeridian has been set to False. Please use degree-based coordinates system if you want to allow antimeridian crossing". Let me know.

climada/util/coordinates.py
luseverin: I just checked and` equal_crs` returns `True` if both are `None`. I simplified the check and clarified the error messages accordingly.

climada/util/coordinates.py
luseverin: I now do not convert the coordinates to be consistent with the default threshold in km and warn that the threshold is now considered as being defined in the input unit. I will set up a new issue to discuss how to improve the default behavior for the distance threshold.

climada/util/coordinates.py
```diff
@@ -1386,8 +1386,10 @@ def _nearest_neighbor_haversine(centroids, coordinates, unit, threshold):
     # first check that unit is in degree
     if unit != "degree":
         raise ValueError(
-            "Only degree unit is supported for nearest neighbor approximation."
-            "Please use euclidean distance for non-degree units."
+            "Only degree unit is supported for nearest neighbor matching using"
+            "'haversine' distance. Please use euclidean distance by setting "
+            "'distance=euclidean' in the match_coordinates or match_centroids "
+            "function when using units that are not degrees."
```
chahank:
```suggestion
            "Only coordinates in degree units are supported for nearest neighbor matching using"
            "'haversine' distance."
```
The method in which this error is raised is `_nearest_neighbor_haversine`. This method should not include mentions to methods using it. This very quickly becomes intransparent.

climada/util/coordinates.py
chahank: Yes, we can try that. If this ends up logging warnings all the time we should remove it again as this would not be the goal of a warning.

climada/util/coordinates.py
```diff
@@ -1318,8 +1316,10 @@ def _nearest_neighbor_approx(
     # first check that unit is in degree
     if unit != "degree":
         raise ValueError(
-            "Only degree unit is supported for nearest neighbor approximation."
-            "Please use euclidean distance for non-degree units."
+            "Only degree unit is supported for nearest neighbor matching using"
+            "'approx' distance. Please use euclidean distance by setting "
+            "'distance=euclidean' in the match_coordinates or match_centroids "
+            "function when using units that are not degrees."
```
chahank:
```suggestion
            "Only degree unit is supported for nearest neighbor matching using"
            "'approx' distance."
```
A method should not know what methods are calling it, hence the error message should not reference these methods, otherwise it gets quickly intransparent. 

Rather, the docstring of `match_coordinates` and `match_centroids` should make this clear.

climada/util/coordinates.py
chahank: Excellent. Can you please open an issue to address this?

chahank:
Please do not forget to update the changelog, thanks!

climada/util/coordinates.py
```diff
@@ -1089,15 +1173,38 @@ def match_coordinates(
 
         # assign remaining coordinates to their geographically nearest neighbor
         if threshold > 0 and exact_assign_idx.size != coords_view.size:
+            # convert to proper units before proceeding to nearest neighbor search
+            if unit == "degree":
+                # check that coords are indeed geographic before converting
```
chahank: What does this mean? No coordinate conversion is performed in this method.

luseverin: I think it is a legacy of when I was converting meters to kilometers. I removed it.

PR #1031, opened 2025-03-23
add bounds

NicolasColombi:
Changes proposed in this PR:

This PR add tropical cyclones basins bounds as a dictionary named `BASINS_BOUNDS` in 0-360 coordinates, following STORM dataset bounds definitions.

This addition can be used to determine whether a track originated in that basin simply by using shapely functionalities:

```
point = Point(lat, lon)
True_or_False = polygon.contains(point)
```

Particularly useful to decrease the memory overload during global wind fields computations by computing fields basin wide.

Preview of the bounds:
image

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Simple yet useful addition!

I would suggest two things:
1- make it a data class instead
2- make it clear that these are the STORM basin definitions. These can vary substantially, and hence cannot be defined as a definitive default

Also, I am not sure to understand the point `Particularly useful to decrease the memory overload during global wind fields computations by computing fields basin-wide.`? 

In what way does this work beyond what is already possible?

chahank:
Thanks for the explanation! I am a bit confused though, because for each track only the centroids near the track are considered (and they are loaded one by one into memory). 

Could you maybe post an example of a code that shows this increase in memory efficiency? That would certainly be useful to see.

NicolasColombi:
I see, then I might have misinterpreted Sam's words. I haven't tested yet if this indeed increases efficiency (it was my hypothesis given the above), but I will try soon. Also, if you know already areas in the Winfield computation that could be optimized memory-wise, let me know. As my memory is currently exploding using global exposure at 450 arcsec with only 200 tracks (>50GB on euler) and >150GB at 150 arcsec resolution.

NicolasColombi:
@chahank I converted it to a dataclass (unsure whether this is how you intended it) and added a method `split_by_basin()` which creates 6 TCTtracks instances (1 per basin) from 1 TCTtrack instance.

chahank:
Thanks! Although the idea of the data class is precisely to not have a macro variable like BASIN. The idea is to have the data in that class directly. Also, the idea is that you can then clearly explain that this is an arbitrary boarder for basins based on one particular publication. 

I am not sure to understand the purpose of the method `split_by_basin()`. Can you please provide an example?

chahank:
Exactly. The data class is there to contain the data. A global variable like BASIN is hard to manage and document, and is usually quite obscure to the user as it does not appear in the documentation.

chahank:
@emanuel-schmid aaa no of course, sorry for the confusion! I meant an `Enum` class.

NicolasColombi:
@emanuel-schmid makes more sense now! Let me know if it is implemented as intended. For consistency, we may also consider adding the basin pressure to that class ?

Thanks @ValentinGebhart. I agree, I switched to -180,+180 as it is consistent with the tracks coordinate range. I had to split the "SP" basin into two polygons to avoid issues with the antimeridian (original reason why I chose longitude between 0-360).

Before I write the test for `subset_by_basin()`, let me know if you have comments on the function.

emanuel-schmid:
@NicolasColombi many thanks. I've got too questions about the subset_by_basin method

1) wouldn't it be preferable to return an array with the basins instead of a dictionary with lists?
2) supposedly there are much more performant ways to do this than to loop over the data - is speed irrelevant here?

NicolasColombi:
> @NicolasColombi many thanks. I've got two questions about the subset_by_basin method
> 
> 1. wouldn't it be preferable to return an array with the basins instead of a dictionary with lists?
> 2. supposedly there are much more performant ways to do this than to loop over the data - is speed irrelevant here?

@emanuel-schmid 
For 1. don't you lose the information on which basin the TCTtracks object belongs to with an array ? I just picked a format where you can access the data by calling the basin name. But mybe there is a better option :upside_down_face:

For 2. Speed is not so much of an issue here, but if you know a more efficient way I am absolutely keen to modify the function. Can you advise on how to optimize it?

NicolasColombi:
> > @NicolasColombi many thanks. I've got two questions about the subset_by_basin method
> > 
> > 1. wouldn't it be preferable to return an array with the basins instead of a dictionary with lists?
> > 2. supposedly there are much more performant ways to do this than to loop over the data - is speed irrelevant here?
> 
> @emanuel-schmid For 1. don't you lose the information on which basin the TCTtracks object belongs to with an array ? I just picked a format where you can access the data by calling the basin name. But mybe there is a better option 🙃
> 
> For 2. Speed is not so much of an issue here, but if you know a more efficient way I am absolutely keen to modify the function. Can you advise on how to optimize it?

@emanuel-schmid Any updates ?

chahank:
Would it make sense to also add the bounds for the impact basins as defined in https://www.nature.com/articles/s41467-022-33918-1 ?

NicolasColombi:
> Would it make sense to also add the bounds for the impact basins as defined in https://www.nature.com/articles/s41467-022-33918-1 ?

Maybe, what do you have in mind ? But at the moment I personally do not see a direct application. So, we can either add it now without a clear idea, or we add it later if needed. I would personally opt for the latter, unless you had something in mind.

chahank:
> Maybe, what do you have in mind ? But at the moment I personally do not see a direct application. So, we can either add it now without a clear idea, or we add it later if needed. I would personally opt for the latter, unless you had something in mind.

I am a bit confused now. I thought your goal was to be able to load tracks for specific basins. You will also use the basins defined in  https://www.nature.com/articles/s41467-022-33918-1 . Hence, why would you not implement these?

NicolasColombi:
Because I do not need them, nor I see a use case aside from defining centroids, which I already have. But of course feel free to add them; I am sure someone will find it useful.

chahank:
Still confused, but ok.

Then please just choose a name that makes obvious that this is only the definition used in STORM (is this really defined by STORM, or did they use the definition from somewhere else?). Currently, it looks like the Basin is a universally defined set of boundaries.

NicolasColombi:
> Still confused, but ok.
> 
> Then please just choose a name that makes obvious that this is only the definition used in STORM (is this really defined by STORM, or did they use the definition from somewhere else?). Currently, it looks like the Basin is a universally defined set of boundaries.

Not really, STORM defined boundaries based on IBTracks, but did not use exactly the same. I wrote it clearly at the beginning of the class description that these boundaries were defined according to STORM. I think it is up to the user to check this information, especially since it is easy to find and basin boundaries are very similar across different publications. Adding "STORM" to the class name would be confusing in my opinion.

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +415,68 @@ def subset(self, filterdict):
...
+        Returns
+        -------
+        dict_tc_basins : dict
+            A dictionary where the keys are basin names (e.g., "NA", "EP", "WP", etc.) and the
+            values are instances of the `TCTracks` class containing the tropical cyclones that
+            belong to each basin.
+
+        Example:
+        --------
+        >>> tc = TCTracks.from_ibtracks("")
```
emanuel-schmid, line 441: as an example this is suboptimal:
```ptyhon
>>> tc = TCTracks.from_ibtracks("")
  File "<stdin>", line 1, in <module>
AttributeError: type object 'TCTracks' has no attribute 'from_ibtracks'
```

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +415,68 @@ def subset(self, filterdict):
...
+
+            # Find the basin that contains the point
+            for basin in Basin:
+                if basin.value.contains(origin_point):
+                    basins_dict[basin.name].append(track)
+                    point_in_basin = True
+                    break
+
+            if not point_in_basin:
+                tracks_outside_basin.append(track.id_no)
```
emanuel-schmid, line 464: @NicolasColombi as to 1. and 2.
I would first define a method that assigns basins to a track:
```python
BASINS_GDF = gpd.GeoDataFrame(
    {'basin': b, 'geometry': b.value} for b in Basin
)

def get_basins(track):  # this is the method I had in mind for 1. and I'd guess it could be a performance boost
    track_coordinates = GeoDataFrame(
        geometry=gpd.points_from_xy(track.lon, track.lat)
    )
    return track_coordinates.sjoin(BASINS_GDF , how='left', predicate='within').basin
```

emanuel-schmid: Having that I'd write the track loop like this:
```python
for track in self.data:
    touched = get_basins(track).dropna().drop_duplicates()
    if touched.size:
        for basin in touched:
            basins_dict[basin].append(track)
    else:
        tracks_outside_basin.append(track)
```

emanuel-schmid (commented):
updates 😄

climada/hazard/tc_tracks.py
NicolasColombi: Thanks @emanuel-schmid for the update, sorry for the (very) late reply. If I understand it correctly, if a track crosses multiple basin it will be present in the output dictionary in all the basins that it crossed, which is fine, just different than attributing the origin basin, as I had in mind. If we want only the origin, I guess we can still use your method but select only the first row of the gdf ? 

@chahank : Conceptually, do we want to assign basin to tracks based on the origin basin or all the basins they cross ?

chahank: I have no strong opinion, but I would say the second makes more sense as a default. Ideally, just allow for both.

NicolasColombi: Ok! I implemented Emanuel version and allowed the flexibility of choosing only the origin basin or not.

NicolasColombi: @chahank I will be on holidays for the next two week, I think this PR should be ready now. If you require additional changes I will be available from Sept.15, or if the changes are really small, I might carve out some time tomorrow.

climada/hazard/tc_tracks.py
```diff
@@ -193,6 +195,102 @@
...
+                    (-180.0, -60.0),
+                ]
+            ),
+        ]
+    )
+
+
+BASINS_GDF = gpd.GeoDataFrame(
+    {"basin": b, "geometry": b.value} for b in Basin_bounds_storm
+)
```
chahank: This beats the purpose of the enum class a bit. Please do without it if possible (I know this was proposed by @emanuel-schmid , but this is just a very minor change and allows for future flexibility). This should be part of the method `get_basins`. The latter should also have a parameter `source` or similar, which at the moment can only have the value `STORM`. The method `subset_by_basin` should also have that.

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +420,59 @@ def subset(self, filterdict):
 
         return out
 
+    def get_basins(track):
```
chahank, line 418: Please add the docstrings.

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +420,59 @@ def subset(self, filterdict):
...
+        origin : bool
+            Either True or False. If True, the outputs basin will contain only the tracks that originated there.
+            If False, every track that crossed a basin will be present in the basin.
+
+        Returns
+        -------
+        dict_tc_basins : dict
+            A dictionary where the keys are basin names (e.g., "NA", "EP", "WP", etc.) and the
+            values are instances of the `TCTracks` class: effectively all tracks that intersected or originated
+            in each basin, depending on the argument "origin".
```
chahank, line 474: I think it makes much more sense to return for each basin a TCTrack object, and not a list of xarray datasets.

climada/hazard/test/test_tc_tracks.py
```diff
@@ -875,6 +875,33 @@ def test_subset_years(self):
...
+        dict_basins, tracks_outside_basin = tc_test.subset_by_basin(origin=True)
+
+        self.assertEqual(len(dict_basins["NA"]), 0)
+        self.assertEqual(len(dict_basins["EP"]), 2)
+        self.assertEqual(len(dict_basins["WP"]), 2)
+        self.assertEqual(len(dict_basins["NI"]), 0)
+        self.assertEqual(len(dict_basins["SI"]), 1)
+        self.assertEqual(len(dict_basins["SP"]), 0)
+        self.assertEqual(len(tracks_outside_basin), 0)
+
```
chahank, line 905: You can do a simpler and more explicit test. Make a TCTrack where you directly assign the basin variable. Then control that this remains correct. Here the test is obscure, as it is not clear what is in the file `TEST_TRACK_EMANUEL`. Then you should also test all basins since now `NI` and `SP` and `out of basin` are not tested.

CHANGELOG.md
```diff
@@ -91,6 +91,7 @@ Removed:
 
 ### Added
 
+- `climada.hazard.tc_tracks.BASINS_BOUNDS` dictionary [#1031](https://github.com/CLIMADA-project/climada_python/pull/1031)
 - `climada.hazard.tc_tracks.TCTracks.subset_years` function [#1023](https://github.com/CLIMADA-project/climada_python/pull/1023)
 - `climada.hazard.tc_tracks.TCTracks.from_FAST` function, add Australia basin (AU) [#993](https://github.com/CLIMADA-project/climada_python/pull/993)
```
chahank: Please complete.

climada/hazard/tc_tracks.py
NicolasColombi: I am not sure to understand: do you want to remove this or move it under`get_basin` ? can you clarify ? for the moment, I moved it.

climada/hazard/tc_tracks.py
NicolasColombi: Agree, I implemented it, though I think the code could be written better.

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +415,87 @@ def subset(self, filterdict):
 
         return out
 
+    def get_basins(track):
+        """Identify the tropical-cyclone basins crossed by a single track.
+
+        The function appends the latitude and longitude coordinates of the input track
+        into a GeoDataFrame of point geometries and performs a spatial join with the
+        global BASINS_GDF polygons (each representing a tropical-cyclone basin).
+        It returns the basin name(s) for every point along the track.
```
chahank:
```suggestion
        Provides the basin name for every point along the track.
        The basins are defined according to the STORM definition:
        https://www.nature.com/articles/s41597-020-0381-2
        The names are:
        WP: West Pacific
        NA: North Atlantic
        NI: North Indian
        SP: South Pacific
        SI: South Indian
        EP: East Pacific
```

chahank: Correct?

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +415,87 @@ def subset(self, filterdict):
...
+            The index matches the index of the input track coordinates."""
+
+        BASINS_GDF = gpd.GeoDataFrame(
+            {"basin": b, "geometry": b.value} for b in Basin_bounds_storm
+        )
+
+        track_coordinates = gpd.GeoDataFrame(
+            geometry=gpd.points_from_xy(track.lon, track.lat)
+        )
+        return track_coordinates.sjoin(BASINS_GDF, how="left", predicate="within").basin
```
chahank:
```suggestion
        basins_gdf = gpd.GeoDataFrame(
            {"basin": b, "geometry": b.value} for b in BasinBoundsStorm
        )

        track_coordinates = gpd.GeoDataFrame(
            geometry=gpd.points_from_xy(track.lon, track.lat)
        )
        return track_coordinates.sjoin(basins_gdf, how="left", predicate="within").basin
```

climada/hazard/tc_tracks.py
```diff
@@ -193,6 +195,97 @@
 dataset using STORM. Scientific Data 7(1): 40."""
 
 
+class Basin_bounds_storm(Enum):
```
chahank:
```suggestion
class BasinBoundsStorm(Enum):
```
We have consistent styles for naming classes, variables, attributes etc. The previous suggestion does not conform to the style.

CHANGELOG.md
```diff
@@ -106,6 +106,7 @@ Removed:
 ### Added
 
 - `climada.entity.impact_funcs.trop_cyclone.ImpfSetTropCyclone.get_impf_id_regions_per_countries` function [#1034](https://github.com/CLIMADA-project/climada_python/pull/1034)
+- `climada.hazard.tc_tracks.Basin_bounds_storm` Enum class `climada.hazard.tc_tracks.subset_by_basin` function [#1031](https://github.com/CLIMADA-project/climada_python/pull/1031)
```
chahank:
```suggestion
- `climada.hazard.tc_tracks.BasinBoundsStorm` Enum class `climada.hazard.tc_tracks.subset_by_basin` function [#1031](https://github.com/CLIMADA-project/climada_python/pull/1031)
```

climada/hazard/tc_tracks.py
```diff
@@ -322,6 +415,87 @@ def subset(self, filterdict):
...
+            else:  # if every basin crossed is of interest (origin = False)
+                touched = TCTracks.get_basins(track).dropna().drop_duplicates()
+                if touched.size:
+                    for basin in touched:
+                        basins_dict[basin.name].append(track)
+                else:
+                    tracks_outside_basin.append(track)
+
+        # return TCTracks objetcs
+        for basin in Basin_bounds_storm:
```
chahank:
```suggestion
        for basin in BasinBoundsStorm:
```

climada/hazard/tc_tracks.py
chahank: I still think the naming is not well chosen because it is only one way to define basins, but let's keep it as is for the moment. Thanks!

CHANGELOG.md
chahank: Thanks!

climada/hazard/tc_tracks.py
NicolasColombi: > Correct?

Yes!

PR #1033, opened 2025-03-26
Remove pandas-datareader

### LINKED ISSUE #987: Evaluate alternatives to pandas-datareader for downloading World Bank data
**Is your feature request related to a problem? Please describe.**
The `pandas-datareader` package has not been updated since 2021-07-13 and is now breaking with Python 3.12 and newer versions due to the removal of `distutils`. This package is currently used in `util.finance.py` for downloading World Bank data and in related tests.

**Describe the solution you'd like**
Identify and migrate to an alternative for downloading World Bank data. The steps could include:
1. Researching viable replacements for `pandas-datareader`, such as `wbdata` ([see here](https://pypi.org/project/wbdata/)) or direct API usage via libraries like `requests` or `pandas`.
2. Refactoring `util.finance.py` to use the new solution.
3. Updating related tests to ensure compatibility with the new implementation.

**Describe alternatives you've considered**
- Continuing to use `pandas-datareader` by forking and maintaining a private version. However, this would increase the maintenance burden on the project.
- Temporarily patching the current implementation to support Python 3.12 and beyond, though this is not a sustainable long-term solution.

For reference, a related issue in `pandas-datareader` highlights these problems: https://github.com/pydata/pandas-datareader/issues/977

peanutfun:
Changes proposed in this PR:

Use JSON/pandas solution for downloading World Bank indicator data.

* Add function `download_world_bank_indicator`.
* Add unit test.
* Remove `pandas-datareader` from requirements.

This PR fixes #987 

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,6 @@ jobs:
     needs: build-and-test
     with:
       core_branch: ${{ github.ref }}
-      petals_branch: develop
+      petals_branch: add-lxml-dependency  # TODO: DO NOT MERGE LIKE THIS
     permissions:
       checks: write
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,6 @@ jobs:
     needs: build-and-test
     with:
       core_branch: ${{ github.ref }}
-      petals_branch: add-lxml-dependency  # TODO: DO NOT MERGE LIKE THIS
+      petals_branch: develop
     permissions:
       checks: write
--- a/climada/util/finance.py
+++ b/climada/util/finance.py
@@ -21,6 +21,7 @@ Finance functionalities.
 
 __all__ = ["net_present_value", "income_group", "gdp"]
 
+import json
 import logging
 import shutil
 import warnings
@@ -31,7 +32,6 @@ import numpy as np
 import pandas as pd
 import requests
 from cartopy.io import shapereader
-from pandas_datareader import wb
 
 from climada.util.constants import SYSTEM_DIR
 from climada.util.files_handler import download_file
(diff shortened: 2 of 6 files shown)


peanutfun:
> Why the change of mind from https://github.com/CLIMADA-project/climada_python/issues/987#issuecomment-2747415765 though?

It felt to me like adding a dependency for a single task, which can also be achieved without it, was too much. I think the solution is short enough to implement it on our end instead of adding `wbdata`, whose entire feature set we currently do not intend to use.

peanutfun:
https://github.com/CLIMADA-project/climada_petals/pull/168 Adds the `lxml` dependency to Petals, which was needed but never added explicitly. `pandas-datareader` introduced it into our dependency tree.

.github/workflows/ci.yml
```diff
@@ -74,6 +74,6 @@ jobs:
     needs: build-and-test
     with:
       core_branch: ${{ github.ref }}
-      petals_branch: develop
+      petals_branch: add-lxml-dependency  # TODO: DO NOT MERGE LIKE THIS
```
peanutfun, line 77: Revert once https://github.com/CLIMADA-project/climada_petals/pull/168 is merged
```suggestion
      petals_branch: develop
```

climada/util/finance.py
```diff
@@ -181,6 +180,68 @@ def gdp(cntry_iso, ref_year, shp_file=None, per_capita=False):
...
+                    "correct country code and indicator ID?"
+                )
+        # If no, we should be fine
+        except KeyError:
+            pass
+
+        # Update the data
+        pages = json_data[0]["pages"]
+        page = page + 1
+        raw_data.append(json_data[1])
```
emanuel-schmid, line 230:
```suggestion
        raw_data.extend(json_data[1])
```

climada/util/finance.py
```diff
@@ -181,6 +180,68 @@ def gdp(cntry_iso, ref_year, shp_file=None, per_capita=False):
...
+
+    # Create dataframe
+    data = pd.concat([pd.DataFrame.from_records(rec) for rec in raw_data])
+
+    # Parse dates
+    if parse_dates:
```
emanuel-schmid, line 236: ideally the caller/user doesn't have to care whether or not the date has to be parsed to a `datetime` object.
I'd suggest to either derive `parse_dates` from the data itself or return it as is.

climada/util/finance.py
```diff
@@ -181,6 +180,68 @@ def gdp(cntry_iso, ref_year, shp_file=None, per_capita=False):
...
+        except KeyError:
+            pass
+
+        # Update the data
+        pages = json_data[0]["pages"]
+        page = page + 1
+        raw_data.append(json_data[1])
+
+    # Create dataframe
+    data = pd.concat([pd.DataFrame.from_records(rec) for rec in raw_data])
```
emanuel-schmid, line 233:
```suggestion
    data = DataFrame(raw_data)
```

emanuel-schmid (changes requested):
Excellent Job! 😃
Just couple of minor remarks from my side.

climada/util/finance.py
peanutfun: The World Bank API data is most often stored in yearly intervals, but sometimes in other intervals, too (quarterly, monthly, etc.). We could first try to parse as `int`, and if that fails, try parsing as `DatetimeIndex` instead. However, I would not remove the parameter, because users may not want to retrieve DatetimeIndex even for integer data 🤔

climada/util/finance.py
peanutfun:
```suggestion
    data = pd.DataFrame.from_records(raw_data)
```

emanuel-schmid:
Emerging from the Jenkins Compatibility test: `worldbank` fails for Anguilla (AIA) because the worldbank API (https://api.worldbank.org/v2/country/AIA/indicators/NY.GDP.MKTP.CD/?format=json) returns 
```json
[{"page":0,"pages":0,"per_page":0,"total":0,"sourceid":null,"lastupdated":null},null]
```
Now this raises a `TypeError` at `.extend` which is not caught in `gdp`, leading to failing test in `climada_petalsl.test.test_black_marble`.

@peanutfun What should we do about it? Raise a "ValueError" in `worldbank_indicator`? Raise a custom error? Return an empty DataFrame (columns but no rows)? Simply catch the exception in `gdp`?

peanutfun:
@emanuel-schmid Interesting! This seems to be a case where there is no data for this particular indicator, but the country code itself is known/valid. I'll check what `pandas-datareader` does in that case, and try to implement the same to stay compatible.

peanutfun:
> Are you sure? I suspect the error used to be thrown in world_bank and caught in gdp

Indeed, good catch. I'll throw a `ValueError` then

peanutfun:
@emanuel-schmid I updated the code accordingly. The failing tests on Jenkins are unrelated. GitHub tests pass.

PR #1037, opened 2025-04-02
Integrates risk trajectories within climada

spjuhel:
Changes proposed in this PR:
- [WIP]
- 

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/trajectories/risk_trajectory.py
+++ b/climada/trajectories/risk_trajectory.py
@@ -1,0 +1,626 @@
+"""
+This file is part of CLIMADA.
+
+Copyright (C) 2017 ETH Zurich, CLIMADA contributors listed in AUTHORS.
+
+CLIMADA is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License as published by the Free
+Software Foundation, version 3.
+
+CLIMADA is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
+
+---
+
+"""
+
+import datetime
+import logging
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from climada.engine.impact_calc import ImpactCalc
+from climada.entity.disc_rates.base import DiscRates
+from climada.trajectories.riskperiod import RiskPeriod
+from climada.trajectories.snapshot import Snapshot, pairwise
+
+LOGGER = logging.getLogger(__name__)
+
+
+class RiskTrajectory:
+
+    _grouper = ["measure", "metric"]
+
+    def __init__(
+        self,
+        snapshots_list: list[Snapshot],
+        risk_disc: DiscRates | None = None,
+        metrics: list[str] = ["aai", "eai", "rp"],
+        return_periods: list[int] = [100, 500, 1000],
+        compute_groups=False,
+        risk_transf_cover=None,
+        risk_transf_attach=None,
+    ):
+        "docstring"
+        self._metrics_up_to_date: bool = False
+        self.metrics = metrics
+        self.return_periods = return_periods
+        self.start_date = min([snapshot.date for snapshot in snapshots_list])
+        self.end_date = max([snapshot.date for snapshot in snapshots_list])
+        self.risk_disc = risk_disc
+        self.risk_transf_cover = risk_transf_cover
+        self.risk_transf_attach = risk_transf_attach
+        LOGGER.debug("Computing risk periods")
+        self.risk_periods = self._calc_risk_periods(snapshots_list)
+        self._update_risk_metrics(compute_groups=compute_groups)
+
+    def _calc_risk_periods(self, snapshots):
+        return [
+            RiskPeriod(start_snapshot, end_snapshot)
+            for start_snapshot, end_snapshot in pairwise(snapshots)
+        ]
+
+    def _update_risk_metrics(self, compute_groups=False):
+        results_df = []
+        for period in self.risk_periods:
+            results_df.append(
+                bayesian_mixer_opti(
+                    period,
+                    self.metrics,
+                    self.return_periods,
+                    compute_groups,
+                    all_groups_name="All",
+                )
+            )
+        results_df = pd.concat(results_df, axis=0)
+
+        # duplicate rows arise from overlapping end and start if there's more than two snapshots
+        results_df.drop_duplicates(inplace=True)
+
+        # reorder the columns (but make sure not to remove possibly important ones in the future)
+        columns_to_front = ["date", "measure", "metric"]
+        if compute_groups:
+            columns_to_front = ["group"] + columns_to_front
+        self._annual_risk_metrics = results_df[
+            columns_to_front
+            + [
+                col
+                for col in results_df.columns
+                if col not in columns_to_front + ["group", "risk", "rp"]
+            ]
+            + ["risk"]
+        ]
+        self._metrics_up_to_date = True
+
+    @staticmethod
+    def _get_risk_periods(
+        risk_periods, start_date: datetime.date, end_date: datetime.date
+    ):
+        return [
+            period
+            for period in risk_periods
+            if (start_date >= period.start_date or end_date <= period.end_date)
+        ]
+
+    def _calc_annual_risk_metrics(self, npv=True):
+        def npv_transform(group):
+            start_date = group.index.get_level_values("date").min()
+            end_date = group.index.get_level_values("date").max()
+            return calc_npv_cash_flows(
+                group.values, start_date, end_date, self.risk_disc
+            )
+
+        if self._metrics_up_to_date:
+            df = self._annual_risk_metrics
+        else:
+            self._update_risk_metrics()
+            df = self._annual_risk_metrics
+
+        if npv:
+            df = df.set_index("date")
+            grouper = self._grouper
+            if "group" in df.columns:
+                grouper = ["group"] + grouper
+
+            df["risk"] = df.groupby(
+                grouper,
+                dropna=False,
+                as_index=False,
+                group_keys=False,
+            )["risk"].transform(npv_transform)
+            df = df.reset_index()
+
+        return df
+
+    @classmethod
+    def _calc_periods_risk(cls, df: pd.DataFrame, time_unit="year", colname="risk"):
+        def identify_continuous_periods(group, time_unit):
+            # Calculate the difference between consecutive dates
+            if time_unit == "year":
+                group["date_diff"] = group["date"].dt.year.diff()
+            if time_unit == "month":
+                group["date_diff"] = group["date"].dt.month.diff()
+            if time_unit == "day":
+                group["date_diff"] = group["date"].dt.day.diff()
+            if time_unit == "hour":
+                group["date_diff"] = group["date"].dt.hour.diff()
+            # Identify breaks in continuity
+            group["period_id"] = (group["date_diff"] != 1).cumsum()
+            return group
+
+        grouper = cls._grouper
+        if "group" in df.columns:
+            grouper = ["group"] + grouper
+
+        df_sorted = df.sort_values(by=cls._grouper + ["date"])
+        # Apply the function to identify continuous periods
+        df_periods = df_sorted.groupby(grouper, dropna=False, group_keys=False).apply(
+            identify_continuous_periods, time_unit
+        )
+
+        # Group by the identified periods and calculate start and end dates
+        df_periods = (
+            df_periods.groupby(grouper + ["period_id"], dropna=False)
+            .agg(
+                start_date=pd.NamedAgg(column="date", aggfunc="min"),
+                end_date=pd.NamedAgg(column="date", aggfunc="max"),
+                total=pd.NamedAgg(column=colname, aggfunc="sum"),
+            )
+            .reset_index()
+        )
+
+        df_periods["period"] = (
+            df_periods["start_date"].astype(str)
+            + " to "
+            + df_periods["end_date"].astype(str)
+        )
+        df_periods = df_periods.rename(columns={"total": f"{colname}"})
+        df_periods = df_periods.drop(["period_id", "start_date", "end_date"], axis=1)
+        return df_periods[
+            ["period"] + [col for col in df_periods.columns if col != "period"]
+        ]
+
+    @property
+    def all_dates_risk_metrics(self):
+        return self._calc_risk_metrics(total=False, npv=True)
+
+    @property
+    def total_risk_metrics(self):
+        return self._calc_risk_metrics(total=True, npv=True)
+
+    def _calc_risk_metrics(self, total=False, npv=True):
+        df = self._calc_annual_risk_metrics(npv=npv)
+        if total:
+            return self._calc_periods_risk(df)
+
+        return df
+
+    def _calc_waterfall_plot_data(self, start_date=None, end_date=None):
+        start_date = self.start_date if start_date is None else start_date
+        end_date = self.end_date if end_date is None else end_date
+        considered_risk_periods = self._get_risk_periods(
+            self.risk_periods, start_date=start_date, end_date=end_date
+        )
+
+        risk_component = {
+            str(period.start_date)
+            + "-"
+            + str(period.end_date): self._calc_risk_component(period)
+            for period in considered_risk_periods
+        }
+        risk_component = pd.concat(
+            risk_component.values(), keys=risk_component.keys(), names=["Period"]
+        ).reset_index()
+        risk_component = risk_component.loc[
+            (risk_component["date"].dt.date >= start_date)
+            & (risk_component["date"].dt.date <= end_date)
+        ]
+        risk_component["Base risk"] = risk_component["Base risk"].min()
+        risk_component[["Change in Exposure", "Change in Hazard (with Exposure)"]] = (
+            risk_component[["Change in Exposure", "Change in Hazard (with Exposure)"]]
+            .replace(0, None)
+            .ffill()
+            .fillna(0.0)
+        )
+        return risk_component
+
+    def _calc_risk_component(self, period: RiskPeriod):
+        imp_mats_H0 = period.imp_mats_0
+        imp_mats_H1 = period.imp_mats_1
+        freq_H0 = period.snapshot0.hazard.frequency
+        freq_H1 = period.snapshot1.hazard.frequency
+        dately_eai_H0, dately_eai_H1 = calc_dately_eais(
+            imp_mats_H0, imp_mats_H1, freq_H0, freq_H1
+        )
+        dately_aai_H0, dately_aai_H1 = calc_dately_aais(dately_eai_H0, dately_eai_H1)
+        prop_H1 = np.linspace(0, 1, num=len(period.date_idx))
+        prop_H0 = 1 - prop_H1
+        dately_aai = prop_H0 * dately_aai_H0 + prop_H1 * dately_aai_H1
+
+        risk_dev_0 = dately_aai_H0 - dately_aai[0]
+        risk_cc_0 = dately_aai - (risk_dev_0 + dately_aai[0])
+        df = pd.DataFrame(
+            {
+                "Base risk": dately_aai - (risk_dev_0 + risk_cc_0),
+                "Change in Exposure": risk_dev_0,
+                "Change in Hazard (with Exposure)": risk_cc_0,
+            },
+            index=period.date_idx,
+        )
+        return df.round(1)
+
+    def plot_dately_waterfall(self, ax=None, start_date=None, end_date=None):
+        if ax is None:
+            _, ax = plt.subplots(figsize=(12, 6))
+        start_date = self.start_date if start_date is None else start_date
+        end_date = self.end_date if end_date is None else end_date
+        risk_component = self._calc_waterfall_plot_data(
+            start_date=start_date, end_date=end_date
+        )
+        risk_component.plot(ax=ax, kind="bar", x="date", stacked=True)
+        # Construct y-axis label and title based on parameters
+        value_label = "USD"
+        title_label = (
+            f"Risk between {start_date} and {end_date} (Annual Average impact)"
+        )
+
+        ax.set_title(title_label)
+        ax.set_ylabel(value_label)
+        return ax
+
+    def plot_waterfall(self, ax=None, start_date=None, end_date=None):
+        start_date = self.start_date if start_date is None else start_date
+        end_date = self.end_date if end_date is None else end_date
+        risk_component = self._calc_waterfall_plot_data(
+            start_date=start_date, end_date=end_date
+        )
+        if ax is None:
+            _, ax = plt.subplots(figsize=(8, 5))
+
+        risk_component = risk_component.loc[
+            (risk_component["date"].dt.date == end_date)
+        ].squeeze()
+
+        labels = [
+            f"Risk {start_date}",
+            f"Exposure {end_date}",
+            f"Hazard {end_date}¹",
+            f"Total Risk {end_date}",
+        ]
+        values = [
+            risk_component["Base risk"],
+            risk_component["Change in Exposure"],
+            risk_component["Change in Hazard (with Exposure)"],
+            risk_component["Base risk"]
+            + risk_component["Change in Exposure"]
+            + risk_component["Change in Hazard (with Exposure)"],
+        ]
+        bottoms = [
+            0.0,
+            risk_component["Base risk"],
+            risk_component["Base risk"] + risk_component["Change in Exposure"],
+            0.0,
+        ]
+
+        ax.bar(
+            labels,
+            values,
+            bottom=bottoms,
+            edgecolor="black",
+            color=["tab:blue", "tab:orange", "tab:green", "tab:red"],
+        )
+        for i in range(len(values)):
+            ax.text(
+                labels[i],
+                values[i] + bottoms[i],
+                f"{values[i]:.0e}",
+                ha="center",
+                va="bottom",
+                color="black",
+            )
+
+        # Construct y-axis label and title based on parameters
+        value_label = "USD"
+        title_label = f"Risk at {start_date} and {end_date} (Annual Average impact)"
+
+        ax.set_title(title_label)
+        ax.set_ylabel(value_label)
+        # ax.tick_params(axis='x', labelrotation=90,)
+        ax.annotate(
+            """¹: The increase in risk due to hazard denotes the difference in risk with future exposure
+and hazard compared to risk with future exposure and present hazard.""",
+            xy=(0.0, -0.15),
+            xycoords="axes fraction",
+            ha="left",
+            va="center",
+            fontsize=8,
+        )
+
+        return ax
+
+
+def calc_npv_cash_flows(cash_flows, start_date, end_date=None, disc=None):
+    # If no discount rates are provided, return the cash flows as is
+    if not disc:
+        return cash_flows
+
+    if not isinstance(cash_flows, pd.Series) or not isinstance(
+        cash_flows.index, pd.DatetimeIndex
+    ):
+        raise ValueError("cash_flows must be a pandas Series with a datetime index")
+
+    # Determine the end date if not provided
+    if end_date is None:
+        end_date = cash_flows.index[-1]
+
+    df = cash_flows.to_frame(name="cash_flow")
+    df["year"] = df.index.year
+
+    # Merge with the discount rates based on the year
+    df = df.merge(
+        pd.DataFrame({"year": disc.years, "rate": disc.rates}), on="year", how="left"
+    )
+
+    # Calculate the discount factors
+    df["discount_factor"] = (1 / (1 + df["rate"])) ** (
+        df.index - start_date
+    ).days / 365.25
+
+    # Apply the discount factors to the cash flows
+    df["npv_cash_flow"] = df["cash_flow"] * df["discount_factor"]
+
+    return df["npv_cash_flow"]
+
+
+def calc_dately_eais(imp_mats_0, imp_mats_1, frequency_0, frequency_1):
+    """
+    Calculate dately expected annual impact (EAI) values for two scenarios.
+
+    Parameters
+    ----------
+    imp_mats_0 : list of np.ndarray
+        List of interpolated impact matrices for scenario 0.
+    imp_mats_1 : list of np.ndarray
+        List of interpolated impact matrices for scenario 1.
+    frequency_0 : np.ndarray
+        Frequency values associated with scenario 0.
+    frequency_1 : np.ndarray
+        Frequency values associated with scenario 1.
+
+    Returns
+    -------
+    tuple
+        Tuple containing:
+        - dately_eai_exp_0 : list of float
+            Dately expected annual impacts for scenario 0.
+        - dately_eai_exp_1 : list of float
+            Dately expected annual impacts for scenario 1.
+    """
+    dately_eai_exp_0 = [
+        ImpactCalc.eai_exp_from_mat(imp_mat, frequency_0) for imp_mat in imp_mats_0
+    ]
+    dately_eai_exp_1 = [
+        ImpactCalc.eai_exp_from_mat(imp_mat, frequency_1) for imp_mat in imp_mats_1
+    ]
+    return dately_eai_exp_0, dately_eai_exp_1
+
+
+def calc_dately_aais(dately_eai_exp_0, dately_eai_exp_1):
+    """
+    Calculate dately aggregate annual impact (AAI) values for two scenarios.
+
+    Parameters
+    ----------
+    dately_eai_exp_0 : list of float
+        Dately expected annual impacts for scenario 0.
+    dately_eai_exp_1 : list of float
+        Dately expected annual impacts for scenario 1.
+
+    Returns
+    -------
+    tuple
+        Tuple containing:
+        - dately_aai_0 : list of float
+            Aggregate annual impact values for scenario 0.
+        - dately_aai_1 : list of float
+            Aggregate annual impact values for scenario 1.
+    """
+    dately_aai_0 = [
+        ImpactCalc.aai_agg_from_eai_exp(eai_exp) for eai_exp in dately_eai_exp_0
+    ]
+    dately_aai_1 = [
+        ImpactCalc.aai_agg_from_eai_exp(eai_exp) for eai_exp in dately_eai_exp_1
+    ]
+    return dately_aai_0, dately_aai_1
+
+
+def calc_freq_curve(imp_mat_intrpl, frequency, return_per=None):
+    """
+    Calculate the frequency curve
+
+    Parameters:
+    imp_mat_intrpl (np.array): The interpolated impact matrix
+    frequency (np.array): The frequency of the hazard
+    return_per (np.array): The return period
+
+    Returns:
+    ifc_return_per (np.array): The impact exceeding frequency
+    ifc_impact (np.array): The impact exceeding the return period
+    """
+
+    # Calculate the at_event make the np.array
+    at_event = np.sum(imp_mat_intrpl, axis=1).A1
+
+    # Sort descendingly the impacts per events
+    sort_idxs = np.argsort(at_event)[::-1]
+    # Calculate exceedence frequency
+    exceed_freq = np.cumsum(frequency[sort_idxs])
+    # Set return period and impact exceeding frequency
+    ifc_return_per = 1 / exceed_freq[::-1]
+    ifc_impact = at_event[sort_idxs][::-1]
+
+    if return_per is not None:
+        interp_imp = np.interp(return_per, ifc_return_per, ifc_impact)
+        ifc_return_per = return_per
+        ifc_impact = interp_imp
+
+    return ifc_impact
+
+
+def calc_dately_rps(imp_mats_0, imp_mats_1, frequency_0, frequency_1, return_periods):
+    """
+    Calculate dately return period impact values for two scenarios.
+
+    Parameters
+    ----------
+    imp_mats_0 : list of np.ndarray
+        List of interpolated impact matrices for scenario 0.
+    imp_mats_1 : list of np.ndarray
+        List of interpolated impact matrices for scenario 1.
+    frequency_0 : np.ndarray
+        Frequency values for scenario 0.
+    frequency_1 : np.ndarray
+        Frequency values for scenario 1.
+    return_periods : list of int
+        Return periods to calculate impact values for.
+
+    Returns
+    -------
+    tuple
+        Tuple containing:
+        - rp_0 : list of np.ndarray
+            Dately return period impact values for scenario 0.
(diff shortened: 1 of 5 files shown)


chahank:
Awesome work! 

Small point, I would use `datetime.datetime` objects instead of `datetime.data` since sometimes hazards are used in the context of weather forecasts with sub-daily resolution.

PS: where should we have discussions on this big additiopn? I think the pull request might not be the best location. Shall we maybe open an issue or a discussion?

chahank:
I would look for another name than `freq`. The word `frequency` is already used to describe the probability/frequency of hazard events. I think this would be quite confusing. Maybe `intervals` ?

chahank:
I think it is a bit unclear to me which class in the end contains the data from the risk trajectories.

chahank:
How about `per_datetime_risk_metrics` as I think you propose once too? Maybe too complicated ...

chahank:
> Add a check that group_id are identical for Exposures or handle that case (how?)

What is the reason for this? Can one not simply compute the impact metrics per group if needed and that is it? I am not sure to see the issues at a first glance (I am sure there are some :D)

spjuhel:
> I would look for another name than `freq`. The word `frequency` is already used to describe the probability/frequency of hazard events. I think this would be quite confusing. Maybe `intervals` ?

The actual name is `time_frequency`, Indeed `interval` is more fitting.

luseverin (changes requested):
Hi @spjuhel, super nice work!
Besides some comments on the juypter notebook (typos and a few suggestions), I have the following comments:
- Not sure about the meaning of the group_id parameter and why it is required. When not specified, RiskTrajectory() fails with a KeyError: 'group_id' . As far as I understood this should be an optional argument, or did I misundertand this?
- What about changes in vulnerability/impact function? I think you mention in the tuto that is complicated due to non-linearity, but the same argument applies to hazard and exposure. Wouldn't it be possible to have the risk delta attributable to changes in vulnerability in the same way as it is done for hazard and vulnerability already? Something like this: RiskDelta(v) = Risk(haz_past exp_past, vul_fut) - Risk(haz_past, exp_past, vul_past).  Not sure this makes sense..

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+            List of interpolated impact matrices for each time points in the specified range.
+        """
+
+        def interpolate_sm(mat_start, mat_end, time, time_points):
+            """Perform exponential interpolation between two matrices for a specified time point."""
+            if time > time_points:
+                raise ValueError("time point must be within the range")
+
+            # Convert matrices to logarithmic domain
+            log_mat_start = np.log(mat_start.toarray() + np.finfo(float).eps)
```
chahank: You might need to specify what exponent you are you using be default. Or allow the user to choose.

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _apply_risk_transfer(
+        self,
+        impacts: tuple[Impact, Impact, Impact, Impact],
+        risk_transf_attach: float | None,
+        risk_transf_cover: float | None,
+        calc_residual: bool,
+    ):
+        """Apply risk transfer to the calculated impacts."""
+        if risk_transf_attach is not None and risk_transf_cover is not None:
```
chahank: I think canonically in the impactcalc class it is fine if either is defined. Any reason to require both here?

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,160 @@
...
+    ):
+        """
+        Calculate either the residual or the risk transfer impact matrix.
+
+        The impact matrix is adjusted based on the total impact for each event.
+        When calculating the residual impact, the result is the total impact minus
+        the risk layer. The risk layer is defined as the minimum of the cover and
+        the maximum of the difference between the total impact and the attachment.
+        If `calc_residual` is False, the function returns the risk layer matrix
+        instead of the residual.
```
chahank: I think this is a bit cryptic to read. What is a risk layer for instance?

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,160 @@
...
+        -------
+        scipy.sparse.csr_matrix
+            The adjusted impact matrix, either residual or risk transfer.
+
+        """
+
+        if risk_transf_attach and risk_transf_cover:
+            imp_mat = copy.deepcopy(imp_mat)
+            # Calculate the total impact per event
+            total_at_event = imp_mat.sum(axis=1).A1
```
chahank: Would it maybe be cleaner to use the methods defined in impactcalc or impact to compute this?

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,160 @@
...
+                numerator = residual_at_event
+            else:
+                numerator = transfer_at_event
+
+            rescale_impact_values = np.divide(
+                numerator,
+                total_at_event,
+                out=np.zeros_like(numerator, dtype=float),
+                where=total_at_event != 0,
+            )
```
chahank: Does this mean that the insurance cover is shared among all exposure points and equally distributed? If yes, I think this needs to be made very clear in the docstring. Otherwise I am confused about what this relative residuals / transfers mean.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+class LinearInterpolation(InterpolationStrategy):
+    """Linear interpolation strategy."""
+
+    def interpolate(self, imp_E0, imp_E1, time_points: int):
+        try:
+            return self.interpolate_imp_mat(imp_E0, imp_E1, time_points)
+        except ValueError as e:
+            if str(e) == "inconsistent shapes":
+                raise ValueError(
+                    "Interpolation between impact matrices of different shapes"
```
chahank: Very cryptic for a user if triggered not in this method specifically, when a method calls this method.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+                )
+            else:
+                raise e
+
+    @staticmethod
+    def interpolate_imp_mat(imp0, imp1, time_points):
+        """Interpolate between two impact matrices over a specified time range.
```
chahank: what is a time range? Here there are only time points...

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+            The impact calculation for the starting time.
+        imp1 : ImpactCalc
+            The impact calculation for the ending time.
+        time_points:
+            The number of points to interpolate.
+
+        Returns
+        -------
+        list of np.ndarray
+            List of interpolated impact matrices for each time points in the specified range.
```
chahank: There is no specified range in this method.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+
+        def interpolate_sm(mat_start, mat_end, time, time_points):
+            """Perform linear interpolation between two matrices for a specified time point."""
+            if time > time_points:
+                raise ValueError("time point must be within the range")
```
chahank: Quite cryptic.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+            ratio = time / (time_points - 1)
+
+            # Convert the input matrices to a format that allows efficient modification of its elements
+            mat_start = lil_matrix(mat_start)
+            mat_end = lil_matrix(mat_end)
+
+            # Perform the linear interpolation
+            mat_interpolated = mat_start + ratio * (mat_end - mat_start)
+
+            return csr_matrix(mat_interpolated)
```
chahank: Might it be a big effiency gain to have the impact matrices to be lil_matrix format all the time instead of crs_matrix? I think the later is mostly a legacy choice.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,141 @@
...
+        time_points:
+            The number of points to interpolate.
+
+        Returns
+        -------
+        list of np.ndarray
+            List of interpolated impact matrices for each time points in the specified range.
+        """
+
+        def interpolate_sm(mat_start, mat_end, time, time_points):
```
chahank: time is a bit a weird variable name, as if I understand correctly it should be a counter right?

climada/trajectories/interpolation.py
chahank: Also, is the array cast not potentially very very inefficient? Would it not be better to work on the .data attribute instead? This would require building an extract log sparse matrix but might avoid blowing up the memory.

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,760 @@
...
+
+    def __init__(
+        self,
+        snapshots_list: list[Snapshot],
+        *,
+        interval_freq: str = "YS",
+        all_groups_name: str = "All",
+        risk_disc: DiscRates | None = None,
+        risk_transf_cover=None,
+        risk_transf_attach=None,
```
chahank: So this would be global covers and attachment for the whole exposure? What about the cover and attachment per exposure point as saved in the exposure?

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,760 @@
...
+        return self._per_period_risk(df)
+
+    def _compute_metrics(
+        self, metric_name: str, metric_meth: str, npv: bool = True, **kwargs
+    ):
+        """Helper method to compute metrics."""
+        df = self._generic_metrics(
+            npv=npv, metric_name=metric_name, metric_meth=metric_meth, **kwargs
+        )
+        return df
```
chahank: Is this method not completely useless?

climada/trajectories/impact_calc_strat.py
spjuhel: I think so, this is legacy code from Victor. I don't really know why he didn't use existing options.

I think I am just going to remove that part, and let it be handled by the ImpactCalc from what is defined in the exposure.

chahank: Ok, understood. I think you will still have to handle this explicitly somewhere as the default in ImpactCalc is (or was?) not well adapted for the adaptation option case, and because it is unclear whether it is the residuals or the transfer.

climada/trajectories/risk_trajectory.py
spjuhel: Only in appearance. It has a use in option_appraisal, where we need "_compute_metrics" to remain the same, but `_generic_metrics` to have a different signature.

chahank:
Just a note: you are modifying 145 files in this branch, I do not think this is all related to this PR ;).

climada/trajectories/impact_calc_strat.py
spjuhel: Should be better now

climada/trajectories/impact_calc_strat.py
spjuhel: Should be clearer now

spjuhel:
- I pushed the unittest and have almost full coverage.
- I am pretty sure the core code is ready for review. 
- I just updated the TODOs on top for tracking some remaining work.

chahank:
Small comments to the tutorial. They are really just from a user’s perspective at the moment.

- Instead of a wall of text, maybe start with a quick example which goes within one or two cells to the results. I can help with that :)
- Is the `snapshot.measure` supposed to be a `MeasureSet` ? Shall we then use the name `measureset` or similar?
- The method `snap.apply_measure()` cannot apply the measure that are already in `snapshot.measure` ? This is a bit unintuitive. Should `snapshot.measure` and `.apply_measure` be private?
- `default_rp` at [50, 100, 500] is probably a bit ambitious. Is it necessary to have a default?
- The `risk_traj.eai_exp` contains the column `coord_id` , but `risk_traj.coord_id` does not exist. It is unclear what one should do with this information.
- `risk_traj.risk_periods` returns a list of python pointers, so quite useless (in particular since it is a property without docstring, so I cannot make sense of that output). Should this be private?
- `risk_traj.risk_components_metrics()` maybe change components to contribution ? Also, there is the `index` column which is unclear.
- `plot_per_date_waterfall()` maybe rename `plot_time_waterfall()` as it is more than just per date?
- I think the handling of the RP is a bit confusing now. There is a default set, but it is unclear how and why. Then there is a method to access them, `isk_traj.return_periods_metrics` which requires the RPs as input. However, the method `risk_traj.per_period_risk_metrics()` (and others) just compute the RP based on the default.
- The use of the method `isk_traj.npv_transform` is unclear. Should this be private? Idem for `risk_traj.identify_continuous_periods`
- In a future, the discounted vs not discounted risk plot might be implemented directly, looks useful.
- Maybe we could make Enum classes to define the choices for the interpolation and impact calculations methods.

Really exciting, overall the use looks very straightforward! (edited)

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,163 @@
...
+CLIMADA is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
+
+---
+
+This modules implements the Snapshot and SnapshotsCollection classes.
```
chahank:
```suggestion
This modules implements the Snapshot impact computation strategies.
```

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def compute_impacts_pre_transfer(
+        self,
+        snapshot0: Snapshot,
+        snapshot1: Snapshot,
+        future: tuple[int, int, int],
+    ) -> Impact:
+        exp = snapshot1.exposure if future[0] else snapshot0.exposure
+        haz = snapshot1.hazard if future[1] else snapshot0.hazard
+        vul = snapshot1.impfset if future[2] else snapshot0.impfset
```
chahank: I am a bit confused by this. `future` is supposed to be an integer, but here you use it only as a boolean.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,246 @@
...
+
+    Returns a list of `interpolation_range` matrices linearly interpolated between
+    `mat_start` and `mat_end`.
+    """
+    res = []
+    for point in range(interpolation_range):
+        ratio = point / (interpolation_range - 1)
+        mat_interpolated = mat_start + ratio * (mat_end - mat_start)
+        res.append(mat_interpolated)
+    return res
```
chahank:
```suggestion
    return [mat_start + prop* (mat_end - mat_start) for prop in np.linspace(0,1,interpolation_range)
    ]
```
I think list comprehension are almost always better memory-wise than loop + append.

chahank: Depending on when / how it is used later, you might also consider having a generator.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,246 @@
...
+    mat_end.data = np.log(mat_end.data + np.finfo(float).eps) / np.log(rate)
+
+    # Perform linear interpolation in the logarithmic domain
+    res = []
+    for point in range(interpolation_range):
+        ratio = point / (interpolation_range - 1)
+        mat_interpolated = mat_start * (1 - ratio) + ratio * mat_end
+        mat_interpolated.data = np.exp(mat_interpolated.data * np.log(rate))
+        res.append(mat_interpolated)
+    return res
```
chahank, line 148: Could also be a list comprehension.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def linear_interp_arrays(arr_start, arr_end):
```
chahank: I do not really understand what this method is supposed to do. What are the dimensions? Why is `interpolation_range = arr_start.shape[0]`?

climada/trajectories/impact_calc_strat.py
spjuhel: You are right, I'll update that

climada/trajectories/impact_calc_strat.py
spjuhel: What do you do if only one is defined in ImpactCalc?

climada/trajectories/interpolation.py
chahank: And curiosity, what does `res` stand for?

climada/trajectories/interpolation.py
spjuhel: This function is used to interpolate between two arrays representing the change in a metric over `array.shape[0]` different dates, with one or more dimensions of risk changing. For instance, assuming you have two arrays representing the evolution of AAIs between *E0-to-E1 with H0V0* and *E0-to-E1 with H1V0*, it returns the evolution of AAIs between E0H0-to-E1H1 with V0. (With E being exposure, H hazard, V vulnerability, and 0 "present", 1 "future")

climada/trajectories/impact_calc_strat.py
chahank: We just apply the one that is defined. The attachment cuts from below, and the cover from above. So,  no attachment is equivalent to attachment = 0, and no cover is equal to cover = infinity.

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,143 @@
...
+class Snapshot:
+    """
+    A snapshot of exposure, hazard, and impact function at a specific date.
+
+    Attributes
+    ----------
+    date : datetime
+        Date of the snapshot.
+    measure: Measure | None
+        The possible measure applied to the snapshot.
```
chahank, line 61: Probably incomplete.

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,143 @@
...
+            The date of the Snapshot, it can be an integer representing a year,
+            a datetime object or a string representation of a datetime object
+            with format YYYY-MM-DD.
+
+        """
+
+        self._exposure = copy.deepcopy(exposure)
+        self._hazard = copy.deepcopy(hazard)
+        self._impfset = copy.deepcopy(impfset)
+        self.measure = None
```
chahank: Any reason to expose the measure and date, but not the exposure, hazard and impfset?

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,143 @@
...
+        -------
+            The Snapshot with the measure applied.
+
+        """
+
+        LOGGER.debug(f"Applying measure {measure.name} on snapshot {id(self)}")
+        exp_new, impfset_new, haz_new = measure.apply(
+            self.exposure, self.impfset, self.hazard
+        )
+        snap = Snapshot(exp_new, haz_new, impfset_new, self.date)
```
chahank:
```suggestion
        snap = Snapshot(*measure.apply(self.exposure, self.impfset, self.hazard), self.date)
```
Not 100% sure, but would the following maybe spare one copy in memory of hazard, exposure and impfset?

climada/trajectories/timeseries.py
```diff
@@ -0,0 +1,138 @@
...
+CLIMADA is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
+
+---
+
+"""
```
chahank: Not so clear whether this file should be commited with this PR.

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+
+"""
+
+import datetime
+import itertools
+import logging
+
+import matplotlib.dates as mdates
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
```
chahank:
```suggestion
import matplotlib.ticker as mticker
```
for consistency?

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+from pandas.tseries.frequencies import to_offset
+
+from climada.entity.disc_rates.base import DiscRates
+from climada.trajectories.interpolation import InterpolationStrategyBase
+from climada.trajectories.riskperiod import (
+    AllLinearStrategy,
+    CalcRiskPeriod,
+    ImpactCalcComputation,
+    ImpactComputationStrategy,
+)
```
chahank: Any reason to not import these modules directly (instead of importing them from `riskperiod`) ?

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+    ImpactCalcComputation,
+    ImpactComputationStrategy,
+)
+from climada.trajectories.snapshot import Snapshot
+from climada.util import log_level
+
+LOGGER = logging.getLogger(__name__)
+
+POSSIBLE_METRICS = ["eai", "aai", "return_periods", "risk_components", "aai_per_group"]
+DEFAULT_RP = [50, 100, 500]
```
chahank: As raised in the discussion, I think there is a conundrum with this default. It should imho be removed.

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+    _grouper = ["measure", "metric"]
+    """Results dataframe grouper"""
+
+    def __init__(
+        self,
+        snapshots_list: list[Snapshot],
+        *,
+        time_resolution: str = "Y",
+        all_groups_name: str = "All",
+        risk_disc: DiscRates | None = None,
```
chahank: No biggy, but one could refrain from inventing a new word here. `disc_rates` or similar would be quite clear.

climada/trajectories/interpolation.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class InterpolationStrategy(InterpolationStrategyBase):
+    """Interface for interpolation strategies."""
+
+    def __init__(
+        self,
+        exposure_interp: Callable,
+        hazard_interp: Callable,
+        vulnerability_interp: Callable,
```
chahank, line 410: It might be useful to know that exposure_interp must be an impact matrix interpolation, while hazard_interp and vulnerability_inter are metrics interpolations.

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+        self._reset_metrics()
+        self._risk_period_up_to_date: bool = False
+        self._snapshots = snapshots_list
+        self._all_groups_name = all_groups_name
+        self._default_rp = DEFAULT_RP
+        self.start_date = min([snapshot.date for snapshot in snapshots_list])
+        self.end_date = max([snapshot.date for snapshot in snapshots_list])
+        self._time_resolution = time_resolution
+        self._risk_disc = risk_disc
+        self._interpolation_strategy = interpolation_strategy or AllLinearStrategy()
```
chahank: Neat! I did not know this way of using or with None works.

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+    @property
+    def default_rp(self) -> list[int]:
+        """The default return period values to use when computing risk period metrics.
+
+        Notes
+        -----
+
+        Changing its value resets the corresponding metric.
+        """
+        return self._default_rp
```
chahank:
```suggestion
```

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+        self._risk_disc = value
+
+    @property
+    def risk_periods(self) -> list[CalcRiskPeriod]:
+        """The computed risk periods from the snapshots."""
+        if self._risk_periods_calculators is None or not self._risk_period_up_to_date:
+            self._risk_periods_calculators = self._calc_risk_periods(self._snapshots)
+            self._risk_period_up_to_date = True
+
+        return self._risk_periods_calculators
```
chahank: Is this supposed to return a method (i.e. calculators) or risk periods (as the name suggests)?

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+        Changing its value resets the metrics.
+        """
+        return self._risk_disc
+
+    @risk_disc.setter
+    def risk_disc(self, value, /):
+        if not isinstance(value, DiscRates):
+            raise ValueError("Risk discount needs to be a `DiscRates` object.")
+
+        self._reset_metrics()
```
chahank, line 224: That is a bit an unclear consequence of setting the discount rate. I think I understand why it is done (after setting the discount rate, the metrics are not interpretable anymore, since it is unclear whether it is discounted or not). But to me this shows that a clearer implementation is needed.

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+            -------
+            >>> list(pairwise([1, 2, 3, 4]))
+            [(1, 2), (2, 3), (3, 4)]
+            """
+            a, b = itertools.tee(container)
+            next(b, None)
+            return zip(a, b)
+
+        LOGGER.debug(f"{self.__class__.__name__}: Calc risk periods")
+        # impfset = self._merge_impfset(snapshots)
```
chahank:
```suggestion
```

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+        """
+
+        def _npv_group(group, disc):
+            start_date = group.index.get_level_values("date").min()
+            return calc_npv_cash_flows(group, start_date, disc)
+
+        df = df.set_index("date")
+        grouper = cls._grouper
+        if "group" in df.columns:
+            grouper = ["group"] + grouper
```
chahank: Why would "group" be in the metrics dataframe?

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+            )
+
+        # Construct the attribute name for storing the metric results
+        attr_name = f"_{metric_name}_metrics"
+
+        tmp = []
+        for calc_period in self.risk_periods:
+            # Call the specified method on the calc_period object
+            with log_level(level="WARNING", name_prefix="climada"):
+                tmp.append(getattr(calc_period, metric_meth)(**kwargs))
```
chahank: List comprehension?
```suggestion
        # Call the specified method on the calc_period object
        with log_level(level="WARNING", name_prefix="climada"):
            tmp = [getattr(calc_period, metric_meth)(**kwargs) for calc_period in self.risk_periods]
```

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+
+        # Notably for per_group_aai being None:
```
chahank, line 341: Why would it be None?

climada/trajectories/risk_trajectory.py
```diff
@@ -0,0 +1,865 @@
...
+
+        tmp = self._compute_metrics(
+            npv=npv,
+            metric_name="risk_components",
+            metric_meth="calc_risk_components_metric",
+            **kwargs,
+        )
+
+        # If there is more than one Snapshot, we need to update the
+        # components from previous periods for for continuity
```
chahank:
```suggestion
        # components from previous periods for continuity
```

climada/trajectories/impact_calc_strat.py
spjuhel: Fixed in 252e68e0

climada/trajectories/interpolated_trajectory.py
spjuhel: In option appraisal we have the risk discount rate (`risk_disc`) and the cost discount rate (`cost_disc`). Should I use `risk_disc_rates` and `cost_disc_rate` ?

Also, this is a legacy from victor, we could decide it does not make sense to have a different one.

climada/trajectories/timeseries.py
spjuhel: Removed

climada/trajectories/interpolated_trajectory.py
spjuhel: Fixed

climada/trajectories/risk_trajectory.py
spjuhel: Is it a forced input? What if people are not interested in RP? Is it empty by default?

climada/trajectories/risk_trajectory.py
spjuhel: Fixed (changes can be debated, but closing this convo)

climada/trajectories/risk_trajectory.py
spjuhel: I think I fixed this, following your suggestion:

The discount rate is always applied if it was provided at init. 
Changing the discount still resets the cached metrics such that they are computed with the new one.

climada/trajectories/risk_trajectory.py
spjuhel: Tidy format?

|    |   group | date   | measure    | metric   |        risk |
|---:|--------:|:-------|:-----------|:---------|------------:|
|  0 |       1 | 2018   | no_measure | aai      | 2.71783e+08 |
|  1 |       2 | 2018   | no_measure | aai      | 1.71122e+07 |
|  2 |       1 | 2019   | no_measure | aai      | 3.15041e+08 |
|  3 |       2 | 2019   | no_measure | aai      | 2.00122e+07 |
|  4 |       1 | 2020   | no_measure | aai      | 3.62029e+08 |

climada/trajectories/interpolated_trajectory.py
spjuhel: If no groups are defined in the exposure?

climada/trajectories/interpolated_trajectory.py
chahank: That would be much clearer. I have no strong opinion on whether risk and cost should be discounted separately or not.

climada/trajectories/risk_trajectory.py
chahank: I do not undersand this response exactly. 

Having arbitrary defined default RPs, which sometimes are applied, and sometimes user input RPs are applied, is confusing, as explained in the discussion to this PR. So, I would remove the default here. You can always set the default in the method that computes the values for the RPs. Then it is clear to the user. I would also choose a bit more conservative values, e.g., [20, 50, 100]. The RP500 is only rarely available in hazard datasets. 

Does this clarify the response?

climada/trajectories/impact_calc_strat.py
spjuhel: Fixed (removed utt**sh*t code)

climada/trajectories/interpolation.py
spjuhel: result / return

chahank: Ok, this is a terrible name, but also not fundamental.

climada/trajectories/snapshot.py
spjuhel: Properties have their own docstring if you are talking about that?

chahank: hmmmm ok, I see. This is quite confusing, since the properties are just wrapper for attributes which are not described here. Reading this, one does not know what is in the snapshot, it looks like there are only dates and measures.

climada/trajectories/snapshot.py
spjuhel: Fixed in b4316d70

climada/trajectories/snapshot.py
spjuhel: Fixed in b4316d70

climada/trajectories/interpolated_trajectory.py
spjuhel: I don't understand the question?

Do you mean `from riskperiod import` instead of `from climada.trajectories.riskperiod` ?

This is virtually the same (slightly more robust). These lines are generated automatically by my LSP when a new import is required.

climada/trajectories/risk_trajectory.py
spjuhel: So how it works now is that the methods do not have RPs has argument any more (simplifies a lot) and always use the .return_periods attribute which has a default value (changed to 20, 50, 100) and can be set at init (to be coded still) or via the property afterwards (and takes care of updating the metrics).

Does that work for you?

climada/trajectories/interpolated_trajectory.py
chahank: Sorry for being unclear. I meant that for instance `ImpactCalcComputation` should be imported from its module, i.e. `from climada.trajectories.impact_calc_strat` right?

climada/trajectories/risk_trajectory.py
chahank: Sounds good! Important is that you can set it in the init and via properties too.

climada/trajectories/interpolated_trajectory.py
spjuhel: Fixed in 3a026cd8

climada/trajectories/risk_trajectory.py
spjuhel: Updated in 3a026cd8

climada/trajectories/interpolated_trajectory.py
spjuhel: Updated/Fixed in:
d694c64e Refactoring continued
e2fd678e Refactoring continued (Also now includes the new files)
2bf77689 Big refactoring to include "static" (non-interpolated) trajectories

climada/trajectories/interpolation.py
spjuhel: How do you handle the change of `.data` within a list comprehension ?

climada/trajectories/interpolation.py
spjuhel: Fixed in 3f0c5f97 origin/feature/risk_trajectory Improves docstrings, typehint

climada/trajectories/interpolated_trajectory.py
spjuhel: Fixed

climada/trajectories/interpolated_trajectory.py
spjuhel: fixed 26bb85d4 origin/feature/risk_trajectory adds list comprehensions, fixes typo

climada/trajectories/risk_trajectory.py
spjuhel: 26bb85d4 origin/feature/risk_trajectory adds list comprehensions, fixes typo

climada/trajectories/impact_calc_strat.py
spjuhel: fixed in 6e12608c origin/feature/risk_trajectory improves docstring for calc strategies

climada/trajectories/interpolation.py
spjuhel: To be thought of if memory use becomes an issue. But I will keep a list comprehension at the moment.

climada/trajectories/interpolation.py
spjuhel: Improved docstring to be more explicit

climada/trajectories/snapshot.py
spjuhel: I agree, but this is the PEP standard.

climada/trajectories/interpolated_trajectory.py
spjuhel: Oh, right! Yes indeed this is not elegant.
In any case, I think I am going to make proxies within `trajectory/__init__.py` to simplify imports.

spjuhel: Fixed

spjuhel:
Can I ask when you plan to continue and finish the review?
The code should be very stable now.

climada/test/test_trajectories.py
```diff
@@ -0,0 +1,2 @@
+# 100% Coverage goal:
+## Integration test for risk period with periods and freq defined
```
chahank: Not sure to understan the purpose of this file. Are integration tests planed for later on?

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,291 @@
...
+    r"""
+    Default impact computation strategy using the core engine of climada.
+
+    This strategy first calculates the raw impact using the standard
+    :class:`ImpactCalc` logic and then applies a global risk transfer mechanism.
+
+    This risk transfer is distinct and non-exclusive with the risk transfer
+    defined through the Exposures object. While in the Exposures case, the transfer
+    is defined on a per coordinate basis, this one is applied total impact on the whole region
+    considered, and proportionally rescales the impacts per coordinate accordingly.
```
chahank:
```suggestion
    is defined on a per coordinate basis, this one is applied total impact on the whole region
    considered, and proportionally rescales the impacts per coordinate accordingly.
```
What does the `rescales the impacts per coordinate accordingly` mean ? Attachment and cover would just apply for the whole exposure right? What is being rescaled?

climada/trajectories/impact_calc_strat.py
```diff
@@ -0,0 +1,291 @@
...
+       (sum of impacts across all exposure points).
+    2. **Transferred Risk per Event**: Defined as:
+       $$\min(\max(0, \text{Total Impact} - \text{attachement}), \text{cover})$$
+    3. **Residual Risk per Event**:
+       $$\text{Total Impact} - \text{Transferred Risk per Event}$$
+    4. **Adjustment**: The original impact per exposure point is scaled
+       by the ratio of (Residual Risk / Total Impact) or
+       (Transferred Risk / Total Impact) for that event.
+       This ensures the risk transfer is shared proportionally among all
+       impacted exposure points.
```
chahank: I guess this is the scaling. Could you please explain what this corresponds too? I am not familiar with this risk transfer spreadind.

climada/test/test_trajectories.py
spjuhel: Oh yeah, forgot about these.

spjuhel: Will do ASAP.

climada/trajectories/impact_calc_strat.py
spjuhel: The original code from Victor changed the value of the impact matrix by proportionally rescaling them such that their sum corresponded to the transfered (residual) risk after this global risk transfer was applied.

chahank: I do not understand this. I think we should consider what makes sense, not necessarily what was implemented tentatively in the past. 
Can we find at least some evidence that it is a good idea to equally distribute global risk transfer money?

spjuhel: I agree with you about the little sense it makes.
But actually, maybe we just don't want that? How do you want us to consider this global transfer?
I mean, either we change the impact matrix or we don't. And rescaling the values is what seems to be the most straightforward if we do change it.

We can also not change the matrix and save the information elsewhere, but then it means we have to completely rethink the way we compute risk metrics.

I personally would go with just removing this feature (at least for now, until it is clearer).

chahank: Good idea. I would keep the code either in it but not accessible, or in a separate branch.

spjuhel:
This pull request has been separated into sub PRs: #1197 #1198 #1199 #1200 #1201

PR #1042, opened 2025-04-07
Remove write example in measures set

chahank:
Changes proposed in this PR:

- Updates the measure set tutorial and removes the write example to avoid writing in a now non-existent folder.

- Pending: change the demo file `EXP_DEMO_H5` to avoid the deprecation warning about `shapely 2.`


spjuhel (commented):
If I am being picky:
- Maybe explicitly put the path argument in the two calls?
- State, somewhere about the ENT_TEMPLATE_XLS constant?
- Why "ingested" and not "read"? 

Other than that, I agree that it is best if tutorials don't need to read/write files.

chahank:
I just removed the write example.

chahank:
@emanuel-schmid : it seems this was closed without merging. Was this on purpose? Otherwise, I would just merge it into develop since now the new documentation is also on develop.

emanuel-schmid:
@chahank : oh! I see. The branch `feature/documentation-restructuring` has been merged into develop but there was the feature/check_meas_tuto that waited to be merged into it and then I deleted `feature/documentation-restructuring`.
Sorry for that. Sometimes I delete merged branches because they are usually just forgotten. I'll make a new PR which targets develop.

chahank:
Thanks for the confirmation!

PR #1038, opened 2025-04-03
Hotfix: plot NaNs in geo_im_from_array

ValentinGebhart:
Context:
In PR #929, we introduced that in `util.plot.geo_im_from_array`, locations with NaN are plotted in gray and locations without a centroid (more precisely, outside the convex hull of the given centroids) are plotted in white. This created problems with `matplotlib.pyplot.pcolormesh`, which, for some proj inputs, created NaNs inside the given data that covered the values that should be plotted.

Changes proposed in this PR:
- The background color of the plot is now gray (NaN color), using axis.set_facecolor, and values without centroid (ie outside convex hull) are plotted in white with cmap.set_under. Like this, locations with only NaNs are plotted in gray, and locations with NaNs and a valid value are plotted with the valid value. Before, NaNs were plotted in gray using cmap.set_bad, and values without centroid were plotted in white with cmap.set_under 
- If there are NaNs in the plotted data, a legend is added to the plot saying that gray means NaN. 

Notes:
- the function `util.plot.geo_im_from_array` is used in
`util.plot.plot_from_gdf`, which is used in `hazard.plot_rp_intensity` and `impact.plot_rp_imp`, and in
`hazard.plot.HazardPlot._event_plot`, which is used in `hazard.plot_intensity` and `hazard.plot_fraction`.

This PR fixes #1028

Example:
````
import numpy as np
from climada.hazard import Hazard
from climada.util import HAZ_DEMO_H5 # CLIMADA's Python file
haz_tc_fl = Hazard.from_hdf5(HAZ_DEMO_H5) # Historic tropical cyclones in Florida from 1990 to 2004
haz_tc_fl.check() # Use always the check() method to see if the hazard has been loaded correctly

centroids_mask = np.array(
[ (i + j > 10) for j in range(50) for i in range(50)]
)
haz_tc_fl.centroids = haz_tc_fl.centroids.select(sel_cen=centroids_mask)
haz_tc_fl.intensity = haz_tc_fl.intensity[:, -2434:]

return_periods, label, column_label = haz_tc_fl.local_return_period([30, 40])

from climada.util.plot import plot_from_gdf
plot_from_gdf(return_periods, colorbar_name=label, title_subplots=column_label)
````
plot_nan

### PR Author Checklist

- [x] Read the [Contribution Guide][contrib]
- [x] Correct target branch selected (if unsure, select `develop`)
- [x] Descriptive pull request title added
- [x] Source branch up-to-date with target branch
[...]

PR #1047, opened 2025-04-15
Add option in raster plot to crop around centroids

ValentinGebhart:
Changes proposed in this PR:
- Add an optional parameter to `geo_im_from_array` such that the user can input a (relative) distance above which plot locations that are further away than this distance from any centroids are masked (ie, no values plotted). This option is then available in all function that call `geo_im_from_array`, i.e., `_event_plot`, `hazard.plot_intensity`, `hazard.plot_fraction`, `plot_from_gdf`, `hazard.plot_rp_intensity`, `impact.plot_rp_imp`
- removed bug about the NaN legend in `geo_im_from_array`

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1178,6 +1178,7 @@ class Impact:
         return_periods=(25, 50, 100, 250),
         log10_scale=True,
         axis=None,
+        mask_rel_distance=None,
         kwargs_local_exceedance_impact=None,
         **kwargs,
     ):
@@ -1242,7 +1243,12 @@ class Impact:
             )
 
         axis = u_plot.plot_from_gdf(
-            impacts_stats, title, column_labels, axis=axis, **kwargs
+            impacts_stats,
+            title,
+            column_labels,
+            axis=axis,
+            mask_rel_distance=mask_rel_distance,
+            **kwargs,
         )
         return axis, impacts_stats_vals
 
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1195,6 +1195,11 @@ class Impact:
             plot impact as log10(impact). Default: True
         smooth : bool, optional
             smooth plot to plot.RESOLUTIONxplot.RESOLUTION. Default: True
+        mask_rel_distance: float, optional
+            Relative distance (with respect to maximal map extent in longitude or latitude) to data
+            points above which plot should not display values. For instance, to only plot values
+            at the centroids, use mask_rel_distance=0.01. If None, the plot is not masked.
+            Default is None.
         kwargs_local_exceedance_impact: dict
             Dictionary of keyword arguments for the method impact.local_exceedance_impact.
         kwargs : dict, optional
--- a/climada/engine/impact.py
+++ b/climada/engine/impact.py
@@ -1178,7 +1178,7 @@ class Impact:
         return_periods=(25, 50, 100, 250),
         log10_scale=True,
         axis=None,
-        mask_rel_distance=None,
+        mask_relative_distance=None,
         kwargs_local_exceedance_impact=None,
         **kwargs,
     ):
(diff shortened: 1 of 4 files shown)


ValentinGebhart:
Default is now not to mask anything, so plots will look as before if user does not specify differently. Up for discussion :) @chahank @spjuhel

climada/engine/impact.py
```diff
@@ -1194,6 +1195,11 @@ def plot_rp_imp(
             plot impact as log10(impact). Default: True
         smooth : bool, optional
             smooth plot to plot.RESOLUTIONxplot.RESOLUTION. Default: True
+        mask_relative_distance: float, optional
+            Relative distance (with respect to maximal map extent in longitude or latitude) to data
+            points above which plot should not display values. For instance, to only plot values
+            at the centroids, use mask_relative_distance=0.01. If None, the plot is not masked.
+            Default is None.
```
emanuel-schmid, line 1202: 🧐 This parameter is really hard to describe. I've read it repeatedly and still need the example images to understand it. 
For me it would be probably easier if it was written in a more "positive" way, something like

```suggestion
        mask_relative_distance: float, optional
            Only regions are plotted that are closer to any of the data points than this distance,
            relative to overall plot size. For instance, to only plot values
            at the centroids, use mask_relative_distance=0.01. If None, the plot is not masked.
            Default is None.
```
Though I'm not sure, whether it's better like that.

Then: don't meant to be fuzzy but the parameter name is a somewhat awkward, don't you think?
Is it not rather `relative_mask_distance`? Or just `mask_distance`? Or what about `plotting_range`? (as in only plot stuff within this plotting range)

emanuel-schmid:
🙌 🎉 
The effect of the new parameter is awesome! Imho should be even the default. 

Only don't like the name and description too much.
And the changelog entry is missing.

climada/util/plot.py
```diff
@@ -477,7 +496,7 @@ def geo_im_from_array(
         )
         # handle NaNs in griddata
         color_nan = "gainsboro"
-        if np.any(np.isnan(x) for x in grid_im):
+        if np.isnan(grid_im).any():
```
emanuel-schmid, line 499: I assume, this is the fix for the bug? 
When does it occur? (How can it be reproduced?)

climada/engine/impact.py
ValentinGebhart: agreed, I will change the description and the parameter name to `mask_distance`. Thanks!

climada/util/plot.py
ValentinGebhart: Yes, the idea was that whenever grid_im has a NaN we want to include a small legend (because we plot nan values in gray). `np.any(np.isnan(x) for x in grid_im)` from before did not actually output a bool but a generator, such that this if condition was always True

emanuel-schmid:
> I would leave the default for now so people are not too confused but happy to adapt that in the future.

As I said, not for me to judge. But I find the default plots atm are actually more confusing than a sudden change.

climada/util/plot.py
emanuel-schmid: @ValentinGebhart Thanks! You don't happen to have an example that you could point me to?

ValentinGebhart: Sure. This would be an example where we want the legend (see PR #1038):
```python
import numpy as np
from climada.util import HAZ_DEMO_H5 # CLIMADA's Python file
haz_tc_fl = Hazard.from_hdf5(HAZ_DEMO_H5) # Historic tropical cyclones in Florida from 1990 to 2004
haz_tc_fl.check() # Use always the check() method to see if the hazard has been loaded correctly

centroids_mask = np.array(
[ (i + j > 10) for j in range(50) for i in range(50)]
)
haz_tc_fl.centroids = haz_tc_fl.centroids.select(sel_cen=centroids_mask)
haz_tc_fl.intensity = haz_tc_fl.intensity[:, -2434:]

return_periods, label, column_label = haz_tc_fl.local_return_period([30, 40])

from climada.util.plot import plot_from_gdf
plot_from_gdf(return_periods, colorbar_name=label, title_subplots=column_label)
```

and here we do not want the legend but in the current develop version it would be printed as well. 
```python
haz_tc_fl.plot_intensity(event=0)
```

spjuhel:
I agree with Emanuel the plots are much clearer. (But it is not a strong opinion)

peanutfun:
I also agree that adding a default `mask_distance` > 0 makes the plots clearer and we can add it immediately!

ValentinGebhart:
Thanks for the feedback! I changed the default to `mask_distance=0.01`. We can always change that back if we don't like it. 

@emanuel-schmid from my side this would be ready to merge. Thanks!

PR #1048, opened 2025-04-17
Add classes for calibrating average ensemble and ensemble of tragedies

peanutfun:
Changes proposed in this PR:
-
-

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+import numpy as np
+import pandas as pd
+from numpy.random import default_rng
+from pathos.multiprocessing import ProcessPool
+from tqdm import tqdm
+
+from ...engine.unsequa.input_var import InputVar
+from ...entity.impact_funcs import ImpactFunc, ImpactFuncSet
+from ..coordinates import country_to_iso
+from .base import Input, Optimizer, Output
```
emanuel-schmid: PEP 8: "Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages)"

Imo, this is probably true for `..`-modules. For siblings relative imports are OK.

```suggestion
from climada.engine.unsequa.input_var import InputVar
from climada.entity.impact_funcs import ImpactFunc, ImpactFuncSet
from climada.util.coordinates import country_to_iso
from .base import Input, Optimizer, Output
```

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+        target=out.target,
+        event_info=event_info_from_input(inp),
+    )
+
+
+@dataclass
+class EnsembleOptimizerOutput:
+    """The collective output of an ensemble optimization"""
+
+    data: pd.DataFrame
```
emanuel-schmid, line 190: This is stretching the concept of a `dataclass` as I see it a lot. For instance, you can't even do this:

```
EnsembleOptimizerOutput(df1) == EnsembleOptimizerOutput(df2)
```

Is there a reason why it has to be a `dataclass`? Woldn't it be more natural to have an ordinary class that uses `from_outputs` as `__init__` ?

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+        )
+        # Cosmetics
+        ax.set_xlabel(f"Intensity [{impf.intensity_unit}]")
+        ax.set_ylabel("Impact")
+        ax.yaxis.set_major_formatter(mticker.PercentFormatter(xmax=1))
+        ax.set_ylim(0, 1)
+        return ax
+
+
+@dataclass
```
emanuel-schmid, line 493: `Input` has no `__eq__`, hence this class has neither. Here too, I'm not sure why it needs to be a dataclass. 
Is it about the implicit immutability of a dataclass object? If so, we maybe even should make it explicit by

```suggestion
@dataclass(frozen=True)
```

Although this would require to drop the `__post_init__` methods and `InitVar`s and replace them by something like

```
@classmethod
def from_initvars(cls, x, y):
    return cls(something(x,y))
```

Even so, it would be nice to allow comparing two such objects meaningfully - and if it means that we have to implement `__eq__` for a `dataclass`. Otherwise I honestly don't see a point of calling it `dataclass`.

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+    def input_from_sample(self, sample: list[tuple[int, int]]):
+        """Shallow-copy the input and update the data"""
+        input = copy(self.input)  # NOTE: Shallow copy!
+        input.data = sample_data(input.data, sample)
+        return input
+
+
+@dataclass
+class TragedyEnsembleOptimizer(EnsembleOptimizer):
+    """An optimizer for the ensemble of tragedies
```
emanuel-schmid: A very short summary that doesn't give much of a clue what this implementation of the `EnsembleOpitmizer` does.

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+                raise ValueError(
+                    "Ensemble size must be smaller than maximum number of samples "
+                    f"(here: {len(self.samples)})"
+                )
+
+            rng = default_rng(random_state)
+            self.samples = rng.choice(self.samples, ensemble_size, replace=False)
+
+    def input_from_sample(self, sample: list[tuple[int, int]]):
+        """Subselect all input"""
```
emanuel-schmid, line 709: 🤔

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+        """Copy settings in the init_kwargs and update for each iteration"""
+        kwargs = copy(init_kwargs)  # Maybe deepcopy?
+        if "random_state" in kwargs:
+            kwargs["random_state"] = kwargs["random_state"] + iteration
+        return kwargs
+
+
+@dataclass
+class AverageEnsembleOptimizer(EnsembleOptimizer):
+    """An optimizer for the average ensemble
```
emanuel-schmid: 🤔 Hard to understand. I wish the description was more elaborate. s.b.

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+        ax.set_xlabel(f"Intensity [{impf.intensity_unit}]")
+        ax.set_ylabel("Impact")
+        ax.yaxis.set_major_formatter(mticker.PercentFormatter(xmax=1))
+        ax.set_ylim(0, 1)
+        return ax
+
+
+@dataclass
+class EnsembleOptimizer(ABC):
+    """Abstract base class for defining an ensemble optimizer
```
emanuel-schmid: What is an ensemble optimizer?

emanuel-schmid (changes requested):
Awesome amendment! Very well written module I'd say.
I just wish the pydoc strings were more elaborate and I wonder about the use of dataclasses.

climada/util/calibrate/ensemble.py
peanutfun: I am using `dataclass` to avoid boilerplate code. In my view, the `__init__` of the class should be as permissive as possible. Therefore, a usual init will look like this:
```python
class MyClass:

    def __init__(attr1, attr2):
        self.attr1 = attr1
        self.attr2 = attr2
```
I think the code is trivial and using `dataclass` to construct is seems reasonable to me. It improves readability, maintainability, and extensibility.

> Woldn't it be more natural to have an ordinary class that uses `from_outputs` as `__init__ `?

I think `from_outputs` is too restrictive for making it the default initialization (although I use it exclusively throughout my code). See my comment above.

> You can't even do this:

A good point! I will implement a custom `__eq__`. But I don't think that's a reason to abandon `dataclass` in this case.

climada/util/calibrate/ensemble.py
peanutfun: Again, I use `dataclass` mostly to avoid a trivial `init`. This is especially useful for derived classes, where you might easily miss that you also need to initialize the base class.

I did not check it, but a default `__eq__` should be created by `dataclass`. I did not really consider freezing the instances, as I think it's not necessary. `Input` probably is the only sensible candidate for freezing, as results are distorted if one changes it.

climada/util/calibrate/ensemble.py
peanutfun: Will clarify in the docs. An optimizer yields a single set of optimal parameters (or impact functions), whereas an ensemble optimizer yields an ensemble of them.

climada/util/calibrate/ensemble.py
peanutfun: I agree that this does not become clear in the docs. The idea is the following: Each `EnsembleOptimizer` only stores a list of `samples`, which contains the indices of the data points that are to be used in each optimization. The concrete optimizer then defines in `input_from_sample`, how a new `Input` object is generated for each sample and thus each optimization. In `TragedyEnsembleOptimizer`, this is used to drastically reduce the amount of hazard data that needs to be processed by each optimization. I'll try to clarify

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,652 @@
...
+    Attributes
+    ----------
+    sample_fraction : float
+        The fraction of data points to use for each calibration
+    ensemble_size : int
+        The number of calibration tasks to perform
+    random_state : int
+        The seed for the random number generator selecting the samples
+    replace : bool
+        If samples of the impact data should be drawn with replacement
```
peanutfun: TODO: Given the structure of `Input.data`, sampling with replacement actually has no effect at the moment!

peanutfun:
@emanuel-schmid I hope the docs are clearer now!

I also found a bug in the implementation where the `replace` parameter in `AverageEnsembleOptimizer` would not have any effect because the data structure would only allow to add a sample once. I fixed this by implementing an additional data structure `data_weights` to `Input`, along with some tests. Effectively, drawing a sample more than once in `AverageEnsembleOptimizer` will multiply the weights of that sample. Other than that, users are now free to weight their data as they see fit (although we did not consider that necessary before)

climada/util/calibrate/ensemble.py
peanutfun: I implemented a custom `__eq__` for `EnsembleOptimizerOutput`

climada/util/calibrate/ensemble.py
peanutfun: Should be better now

climada/util/calibrate/ensemble.py
peanutfun: I hope I clarified

climada/util/calibrate/ensemble.py
peanutfun: I hope I clarified

climada/util/calibrate/ensemble.py
peanutfun: Done!

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -616,11 +616,13 @@ def __post_init__(self, random_state, allow_duplicate_points, bayes_opt_kwds):
...
-    def _target_func(self, data: pd.DataFrame, predicted: pd.DataFrame) -> Number:
+    def _target_func(
+        self, data: np.ndarray, predicted: np.ndarray, weights: np.ndarray | None
+    ) -> Number:
         """Invert the cost function because BayesianOptimization maximizes the target"""
-        return -self.input.cost_func(data, predicted)
+        return -self.input.cost_func(data, predicted, weights)
 
-    def run(self, controller: BayesianOptimizerController) -> BayesianOptimizerOutput:
+    def run(self, **opt_kwargs) -> BayesianOptimizerOutput:
```
emanuel-schmid, line 625: Why is the signature not just like the pydoc describes it?

climada/util/calibrate/bayesian_optimizer.py
```diff
@@ -631,15 +633,25 @@ def run(self, controller: BayesianOptimizerController) -> BayesianOptimizerOutpu
...
             on the sampled parameter space.
         """
+        # Take the controller
+        try:
+            controller = opt_kwargs.pop("controller")
+        except KeyError as err:
+            raise RuntimeError(
+                "BayesianOptimizer.run requires 'controller' as keyword argument"
+            ) from err
+
```
emanuel-schmid, line 654: all of this wouldn't be necessary with a signature as described in the pydoc

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,729 @@
...
+
+    return data_sampled
+
+
+def sample_weights(weights: pd.DataFrame, sample: list[tuple[int, int]]):
+    """
+    Return an updated DataFrame containing the appropriate weights for a sample.
+
+    Weights that are not in ``sample`` are set to zero, whereas weights that are sampled
+    multiple times will effectively multiplied by their occurrence in ``sample``.
```
emanuel-schmid:
```suggestion
    multiple times are effectively multiplied by their occurrence in ``sample``.
```

climada/util/calibrate/ensemble.py
emanuel-schmid: not really convinced given the need to implement `__eq__` and `__post__` 
\- but I guess it doesn't matter much. :sunglasses:

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,729 @@
...
+from climada.engine.unsequa.input_var import InputVar
+from climada.entity.impact_funcs import ImpactFunc, ImpactFuncSet
+from climada.util.coordinates import country_to_iso
+
+from .base import Input, Optimizer, Output
+
+LOGGER = logging.getLogger(__name__)
+
+
+def sample_data(data: pd.DataFrame, sample: list[tuple[int, int]]):
```
emanuel-schmid, line 46: function _could_ be private if we wanted (with regard to possible Extensions better not though)

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,729 @@
...
+    data_sampled = pd.DataFrame(np.nan, columns=data.columns, index=data.index)
+
+    # Extract sample values from data
+    for row, col in sample:
+        data_sampled.iloc[row, col] = data.iloc[row, col]
+
+    return data_sampled
+
+
+def sample_weights(weights: pd.DataFrame, sample: list[tuple[int, int]]):
```
emanuel-schmid, line 78: dito

climada/util/calibrate/ensemble.py
```diff
@@ -0,0 +1,729 @@
...
+    weights_sampled = pd.DataFrame(0.0, columns=weights.columns, index=weights.index)
+
+    # Add weights for each sample
+    for row, col in sample:
+        weights_sampled.iloc[row, col] += weights.iloc[row, col]
+
+    return weights_sampled
+
+
+def event_info_from_input(inp: Input) -> dict[str, Any]:
```
emanuel-schmid, line 108: dito

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Same argument as we had in the discussion about the unsequa abstract base class:

The `Optimizer` abstract base class defines an interface that should be kept for all derived classes. However, the derived classes can, or in this case *must*, take additional, class specific arguments. The only elegant (?) way I see for doing this is simply going with a variadic `kwargs` argument and then documenting the method like it had the actual arguments that are needed here.

tldr: The linter would complain: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/arguments-differ.html

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Yes, but see above.

climada/util/calibrate/bayesian_optimizer.py
peanutfun: Note that we do the same thing in the ScipyOptimizer: https://github.com/CLIMADA-project/climada_python/blob/93aa8b284f80cd566ec9ca53bc78d4ff2fa1eb59/climada/util/calibrate/scipy_optimizer.py#L76

emanuel-schmid: :facepalm: aah, right! :+1:

PR #1051, opened 2025-04-30
Feature/exposures hdf5 io

emanuel-schmid:
Changes proposed in this PR:

- By default convert geometry data to wkb format before writing Exposures to HDF5 files.
- Add a flag to `Exposures.write_hdf5` that allows to keep the previous behavior.
- Adapt the `from_hdf5` methods accordingly.

The data is still be pickled but pickled byte arrays are unlikely to be subject to change.
This is not a performance boost. Quite likely there are faster ways to store geometry data as wkb arrays,
but they are not obvious and maybe hard to get right. In light of future plans to introduce the parquet file format as standard in climada, any more efforts to optimize have been omitted.

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -1145,7 +1145,7 @@ class Exposures:
                     pandas_df[col] = np.asarray(self.data[col])
                 else:
                     wkb_data[col] = to_wkb_store(self.geometry)
-                    pandas_df.drop(columns=["geometry"])
+                    pandas_df.drop(columns=[col], inplace=True)
 
         # Avoid pandas PerformanceWarning when writing HDF5 data
         with warnings.catch_warnings():
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -29,6 +29,7 @@ from pathlib import Path
 
 import cartopy.crs as ccrs
 import contextily as ctx
+import geopandas as gpd
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
@@ -37,7 +38,6 @@ from deprecation import deprecated
 from geopandas import GeoDataFrame, GeoSeries, points_from_xy
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from rasterio.warp import Resampling
-from xarray import DataArray
 
 import climada.util.coordinates as u_coord
 import climada.util.hdf5_handler as u_hdf5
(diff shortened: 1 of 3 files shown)


climada/entity/exposures/base.py
```diff
@@ -1121,20 +1122,31 @@ def plot_basemap(
...
         store = pd.HDFStore(file_name, mode="w")
         pandas_df = pd.DataFrame(self.gdf)
+        wkb_columns = []
         for col in pandas_df.columns:
             if str(pandas_df[col].dtype) == "geometry":
-                pandas_df[col] = np.asarray(self.gdf[col])
+                if pickle_geometry_as_shapely:
+                    pandas_df[col] = np.asarray(self.gdf[col])
+                else:
+                    pandas_df[col] = gpd.GeoSeries.to_wkb(pandas_df[col])
```
peanutfun, line 1148: `GeoSeries.to_wkb` calls `shapely.to_wkb` on `self`.  Are you sure this does what it's supposed to do?
```suggestion
                    pandas_df[col] = gpd.GeoSeries(pandas_df[col]).to_wkb()
```
It might be more elegant to use [`GeoDataFrame.to_wkb`](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_wkb.html) on `self`, which already returns a `pandas.DataFrame`.

peanutfun (changes requested):
Thanks for the update! I am not sure that the current version actually converts the data to WKB, see my comment. Is there a particular reason why we need to keep the old behavior of `write_hdf5` at all? We could simply switch to the new one, and avoid adding a new parameter (which will need to be deprecated at some point etc...)

climada/entity/exposures/base.py
emanuel-schmid: Thanks for the review!

>  Is there a particular reason why we need to keep the old behavior of write_hdf5 at all?

Two reasons:
- I'm afraid the new way is slower, so the old way would be preferable if files don't have to be kept forever.
- For testing readability of old style files.

emanuel-schmid:
Thanks for the review!

>  Is there a particular reason why we need to keep the old behavior of write_hdf5 at all?

Two reasons:

- I'm afraid the new way is slower, so the old way would be preferable if files don't have to be kept forever.
- For testing readability of old style files.

peanutfun:
> I'm afraid the new way is slower, so the old way would be preferable if files don't have to be kept forever.

If it's only a little, I would still be in favor of dropping it. If it's considerable, we need to tell that in the docs and probably change the default value?

> For testing readability of old style files.

I only want to drop the support of *writing* old-style files. Reading should "silently" read both types.

climada/entity/exposures/base.py
peanutfun: Because you are calling a method of an uninstantiated object. Maybe use `shapely.to_wkb(pandas_df[col])` directly?

emanuel-schmid:
> If it's only a little, ..

I've checked for Litpop from Russia: wkb is 5 (reading) to 10 (writing) times faster than shapely pickles.
So I agree, probably it's better to drop it.

For the test it's a pity though. Proof that the result is still the same. depends on a file that has been written with the deprecated method.
I'm tempted to skip that regression test as well and just trust the equality of the current implementation. 🤔

climada/entity/exposures/base.py
emanuel-schmid: I'll keep it as it is (`pandas_df[col] = gpd.GeoSeries(pandas_df[col]).to_wkb()`) so geopandas can take care of it.

climada/entity/exposures/base.py
```diff
@@ -1132,9 +1133,11 @@ def write_hdf5(self, file_name):
         LOGGER.info("Writing %s", file_name)
         store = pd.HDFStore(file_name, mode="w")
         pandas_df = pd.DataFrame(self.gdf)
+        wkb_columns = []
         for col in pandas_df.columns:
             if str(pandas_df[col].dtype) == "geometry":
-                pandas_df[col] = np.asarray(self.gdf[col])
+                pandas_df[col] = gpd.GeoSeries(pandas_df[col]).to_wkb()
+                wkb_columns.append(col)
```
peanutfun, line 1140: We can simplify this with `GeoDataFrame.to_wkb()`:
```python
wkb_columns = self.gdf.columns[self.gdf.dtypes == "geometry"]
pandas_df = self.gdf.to_wkb()  # Done!
```
See https://github.com/geopandas/geopandas/blob/7255c57371eafbff5a09f881db2b905d1f9b28a7/geopandas/geodataframe.py#L1254
(I cannot make this a suggestion on GitHub because there is a deleted line 🤷 )

emanuel-schmid: Hm. I thought I have checked that and found that `gdf.to_wkb()` only converts the 'geometry' column itself, leaving all other columns of type "geometry" untouched. Wrong?

peanutfun: Code and docs say it converts all geometry-type columns. The code I linked above is equivalent to our original one except for the explicit `str` cast on the returned `dtype`, which apparently is not necessary.

PR #1052, opened 2025-05-05
remove geopandas.datasets

emanuel-schmid:
Changes proposed in this PR:

- `gopandas.datasets` is deprecated and has been removed from geopandas >= 1.0
- in climada, the module is used exactly once, in test_tc_tacks, to create an exposures object from 110m admin 0 cultural vectors
- The nearly very same data can be downloaded via cartopy.io.shapereader
- In particular, for Cuba, which is the only relevant country in this test, the data is exactly equal.

This PR renders upgrading geopandas to 1.0 possible.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/test/test_tc_tracks.py
+++ b/climada/hazard/test/test_tc_tracks.py
@@ -27,7 +27,8 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 import xarray as xr
-from shapely.geometry import LineString, MultiLineString, Point
+from cartopy.io import shapereader
+from shapely.geometry import LineString, MultiLineString
 
 import climada.hazard.tc_tracks as tc
 import climada.util.coordinates as u_coord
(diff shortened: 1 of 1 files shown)


peanutfun (changes requested):
Good call, upgrading to GeoPandas v1 is very desirable! 👏 

But why use the shape reader at all if the only use for it is to test track filtering? We do not need a sensible exposure for that. We could simply define an exposure consisting of a few custom lat/lon points located in Cuba and then run the test, right?

emanuel-schmid:
Thanks for the review! 😄 

> But why use the shape reader at all if the only use for it is to test track filtering? We do not need a sensible exposure for that. We could simply define an exposure consisting of a few custom lat/lon points located in Cuba and then run the test, right?

True. On the other hand, the updated test uses quite exactly the same data as before and may serve as a template for similar cases.

PR #1054, opened 2025-05-13
Make use of ABC abstract class in unsequa module

luseverin:
Changes proposed in this PR:
-Make use of ABC abstract class in the unsequa module

I updated the `unsequa` module so that it uses the [abstract (ABC) class](https://docs.python.org/3/library/abc.html) concept. The idea is to be able to readily adapt the unsequa module to other use cases, using the related abstract class. In my case I wanted to have unsequa work with the [network module](https://github.com/CLIMADA-project/climada_petals/tree/feature/nw_minimal/climada_petals) from petals. I splitted this task in two where 1) the unsequa module is made abstract (this PR); 2) the adaptation of the unsequa module to the network module use case is implemented (in branch [feature/unsequa_cascade_abstract](https://github.com/CLIMADA-project/climada_python/tree/feature/unsequa_cascade_abstract), work in progress). 

As @emanuel-schmid and @chahank nicely designed the unsequa module so that it is already quite abstract in the first place, this did not require too much work. However, I was not sure of whether the `__init__`  of the `UncOutput` class should be made abstract or not, as making it abstract is preventing `__init__` to be used by subclasses using `super()`..

### PR Author Checklist

### PR Reviewer Checklist


peanutfun (commented):
What does this PR achieve that was not possible before? The new abstract methods do not serve any purpose if they do not define an interface (through parameters). But defining an interface also means that derived methods must use the same parameters. The current structure of the unsequa module already violates this, and Jenkins complains, rightfully so. See https://pylint.pycqa.org/en/latest/user_guide/messages/warning/arguments-differ.html 

So, if you want derived classes to change the interface, you don't need an interface in the first place. On the other hand, we could make an effort here to define a proper interface for the unsequa base classes that conforms to all derived use cases.

luseverin:
@peanutfun Thanks for the feedback! I think the idea here was not to add a functionality, but to make use of abstract classes as it considers as "best practice"? But yes in theory you can implement other subclasses in the unsequa module without the use of an abstract class, as it has been done before for e.g. `calc_delta_climate` or `calc_cost_benefit`. But now I understand how this might be problematic as the derived classes indeed require additional parameters.. Would it be beneficial in your opinion to make this effort to define a proper interface for the unsequa base class that will be conform with the derived classes? Not sure of how much additional work this would require..

climada/engine/unsequa/unc_output.py
```diff
@@ -107,16 +108,20 @@ class UncOutput:
         "sensitivity_kwargs",
     ]
 
+    # @abstractmethod
```
emanuel-schmid, line 111: making `__init__` abstract looks beyond the target to me. Is there a use case?

climada/engine/unsequa/calc_base.py
```diff
@@ -423,6 +425,16 @@ def sensitivity(
 
         return sens_output
 
+    @abstractmethod
+    def uncertainty(self):
+        """Compute the uncertainty of the output values"""
+        pass
+
+    @abstractmethod
```
emanuel-schmid, line 433: private _and_ abstract seems somehow contradicting. if we go for it, `compute_metrics` ought to be public.

emanuel-schmid:
I kind of agree with Lukas. Up to this point the abstraction is not justified yet (and not yet according to all rules).
However - I can see (or hunch) its potential. (; 
Only, in this case, I'd rather combine the implementation of the abstraction and its use in one PR.

chahank:
@emanuel-schmid  and @peanutfun I am not sure what you want to happen here. I think adding the abstract structure would make it much simpler for future developers to add functionalities, in the same way as @luseverin is currently doing.

chahank:
@luseverin : small point, but it seems that some modification of the tutorial happened that might not have been intended (2,182 additions, 478 deletions)

peanutfun:
> I think adding the abstract structure would make it much simpler for future developers to add functionalities, [...]

@chahank Please refer to my earlier comment: https://github.com/CLIMADA-project/climada_python/pull/1054#pullrequestreview-2840007838. The only way to make things simpler is to define a proper interface in the base class(es). But this is neither done here nor was it done originally.

chahank:
Thanks, but this does not answer the question: what do you want to happen here? I think the current comments do not allow @luseverin to either improve, or close this PR.

climada/engine/unsequa/calc_base.py
emanuel-schmid: @peanutfun Technically speaking, you're absolutely right. In this particular case I still think `compute_metrics` ought to be public.

emanuel-schmid:
Sorry for not being clear. In order to continue with this PR, I would want

- a proper definition of the interface
- a use case where this interface is used / called, which justifies the abstraction

peanutfun:
After personal discussion with @luseverin:

The only option that does not violate the substitution principle is to define an abstract method that takes only `**kwargs` as arguments:
```python
@abstractclass
class Calc:

    @abstractmethod
    def uncertainty(**kwargs):
        pass
```
This would tell developers that they have to implement the function in a derived class. However, it would make the function interface for users much more obscure. I think we should cater more towards users than developers. We can simply tell developers in the documentation what function to implement in a derived class.

An abstract base class `UncOutput` does not make any sense, because the derived classes only define new class attributes. Abstract methods, however, are intended to encapsulate functionality (only). Note that `__init__` methods should never be made abstract. Instead, derived classes should always define their own `__init__` and call the init of the base class (among potentially other things).

PR #1065, opened 2025-07-01
Remove intensity_thres attribute from Hazard class

### LINKED ISSUE #1007: Unclear attribute of Hazard class
The hazard class has currently an attribute called `self.intensity_thres` that is silently set to the value 10. It is very unclear what the use of this attribute is, and even more unclear how the value of 10 was chosen. Moreover, it appears that the value cannot be set in the init. I propose to remove it entirely.

ValentinGebhart:
Changes proposed in this PR:
- Removed hard-coded `intensity_thres` attribute from Hazard class (including minor changes in some local exceedance functions that made use of this attribute).
- The attribute still exists in the classes `StormEurope` and `TropCyclone`, as it is used in data storing there. The handling of the attribute was different in both classes, which I have adapted such that both now use the handling as in`TropCyclone`.

This PR fixes #1007.

### PR Author Checklist

### PR Reviewer Checklist


chahank:
ok, then we should probably leave it as is in the TC and winterstorm models. Maybe we can add a note in the docstring somewhere to make sure people know about it?

climada/hazard/base.py
```diff
@@ -485,7 +481,7 @@ def local_exceedance_intensity(
         self,
         return_periods=(25, 50, 100, 250),
         method="interpolate",
-        min_intensity=None,
+        min_intensity=0,
```
chahank, line 484: Does this mean that all sparse data is projected onto a dense matrix? This might lead to problems no?

ValentinGebhart: The calculation of local exceedance intensity is centroid wise, so yes, the sparse matrix is converted to a dense matrix column by column. The min_intensity parameter being zero or positive does not change this. 

It might be that a positive min_intensity parameter here would result in larger geodataframes (ie. centroids with only very small intensities would still lead to a computation and nonzero values, instead of faster handling if all intensities are considered zero). However, as we do not have a minimum intensity for a general hazard object, I think this should be set to zero per default, and can be adapted by the user if wanted.

Does this answer your question?

chahank: Yes, thanks!

chahank:
> Sounds good. We could add some info to the docstring of the two classes (StormEurope and TropCyclone)?

This sounds good.

> Or do you mean the (class) methods that make use of these attributes? These should already have some info in the docstrings.

climada/hazard/storm_europe.py
```diff
@@ -79,10 +85,15 @@ class StormEurope(Hazard):
...
+    Class Attributes
+    ----------------
+    intensity_thres : float
+        Threshold below which wind speeds (in m/s) are stored as 0. Defaults to 14.6.
     """
 
-    intensity_thres = 14.7
-    """Intensity threshold for storage in m/s; same as used by WISC SSI calculations."""
+    intensity_thres = DEF_INTENSITY_THRES
+    """Intensity threshold for storage in m/s."""
```
chahank: Sorry for the late extra comment, but is the class attribute then still needed? It looks like the intensity_thresh is now a parameter of all relevant methods no?

ValentinGebhart: Yes, `intensity_thresh` is an optional parameter of all relevant methods. This class attribute is used as the default values for these methods. If you change the attribute for a specific instance, the default value of all corresponding methods changes, and you do not have to change it manually each time.

Does this make sense? If there is a smarter implementation, happy to change this.

chahank: It makes sense, although I am not sure the current implementation is doing what you want. 

The code now has an class attribute, but actually never uses it, since the value of this attribute is always passed as parameter to all relevant methods. Thus, the class attribute seems redundant to me. What do you think?

ValentinGebhart: Yes, almost agree. There is one place in `StormEurope` and one in `TropCyclone` where the correpsonding class attributes are used (called on class instances though):
In `StormEurope`: in the function `calc_ssi` 
In `TropCyclone`: in the function `set_from_tracks` which has a deprecation warning. 

It looks to me as if we can remove these uses and then also remove the class attributes. I have to think about the `calc_ssi` though. If you confirm that it is better to remove them, I can try to implement that :)

chahank: How about we add the parameter to the methods `calc_ssi`, remove the method `set_from_tracks`, adjust the method `from_tracks` (remove the `haz.intensity_thres = intensity_thres`), and remove the class parameter?

ValentinGebhart: I removed the class attribute from `StormEurope` and `TropCyclone`, and the deprecated `TropCyclone.set_from_tracks`. In the `StormEurope.calc_ssi` method, there was already an optional parameter `threshold` for the calculation of SSI. Before, the method checked if this parameter was larger than the class attribute `intensity_thres` and, if not, gave an error "threshold cannot be below threshold upon read_footprint". I don't understand why it should be a problem though if the calculation threshold is below the read-in threshold...

Let me know what you think, for me this looks good.

ValentinGebhart:
I have tried to check if there there is any problem with writing or reading "old" hazard files after removing the `intensity_thres` class attribute. I realized that, in the old version,
- the `write_hdf5` method only saves instance attributes, not class attributes. Thus, the `intensity_thres` was only stored if it was changed manually. The default class attribute was not stored.
- the `from_hdf5` method did not read any `intensity_thres` attribute, even if it was written to the file. This is because the method first creates an empty instance of the corresponding class (e.g. StormEurope) and then reads all attributes of the instance (but NOT class attributes).

This is code to see this in action, use e.g., main or develop branches. The first print command resulted in 18., the second in 14.7 (the default class attribute).
```python
from climada.hazard import StormEurope
from climada.util.constants import WS_DEMO_NC

storm_instance = StormEurope.from_footprints(WS_DEMO_NC)
storm_instance.intensity_thres = 18.  # change intensity threshold
print(storm_instance.intensity_thres)

storm_instance.write_hdf5('storm_with_attribute.h5')
storm_instance = StormEurope.from_hdf5('storm_with_attribute.h5')
print(storm_instance.intensity_thres)
```

Of course, in the new version, there is simply no `intensity_thres` attribute anymore. I assume that it should not create any problems, as 1) it was usually never saved, and 2) even if it was written before to the file, it was usually never read.

@chahank Would this be enough check and can we merge?

climada/hazard/storm_europe.py
```diff
@@ -79,10 +85,12 @@ class StormEurope(Hazard):
...
     ssi : np.array, float
         SSI as set by set_ssi; uses the Dawkins definition by default.
-    """
 
-    intensity_thres = 14.7
-    """Intensity threshold for storage in m/s; same as used by WISC SSI calculations."""
+    Class Attributes
+    ----------------
+    intensity_thres : float
+        Threshold below which wind speeds (in m/s) are stored as 0. Defaults to 14.6.
```
chahank:
```suggestion
```
Correct?

climada/hazard/trop_cyclone/trop_cyclone.py
```diff
@@ -88,10 +88,12 @@ class TropCyclone(Hazard):
...
         matrix of shape (npositions, ncentroids * 2) that can be reshaped to a full ndarray of
         shape (npositions, ncentroids, 2).
-    """
 
-    intensity_thres = DEF_INTENSITY_THRES
-    """intensity threshold for storage in m/s"""
+    Class Attributes
+    ----------------
+    intensity_thres : float
+        Threshold below which wind speeds (in m/s) are stored as 0. Defaults to 17.5.
```
chahank:
```suggestion
```
Correct?

chahank:
> Also, 3 tests are failing, one for the supply chain module and one for the unsequa module. Not sure if this is related atm.

The supply chain one looks unrelated; there is no hazard involved as far as I can see. For unsequa it could be due to the reading of the tropical cyclone datasets `tc_fl_1990_2004.h5`.

chahank:
> Thanks for checking the docstrings, this is now up to date. For the test, I don't know, for me locally it runs through.

Ok, let's wait for the last round of tests. Do also the integration tests work for you locally?

ValentinGebhart:
To me the failing test don't seem related to the changes, right? And when running the integration tests locally I get a few errors but they seem all to be related to LitPop.

chahank:
I do not see why the failures would be related to litpop, as the build seem to fail before the tests even start. But I agree that this looks not related. I would thus merge.

ValentinGebhart:
Ok great, thanks! Yes, I was saying that when running the integration tests locally there is some issue with Litpop data, but maybe that is some problem to my local setup etc

PR #1066, opened 2025-07-01
Hotfix: assign centroids in cost benefit impact calculation if in new exposure no centroids were assigned

### LINKED ISSUE #1056: CostBenefit measures with new exposure need assigned centroids
**Describe the bug**
In the current Cost Benefit module, when you implement a measure including a new exposure with
```python
measure.exposures_set = new_expsoure
```
with a new exposure object `new_expsoure` that does not have assigned hazard centroids, the cost benefit calculation (that includes the above measure in the entity object) like 
```python
cb = CostBenefit()
cb.calc(hazard, entity)
```
results in a quite cryptic error saying `ValueError: 'assign_centroids' is set to 'False' but no centroids are assigned for the given hazard type (AD)`. The `calc` method even has a `assign_centroid` which however does not change the above error and is thus confusing. Also `entity.exposures` having assigned centroids does not help avoiding the error,

**Suggestions**
1. easy but not efficient: Change one line in `CostBenefit._calc_impact_measures` to 
```python
imp_tmp, risk_transf = measure.calc_impact(
                exposures, imp_fun_set, hazard, assign_centroids=True
            )
```
instead of 
```python
imp_tmp, risk_transf = measure.calc_impact(
                exposures, imp_fun_set, hazard, assign_centroids=False
            )
```
which redoes the assignment but of of course is suboptimal if already assigned.
2. Check if the new exposure object has assigned centroids
3. Do nothing because soon there will be the new Cost Benefit module that solves this problem as well.

ValentinGebhart:
Changes proposed in this PR:
- If a Measure included a new exposure object with no hazard centroids assigned, this led to an error message that was not clear how to resolve, see #1056. Now the method checks if there are assigned centroids and assigns them if not.

This PR fixes #1056

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/entity/measures/base.py
+++ b/climada/entity/measures/base.py
@@ -179,7 +179,7 @@ class Measure:
         u_check.size(2, self.mdd_impact, "Measure.mdd_impact")
         u_check.size(2, self.paa_impact, "Measure.paa_impact")
 
-    def calc_impact(self, exposures, imp_fun_set, hazard, assign_centroids=True):
+    def calc_impact(self, exposures, imp_fun_set, hazard):
         """
         Apply measure and compute impact and risk transfer of measure
         implemented over inputs.
@@ -192,12 +192,6 @@ class Measure:
             impact function set instance
         hazard : climada.hazard.Hazard
             hazard instance
-        assign_centroids : bool, optional
-            indicates whether centroids are assigned to the self.exposures object.
-            Centroids assignment is an expensive operation; set this to ``False`` to save
-            computation time if the hazards' centroids are already assigned to the exposures
-            object.
-            Default: True
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


climada/entity/measures/base.py
```diff
@@ -206,7 +200,15 @@ def calc_impact(self, exposures, imp_fun_set, hazard, assign_centroids=True):
         """
 
         new_exp, new_impfs, new_haz = self.apply(exposures, imp_fun_set, hazard)
-        return self._calc_impact(new_exp, new_impfs, new_haz, assign_centroids)
+        # assign centroids if missing
+        if new_haz.centr_exp_col not in new_exp.gdf.columns:
+            LOGGER.warning(
+                "No assigned hazard centroids in exposure object after the "
+                "application of the measure. The centroids were assigned in impact calcualtion."
```
peanutfun: I think we should clarify how to avoid the warning and what it implies:
```suggestion
                "No assigned hazard centroids in exposure object after the "
                "application of the measure. The centroids will be assigned during impact "
                "calculation. This is potentiall costly. To silence this warning, make sure "
                "that centroids are assigned to all exposures."
```

peanutfun (changes requested):
Thank you for the fix!

I would appreciate a test. Could you just add the [minimal example](https://github.com/CLIMADA-project/climada_python/issues/1056#issuecomment-2885778450) from the issue as test case? The test should make sure that the computation runs even when no centroids are assigned, and that the warning is issued. Alternatively, it might also be easy to add such checks to the existing unit tests for `Measures` and `CostBenefit`.

ValentinGebhart:
@peanutfun I have know changed the case when we check that the centroid warning should NOT be raised, as we discussed. I checked that it works (ie raises an error if the centroid warning is raised wrongly, and passes if no warning is raised). It somehow looks a bit awkward to me, but if you think this is good enough, we can merge.

PR #1075, opened 2025-07-17
Updates to impact yearset module and review of tutorial

ValentinGebhart:
While checking the impact yearsets tutorial, I noticed a few potential improvements of the yearset module. The main conceptual point was that the impact events were sampled without replacement. This seemed to work well if all events had the same frequency but resulted in strange behavior if the frequencies vary a lot (see example in comment below). Also, the frequency of the final impact object was unclear.

Changes proposed in this PR:
- When sampling without replacement (default behavior), if the frequency of the different events is different, a warning is issued.
- The yearly impact objects now has a frequency of 1/n_years, where n_years is the number of sampled years.
- The yearly impact object was a deep copy of the original impact object with a few parameters changed (eg, the impact matrix was transfered, which I think dows not make sense). Now the yearly impact object is a fresh impact object filled with relevant attributes only.
- The correction factor is now applied by multiplication and not division (such that the yearly impact aai_agg is the same as the aai_agg of the original impact object). 
- Minor changes in the tutorial.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/util/yearsets.py
+++ b/climada/util/yearsets.py
@@ -25,7 +25,7 @@ LOGGER = logging.getLogger(__name__)
 
 
 def impact_yearset(
-    imp, sampled_years, lam=None, correction_fac=False, with_replacement=True, seed=None
+    imp, sampled_years, lam=None, correction_fac=True, with_replacement=False, seed=None
 ):
     """Create a yearset of impacts (yimp) containing a probabilistic impact for each year
     in the sampled_years list by sampling events from the impact received as input with a
@@ -45,13 +45,13 @@ def impact_yearset(
             events per year. If no lambda value is given, the default lam = sum(imp.frequency)
             is used. Default is None.
         correction_fac : boolean, optional
-            If True a correction factor is applied to the resulting yimp. It is
+            If True, a correction factor is applied to the resulting yimp. It is
             scaled in such a way that the expected annual impact (eai) of the yimp
-            equals the eai of the input impact. Defaults to False.
+            equals the eai of the input impact. Defaults to True.
         with_replacement : bool, optional
             If True, impact events are sampled with replacement. If False, events are sampled
             without replacement. Sampling without replacement can yield distorted samples if
-            frequencies of different events are unqual. Defaults to True.
+            frequencies of different events are unqual. Defaults to False.
         seed : Any, optional
             seed for the default bit generator
             default: None
@@ -86,7 +86,7 @@ def impact_yearset(
 
 
 def impact_yearset_from_sampling_vect(
-    imp, sampled_years, sampling_vect, correction_fac=False
+    imp, sampled_years, sampling_vect, correction_fac=True
 ):
     """Create a yearset of impacts (yimp) containing a probabilistic impact for each year
     in the sampled_years list by sampling events from the impact received as input following
@@ -112,7 +112,7 @@ def impact_yearset_from_sampling_vect(
         correction_fac : boolean, optional
             If True a correction factor is applied to the resulting yimp. It is
             scaled in such a way that the expected annual impact (eai) of the yimp
-            equals the eai of the input impact. Defaults to False.
+            equals the eai of the input impact. Defaults to True.
 
     Returns
     -------
@@ -170,7 +170,7 @@ def sample_from_poisson(n_sampled_years, lam, seed=None):
     return np.round(np.random.poisson(lam=lam, size=n_sampled_years)).astype("int")
 
 
-def sample_events(events_per_year, freqs_orig, with_replacement=True, seed=None):
+def sample_events(events_per_year, freqs_orig, with_replacement=False, seed=None):
     """Sample events uniformely from an array (indices_orig) without replacement
     (if sum(events_per_year) > n_input_events the input events are repeated
     (tot_n_events/n_input_events) times, by ensuring that the same events doens't
@@ -185,7 +185,7 @@ def sample_events(events_per_year, freqs_orig, with_replacement=True, seed=None)
         with_replacement : bool, optional
             If True, impact events are sampled with replacement. If False, events are sampled
             without replacement. Sampling without replacement can yield distorted samples if
-            frequencies of different events are unqual. Defaults to True.
+            frequencies of different events are unqual. Defaults to False.
         seed : Any, optional
             seed for the default bit generator.
             Default: None
@@ -209,7 +209,7 @@ def sample_events(events_per_year, freqs_orig, with_replacement=True, seed=None)
             LOGGER.warning(
                 "The frequencies of the different events are not equal. This can lead to "
                 "distorted sampling if the frequencies vary significantly. To avoid this, "
-                "please set with_replacement=True to sample with replacement instead."
+                "please set `with_replacement=True` to sample with replacement instead."
             )
 
         indices = indices_orig
@@ -226,7 +226,7 @@ def sample_events(events_per_year, freqs_orig, with_replacement=True, seed=None)
                 )
 
             # if not enough events remaining, append original events
-            if len(indices) < amount_events or len(indices) == 0:
+            if len(np.unique(indices)) < amount_events or len(indices) == 0:
                 indices = np.append(indices, indices_orig)
                 freqs = np.append(freqs, freqs_orig)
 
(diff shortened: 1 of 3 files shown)


climada/util/yearsets.py
```diff
@@ -203,47 +200,64 @@ def sample_events(events_per_year, freqs_orig, seed=None):
...
-        probab_dis = freqs[np.unique(indices, return_index=True)[1]] / (
-            np.sum(freqs[np.unique(indices, return_index=True)[1]])
-        )
+    else:
+        # easier method if we allow for replacement sample with replacement
+        if with_replacement is False:
+            LOGGER.warning(
+                "Sampling without replacement not implemented for events with varying "
+                "frequencies. Events are sampled with replacement."
+            )
```
chahank: This is a very confusing way to implement things. Having parameter combinations that work and others which do not is to be avoided if possible.

Also, I am a bit confused here. This is just a warning, while I think it should be an error since you want the code to not run for this combination of input parameters. Silently changing inputs upon calculations is also to be avoided. 

But more generally, why raise an error if the frequencies are not equal? Would it not make sense to have a default behaviour (unequal frequencies with replacement, equal frequencies without replacement), and the user can overwrite this if they want? Or even more simply, always sample with replacement?

chahank (changes requested):
Good catch! I think the implementation goes in the right direction, but I would improve the way it is presented to the user a bit. 

Also, the other improvements look good to me. I might look at the code in detail later to be sure it does what it should do if you want.

climada/util/yearsets.py
ValentinGebhart: Thanks for the feedback :) 

I see your point. Actually, I also thought the easiest way is just always to sample with replacement but maybe there was a reason to implement sampling without replacement that I do not realize.

I suggest this implementation: 
- default: sample with replacement
- Optional parameter `with_replacement` can be set to `False` to sample without replacement. If frequencies of the events are not equal, we raise an error. 

What do you think?

chahank: > Optional parameter with_replacement can be set to False to sample without replacement. If frequencies of the events are not equal, we raise an error.

Would you say that it is always wrong to do with replacement when the frequencies are unequal? I thought to understand that it is only a problem in certain cases. So maybe rather raise a warning and explain the potential problem?

ValentinGebhart: Hm good question. I think strictly speaking, sampling with replacement (how it is implemented) is always wrong if the frequencies are unequal. However I guess the error is small if the frequencies are not super different, and only becomes significant in cases like the above. 
For me it would be okay to just raise a warning, as you suggest. Should I implement that?

chahank: Sounds like a good compromise. Anyway, the default will be with replacement, and thus most users won't have problems.

ValentinGebhart: I now changed the default to sample with replacement, and issue a warning when sampling without replacement for unequal event frequencies. 
I also changed the default to not applying the correction factor, such that the default is that the yearly impacts are actual samples of the original impact object.

chahank: Great! Just so that we do not introduce a confusing default change, can @carmensteinmann please confirm that setting the correction factor application to false by default is a good idea?

climada/util/yearsets.py
```diff
@@ -68,34 +74,20 @@ def impact_yearset(imp, sampled_years, lam=None, correction_fac=True, seed=None)
...
+    yimp = impact_yearset_from_sampling_vect(
+        imp, sampled_years, sampling_vect, correction_fac=correction_fac
     )
 
     return yimp, sampling_vect
 
 
 def impact_yearset_from_sampling_vect(
-    imp, sampled_years, sampling_vect, correction_fac=True
+    imp, sampled_years, sampling_vect, correction_fac=False
```
chahank: If we keep it as is (pending confirmation by @carmensteinmann), we should adjust the docstring. I unfortunately cannot suggest a change here directly in the comment.

chahank:
So that we do not forget, the changelog needs to be updated once done.

climada/util/yearsets.py
```diff
@@ -135,23 +124,24 @@ def impact_yearset_from_sampling_vect(
...
+
+    # create yearly impact object
+    yimp = Impact(
+        at_event=imp_per_year,
+        date=u_dt.str_to_date([str(date) + "-01-01" for date in sampled_years]),
+        event_name=np.arange(1, len(sampled_years) + 1),
+        event_id=np.arange(1, len(sampled_years) + 1),
+        unit=imp.unit,
+        haz_type=imp.haz_type,
+        frequency=np.ones(len(sampled_years)) / len(sampled_years),
```
chahank:
```suggestion
        frequency=np.ones(len(sampled_years)) / len(sampled_years),
        frequency_unit='1/year'
        aai_agg=np.sum(yimp.frequency * yimp.at_event)
```

climada/util/yearsets.py
```diff
@@ -135,23 +124,24 @@ def impact_yearset_from_sampling_vect(
...
     )
 
+    yimp.aai_agg = np.sum(yimp.frequency * yimp.at_event)
```
chahank:
```suggestion
```

climada/util/yearsets.py
chahank: And `eai_exp` is not defined?

chahank:
Small cosmetic suggestion and a question regarding the change in the `eai_exp` value compared to the previous implementation. 

Once resolved this last question, good to go!

climada/util/yearsets.py
ValentinGebhart: to your suggestion: I think `yimp` is not yet defined, so I cannot use it in its own initiation (when you define `aai_agg`), right? I could define it in the initialtion using `aai_agg=np.sum(imp_per_year )/len(sampled_years)`, would that be better? Then the frequency would be hard coded to be constant though

to `eai_exp`: I believe the impact year set does not have spatial information anymore: they build upon `imp.at_event` of the original impact object, so it is aggregated over the locations. We could add `eai_exp` with one entry equal to the `aai_agg`, if you think that would be useful?

chahank: aaa correct. Yes, just use the frequency definition instead, so 
```suggestion
        frequency=frequency,
        aai_agg = np.sum(frequency * imp_per_year)
```
and somewhere above `frequenvy=np.ones(len(sampled_years)) / len(sampled_years)`

chahank: In the previous implementation, `eai_exp` remains whatever was in `imp.eai_exp`. Not that this was a good idea necessarily, but the current implementation changes the output.

ValentinGebhart: yes, in the previous implementation, `yimp` was simply a copy of the original impact object, and just a few attributes were changed. Eg, also the impact matrix was transferred if the original impact object had one. I think this results in a quite confusing output object (kind of hybrid between original events and sampled yearly impacts). That is why I changed it such that `yimp` only has attributes that correspond to the yearly impacts, and yes, that output can look quite different to before.

Does it make sense to change the output in this way? Or are you saying that it is problematic to change the output?

ValentinGebhart: or are you saying that it would be valuable to additionally tranfer just `imp.eai_exp` as there is still some spatial info in here? even tough the numbers of `imp.eai_exp` would then not add up to `imp.agg_aai` if we do not use the correction factor.

chahank: I see, it is much clearer now without carrying over the arguments from before, I agree. Maybe mention it in the Changelog and we are good. 

The `eai_exp` should be computable from the impact matrix and the sampling vector right? But this would probably be a separate PR.

ValentinGebhart: Perfect, I'll adapt the changelog.

Yes, using the original impact matrix and sampling vector one could compute `eai_exp` of the sampled impacts. Still, the `yimp` does not have any location information anymore (e.g. centroids), so maybe it would also be confusing to save the `eai_exp` in the `yimp` object. 
A different way would be to compute a spatially explicit yearly impact set, i.e., where the sampled yearly impacts are saved per centroid (and not aggreagted). There the `eai_exp` would make a lot of sense. If we think this would be interesting, we can do it in a separate PR :)

chahank: I think this would be interesting in the context of the multi-hazard impact module https://github.com/CLIMADA-project/climada_python/discussions/1077 . I would suggest focusing on the multi-hazard impact module instead of tweaking the existing yearset util methods.

ValentinGebhart:
After discussion with Carmen, we decided to keep the default parameters as before:
- **sampling WITHOUT replacement:** better inclusion of all original events in the sampling, hoping that situations with very unbalanced frequencies of events are "uncommon"
- **applying the correction factor**: aai of yearset and original imp object are equal, even though sampled impacts have been changed.

I adapted the docstrings, tutorial and changelog accordingly.

Issue #1055, opened 2025-05-15
Hazard.write_hdf5 and from_hdf5 don't scale since Climada 5.0


emanuel-schmid:
Since the refactoring of `Centroids` in Climada 5.0, their coordinates are stored as a `GeoDataFrame.geometyr` and not as `numpy` arrays, lat/lon anymore.

This has the following consequences:

- the files produced by `write_hdf5` are much bigger than they used to be. Here's an example for Centroids in a grid of size 5'760'000:

| geometry saved as | uncompressed | compressed |
| - | - | - |
| `shapely.Point` (current) | 210M | 167M |
| wkb, byte array (planned) | 177M | 134M |
| x and y (no geometry) | 134M | 15M |

- when reading the hdf files with pickled `Point`s, the risk for exceeding memory limitations is quite high. With a memory limit of 4G, I have not been able to read them without killing the kernel.

It has already been the plan to store `Centroids.gdf` geometries in _wkb_ format, like the ones in `Exposures.gdf`. 
This would alleviate the problem somewhat: lower memory requirements, smaller files (-20%)

However: if we converted the geometry column to x/y columns prior of storing and vice versa after reading, the files would be 90% smaller, and reading/writing faster.
This only works if the `Centroids` are really points and not another type of geometry.

peanutfun:
What about the file formats suggested by GeoPandas? Could you try saving the `Centroids.gdf` directly with [`to_parquet`](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_parquet.html#geopandas.GeoDataFrame.to_parquet) or [`to_arrow`](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_arrow.html#geopandas.GeoDataFrame.to_arrow) and see how that performs in terms of storage size and memory requirements?

emanuel-schmid:
👍  I've tried `to_parquet` and `read_parquet`
Without further ado it writes to a file of 76M in about 3 seconds, (gzip compressed 47M in about 7 seconds).
The geometries are stored in wkb format, btw)

Using parquet as canonical file format in climada is already the plan, but I guess we still need to make it work with hdf5.

PR #1076, opened 2025-07-23
pickle free centroids hdf

emanuel-schmid:
Changes proposed in this PR:

- centroids are not stored as pickled shaply objects in hdf5
- a compression is introduced
- in case centroids are strictly points, they are stored as x and y columns
- otherwise as (pickled) wkb, same as for exposures (#1051)

This PR is a pragmatic workaround of #1055

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/centroids/centr.py
+++ b/climada/hazard/centroids/centr.py
@@ -906,11 +906,20 @@ class Centroids:
             (path and) file name to write to.
         """
         LOGGER.info("Writing %s", file_name)
-        store = pd.HDFStore(file_name, mode=mode)
+        xycols = []
+        wkbcols = []
+        store = pd.HDFStore(file_name, mode=mode, complevel=9)
         pandas_df = pd.DataFrame(self.gdf)
         for col in pandas_df.columns:
             if str(pandas_df[col].dtype) == "geometry":
-                pandas_df[col] = np.asarray(self.gdf[col])
+                try:
+                    pandas_df[col + ".x"] = self.gdf[col].x
+                    pandas_df[col + ".y"] = self.gdf[col].y
+                    pandas_df.drop(col, inplace=True)
+                    xycols.append(col)
+                except ValueError:
+                    pandas_df[col] = self.gdf[col].to_wkb()
+                    wkbcols.append(col)
 
         # Avoid pandas PerformanceWarning when writing HDF5 data
         with warnings.catch_warnings():
@@ -918,10 +927,12 @@ class Centroids:
             # Write dataframe
             store.put("centroids", pandas_df)
 
-        store.get_storer("centroids").attrs.metadata = {
-            "crs": CRS.from_user_input(self.crs).to_wkt()
-        }
-
+        centroids_metadata = {"crs": CRS.from_user_input(self.crs).to_wkt()}
+        if xycols:
+            centroids_metadata["xy_columns"] = xycols
+        if wkbcols:
+            centroids_metadata["wkb_columns"] = wkbcols
+        store.get_storer("centroids").attrs.metadata = centroids_metadata
         store.close()
 
     @classmethod
@@ -951,6 +962,14 @@ class Centroids:
                 # the CRS was stored in '_crs'/'crs'
                 crs = metadata.get("crs")
                 gdf = gpd.GeoDataFrame(store["centroids"], crs=crs)
+                for xycol in metadata.get("xy_columns", []):
+                    gdf[xycol] = gpd.points_from_xy(
+                        x=gdf[xycol + ".x"], y=gdf[xycol + ".y"], crs=crs
+                    )
+                    gdf.drop([xycol + ".x", xycol + ".y"], inplace=True)
+                for wkbcol in metadata.get("wkb_columns", []):
+                    gdf[wkbcol] = gpd.GeoSeries.from_wkb(gdf[wkbcol], crs=crs)
+
         except TypeError:
             with h5py.File(file_name, "r") as data:
                 gdf = cls._gdf_from_legacy_hdf5(data.get("centroids"))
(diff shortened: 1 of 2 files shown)


chahank (changes requested):
Good changes, although some tests are failing now. 

As a note, centroids are currently never supposed to be anything else than points, as the impact calculations cannot deal with other geometries in the centroids at the moment. I think it does not hurt to have the code ready.

climada/hazard/centroids/centr.py
```diff
@@ -906,23 +906,36 @@ def write_hdf5(self, file_name, mode="w"):
...
-            store.put("centroids", pandas_df)
-
-        store.get_storer("centroids").attrs.metadata = {
-            "crs": CRS.from_user_input(self.crs).to_wkt()
-        }
-
-        store.close()
+        xycols = []
+        wkbcols = []
+        store = pd.HDFStore(file_name, mode=mode, complevel=9)
```
peanutfun, line 915: For zlib, it seems like looks like high compression levels only slightly reduce the file size while costing much performance. A lower value seems more advisable to me. See https://www.pytables.org/usersguide/optimization.html#compression-issues

```suggestion
        store = pd.HDFStore(file_name, mode=mode, complevel=3)
```

climada/hazard/centroids/centr.py
```diff
@@ -906,23 +906,36 @@ def write_hdf5(self, file_name, mode="w"):
...
+        try:
+            pandas_df = pd.DataFrame(self.gdf)
```
peanutfun, line 917: Is it really necessary to copy the entire data into a new data frame?

climada/hazard/centroids/centr.py
```diff
@@ -906,23 +906,36 @@ def write_hdf5(self, file_name, mode="w"):
...
+            for col in pandas_df.columns:
+                if str(pandas_df[col].dtype) == "geometry":
+                    try:
+                        pandas_df[col + ".x"] = self.gdf[col].x
+                        pandas_df[col + ".y"] = self.gdf[col].y
+                        pandas_df.drop(columns=[col], inplace=True)
+                        xycols.append(col)
+                    except ValueError:
```
peanutfun: Make clear that you need points for this
```suggestion
                    if (self.gdf.geometry.geom_type == "Point").all():
                        pandas_df[col + ".x"] = self.gdf[col].x
                        pandas_df[col + ".y"] = self.gdf[col].y
                        pandas_df.drop(columns=[col], inplace=True)
                        xycols.append(col)
                    else:
```

climada/hazard/centroids/centr.py
```diff
@@ -906,23 +906,36 @@ def write_hdf5(self, file_name, mode="w"):
...
+                        pandas_df[col] = self.gdf[col].to_wkb()
+                        wkbcols.append(col)
```
peanutfun, line 931: The WKB-case is not covered by the tests

climada/hazard/centroids/centr.py
```diff
@@ -906,23 +906,36 @@ def write_hdf5(self, file_name, mode="w"):
...
-        }
-
-        store.close()
+        xycols = []
+        wkbcols = []
+        store = pd.HDFStore(file_name, mode=mode, complevel=9)
+        try:
+            pandas_df = pd.DataFrame(self.gdf)
+            for col in pandas_df.columns:
+                if str(pandas_df[col].dtype) == "geometry":
```
peanutfun, line 923: Make clear that you do not want to iterate over all columns:
```suggestion
            for col in filter(lambda x: str(x.dtype) == "geometry", pandas_df.columns):
```
(Suggestion won't work because the following code needs to be indented less)

climada/hazard/centroids/centr.py
```diff
@@ -950,7 +963,15 @@ def from_hdf5(cls, file_name):
                 # in previous versions of CLIMADA and/or geopandas,
                 # the CRS was stored in '_crs'/'crs'
                 crs = metadata.get("crs")
-                gdf = gpd.GeoDataFrame(store["centroids"], crs=crs)
+                gdf = gpd.GeoDataFrame(store["centroids"])
+                for xycol in metadata.get("xy_columns", []):
```
peanutfun: Please also add a test with multiple xy_columns/wkb_columns to be stored and read.

peanutfun (changes requested):
Thank you for tackling this issue! I have some concerns regarding the readability and performance of the code, especially because the entire data needs to be copied. I also think the tests need to be adapted a bit. See my comments.

climada/hazard/centroids/centr.py
emanuel-schmid: Certainly not, creating a DataFrame from a GeoDataFrame doesn't copy but refer.

climada/hazard/centroids/centr.py
emanuel-schmid: elegant suggestion - but I leave it as it is. it's more "climada style" like that.

climada/hazard/centroids/centr.py
emanuel-schmid: I leave it as it is: in an arbitrary test, the cpu decrease was 0.4 seconds, about 15%, the size increase 2M, about 10%. From a $ point of view complevel 9 seems justified.

climada/hazard/centroids/centr.py
peanutfun: Indeed. Thanks!

climada/hazard/centroids/centr.py
peanutfun: Then please add a comment, what you call "climada style" confused me quite a bit 😕 
```suggestion
            # Iterate over geometry columns (only)
            for col in pandas_df.columns:
                if str(pandas_df[col].dtype) == "geometry":
```

climada/hazard/centroids/test/test_centr.py
```diff
@@ -725,6 +725,51 @@ def test_read_write_hdf5(self):
...
+                            shapely.LineString([(0, 1), (1, 2)]),
+                        ]
+                        * 8
+                    }
+                ).set_geometry("more_shapes", crs=DEF_CRS)
+            )
+        )
+        centroids_w.write_hdf5(tmpfile)
+        centroids_r = Centroids.from_hdf5(tmpfile)
+        self.assertTrue(centroids_w == centroids_r)
```
peanutfun:
```suggestion
        self.assertEqual(centroids_w, centroids_r)
```

climada/hazard/centroids/centr.py
peanutfun: 15% decrease vs. 10% increase seems like an argument for a lower complevel, from my point of view. But I guess it's not that relevant. In case we run into some issues, we might consider making this a method kwarg in the future.

climada/hazard/centroids/test/test_centr.py
emanuel-schmid: (this was actually done in purpose - the idea was to make sure the overridden equality operator does what it ought to do, regardless of what exactly happens inside `assertEqual`, about which I have no clue)

PR #1080, opened 2025-08-05
Feature/relative matching distance

### LINKED ISSUE #1079: Improve default behavior of nearest neighbor matching to avoid using a fixed distance threshold
**Description of the issue**
When two sets of coordinates are tried to be matched in `util.match_coordinates`, non-exact matching is allowed only if the distance between the points to be matched is smaller than a fixed `threshold` parameter. This `threshold` parameter has a default value of 100 (km) and can be updated by the user. However, using a fixed-value threshold as a default is sub-optimal as it results in always using the same default limit value of 100 regardless of the unit of the input coordinates system. Furthermore, it can lead to frequent warnings when hazard and exposures data with differing resolutions are tried to be matched.

**Proposed fix**
I suggest using a relative threshold i.e. that is based on the resolution of the input data instead of a fixed threshold. E.g. we could get the resolutions of the input data using `util.get_resolution` and then set a default threshold in function of that and alternatively warn if the difference of resolution between the two input data is too large.

**Additional context**
This issue has been discussed in the context of improving functions to also allow for non-geographic coordinates to be used as input (see [#1020 ]).

chahank:
Changes proposed in this PR:
- This PR contains the changes from #1020 
- In addition, it adds a relative matching threshold based on the equivalent grid spacing of the centroids.

This PR fixes #1079.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/util/coordinates.py
+++ b/climada/util/coordinates.py
@@ -2199,7 +2199,7 @@ def get_resolution_1d(coords, min_resol=1.0e-8):
     res = np.diff(np.unique(coords))  # always positive because unique is sorted
     diff = np.abs(np.diff(coords))
     mask = (res > min_resol) & np.isin(res, diff)
-    if not len(mask):
+    if len(mask) == 0:
         raise ValueError(
             "Cannot find a resolution higher than `min_resol` for the given coordinates"
         )


climada/util/coordinates.py
```diff
@@ -85,6 +85,79 @@
...
+    ------
+    ValueError : if lat or lon out of likely geographical coordinate bounds
+
+    Returns
+    -------
+    test : bool
+        True if lat/lon ranges seem to be in the geographic coordinates range, otherwise False.
+    """
+    lat = np.array(lat)
+    lon = np.array(lon)
```
peanutfun: This creates copies. Use `asarray` to avoid copies if numpy arrays are passed

climada/util/coordinates.py
```diff
@@ -85,6 +85,79 @@
...
+        Latitudes and longitudes of points.
+
+    Raises
+    ------
+    ValueError : if lat or lon out of likely geographical coordinate bounds
+
+    Returns
+    -------
+    test : bool
+        True if lat/lon ranges seem to be in the geographic coordinates range, otherwise False.
```
peanutfun, line 104: This function never returns `False`. It either returns `True` or raises an error. I think you should split that functionality: Have a function `is_geo_coords` (or similar) that returns true/false, and then have a function `check_XZY`, that raises an error or returns None.

climada/util/coordinates.py
```diff
@@ -85,6 +85,79 @@
...
+    if not like_geo:
+        raise ValueError(
+            "Input lat and lon coordinates do not seem to correspond"
+            " to geographic coordinates in degrees. This can be because"
+            " total extents are > 180 for lat or > 360 for lon, lat coordinates"
+            " are outside of -90<lat<90, or lon coordinates are outside of -540<lon<540."
+            " If you use degree values outside of these ranges,"
+            " please shift the coordinates to the valid ranges."
+        )
+    return True
```
peanutfun: ?
```suggestion
    return like_geo
```

climada/util/coordinates.py
```diff
@@ -1059,10 +1147,13 @@ def match_coordinates(
     if np.array_equal(coords, coords_to_assign):
         assigned_idx = np.arange(coords.shape[0])
     else:
+        if threshold is None:
+            threshold = 2 * max(abs(r) for r in get_resolution(coords_to_assign.T))
```
peanutfun: `get_1d_resolution` can be extremely costly. It takes unique coordinates and computes differences between them. Shouldn't we raise a warning that this is suboptimal?

https://github.com/CLIMADA-project/climada_python/blob/08ec51bb958436cd18e77cf8f4178bc05a295d78/climada/util/coordinates.py#L1836

climada/util/test/test_coordinates.py
```diff
@@ -230,6 +236,47 @@ def def_ref_50():
     )
 
 
+class TestGetUnitCoords(unittest.TestCase):
+    def create_mock_gdf(self, crs, num_points=5):
+        geometry = [
+            shapely.Point(x, y) for x, y in zip(range(num_points), range(num_points))
+        ]
+        gdf = gpd.GeoDataFrame(geometry=geometry, crs=crs)
+        return gdf
```
peanutfun, line 254: This is called a *test fixture*. Please use `setUp` to create fixtures and save them in the `TestCase` object. You can then retrieve them from the `self` object within each test case. See https://docs.python.org/3/library/unittest.html#organizing-test-code

peanutfun (changes requested):
Stumbled upon this and had the time, so I left a review. This looks good already, but I found some issues with data being copied and default operations that might be very costly.

climada/util/coordinates.py
chahank: Yes, I agree.

climada/util/coordinates.py
chahank: I will try to estimate the computation cost and see whether optimization is needed.

chahank:
Ok, all tests are passing, I think the implementation works now. 

TO DO:
- work-in the comments
- improve / adjust all the docstrings where needed
- update the documentation

chahank:
@luseverin : it would be good that you review my changes once to check functionality as you made the original PR https://github.com/CLIMADA-project/climada_python/pull/1020 this is based upon.

climada/util/test/test_coordinates.py
chahank: As I understand it, It is not really a fixture, since it has a variable as input. The idea is to change the value of crs for the tests. 

Or maybe I misunderstood how fixtures work.

climada/util/coordinates.py
chahank: Done.

climada/util/coordinates.py
chahank: Done.

climada/util/coordinates.py
chahank: Testing, for a regular grid with 65'000 points (world at 1° resolution), it takes 13ms, for 6'500'000 points (world at 0.1° resolution) it takes 1.5s. Is this acceptable?

Screenshot 2025-08-27 at 15 07 09

chahank:
Note: the method `lon_normalize` is apparently used in several tests to change longitudinal values that do not conform to the constraints of the new `check_if_geo_coordinates` method. Hence, one can either accept that `lon_normalize` does also work with non-conform coordinates, or one would have to update the several tests in python core and petals. @luseverin ?

climada/engine/impact_calc.py
```diff
@@ -92,6 +92,11 @@ def impact(
             Default: True
         assign_centroids : bool, optional
             indicates whether centroids are assigned to the self.exposures object.
+            The default centroids assignement uses euclidean distance and a threshold equal to
+            twice the highest resolution of the centroids (assuming a regular grid).
```
luseverin: I am a bit confused with the "highest" here; does it refer to the fact that we take the max betwen the longitudinal vs latitudinal resolutions or does it refer to the fact that the get_resolution_1d function takes the minimum grid spacing? In either case I would reformulate

climada/entity/exposures/base.py
```diff
@@ -591,65 +591,81 @@ def assign_centroids(
...
         threshold : float
-            If the distance (in km) to the nearest neighbor exceeds `threshold`,
+            Assignement threshold in units of the exposure's crs.
+            If the distance to the nearest neighbor exceeds `threshold`,
             the index `-1` is assigned.
-            Set `threshold` to 0, to disable nearest neighbor matching.
-            Default: 100 (km)
+            Set `threshold` to 0, to disable nearest neighbor matching and enforce
+            exact matching.
+            Default: twice the highest resolution of the hazard centroids
```
luseverin: Same as above, not sure "highest" is the best formulation..

climada/util/coordinates.py
```diff
@@ -1016,9 +1227,14 @@ def match_coordinates(
     distance : str, optional
         Distance to use for non-exact matching. Possible values are "euclidean", "haversine" and
         "approx". Default: "euclidean"
+    crs : str or pyproj crs, optinal
```
luseverin:
```suggestion
    crs : str or pyproj crs, optional
```

climada/util/coordinates.py
```diff
@@ -1305,49 +1602,74 @@ def _nearest_neighbor_euclidean(
...
         First column contains latitude, second column contains longitude. Each
         row is a geographic point
+    crs : pyproj crs
+        Coordinate reference system (crs) of the input coordinates and centroids.
     threshold : float
-        distance threshold in km over which no neighbor will be found. Those
-        are assigned with a -1 index
+        Assignement threshold in units of the provided crs.
+        If the distance to the nearest neighbor exceeds `threshold`, the index `-1` is assigned.
     check_antimedirian: bool, optional
```
luseverin:
```suggestion
    check_antimeridian: bool, optional
```

climada/util/coordinates.py
```diff
@@ -1170,11 +1445,11 @@ def _dist_sqr_approx(lats1, lons1, cos_lats1, lats2, lons2):
...
 
 def _nearest_neighbor_approx(
-    centroids, coordinates, threshold, check_antimeridian=True
+    centroids, coordinates, crs, threshold, check_antimeridian=True
 ):
     """Compute the nearest centroid for each coordinate using the
-    euclidean distance d = ((dlon)cos(lat))^2+(dlat)^2. For distant points
-    (e.g. more than 100km apart) use the haversine distance.
+    squared equirectangular approximation distance d = ((dlon)cos(lat))^2+(dlat)^2.
+    For distant points (e.g. more than 100km apart) use the haversine distance.
```
luseverin, line 1466:
```suggestion
Requires geographic coordinates in degrees.
```

climada/util/coordinates.py
```diff
@@ -1233,13 +1519,13 @@ def _nearest_neighbor_approx(
...
-            centroids, coordinates, threshold, assigned
+            centroids, coordinates, threshold, assigned, crs=crs
         )
 
     return assigned
 
 
-def _nearest_neighbor_haversine(centroids, coordinates, threshold):
+def _nearest_neighbor_haversine(centroids, coordinates, crs, threshold):
     """Compute the neareast centroid for each coordinate using a Ball tree with haversine distance.
```
luseverin, line 1543:
```suggestion
Requires geographic coordinates in degrees.
```

climada/util/coordinates.py
```diff
@@ -1305,49 +1602,74 @@ def _nearest_neighbor_euclidean(
...
+            centroids, coordinates[idx], threshold, assigned, crs=crs
         )
 
     # Copy result to all exposures and return value
     return assigned[inv]
 
 
-def _nearest_neighbor_antimeridian(centroids, coordinates, threshold, assigned):
+def _nearest_neighbor_antimeridian(centroids, coordinates, threshold, assigned, crs):
     """Recompute nearest neighbors close to the anti-meridian with the Haversine distance
```
luseverin, line 1687:
```suggestion
Requires geographic coordinates in degrees.
```

climada/util/coordinates.py
```diff
@@ -1369,6 +1691,17 @@ def _nearest_neighbor_antimeridian(centroids, coordinates, threshold, assigned):
...
+            "Only geographic crs are supported for nearest neighbor matching using"
+            "'haversine' distance at the antimeridian."
+        )
+    # first check that unit is in degree
+    if get_crs_unit(crs) != "degree":
+        raise ValueError(
+            "Only coordinates in degree units are supported for nearest neighbor matching using"
+            "'haversine' distance at the antimeridian."
+        )
     lon_min = min(centroids[:, 1].min(), coordinates[:, 1].min())
```
luseverin, line 1718:
```suggestion
    # check that coords are indeed geographic degrees (in case we do no trust the crs)
    check_if_geo_coords(centroids[:, 0], centroids[:, 1])
    check_if_geo_coords(coordinates[:, 0], coordinates[:, 1])
    lon_min = min(centroids[:, 1].min(), coordinates[:, 1].min())
```

climada/util/coordinates.py
```diff
@@ -1379,7 +1712,7 @@ def _nearest_neighbor_antimeridian(centroids, coordinates, threshold, assigned):
     mid_lon = 0.5 * (lon_max + lon_min)
     antimeridian = mid_lon + 180
 
-    thres_deg = np.degrees(threshold / EARTH_RADIUS_KM)
+    thres_deg = threshold
```
luseverin, line 1728: Why is that necessary?

climada/util/coordinates.py
```diff
@@ -982,11 +1070,138 @@ def assign_coordinates(*args, **kwargs):
...
+        If the input `coords_to_assign` contains fewer than two coordinates,
+        as resolution cannot be determined from a single point.
+
+    See Also
+    --------
+    get_resolution : Helper function to calculate the resolution of coordinates.
+
+    Notes
+    -----
+    The resolution is defined as the smallest non-zero difference between adjacent
```
luseverin, line 1117: Why do we take the smallest differences? Wouldn't the mean difference be better? What happened if the input coordinates are not gridded?

luseverin (commented):
Thanks @chahank! I only have some minor comments: 

- First a clarification; I am not exactly sure of what is supposed to happen in the case of non-regular coordinates for the determination of the distance threshold, and I am also a bit confused with the "highest resolution" formulation in the description of the threshold definition, see my comments.
- Some tests are missing in  `assign_centroids` (when crs are not equal), in `get_gridcellarea` (when the unit is unknown), in `get_crs_unit`, in `dist_to_coast`, and in `estimate_matching_threshold`.

Otherwise, based on your changes and the tests, it seems to be in order in terms of functionality.

climada/util/coordinates.py
peanutfun: Yes, that seems acceptable to me! Thanks a lot for the testing 🙌

climada/util/test/test_coordinates.py
peanutfun: Indeed, I was wrong about the usage of a fixture here. What you do is more of a [parameterized test](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#parametrize-basics), which unittest cannot really facilitate 🤷

chahank:
> > Note: the method `lon_normalize` is apparently used in several tests to change longitudinal values that do not conform to the constraints of the new `check_if_geo_coordinates` method. Hence, one can either accept that `lon_normalize` does also work with non-conform coordinates, or one would have to update the several tests in python core and petals. @luseverin ?
> 
> Is that an in issue? As far as I understand, the `check_geo_coordinates` function tests that the input coordinates are likely geographic, whereas the `lon_normalize` only makes sure that input coordinates (in degrees) are normalized to a -180<lon<180 range. So in the end, they do different things, but it is still consistent as we still want that the input coordinates are likely geographic before attempting any normalization. Or did I misunderstand that? Maybe you could indicate in which tests this is an issue.

`lon_normalize` idea is that if coordinates are say 0-360, they are moved to -180, 180.  However, the method now accepts any values and simply normalizes them to be in the -180 to 180 range. Thus faulty meter values can be entered, or also faulty degree ranges. The latter is done in several tests, and has also led to incorrect results in the past. Basically, you can have values say 0 to 400. Do we want to prevent this?

chahank:
> Ok, I see. Then it means that we should be raising a `ValueError` with the `check_if_geo_coords` function in those instances. Should we simply call `check_if_geo_coords` within `lon_normalize` and adapt the tests accordingly?

I think it is really worth trying. I tried it quickly, and many tests failed, and I did not yet formed an opinion whether this means that we should adapt the tests or do something else.

chahank:
> Ok so I added `check_if_geo_coords` in `lon_normalize` and updated the couple tests that could be easily updated (essentially, removed test data that was not consistent with our definition of geographic coordinates). A few other things happened in addition:
> 
>     * `test_set_one_file_pass` was failing because an empty lon array was inputed to `lon_normalize`within the call to `tropCyclone.from_tracks(tc_track, centroids=CENTR_TEST_BRB) `. I fixed it by adding an explicit handling of empty arrays in `is_geo_coords`, where empty arrays are considered as valid geographic coordinates. 

Excellent!

> I do not know if this is a desirable feature, and I also do not know if the empty lon array in `tropCyclone.from_tracks(tc_track, centroids=CENTR_TEST_BRB) ` is a bug or not. It seems that the `idx_centr_filter` variable becomes an empty array at some point, leading to an empty centroids array, but I did not investigate this further..
> 
>     * In `test_ibtracs_with_basin`, the track downloaded with` tc_track = tc.TCTracks.from_ibtracs_netcdf(year_range=(2002, 2003), basin="WP", genesis_basin="EP")` has longitudinal value that have extent > 360 (lon_min=-179, lon_max=182.6). I increased the tolerance in `is_geo_coords` for the longitudinal extent from 361 to 541 as a fix. It seems to be working for now and I hope this should work for all tracks that one can get from IBtracks but this is only a guess..

Great. Weird that IBTrACS basically has invalid coordinates, but it works ;). 
Could once try to load all IBTrACS?

> 
>     * A couple tests from the LitPop module are now failing:  `test_switzerland300_pass` and `test_Liechtenstein_15_lit_pass`. It seems that some parameters of the Affine rasterio transform have been changed, e.g. `ent.meta["transform"][4] == 0.08333333000000209` instead of `-0.08333333333333333`, and `ent.meta["transform"][5] == 45.833333335` instead of `47.75`. @chahank any idea where this might come from?

Was this changed in this PR? Then it might have been a mistake on my side. Please feel free to revert these changes.

climada/engine/impact_calc.py
chahank:
```suggestion
            The default centroids assignment uses Euclidean distance and a threshold equal to
            twice the highest resolution (smallest distance between any two nearest neighbours)
            of the centroids (assuming a regular grid).
```

chahank: @luseverin Better? If yes, please commit.

climada/engine/impact_calc.py
```diff
@@ -114,10 +119,20 @@ def impact(
             >>> imp = impcalc.impact(insured=True)
             >>> imp.aai_agg
 
+            >>> haz = Hazard.from_hdf5(HAZ_DEMO_H5)  # Set hazard
+            >>> impfset = ImpactFuncSet.from_excel(ENT_TEMPLATE_XLS)
+            >>> exp = Exposures(pd.read_excel(ENT_TEMPLATE_XLS))
+            >>> #Adjust threshold and distance to centroids/exposure resolution
+            >>> exp.assign_centroids(hazard, threshold=1.5, distance='haversine')
```
chahank:
```suggestion
            >>> exp.assign_centroids(hazard, threshold=1.5, distance='euclidean')
```
I think I would not recommend to always use haversine as this can be quite computationally expensive. If you agree @luseverin , please commit.

climada/entity/exposures/base.py
chahank:
```suggestion
            Default: twice the highest resolution (smallest distance between any two nearest neighbours) of the hazard centroids
```

climada/util/coordinates.py
```diff
@@ -1848,10 +2196,14 @@ def get_resolution_1d(coords, min_resol=1.0e-8):
...
         Resolution of given grid.
     """
-    res = np.diff(np.unique(coords))
-    diff = np.diff(coords)
-    mask = (res > min_resol) & np.isin(res, np.abs(diff))
-    return diff[np.abs(diff) == res[mask].min()][0]
+    res = np.diff(np.unique(coords))  # always positive because unique is sorted
+    diff = np.abs(np.diff(coords))
+    mask = (res > min_resol) & np.isin(res, diff)
+    if not len(mask):
```
chahank:
```suggestion
    if len(mask) == 0:
```

climada/util/coordinates.py
chahank: You are right, not needed, just lazy legacy.

chahank:
Oh, one last thing we should do, since this affects the core climada computation, is to check whether the tutorials still work.

PR #1086, opened 2025-08-19
fix and update: eca_san_salvador application within the script folder

### LINKED ISSUE #1085: Reading hazard files in .mat format is not supported
**Describe the bug**
When trying to run the eca_san_salvador application notebooks, reading hazard using "from_mat" gives an error.

**To Reproduce**
1. Running any of the notebook

Code example:
```python

HAZ_FILE = "Salvador_hazard_FL_2015.mat"

haz_2015 = Hazard.from_mat(HAZ_FILE)
```

**Expected behavior**
climada.hazard.io - INFO - Reading Salvador_hazard_FL_2015.mat 

**Screenshots**

Image

**Climada Version:** 6.0.1

 - Operating system and version: macOS 14.3.1
 - Python version:  3.11.13
 (to obtain this information execute > import sys  >print(sys.version))

**Additional context**

rabinatwayana:
Updates to the eca_san_salvador application within the script folder:
- Replaced Salvador_hazard_FL_2015.mat and Salvador_hazard_FL_2040_extreme_cc.mat with their corresponding .hdf5 files. 
- Updated the "cover" column in FL_entity_Acelhuate_parametric.xlsx (i.e. changed values from 0 to 500)
- Modified the adaptation, parametric, and risk notebooks to read hazards from .hdf5 files instead of .mat files
- Switched the base map in plot_salvador_ma (inside functions_ss.py) from Stamen Terrain to OpenStreetMap Mapnik.

**Supporting Resources:**
File conversion from .mat to .hdf5 can be performed in two ways:
      1. Manually ([link](https://github.com/Omowonuola-Akintola/ECA_CLIMADA_V6.0.1)) 
      2. Using an older version of CLIMADA ([link](https://github.com/rabinatwayana/ECA_scripts_climada_6.0.1))

This PR fixes #1085 

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Thanks for the updating of the script @rabinatwayana ! 

Can you quickly explain why the cover in `FL_entity_Acelhuate_parametric.xlsx` needs to be changed?

ChrisFairless:
Thanks a lot Rabina!

The context for this PR is that the team at UNU-EHS like to use the San Salvador case study for teaching, but the case study doesn't run on more recent versions of CLIMADA due to the issue that Rabina raises.

The issue to the Cover column was discovered during the fix, possibly CLIMADA changed how it treated zero-valued covers at some point. So this changes the Cover to be the value of the exposure (500 everywhere).

One other thing – the Jenkins pre-commit hook is failing here, and I can't interpret the error. Is it because something is wrong with black-jupyter hook, or is it because Jenkins can't commit its changes when the pull request comes from a fork?

script/applications/eca_san_salvador/functions_ss.py
```diff
@@ -38,7 +38,10 @@ def plot_salvador_ma():
     ax = shape.plot(figsize=(10, 10), alpha=0.5)
     ax.set_xlim(-9943223.896891385, -9911000.065720687)
     ax.set_ylim(1530712.637786494, 1555600.2891258441)
-    ctx.add_basemap(ax, zoom=12, url=ctx.providers.Stamen.Terrain)
+    # ctx.add_basemap(ax, zoom=12, url=ctx.providers.Stamen.Terrain)
+    ctx.add_basemap(ax, zoom=12, source=ctx.providers.OpenStreetMap.Mapnik)
+
+
```
peanutfun: Remove commented-out code and fix failing pre-commit hook check
```suggestion
    ctx.add_basemap(ax, zoom=12, source=ctx.providers.OpenStreetMap.Mapnik)

```

rabinatwayana:
Hi Chahank, 
I hope the following test will make it clear about the issue with the cover.

[San_Salvador_Parametric_existing_in_develop_branch](https://github.com/user-attachments/files/21913155/San_Salvador_Parametric_existing_in_develop_branch.ipynb)

[San_Salvador_Parametric_result_with_existing_cover_value_with_0.ipynb](https://github.com/user-attachments/files/21913196/San_Salvador_Parametric_existing_cover_value_with_0.ipynb)

[San_Salvador_Parametric_result_after_updateing_cover_value_to_500](https://github.com/user-attachments/files/21913198/San_Salvador_Parametric_updated_cover_value_to_500.ipynb)

Please let me know if it is fine or if you want to make any changes or do other tests.

ChrisFairless:
re: changing the cover
The existing Excel file, used for demonstrating parametric insurance, provides an update to the main San Salvador exposures. It sets the value of each asset to a constant 500 USD and the cover to zero (in the main notebook the cover was undefined). It describes this as writing property-level insurance up to the level of 500 USD for each property.

(That's not how I would have implemented it: I would have left the asset values unchanged and set the cover to 500, but whatever, this PR is to fix the bug, not to change the analysis.)

In the past this must have worked, and my assumption is that when the notebook was created, CLIMADA treated 0 cover and NA cover as equivalent. Running the code now, it treats zero cover as exactly that: uninsured. I'm not very interested in finding out when and where that change was made, because I think this is now the correct behaviour. But it means that the Excel file needs to be updated to preserve its intended behaviour of generating an EP curve when each property is insured with 500 USD. That means setting the value and the cover of each asset to 500 USD.

ChrisFairless:
Re: moving to CLIMADA Petals
Could we keep the analysis here for now? It's part of different UNU teaching materials. We like this because it (1) uses Excel inputs, (2) isn't TC, (3) is return-period based, and (4) is small data and at a community level.

Part of our upcoming work includes proposing and updating the Entity Excel file templates (and reworking Excel inputs in general) to align with the upcoming changes in the Measures and OptionsAppraisal modules. Once that happens the San Salvador notebook will either need to be retired, updated or replaced, so that would be a good time to discuss its future.

chahank:
Thanks @ChrisFairless for all the clarifications! That is all good for me then.

chahank (approved):
Thanks for the quick update!

PR #1091, opened 2025-09-04
Allow Exposures.from_raster to set Exposure metadata

ChrisFairless:
Changes proposed in this PR:
- Add an `attrs` parameter to the `Exposures.from_raster` class method, analogous to the `Hazard.from_raster` method. In this method the `attrs` parameter is a dictionary that is used as additional keyword arguments in the subsequent call of the `__init__`.
- However, to limit unexpected behaviour, I've limited the input dictionary to keys contained in the `Exposures._metadata` list `["description", "ref_year", "value_unit"]`. Other arguments taken by `Exposures.__init__` method are either already provided by the `from_raster` method, or could make a mess (e.g. `lat`, `lon`, `value`). In the current implementation, they are dropped quietly – should there be a warning too?
- While the `Hazard.from_raster` method takes `attrs` as one of its first parameters, here I've added `attrs` as an additional, final parameter. I think it i's less likely to break stuff and the end is usually where a `**kwargs` goes anyway.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/entity/exposures/base.py
+++ b/climada/entity/exposures/base.py
@@ -707,6 +707,7 @@ class Exposures:
         width=None,
         height=None,
         resampling=Resampling.nearest,
+        attrs=None
     ):
         """Read raster data and set latitude, longitude, value and meta
 
@@ -734,11 +735,19 @@ class Exposures:
         resampling : rasterio.warp,.Resampling optional
             resampling
             function used for reprojection to dst_crs
+        attrs : dict
+            dictionary giving kwargs passed to the Exposures class. Keys must 
+            be in the Exposures._metadata list, i.e. "description", "ref_year",
+            "value_unit"
 
         returns
         --------
         Exposures
         """
+        if not attrs:
+            attrs = {}
+        attrs = {key: attrs[key] for key in set(attrs) & set(cls._metadata)}
+
         meta, value = u_coord.read_raster(
             file_name,
             [band],
@@ -765,6 +774,7 @@ class Exposures:
             },
             meta=meta,
             crs=meta["crs"],
+            **attrs
         )
 
     def plot_scatter(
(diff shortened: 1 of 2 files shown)


ChrisFairless:
Tests are failing but I'm not sure they're related to the proposed change

chahank:
@ChrisFairless thanks for the suggestion.

- Can you please also update the changelog?

- Question: With this fix, you can only add `_metadata` attributes to the Exposure object, but not to the underlying geodataframe. Is this what is intended? So you could for instance not include a `cover`. Is this a problem or are we fine with it?

ChrisFairless:
Changelog updated!

Regarding your question about whether we should be able to set non-metadata properties through this new parameter: my thinking was to restrict it to the metadata parameters so that we don't have to write a lot of code to police the unexpected things someone might provide through this call.

However, reading the `Exposures.__init__` method more carefully, these checks are already made.

So ... let's remove the restriction? The danger is that someone provides e.g. cover and deductible info through `data` that don't align with the raster, and as long as there are the right number of rows they'll never know. But I guess it's not our job to tell them to be careful?

I'll make the change, but will revert it if anyone wants to be more careful.

chahank:
> Changelog updated!
> 
> Regarding your question about whether we should be able to set non-metadata properties through this new parameter: my thinking was to restrict it to the metadata parameters so that we don't have to write a lot of code to police the unexpected things someone might provide through this call.
> 
> However, reading the `Exposures.__init__` method more carefully, these checks are already made.
> 
> So ... let's remove the restriction? The danger is that someone provides e.g. cover and deductible info through `data` that don't align with the raster, and as long as there are the right number of rows they'll never know. But I guess it's not our job to tell them to be careful?
> 
> I'll make the change, but will revert it if anyone wants to be more careful.

Good! 

@emanuel-schmid any objections to this solution? If not, please feel free to merge. Thanks!

PR #1185, opened 2025-12-09
Add select method to HazardForecast

### LINKED ISSUE #1157: Implement `HazardForecast.select`
```python
def select(member=None,
        lead_time=None,
        event_names=None,
        event_id=None,
        date=None,
        orig=None,
        reg_id=None,
        extent=None,
        reset_frequency=False):
    event_id = ...
    return super().select(...)
```

Create event IDs from member & lead time input (#1159), merge with given event IDs, pass to base class.

If base class select does not work (#1155), reimplement it in the derived class.

ValentinGebhart:
Changes proposed in this PR:
- add select method that first select upon member and lead_time argurments, encodes this in event_id, and then call Hazard.select with the new event_id
- expand tests

This PR fixes #1157

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/forecast.py
+++ b/climada/hazard/forecast.py
@@ -147,11 +147,13 @@ class HazardForecast(Forecast, Hazard):
                 if lead_time is not None
                 else np.full_like(self.lead_time, True, dtype=bool)
             )
-            mask_event_id = np.asarray(self.event_id)[(mask_member & mask_lead_time)]
+            event_id_from_forecast_mask = np.asarray(self.event_id)[
+                (mask_member & mask_lead_time)
+            ]
             event_id = (
-                np.intersect1d(event_id, mask_event_id)
+                np.intersect1d(event_id, event_id_from_forecast_mask)
                 if event_id is not None
-                else mask_event_id
+                else event_id_from_forecast_mask
             )
 
         return super().select(
--- a/climada/hazard/forecast.py
+++ b/climada/hazard/forecast.py
@@ -20,6 +20,7 @@ Define Forecast variant of Hazard.
 """
 
 import logging
+from typing import Self
 
 import numpy as np
 
@@ -107,6 +108,8 @@ class HazardForecast(Forecast, Hazard):
 
     def select(
         self,
+        member=None,
+        lead_time=None,
         event_names=None,
         event_id=None,
         date=None,
@@ -114,9 +117,26 @@ class HazardForecast(Forecast, Hazard):
         reg_id=None,
         extent=None,
         reset_frequency=False,
-        member=None,
-        lead_time=None,
-    ):
+    ) -> Self:
+        """Select entries based on the parameters and return a new instance.
+
+        The selection will contain the intersection of all given parameters.
+
+        Parameters
+        ----------
+        member : Sequence of ints
+            Ensemble members to select
+        lead_time : Sequence of numpy.timedelta64
+            Lead times to select
+
+        Returns
+        -------
+        HazardForecast
+
+        See Also
+        --------
+        :py:meth:`~climada.hazard.base.Hazard.select`
+        """
         if member is not None or lead_time is not None:
             mask_member = (
                 self.idx_member(member)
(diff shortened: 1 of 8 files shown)


climada/hazard/forecast.py
```diff
@@ -104,3 +104,62 @@ def _check_sizes(self):
...
+                self.idx_member(member)
+                if member is not None
+                else np.full_like(self.member, True, dtype=bool)
+            )
+            mask_lead_time = (
+                self.idx_lead_time(lead_time)
+                if lead_time is not None
+                else np.full_like(self.lead_time, True, dtype=bool)
+            )
+            mask_event_id = np.asarray(self.event_id)[(mask_member & mask_lead_time)]
```
chahank: Technically, this is probably not a mask anymore, but the event id you select right? Just to be sure this is the wanted behaviour (which I think it is).

climada/hazard/test/test_forecast.py
```diff
@@ -107,46 +107,81 @@ def test_hazard_forecast_concat(haz_fc, lead_time, member):
...
+            haz_fc.intensity.todense()[ordered_select_mask],
+        )
+        npt.assert_array_equal(
+            haz_fc_sel.fraction.todense(),
+            haz_fc.fraction.todense()[ordered_select_mask],
+        )
+
+        assert haz_fc_sel.centroids == haz_fc.centroids
+
+    def test_derived_select(self, haz_fc, lead_time, member, haz_kwargs):
```
chahank: I would here also use the  @pytest.mark.parametrize . This makes the eventual error messages much clearer.

climada/hazard/test/test_forecast.py
```diff
@@ -107,46 +107,81 @@ def test_hazard_forecast_concat(haz_fc, lead_time, member):
...
+        )
+        npt.assert_array_equal(haz_fc_select.event_id, haz_fc.event_id[np.array([1])])
+
+        # Test "outer"
+        haz_fc2 = HazardForecast(
+            lead_time=lead_time, member=np.zeros_like(member, dtype="int"), **haz_kwargs
+        )
+        haz_fc_select = haz_fc2.select(event_id=[1, 2, 4], member=[0])
+        npt.assert_array_equal(haz_fc_select.event_id, [1, 2, 4])
+        npt.assert_array_equal(haz_fc_select.member, [0, 0, 0])
```
ValentinGebhart, line 184: @peanutfun Should we add a test for "no match"? 
Something like
`haz_fc_empty = haz_fc.select(event_names = haz_fc.event_name[:2], member=member[-2:])`
will raise
`IndexError: arrays used as indices must be of integer (or boolean) type`
that we could test for. Not sure if this is behaviour that we want, though, I would rather expect an empty HazardForecast. But it is the same behaviour as Hazard.select()

climada/hazard/forecast.py
ValentinGebhart: Agreed, I'll adapt the naming!

climada/hazard/test/test_forecast.py
ValentinGebhart: actually, I'm a bit confused about the parameterize. Above it seems to be for running the test multiple times with a different select variable. Here we are also testing on different combinations of select variables and their interactions, so I'm not sure how to make use of the parametrize here..

chahank: Another way is then to make one test/function for each combination.

climada/hazard/test/test_forecast.py
peanutfun: You are correct, there should be a test. But we should not go further than the functionality of `Hazard.select`. If that's the error from the base class, I'm fine with it being raised here, too.

climada/hazard/test/test_forecast.py
peanutfun: One function for separate leadtime/member selection, one function for the combined selections, one function for "emtpy" selections (as discussed below)

PR #1197, opened 2025-12-17
Risk Trajectory Split 1 : Snapshots

spjuhel:
As the Risk trajectory PR is too substantial, this is a first split.

This PR implements `Snapshots` as the foundational object for risk trajectories.

These objects regroup `Exposures`, `Hazard` and `ImpactFuncSet` instances for a specific `date`.

The rationale is to provide an immutable snapshot of risk. As such, `Exposures`, `Hazard`, `ImpactFuncSet` and `date` inputs are deep-copied, and "protected" behind read-only properties.

Likewise, to avoid any ambiguity regarding adaptation, i.e., as a user, if I provide an adaptation measure, should I give the base (`Exposures`, `Hazard`, `ImpactFuncSet`) triplet to which the measure is applied ? Or the triplet where the measure as already been applied, a measure is always None and can only be set through `apply_measure()` which returns a new object (with the measure set).

### PR Author Checklist

### PR Reviewer Checklist


climada/trajectories/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+This module implements risk trajectory objects which enable computation and
+possibly interpolation of risk metric over multiple dates.
+
+"""
+
+from .snapshot import Snapshot
+
+__all__ = [
+    "Snapshot",
+]
```
peanutfun, line 28: Excellent 👏  We will need this everywhere, eventually.

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+import pandas as pd
+
+from climada.entity.exposures import Exposures
+from climada.entity.impact_funcs import ImpactFuncSet
+from climada.entity.measures.base import Measure
+from climada.hazard import Hazard
+
+LOGGER = logging.getLogger(__name__)
+
+__all__ = ["Snapshot"]
```
peanutfun, line 41: Probably not necessary here, because the import of the __init__ should suffice?

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+        Returns
+        -------
+            The Snapshot with the measure applied.
+
+        """
+
+        LOGGER.debug(f"Applying measure {measure.name} on snapshot {id(self)}")
+        exp, impfset, haz = measure.apply(self.exposure, self.impfset, self.hazard)
+        snap = Snapshot(exposure=exp, hazard=haz, impfset=impfset, date=self.date)
+        snap._measure = measure
```
peanutfun: This should raise a linter error? You are accessing a private attribute of a different object. Measure should be passed via the init, see comment above.

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def __init__(
+        self,
+        *,
+        exposure: Exposures,
+        hazard: Hazard,
+        impfset: ImpactFuncSet,
+        date: int | datetime.date | str,
+        ref_only: bool = False,
+    ) -> None:
```
peanutfun, line 102: You need to be able to pass `measure` in the init, otherwise you have to modify a private attribute when applying a measure

climada/trajectories/__init__.py
```diff
@@ -0,0 +1,28 @@
+"""
```
peanutfun, line 1: You need to add this module to the docs

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+    ref_only : bool, default False
+        Should the `Snapshot` contain deep copies of the Exposures, Hazard and Impfset (False)
+        or references only (True).
+
+    Attributes
+    ----------
+    date : datetime
+        Date of the snapshot.
+    measure: Measure | None
+        The possible measure applied to the snapshot.
```
peanutfun: Classes don't have parameters, their methods do (the init in particular)

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @property
+    def measure(self) -> Measure | None:
+        """(Adaptation) Measure data for the snapshot."""
+        return self._measure
+
+    @property
+    def date(self) -> datetime.date:
+        """Date of the snapshot."""
+        return self._date
```
peanutfun, line 132: This is a lot of boilerplate code. Maybe just create a custom `__setattr__`, like outlined in option 4 here: https://www.sqlpey.com/python/top-5-methods-to-create-immutable-objects-in-python/

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Parameters
+    ----------
+    exposure : Exposures
+    hazard : Hazard
+    impfset : ImpactFuncSet
+    date : int | datetime.date | str
+        The date of the Snapshot, it can be an integer representing a year,
+        a datetime object or a string representation of a datetime object
+        with format "YYYY-MM-DD".
```
peanutfun: Using a date instead of a datetime might be overly restrictive. `datetime64` can encode each date and is compatible with pandas datetime index / timestamp

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+            # Try to parse the string as a date
+            try:
+                return datetime.datetime.strptime(date_arg, "%Y-%m-%d").date()
+            except ValueError:
+                raise ValueError("String must be in the format 'YYYY-MM-DD'")
+        elif isinstance(date_arg, datetime.date):
+            # Already a date object
+            return date_arg
+        else:
+            raise TypeError("date_arg must be an int, str, or datetime.date")
```
peanutfun: 1. Unnecessary elif/else after return statements
2. Use `isoformat`
```suggestion
        if isinstance(date_arg, str):
            # Try to parse the string as a date
            try:
                return datetime.date.fromisoformat(date_arg)
            except ValueError:
                raise ValueError("String must be in the format 'YYYY-MM-DD'")
        if isinstance(date_arg, datetime.date):
            # Already a date object
            return date_arg

        raise TypeError("date_arg must be an int, str, or datetime.date")
```

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+    ----------
+    date : datetime
+        Date of the snapshot.
+    measure: Measure | None
+        The possible measure applied to the snapshot.
+
+    Notes
+    -----
+
+    The object creates deep copies of the exposure hazard and impact function set.
```
peanutfun:
```suggestion
    If ``ref_only`` is True (default), the object creates deep copies of the
    exposure, hazard, and impact function set.
```

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+        """(Adaptation) Measure data for the snapshot."""
+        return self._measure
+
+    @property
+    def date(self) -> datetime.date:
+        """Date of the snapshot."""
+        return self._date
+
+    @property
+    def impact_calc_data(self) -> dict:
```
peanutfun: Make clear that this can be plugged in as kwargs?
```suggestion
    def impact_calc_kwargs(self) -> dict:
```

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+Snapshot are used to store a snapshot of Exposure, Hazard and Vulnerability
+at a specific date.
+
+"""
+
+import copy
+import datetime
+import logging
+
+import pandas as pd
```
peanutfun, line 32: Unused import

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+        measure : Measure
+            The measure to be applied to the snapshot.
+
+        Returns
+        -------
+            The Snapshot with the measure applied.
+
+        """
+
+        LOGGER.debug(f"Applying measure {measure.name} on snapshot {id(self)}")
```
peanutfun: Do not use format strings in logger messages, let logger format the string. See https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,132 @@
...
+        # We want a new reference
+        self.assertIsNot(snapshot.exposure, self.mock_exposure)
+        self.assertIsNot(snapshot.hazard, self.mock_hazard)
+        self.assertIsNot(snapshot.impfset, self.mock_impfset)
+
+        # But we want equality
+        pd.testing.assert_frame_equal(snapshot.exposure.gdf, self.mock_exposure.gdf)
+
+        self.assertEqual(snapshot.hazard.haz_type, self.mock_hazard.haz_type)
+        self.assertEqual(snapshot.hazard.intensity.nnz, self.mock_hazard.intensity.nnz)
```
peanutfun: Maybe check the matrix itself?

```suggestion
        np.testing.assert_array_equal(snapshot.hazard.intensity.toarray(), self.mock_hazard.intensity.toarray())
```

climada/trajectories/snapshot.py
peanutfun: Also: Does `ref_only` not apply for `measure`?

climada/trajectories/snapshot.py
```diff
@@ -0,0 +1,167 @@
...
+        exp, impfset, haz = measure.apply(self.exposure, self.impfset, self.hazard)
+        snap = Snapshot(exposure=exp, hazard=haz, impfset=impfset, date=self.date)
```
peanutfun: "inherit" `ref_only`?
```suggestion
        snap = Snapshot(exposure=exp, hazard=haz, impfset=impfset, date=self.date, ref_only=self.ref_only)
```

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,132 @@
...
+            impfset=self.mock_impfset,
+            date=2023,
+        )
+        new_snapshot = snapshot.apply_measure(self.mock_measure)
+
+        self.assertIsNotNone(new_snapshot.measure)
+        self.assertEqual(new_snapshot.measure.name, "Test Measure")  # type: ignore
+        self.assertEqual(new_snapshot.exposure, self.mock_modified_exposure)
+        self.assertEqual(new_snapshot.hazard, self.mock_modified_hazard)
+        self.assertEqual(new_snapshot.impfset, self.mock_modified_impfset)
```
peanutfun:
```suggestion
        self.assertIs(new_snapshot.exposure, self.mock_modified_exposure)
        self.assertIs(new_snapshot.hazard, self.mock_modified_hazard)
        self.assertIs(new_snapshot.impfset, self.mock_modified_impfset)
```

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,132 @@
...
+        snapshot = Snapshot(
+            exposure=self.mock_exposure,
+            hazard=self.mock_hazard,
+            impfset=self.mock_impfset,
+            date=date_obj,
+        )
+        self.assertEqual(snapshot.date, date_obj)
+
+    def test_init_with_invalid_date(self):
+        with self.assertRaises(ValueError):
```
peanutfun: Check for exact error message
```suggestion
        with self.assertRaisesRegex(ValueError, "String must be in the format"):
```

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,132 @@
...
+            Snapshot(
+                exposure=self.mock_exposure,
+                hazard=self.mock_hazard,
+                impfset=self.mock_impfset,
+                date="invalid-date",
+            )
+
+    def test_init_with_invalid_type(self):
+        with self.assertRaises(TypeError):
```
peanutfun: Check error message (not sure my raw string is escaped correctly)
```suggestion
        with self.assertRaisesRegex(TypeError, r"date\_arg must be an int, str, or datetime\.date"):
```

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+    def test_init_with_int_date(self):
+        snapshot = Snapshot(
+            exposure=self.mock_exposure,
+            hazard=self.mock_hazard,
+            impfset=self.mock_impfset,
+            date=2023,
+        )
+        self.assertEqual(snapshot.date, datetime.date(2023, 1, 1))
```
peanutfun: Is it really useful to support ints? This result could be a bit confusing

peanutfun (changes requested):
Looks good, but I somehow still have 19 comments 🙈 The linter pipeline does not seem to work, and some comments are related to would-be linter issues.

As discussed, feel free to use pytest when writing new test modules.

We also briefly discussed writing this as a frozen dataclass, but I think the `ref_only` parameter would make the initialization for a dataclass too cumbersome. You can leave it as is, but there might be a way to remove the property definitions (see comment).

climada/trajectories/snapshot.py
spjuhel: I agree it is not required, but I thought it was good practice to formalize what is in __all__ for each module?

climada/trajectories/snapshot.py
spjuhel: I thought about it, after the new factory method, 

The new triplet comes from applying the measure to the original one, so the `Exposure`, `Hazard`, `Impfset` objects are new anyway.
So here we actually want references always (or else we would copy objects not referenced elsewhere)

climada/trajectories/test/test_snapshot.py
spjuhel: Well, the default thing people will use are years, so I would say yes.

peanutfun: But string years are compatible with pandas and numpy:
```python
>>> pandas.Timestamp("2022")
pandas.Timestamp('2022-01-01 00:00:00')
>>> numpy.datetime64("2022")
numpy.datetime64('2022')
>>> numpy.datetime64("2022", "D")  # Force interval 'day'
numpy.datetime64('2022-01-01')
```

climada/trajectories/snapshot.py
peanutfun: I see, then a copy is always unnecessary.

climada/trajectories/snapshot.py
peanutfun: Correct, I guess it does no harm 🙃

climada/trajectories/test/test_snapshot.py
```diff
@@ -0,0 +1,172 @@
...
+            date="2001",
+        )
+
+
+@pytest.mark.parametrize(
+    "input_date,expected",
+    [
+        ("2023", pd.Timestamp(2023, 1, 1)),
+        ("2023-01-01", pd.Timestamp(2023, 1, 1)),
+        (datetime.date(2023, 1, 1), pd.Timestamp(2023, 1, 1)),
```
peanutfun, line 73: Add cases for numpy.datetime64 and pd.Timestamp as input

peanutfun (changes requested):
From personal discussion:
* Remove from_triplet. Only use init and add warning or other instructions to docs.
* Clarify in parameter docs: measure added to init is *not* applied onto the other parameters. Needs to be applied before.
* Urge users to use apply_measure
* Small test adaptation, see comment

spjuhel:
@peanutfun Tell me if the docstring is clear enough.

I think I resolved all other comments.

PR #1199, opened 2025-12-17
Risk Trajectory Split 3 : Impact Computation Strategies

spjuhel:
As the Risk trajectory PR is too substantial, this is a third split.

This PR implements the concept of impact computation strategies.
An impact computation strategy defines what underlying process to follow to get from (exp,haz,vul) triplet to an impact.

The idea is to ease the extension of the trajectories to also allow non `ImpactCalc` based impacts (think network, unsequa, etc.)

Currently, it really is just a wrapper around ImpactCalc().

It has proved helpful in tests/prototypes of risk trajectory / option appraisal to, for instance:
- "custom" compute impacts on polygons / lines 
- pre-assign centroid or not

### PR Author Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] Correct target branch selected (if unsure, select `develop`)
- [ ] Descriptive pull request title added
- [ ] Source branch up-to-date with target branch
- [ ] [Documentation](https://climada-python.readthedocs.io/en/latest/guide/Guide_PythonDos-n-Donts.html#2.--Commenting-&-Documenting) updated
- [ ] [Tests][testing] updated
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]
- [ ] [Changelog](https://github.com/CLIMADA-project/climada_python/blob/main/CHANGELOG.md) updated

### PR Reviewer Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/development/Guide_Review.html) passed
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]

[contrib]: https://github.com/CLIMADA-project/climada_python/blob/main/CONTRIBUTING.md
[testing]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html
[linter]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html#static-code-analysis

Issue #1081, opened 2025-08-05
`SALib.analyze.rbd_fast` shows machine depending behavior


emanuel-schmid:
On two different Machines with exactly the same virtual environment the results from that method differ quite significantly.

**Steps to reproduce**

```bash
python salib_demo.py
```
(with salib_demo.py as in [salib_demo.py.txt](https://github.com/user-attachments/files/21603735/salib_demo.py.txt))

has one of the following outputs, each one consistent on the respective machine:

Machine 1 (my computer, apparently GitHub action runners):

```
{'S1': [0.1089541115226481, 0.7049211232268789, 0.015075727433210298, 0.07117904961243285], 'S1_conf': [0.15260859689652703, 0.07956660882184943, 0.07877370534246698, 0.13270860395201967], 'names': ['x_exp', 'x_paa', 'x_mdd', 'x_haz']}
```

Machine 2 (jenkins server):

```
{'S1': [0.12359332237072101, 0.7077530937494074, -0.009256312995937502, 0.05828787358778384], 'S1_conf': [0.14814694067816694, 0.08041428188470545, 0.08202268215002247, 0.12927511001940797], 'names': ['x_exp', 'x_paa', 'x_mdd', 'x_haz']}
```

On both machines the conda packages are exactly the same (see [env-spec.txt](https://github.com/user-attachments/files/21603637/env-spec.txt)).

The consequence of this discrepancy is that the unit test `climada.engine.unsequa.test.test_unsequa.TestCalcImpact.test_calc_sensitivity_all_pass` started to fail around July 24, 2025.

It's suspicious that the problem started on Machine 2 with a whole bunch of other issues. Possibly a system update went bad. However - everything else runs fine again.

chahank:
That is very strange. Can it be that the random number sequence is different?

emanuel-schmid:
This problem got worse:
After upgrading from salib 1.5.1 to salib 1.5.2, the expected and actual results in `climada.engine.unsequa.test.test_unsequa.TestCalcImpact.test_calc_sensitivity_all_pass` has significant differences throughout _all_ `salib` methods, some of them dramatic (~ 20%).

We need to further analyze whether the new results are still sound.
If they are, we can adapt the test. Otherwise we have to pin the salib version.

luseverin:
[...]

PR #1200, opened 2025-12-18
 Risk Trajectory Split 4 : Non-interpolated (Static) trajectories

spjuhel:
As the Risk trajectory PR is too substantial, this is a fourth split.

This PR builds on PRs #1197 and #1199, introduces `RiskTrajectory`, the abstract basis for trajectory objects, and the concrete class `StaticRiskTrajectory` which aims at facilitating the computation of risk metrics over multiple `Snapshot` objects.

The code is split between a high level interface `StaticRiskTrajectory` and a low level computing class `CalcRiskMetricsPoint`.

`RiskTrajectory`
=================

The rationale behind this class is to hold the common code for any type of risk trajectory:
- a collection of Snapshot
- an optional risk discount rate, and associated methods to compute NPV.
- (abstract) access to risk metrics

`CalcRiskMetricPoint`
=================

The rationale behind this class is to handle the computation and "data formatting" of different risk metrics over a collection of `Snapshot`, following an `ImpactComputationStrategy` (see PR #1199).  In this case, without any interpolation, thus it is essentially a wrapper for multiple calls to `ImpactCalc` or equivalent in `ImpactComputationStrategy`.

The class contains:
- `@lazy_property` properties, dictating how the foundational elements for risk metric should be computed (see below)
- `calc_<risk_metric>` methods (which are the ones called by the `StaticRiskTrajectory` object): these make use of the above and put them in nice and tidy dataframes.

`@lazy_property` explained
----------------------------------------

We want to avoid recomputing when unnecessary, thus we have computed metrics and "impact data" as `@lazy_property`, which cache the results of the computation. For instance, calling `static_traj.per_date_aai` the first time computes the different required `Impact` objects, stores them, and then the `.aai_agg` for each `Impact` and also store them.

A subsequent call to `.per_date_aai` return the previously stored value.
In addition, a subsequent call to `.per_date_eai` will reuse computed `Impact` objects. 

The `@lazy_property` decorator implements this memoization flow: it returns the cached `_metric` attribute if available. If not, it computes the value, stores it in `_metric`, and returns the result.

Attributes that would change metric results (such as the computation strategy or the return periods to compute) make use of `@property` to reset the cached data when their value is changed by calling `_reset_impact_data`.

`StaticRiskTrajectory`
=================

This class is the interface with the user. It receives the list of snapshot to compute metrics for, and gives access to computed risk metrics and plots. It also handles the computation of net present values if a risk discount rate is provided.
It also uses a memoization flow for the "final" results.

### PR Author Checklist

### PR Reviewer Checklist


climada/test/common_test_fixtures.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def reusable_intensity_mat(max_intensity=HAZARD_MAX_INTENSITY):
+    # Choosen such that:
+    # - 1st event has 0 intensity
+    # - 2nd event has max intensity in first exposure point (defaulting to 0 value)
+    # - 3rd event has 1/2* of max intensity in second centroid
+    # - 4th event has 1/4* of max intensity everywhere
+    # *: So that you can double intensity of the hazard and expect double impacts
+    return csr_matrix(
```
chahank, line 132: What would happen with negative intensities? Probably easy to test using a negative intensity factor. Although then the reusable impact function set would need adjustments.

climada/test/common_test_fixtures.py
```diff
@@ -0,0 +1,205 @@
...
+        [
+            [0, 0, 0, 0],
+            [max_intensity, 0, 0, 0],
+            [0, max_intensity / 2, 0, 0],
+            [
+                max_intensity / 4,
+                max_intensity / 4,
+                max_intensity / 4,
+                max_intensity / 4,
+            ],
```
chahank, line 142: Formatting.

climada/test/test_trajectories.py
```diff
@@ -0,0 +1,278 @@
...
+            date=self.FUTURE_DATE,
+        )
+
+        self.expected_base_imp = ImpactCalc(
+            **self.base_snapshot.impact_calc_data
+        ).impact()
+        self.expected_future_imp = ImpactCalc(
+            **self.future_snapshot.impact_calc_data
+        ).impact()
+        # self.group_vector = self.base_snapshot.exposure.gdf[GROUP_ID_COL_NAME]
```
chahank:
```suggestion
```

climada/trajectories/test/test_calc_risk_metrics.py
```diff
@@ -0,0 +1,346 @@
...
+                date=str(present_date),
+            ),
+            Snapshot(
+                exposure=exp_future,
+                hazard=haz_future,
+                impfset=impfset_future,
+                date=str(future_date),
+            ),
+        ],
+        "expected_eai": np.array(
```
chahank, line 69: Not sure whether this is a good idea to have so many meaningless digits after the comma for test that cannot be verified to be correct.

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+LOGGER = logging.getLogger(__name__)
+
+__all__ = [
+    "CalcRiskMetricsPoints",
+]
+
+_CACHE_SETTINGS = {"ENABLE_LAZY_CACHE": False}
+
+
+def lazy_property(method):
```
chahank, line 62: Does this maybe belong into the utils instead?

chahank: Also, is there a reason to not use [@functools.cached_property(func)](https://docs.python.org/3/library/functools.html#functools.cached_property)

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+        ----------
+        snapshots : List[Snapshot]
+            The `Snapshot` list to compute risk for.
+        impact_computation_strategy: ImpactComputationStrategy, optional
+            The method used to calculate the impact from the (Haz,Exp,Vul) of the two snapshots.
+            Defaults to ImpactCalc
+
+        """
+
+        self._reset_impact_data()
```
chahank: Why is this needed in the __init__ ?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+                        snap.exposure.gdf[GROUP_ID_COL_NAME]
+                        for snap in self.snapshots
+                        if GROUP_ID_COL_NAME in snap.exposure.gdf.columns
+                    ]
+                )
+            )
+        except ValueError as exc:
+            error_message = str(exc).lower()
+            if "need at least one array to concatenate" in error_message:
+                self._group_id = np.array([])
```
chahank, line 164: This seems to not generate actionable error messages. How should the user know what to do after getting this error?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+        """The method used to calculate the impact from the (Haz,Exp,Vul)
+        of the snapshots.
+
+        """
+        return self._impact_computation_strategy
+
+    @impact_computation_strategy.setter
+    def impact_computation_strategy(self, value, /):
+        if not isinstance(value, ImpactComputationStrategy):
+            raise ValueError("Not an impact computation strategy")
```
chahank: Not a clear message for the user. Please write error messages that allow the user to solve their problem.

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+        rp_df[MEASURE_COL_NAME] = (
+            self.measure.name if self.measure else NO_MEASURE_VALUE
+        )
+        rp_df[UNIT_COL_NAME] = self.snapshots[0].exposure.value_unit
+        return rp_df
+
+    def apply_measure(self, measure: Measure) -> "CalcRiskMetricsPoints":
+        """Creates a new `CalcRiskMetricsPoints` object with a measure.
+
+        The given measure is applied to both snapshot of the risk period.
```
chahank: That is quite unclear.

climada/trajectories/calc_risk_metrics.py
```diff
@@ -0,0 +1,358 @@
...
+        Parameters
+        ----------
+        measure : Measure
+            The measure to apply.
+
+        Returns
+        -------
+
+        CalcRiskPeriod
+            The risk period with given measure applied.
```
chahank: Do you mean a `CalcRiskMetricsPoints` ?

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+class StaticRiskTrajectory(RiskTrajectory):
+    """This class implements static risk trajectories, objects that
+    regroup impacts computations for multiple dates.
+
+    This class computes risk metrics over a series of snapshots,
+    optionally applying risk discounting. It does not interpolate risk
+    between the snapshot and only provides results for each snapshot.
+
```
chahank: This could be made clearer.

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+
+    POSSIBLE_METRICS = [
+        EAI_METRIC_NAME,
+        AAI_METRIC_NAME,
+        RETURN_PERIOD_METRIC_NAME,
+        AAI_PER_GROUP_METRIC_NAME,
+    ]
+    """Class variable listing the risk metrics that can be computed.
```
chahank, line 81: Arent't the full impact matrices computed and accessible?

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+            The list of `Snapshot` object to compute risk from.
+        return_periods: list[int], optional
+            The return periods to use when computing the `return_periods_metric`.
+            Defaults to `DEFAULT_RP` ([20, 50, 100]).
+        risk_disc_rates: DiscRates, optional
+            The discount rate to apply to future risk. Defaults to None.
+        impact_computation_strategy: ImpactComputationStrategy, optional
+            The method used to calculate the impact from the (Haz,Exp,Vul)
+            of the two snapshots. Defaults to :class:`ImpactCalcComputation`.
+
```
chahank, line 123: What is meant by the "two snapshots". The input is a list of snapshots.

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+    @property
+    def impact_computation_strategy(self) -> ImpactComputationStrategy:
+        """The approach or strategy used to calculate the impact from the snapshots."""
+        return self._risk_metrics_calculators.impact_computation_strategy
+
+    @impact_computation_strategy.setter
+    def impact_computation_strategy(self, value, /):
+        if not isinstance(value, ImpactComputationStrategy):
+            raise ValueError("Not an interpolation strategy")
```
chahank: Unclear error message for the user. This is likely raised at the init of the object.

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+
+        tmp = tmp.set_index(
+            [DATE_COL_NAME, GROUP_COL_NAME, MEASURE_COL_NAME, METRIC_COL_NAME]
+        )
+        if COORD_ID_COL_NAME in tmp.columns:
+            tmp = tmp.set_index([COORD_ID_COL_NAME], append=True)
+
+        # When more than 2 snapshots, there might be duplicated rows, we need to remove them.
+        # Should not be the case in static trajectory, but in any case we really don't want
+        # duplicated rows, which would mess up some dataframe manipulation down the road.
```
chahank, line 230: Should this not be made clear to the user in case it happens? Since some data is arbitrarily discarded.

climada/trajectories/static_trajectory.py
```diff
@@ -0,0 +1,315 @@
...
+    def eai_metrics(self, **kwargs) -> pd.DataFrame:
+        """Return the estimated annual impacts at each exposure point for each date.
+
+        This method computes and return a `DataFrame` with eai metric
+        (for each exposure point) for each date.
+
+        Notes
+        -----
+
+        This computation may become quite expensive for big areas with high resolution.
```
chahank:
```suggestion
        This computation may become quite expensive for exposures with many points (e.g., big areas with high resolution).
```

climada/trajectories/trajectory.py
```diff
@@ -0,0 +1,261 @@
...
+        -----
+
+        Changing its value resets all the metrics.
+        """
+        return self._risk_disc_rates
+
+    @risk_disc_rates.setter
+    def risk_disc_rates(self, value, /):
+        if value is not None and not isinstance(value, (DiscRates)):
+            raise ValueError("Risk discount needs to be a `DiscRates` object.")
```
chahank: What is Risk discount?

climada/util/dataframe_handling.py
```diff
@@ -0,0 +1,63 @@
...
+
+---
+
+Define functions to handle with coordinates
+"""
+
+import pandas as pd
+
+
+def reorder_dataframe_columns(
```
chahank, line 25: A simple test would be useful as this has the potential to screw up outputs if pandas changes the way columns are indexed / selected.

climada/test/common_test_fixtures.py
spjuhel: ?
Black does not complain with me?

climada/test/common_test_fixtures.py
spjuhel: I think that is beyond the scope of this PR.

climada/trajectories/test/test_calc_risk_metrics.py
spjuhel: What do you suggest?

climada/trajectories/calc_risk_metrics.py
spjuhel: Possibly, I didn't know/find this feature (and wanted to avoid lru_cache).
I'll have a look.

climada/trajectories/calc_risk_metrics.py
spjuhel: Otherwise, "_impacts", "_eai_gdf", ... are not existing attribute member.

We could just set this attribute to None, but then one needs to carefully update `_reset_impact_data` if changes are made to the computed data attributes.

For clarity, I changed the method's name to `_init_impact_data`, and have `_reset_impact_data` as an alias.

climada/trajectories/calc_risk_metrics.py
spjuhel: This **catches** the error caused by the absence of a group_id column in all snapshots and acts accordingly (defining the group_id as an empty array). No error are raised here. 
We check the content of the error to only handle that case.

I did forget to raise if this is not the case. Will fix this.

climada/trajectories/calc_risk_metrics.py
spjuhel: Legacy docstring from before, corrected

climada/trajectories/static_trajectory.py
spjuhel: How/What is unclear?

climada/test/common_test_fixtures.py
chahank: Yes, that is why it is not readable. One needs to turn of Black for good formatting :D

climada/trajectories/static_trajectory.py
spjuhel: Not at that high-level class.
This class's purpose is to offer access to EAI and all its aggregations in tidy DataFrames, following a consistent framework.

I'm not sure, I see the point of giving access to impact matrices here.
Users can use `ImpactCalc` / `Snapshot` directly if they want that level of information.

climada/trajectories/static_trajectory.py
spjuhel: Legacy description from when there was only interpolated trajectories (and pairs of snapshots)
Corrected.

climada/trajectories/static_trajectory.py
spjuhel: Improved

climada/trajectories/static_trajectory.py
spjuhel: I added a note in the docstring and a warning.

climada/trajectories/trajectory.py
spjuhel: Changed to "The discount rate applied to risk values"

climada/util/dataframe_handling.py
spjuhel: Test added

climada/trajectories/calc_risk_metrics.py
spjuhel: So toggling caching makes the decorator as complex if not more with `cached_property`. I'd avoid making the change.

climada/test/test_trajectories.py
spjuhel: Fixed

climada/trajectories/calc_risk_metrics.py
spjuhel: Fixed

climada/trajectories/calc_risk_metrics.py
spjuhel: Hopefully fixed.

climada/trajectories/calc_risk_metrics.py
spjuhel: Does this clarify:

```
        """Creates a new `CalcRiskMetricsPoints` object by applying the effects
        of the given measure.

        The effects of the measure are applied to all the snapshots contained
        in the initial `CalcRiskMetricsPoints` and a new `CalcRiskMetricsPoints`
        containing the modified snapshots is returned.

        Parameters
        ----------
        measure : Measure
            The measure to apply.

        Returns
        -------

        CalcRiskMetricsPoints
            The risk period with given measure applied.

        """
```

climada/trajectories/static_trajectory.py
spjuhel: Suggestion implemented

climada/trajectories/test/test_calc_risk_metrics.py
chahank: Can one get results that are reproducible / tractable?

climada/test/common_test_fixtures.py
chahank: hmmm when could this be done? Because negative intensities are valid in CLIMADA, and it would be quite problematic to break this without good reason.

climada/trajectories/calc_risk_metrics.py
chahank: Good for me.

climada/trajectories/static_trajectory.py
chahank: I see. But this matters for the non-static trajectories right? Because interpolation of impact matrics is not accessible with ImpactCacl/Snapshot.

climada/test/common_test_fixtures.py
spjuhel: I'll have a look.

climada/test/common_test_fixtures.py
spjuhel: Ok! I disagree with it being not readable, but I guess that is subjective.

Anyway that file disappears in split 5 (it is actually not the proper way to use common fixture). I'll improve it there.

climada/trajectories/test/test_calc_risk_metrics.py
spjuhel: I've now used test data similar to the one in the integration test. 

Notes: I have added a `conftest.py` that contains the common fixtures for `pytest` (they are not to be imported, pytest knows to use the names defined in `conftest.py`)

climada/trajectories/static_trajectory.py
spjuhel: Yes? But again, the idea of this module is not really to give access to interpolated risk matrices (because we can't do that for all dimensions of risk, only for exposure really)

The `CalcRiskMetricsPeriod` object defined in split 5 takes care of the interpolation process, and does hold the various arrays of interpolated impact matrices between the pair of snapshots it is defined for, arrays needed to compute the risk metrics. As well as generic methods/functions to compute these arrays.

Again, you cannot have interpolated impact matrices between two snapshots, only between some of the edges of the risk cubes (the ones that share the same hazard), and I would rather not make a direct interface to that.

climada/test/common_test_fixtures.py
spjuhel: OK, sorry I got confused actually, my initial comment on "this is beyond this PR" applies.

This PR has absolutely no impact on how negative intensities are handled by CLIMADA, it will work the same as before, we did not modify anything related to impact computation here, everything is just a wrapper.

I do have to check if interpolation cause problems with negative value (but that's split 5 not this one).

What also might change something is the option appraisal refactoring (next big project in line) which does change the `measure` module, and hopefully fixes the problem mentioned by Chris in #1257.

spjuhel:
@chahank 
I am pretty sure I addressed all comments, see the one on negative intensities and resolve if you are OK.

The failing tests are related to #1263 #1264 (and #1081) and are being addressed, and not related to this PR.

I'll merge once you greenlight me :)

PR #1198, opened 2025-12-17
Risk Trajectory Split 2 : Interpolation Strategies

spjuhel:
As the Risk trajectory PR is too substantial, this is a second split.

This PR implements the concept of interpolation strategies to journey along the Risk Cube.

It:
- defines functions for linear and exponential interpolation for both sparse matrices and numpy arrays of metrics
- Abstract class for interpolation strategies, which dictate how to interpolate along the different dimensions of the cube.
- Two concrete classes, one with all linear and one with exponential interpolation for exposure.
 
A strategy is a triplet of callable:
- `exposure_interp`: "how do I generate a range of _given number of dates_ impact matrices, between present date and future date, accounting for changing exposure but constant hazard and vulnerability?"
- `hazard_interp`: "how do I interpolate between two arrays of risk metrics, the first one corresponding to present hazard and evolving exposure, and the second one to future hazard and evolving exposure, to _incorporate_ the evolution due to the changing hazard"
- `vulnerability_interp`: same as `hazard_interp` but for vulnerability.

### PR Author Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] Correct target branch selected (if unsure, select `develop`)
- [ ] Descriptive pull request title added
- [ ] Source branch up-to-date with target branch
- [ ] [Documentation](https://climada-python.readthedocs.io/en/latest/guide/Guide_PythonDos-n-Donts.html#2.--Commenting-&-Documenting) updated
- [ ] [Tests][testing] updated
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]
- [ ] [Changelog](https://github.com/CLIMADA-project/climada_python/blob/main/CHANGELOG.md) updated

### PR Reviewer Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/development/Guide_Review.html) passed
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]

[contrib]: https://github.com/CLIMADA-project/climada_python/blob/main/CONTRIBUTING.md
[testing]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html
[linter]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html#static-code-analysis

PR #1201, opened 2025-12-18
 Risk Trajectory Split 5: Interpolated Risk Trajectories

spjuhel:
As the Risk trajectory PR is too substantial, this is the fifth (and final) split.

This PR builds on all previous splits (#1197 #1198 #1199 #1200).

It implements the high level `InterpolatedRiskTrajectory` class and the low level computing class `CalcRiskMetricsPeriod`, similarly as Split 4, but with interpolation in between pairs of consecutive `Snapshot`. The additional complexity and the differences with Split 4 are summarized below.

It also ships the tutorial for the `trajectory` module.

Core differences
==============

Instead of having one `CalcRiskMetricsPoints` for all snapshot, the high level class holds a list of `CalcRiskMetricsPeriod` objects (one for each pair of consecutive snapshots), each having the role of computing the metrics for the period defined by the two snapshots.

`CalcRiskMetricsPeriod` thus takes a starting (0) and ending (1) `Snapshot` instead of a list of them.
It also takes an `InterpolationStrategy` in addition to the `ImpactComputationStrategy`, and a "time resolution", which defines the number of time points to have for the metrics (including start and end).

Interpolation conceptual flow (in `calc_risk_metrics`)
----------------------------------------------------------------------------

Instead of "simply" computing the impacts for each snapshot, it computes all corners of the so-called Risk Cube™ (all 2³ convolutions of (exp,haz,vul) Impact triplet), and the interpolated risk metrics based on these.

The eight corners are `@lazy_property`, and labelled `EiHjVk` (where i,j,k are 0 or 1), such that for instance `E0H0V0` is the impact object corresponding to Exposure, Hazard and Vulnerability of Snapshot 0.

From then, arrays of impact matrices are interpolated between corners for as many points as needed by the given time resolution. (i.e. if start date is 2020, end date is 2025 and resolution is yearly, arrays of 5 impact matrices will be computed).

As it is not feasible (in general) to interpolate between corners with a different hazard (shapes of impact matrices may differ), this interpolation of impact matrices is only done "along the exposure dimension", that is for the four edges of the Risk Cube™ where hazard and vulnerability remain constant along the edge. (side note here, it should actually also be possible to do that for vulnerability, will see for future updates)

These four arrays of matrices are labelled `imp_mats_HiVj`, to be read as "Evolution of impacts matrices with hazard constant to snapshot i, vulnerability constant to snapshot j, and exposure changing from snapshot 0 to 1".

From then on, interpolation is made on the metrics themselves and not the matrices. 
We compute both the EAIs for each of the four arrays (`per_date_eai_HiVj`), from which the corresponding AAIs can be computed (`per_date_aai_HiVj`), and the "return period impacts".

Then we can compute the "fully interpolated metrics basis", `per_date_eai`, `per_date_aai` (return periods have a special treatment given they rely on an argument), and cache them using `@lazy_property`. Computation is done by interpolating first along the hazard dimension and then along the vulnerability one.

Finally, we can have the `calc_<risk_metric>` methods which take care of tidying the data in nice dataframes.

Differences in interface class
------------------------------------------

`InterpolatedRiskTrajectory` handle the additional "time resolution" attribute, ships `per_period_<risk_metric>` methods in addition to the `per_date` ones, where average values of the metric for each period (consecutive pairs of snapshots) are computed.

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Just as a reminder for later: 
- I would group all the adaptation option appraisal stuff in a separate entry in the documentation toc (instead of it being under impact).
- The trajectories and other tutorial should ideally have a quick and dirty example at the very top (and not after a lenghty wall of text and introduction to all the components). This makes it much easier to enter in the topic.

spjuhel:
@chahank 

- The adaptation option appraisal stuff will be in a different set of PR, indeed, they'll have their own section.
- I've updated the tuto with a quickstart, (it is minimalistic though)

chahank:
It still indicates many files as new or as modified which I think were addressed in the pull request 1-4. Can you please check why (it makes it very hard to review the relevant changes made in this PR alone)? Maybe another develop merge is needed?

spjuhel:
I think these relate to #1254 (I thought I would have merged by then)

spjuhel:
You can review these files:
- climada/trajectories/interpolated_trajectory.py
- climada/trajectories/calc_risk_metrics.py
- doc/user-guide/climada_trajectories.ipynb
- climada/trajectories/test/*

chahank:
> You can review these files:
> 
>     * climada/trajectories/interpolated_trajectory.py
> 
>     * climada/trajectories/calc_risk_metrics.py
> 
>     * doc/user-guide/climada_trajectories.ipynb
> 
>     * climada/trajectories/test/*

Excellent, thanks for the guidance!

chahank:
I have so far looked at the tutorial. Great job, it is very comprehensive, and the quick intro at the beginning is very useful. I would suggest the following small improvements.

Quick example:
- (minor point) Could one replace the  
```
exp_future = Exposures.from_hdf5(EXP_DEMO_H5)
exp_future.gdf["value"] *= 1.1
```
with a direct future exposure file? I recall a few discussion with users that seem to believe that scaling the value is the only way to get future exposures, and I think this is because of such examples in our documentation.
- Can you add the `interpolated_risk_traj.plot_waterfall()`, otherwise it is a bit confusing to see only relative impact changes.

Others:
- Can you explain when one would use `StaticRiskTrajectory` and when the `InterpolatedRiskTrajectory`? Now it is said what the technical difference is, but not what the use-case difference is.
- Overall, it is a bit verbose a points for my taste. Sentences like `As always users should carefully consider if the tool fits the purpose and if the limitations remain acceptable, even more so when used to design Disaster Risk Reduction or Climate Change Adaptation measures.` I think do not add much to a tutorial, but make it longer and thus decrease the probability of users reading it all.
- I like the distinction between simple and advanced uses. Could you make a TOC at the beginning to make navigating and finding the write information easier?

spjuhel:
Is this not enough?
image

I'm very reluctant to add custom-made ToC...

chahank:
> Is this not enough? 
> 
> I'm very reluctant to add custom-made ToC...

I agree that adding a custom made TOC is not ideal. Can you adjust the current headings such that the subsections in the advance usage are shown? I think these need to be easily accessible without having to scroll through the whole thing.

chahank:
Thanks for the updates already! Are you @spjuhel already satisfied with the changes in the tutorial or is it planned to implement more of the suggestions above?

And a little note, for the quick example, can one also replace the future hazard by a file instead of the multiplication of the frequency?

I will come to the content afterwards.

spjuhel:
Not entirely,

The ToC thing I want to have a look, I am pretty sure making deeper title visible will be messy (wrt the rest of documentation), but I think I can adjust that by just removing the "Using the `trajectory` module title".

For the `Hazard`, the thing is we don't have a future hazard demo file (Or I haven't found it).

I think I addressed everything else?

chahank:
> Not entirely,
> 
> The ToC thing I want to have a look, I am pretty sure making deeper title visible will be messy (wrt the rest of documentation), but I think I can adjust that by just removing the "Using the `trajectory` module title".

Yes, I think this is a good solution (the title is anyway a bit redundant for a tutorial). I would also remove the 

> 
> For the `Hazard`, the thing is we don't have a future hazard demo file (Or I haven't found it).

Then we need to make one ;).

> 
> I think I addressed everything else?

Yes thanks! Just maybe I missed it, but I do not see "when one would use StaticRiskTrajectory and when the InterpolatedRiskTrajectory?".

climada/trajectories/calc_risk_metrics.py
```diff
@@ -26,17 +26,29 @@
...
 import pandas as pd
+from scipy.sparse import csr_matrix
 
-from climada.engine.impact import Impact
+from climada.engine.impact import Impact, ImpactFreqCurve
+from climada.engine.impact_calc import ImpactCalc
 from climada.entity.measures.base import Measure
 from climada.trajectories.constants import (
     AAI_METRIC_NAME,
+    CONTRIBUTION_BASE_RISK_NAME,
```
chahank, line 43: Do you think that these should be configurable? If yes, they probably would fit better in one of the climada config files. I doubt whether many are using these config files for custom uses at the moment, so for sure this is not necessary.

spjuhel:
> Yes thanks! Just maybe I missed it, but I do not see "when one would use StaticRiskTrajectory and when the InterpolatedRiskTrajectory?".

https://climada-python--1201.org.readthedocs.build/en/1201/user-guide/climada_trajectories.html#when-to-use-static-vs-interpolated-trajectories

Will work ASAP on your other comments.

climada/trajectories/calc_risk_metrics.py
spjuhel: I think it would be terrible if they were hard-coded as they appear in a lot of the code (and also in the Appraisal objects in future PRs) so it is paramount that they are defined as constants.

I also doubt it is relevant to have it in the config files.

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,920 @@
...
+            of the two snapshots. Defaults to :class:`ImpactCalcComputation`.
+
+        """
+        super().__init__(
+            snapshots_list,
+            return_periods=return_periods,
+            risk_disc_rates=risk_disc_rates,
+        )
+        self.start_date = min((snapshot.date for snapshot in snapshots_list))
+        self.end_date = max((snapshot.date for snapshot in snapshots_list))
```
chahank:
```suggestion
        self.start_date = min(snapshot.date for snapshot in self._snapshots)
        self.end_date = max(snapshot.date for snapshot in self._snapshots)
```

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,920 @@
...
+        """Calculate the difference between consecutive dates."""
+
+        if time_unit == "year":
+            group["date_diff"] = group[DATE_COL_NAME].dt.year.diff()
+        if time_unit == "month":
+            group["date_diff"] = group[DATE_COL_NAME].dt.month.diff()
+        if time_unit == "day":
+            group["date_diff"] = group[DATE_COL_NAME].dt.day.diff()
+        if time_unit == "hour":
+            group["date_diff"] = group[DATE_COL_NAME].dt.hour.diff()
```
chahank: And if `time_unit` is anything else? Does an error need to be raised?

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,920 @@
...
+                labels[i],  # type: ignore
+                val + bottoms[i],
+                f"{val:.0e}",
+                ha="center",
+                va="bottom",
+                color="black",
+            )
+
+        # Construct y-axis label and title based on parameters
+        value_label = "USD"
```
chahank: Should the unit not come from the data instead of being hard coded?

climada/trajectories/interpolated_trajectory.py
```diff
@@ -0,0 +1,920 @@
...
+    ) -> pd.DataFrame:
+        """Return a tidy dataframe of the risk metrics with the total
+        for each different period (pair of snapshots).
+
+        """
+
+        metric_df = self.per_date_risk_metrics(metrics=metrics, **kwargs)
+        return self._date_to_period_agg(
+            metric_df, grouper=self._grouper + [UNIT_COL_NAME], **kwargs
+        )
```
chahank: The same kwargs are passed to two different methods. This is not very clear. Also, the docstring does not explain this.

climada/trajectories/test/test_interpolated_risk_trajectory.py
```diff
@@ -0,0 +1,1419 @@
...
+        pd.testing.assert_frame_equal(
+            result, self.expected_npv_aai
+        )  # Final result is from NPV transform
+
+        # Check internal storage
+        stored_df = getattr(rt, "_aai_metrics")
+        # Assert that the stored DF is the one *before* NPV transformation
+        pd.testing.assert_frame_equal(
+            stored_df.reset_index(drop=True),
+            self.expected_npv_aai.reset_index(drop=True),
```
chahank: Is this now before or after NPV? The variable is now expect_npv_aai...

climada/trajectories/test/test_interpolated_risk_trajectory.py
```diff
@@ -0,0 +1,1419 @@
...
+
+        self.mock_snapshot3 = MagicMock(spec=Snapshot)
+        self.mock_snapshot3.date = datetime.date(2025, 1, 1)
+
+        self.snapshots_list: list[Snapshot] = [
+            self.mock_snapshot1,
+            self.mock_snapshot2,
+            self.mock_snapshot3,
+        ]
+        # self.snapshots_list = cast(list[Snapshot], self.snapshots_list)
```
chahank: Relevant?

climada/trajectories/test/test_interpolated_risk_trajectory.py
```diff
@@ -0,0 +1,1419 @@
...
+        rt = InterpolatedRiskTrajectory(self.snapshots_list)
+        with self.assertRaises(ValueError):
+            rt._generic_metrics(metric_name=None, metric_meth="some_method")
+        with self.assertRaises(ValueError):
+            rt._generic_metrics(metric_name=AAI_METRIC_NAME, metric_meth=None)
+
+    @patch.object(
+        InterpolatedRiskTrajectory, "_reset_risk_metrics_calculators", new_callable=Mock
+    )
+    # @patch.object(InterpolatedRiskTrajectory, "npv_transform", new_callable=Mock)
```
chahank: Relevant?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -365,3 +387,810 @@ def apply_measure(self, measure: Measure) -> "CalcRiskMetricsPoints":
...
+
+        This initializes and instantiate a new `CalcRiskMetricsPeriod` object.
+        No computation is done at initialisation and only done "just in time".
+
+        Parameters
+        ----------
+        snapshot0 : Snapshot
+            The `Snapshot` at the start of the risk period.
+        snapshot1 : Snapshot
+            The `Snapshot` at the end of the risk period.
```
chahank: Naming of variable :).

climada/trajectories/calc_risk_metrics.py
```diff
@@ -365,3 +387,810 @@ def apply_measure(self, measure: Measure) -> "CalcRiskMetricsPoints":
...
+        self, invariant: str, return_periods: list[int]
+    ) -> np.ndarray:
+        return calc_per_date_rps(
+            self._imp_mats(invariant=invariant),
+            (
+                self.snapshot_start.hazard.frequency
+                if "H0" in invariant
+                else self.snapshot_end.hazard.frequency
+            ),
+            self.date_idx.freqstr[0],
```
chahank: Is it always a single digit that is relevant?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -365,3 +387,810 @@ def apply_measure(self, measure: Measure) -> "CalcRiskMetricsPoints":
...
+        Parameters
+        ----------
+
+        return_periods : list of int
+            The return periods to estimate impacts for.
+
+        """
+
+        # currently mathematicaly wrong, but approximatively correct,
+        # to be reworked when concatenating the impact matrices for the interpolation
```
chahank: Should this be documented somewhere clear or is it addressed in the cascade of PRs?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -365,3 +387,810 @@ def apply_measure(self, measure: Measure) -> "CalcRiskMetricsPoints":
...
+    def _imp_mats(self, invariant: str) -> list:
+        """List of `time_points` impact matrices with changing
+        exposure, and invariant hazard and vulnerability.
+
+        """
+        if re.match(r"H[01]V[01]", invariant):
+            return self._interp_mats(f"E0{invariant}", f"E1{invariant}")
+
+        if re.match(r"E[01]H[01]V[01]", invariant):
+            return [getattr(self, invariant).imp_mat] * self.time_points
```
chahank: Should this return independent copies of `imp_mat` or not? If yes, this probably does not do that.

chahank (approved):
Some modest changes and I think otherwise the code is already very nice. Maybe check merge conflicts should be checked though ;).

climada/trajectories/test/test_interpolated_risk_trajectory.py
spjuhel: Removed

climada/trajectories/test/test_interpolated_risk_trajectory.py
spjuhel: After NPV, comment was from a time where I cached before NPV transform, that's no longer the case. (modified the comment)

climada/trajectories/test/test_interpolated_risk_trajectory.py
spjuhel: Removed

climada/trajectories/calc_risk_metrics.py
spjuhel: Fixed

climada/trajectories/calc_risk_metrics.py
spjuhel: Good catch!

climada/trajectories/calc_risk_metrics.py
spjuhel: Good catch again!

climada/trajectories/calc_risk_metrics.py
spjuhel: I should document that. I don't think it is relevant for users though. The approximation remains good.
I can put a note in this docstring for future developments.

climada/trajectories/interpolated_trajectory.py
spjuhel: Fixed

climada/trajectories/interpolated_trajectory.py
spjuhel: Fixed

climada/trajectories/interpolated_trajectory.py
spjuhel: Good catch!

Changed, improved and clarified.

spjuhel:
Ok, I'm merging as soon as the tests are green :)

spjuhel:
OK the failures are related to #1292, nothing to worry.

@emanuel-schmid the integration tests ran fine locally, please tell me if anything is wrong on jenkins (no reason a priori)

chahank:
Fantastic, great job @spjuhel

PR #1252, opened 2026-02-24
Add inter/extrapolate features to ImpactFreqCurve

ValentinGebhart:
Changes proposed in this PR:
- add features of util interpolation module (used in local_exceedance_impact methods etc.) to the class `ImpactFreqCurve`, adding the `interpolate` method to evaluate ImpactFreqCurve at new return periods with inter- and extrapolation options 
- add test for `interpolate` method
- renamed variable in interpolation module to avoid confusion
- add deprecation warning if user calls `impact.calc_freq_curve()` with given return periods as inputs

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


climada/engine/test/test_impact.py
```diff
@@ -321,6 +321,42 @@ def test_ref_value_rp_pass(self):
...
+    def test_evaluate_freq_curve(self):
+        """Test evaluate method of freq curve"""
+        imp = Impact()
+        imp.frequency = np.ones(4) * 0.1
+        imp.at_event = np.zeros(4)
+        imp.at_event[0] = 0.0
+        imp.at_event[1] = 100.0
+        imp.at_event[2] = 50.0
+        imp.at_event[3] = 110.0
+        imp.unit = "USD"
```
chahank: Any reason to make it complicated? In general, always initialize complete python objects instead of modifying them in place.

```suggestion
        imp.at_event = np.array(0.0, 100.0, 50.0, 110.0)
        imp.unit = "USD"
```

climada/engine/test/test_impact.py
```diff
@@ -321,6 +321,42 @@ def test_ref_value_rp_pass(self):
...
+        )
+        npt.assert_array_almost_equal(
+            ifc.evaluate(
+                [1.0, 2.5, 4, 20],
+                method="extrapolate",
+                log_frequency=False,
+                log_impact=False,
+            ),
+            [-300.0, 0.0, 75.0, 115.0],
+        )
```
chahank, line 377: Can you make a brief comment to make it clear that these are the values to expect? 

Also, would it not be better to not have an event with `0` impact to test also the extrapolation at low values?

climada/engine/impact.py
```diff
@@ -732,15 +723,6 @@ def local_return_period(
             self.frequency_unit
         )
 
-        # check method
```
chahank: This looks unrelated to this PR.

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+        self,
+        return_period,
+        *,
+        method="interpolate",
+        log_frequency=True,
+        log_impact=True,
+        min_impact=0,
+        bin_decimals=None,
+    ):
+        """Evaluate impact frequency curve with different extrapolation options.
```
chahank:
```suggestion
        """Evaluate impact frequency curve with different interpolation and extrapolation options.
```

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+
+        Parameters
+        ----------
+        return_period : Iterable[float]
+            return periods for which to evaluate the impact frequency curve
+            axis to use
```
chahank:
```suggestion

```

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+        bin_decimals : int, optional
+            Number of decimals to group and bin impact values. Binning results in smoother (and
+            coarser) interpolation and more stable extrapolation. For more details and sensible
+            values for bin_decimals, see Notes. If None, values are not binned. Defaults to None.
+
+        Returns
+        -------
+        np.array
+            array of exceeded impacts at given return periods
+        """
```
chahank, line 2367: Maybe add the See Also to the util methods.

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+        # sort return periods of ImpactFreqCurve
+        sorted_idxs = np.argsort(self.return_per)
+        impacts = np.squeeze(np.array(self.impact)[sorted_idxs])
+        rps = np.asarray(self.return_per)[sorted_idxs]
+
+        frequency = np.diff(1 / np.array(rps)[::-1], prepend=0)
+
+        return u_interp.preprocess_and_interpolate_ev(
+            exceedance_frequency,
+            None,
```
chahank, line 2377: Is this always required?

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+            frequency,
+            impacts,
+            log_frequency=log_frequency,
+            log_values=log_impact,
+            value_threshold=min_impact,
+            method=method,
+            y_asymptotic=0.0,
```
chahank: Is this always required for return period curves?

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+    def plot_extended(
+        self,
+        *,
+        return_period_range=None,
+        axis=None,
+        log_frequency=False,
+        kwargs_interp=None,
+        **kwargs,
+    ):
+        """Plot impact frequency curve with extrapolation options.
```
chahank:
```suggestion
        """Plot impact frequency curve with interpolation or extrapolation options.
```

climada/util/interpolation.py
```diff
@@ -46,11 +46,11 @@ def preprocess_and_interpolate_ev(
...
     ----------
-    test_frequency : array_like
-        1-D array of test frequencies for which values (e.g., intensities or impacts) should be
+    exceedance_frequency : array_like
+        1-D array of test exceedance frequencies for which values (e.g., intensities or impacts) should be
         assigned. If given, test_values must be None.
     test_values : array_like
-        1-D array of test values (e.g., intensities or impacts) for which frequencies should be
+        1-D array of test values (e.g., intensities or impacts) for which exceedance frequencies should be
         assigned. If given, test_frequency must be None.
```
spjuhel:
```suggestion
        assigned. If given, exceedance_frequency must be None.
```

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+                If set to True, impact values are converted to log scale before
+                inter- and extrapolation. Defaults to True.
+            bin_decimals : int, optional
+                Number of decimals to group and bin impact values. Binning results in smoother (and
+                coarser) interpolation and more stable extrapolation. For more details and sensible
+                values for bin_decimals, see Notes. If None, values are not binned. Defaults to None.
+        kwargs : dict, optional
+            arguments for plot matplotlib function, e.g. color='b'
+
+        Returns
```
chahank: Can you add the See Also for all the relevant util and other methods?

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+            impacts,
+            log_frequency=log_frequency,
+            log_values=log_impact,
+            value_threshold=min_impact,
+            method=method,
+            y_asymptotic=0.0,
+            bin_decimals=bin_decimals,
+        )
+
+    def plot_extended(
```
chahank: Can we find a word that is the same for `evaluate` and `plot_extended`?

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+        } | kwargs_interp
+
+        if return_period_range is None:
+            return_periods = np.linspace(
+                0.5 * min(self.return_per), 1.2 * max(self.return_per), 500
+            )
+        else:
+            return_periods = np.linspace(
+                min(return_period_range), max(return_period_range), 500
+            )
```
chahank: I would maybe let the user decide on the number of interpolation points. Also, a linear interpolation range for log data would look a bit strange no?

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+            return_periods,
+            impacts,
+            axis,
+            log_frequency,
+            self.frequency_unit,
+            self.unit,
+            self.label,
+            **kwargs,
+        )
+        # axis = self._plot(self.return_per, self.evaluate(self.return_per, **(kwargs_interp|{"method": "interpolate"})), axis, log_frequency, self.frequency_unit, self.unit, self.label, **kwargs)
```
chahank:
```suggestion

```

climada/engine/impact.py
```diff
@@ -2299,15 +2281,208 @@ def plot(self, axis=None, log_frequency=False, **kwargs):
...
+        for rp in [min(self.return_per), max(self.return_per)]:
+            axis.axvline(x=1 / rp if log_frequency else rp, linestyle="--", c="gray")
+
+        return axis
+
+    @classmethod
```
chahank: Should this not be a @staticmethod ? I do not see where the cls is used.

chahank (changes requested):
Great, small tweaks. 

Changelog needs to be updated too.

climada/engine/test/test_impact.py
ValentinGebhart: right, I'll add some comments.

The inter/extrapolation has a keyword "min_impact" which is set to 0 by default, meaning it only considers impacts > 0. The 0 is here to test that it does indeed neglect it (now also clearer by new test frequencies).

climada/engine/impact.py
ValentinGebhart: The util functions `u_interp.preprocess_and_interpolate_ev` takes four arguments
```
def preprocess_and_interpolate_ev(
    exceedance_frequency,
    test_values,
    frequency,
    values,
    ...)
```
The third and forth are the given data values used for "training" the interpolation function. The "test" values at which to interpolate are given either as exceedance_frequencies (first slot) or test_values (second slot). One of the two slots must always be None.

Does this answer your question? I also feel that this looks a bit awkward but it was what we had decided when writing the functions.

climada/engine/impact.py
ValentinGebhart: hm good question. This is defining the impact value assigned to return periods smaller than the data range, and is used only if method is not interpolation (there the value assigned is NaN). 
I would not know which value assign if not 0 or NaN. Or would you rather have the user decide?

climada/engine/impact.py
ValentinGebhart: hm, this is "only" about the plotting, we do not return values here. We could let the user decide how many points to use for the line plot?

Here the linspace is used because the user sees the return periods on a linear scale in the plot. But yes, if the user changes the x axes to log scale, or uses the log of exceedance frequencies on the x axes, this might look strange.

Should we change it or is it fine?

ValentinGebhart:
This is what the method new methods would look like in the example at the bottom of the [10min intro](https://climada-python.readthedocs.io/en/stable/user-guide/0_10min_climada.html). 

`imp.calc_freq_curve().plot()`
image

````python
freq_curve = imp.calc_freq_curve()
freq_curve_extra = freq_curve.interpolate(return_periods=np.linspace(10,300,100), method="extrapolate")
freq_curve_inter = freq_curve.interpolate(return_periods=np.linspace(10,300,100))
ax = freq_curve_inter.plot(color="blue")
freq_curve_extra.plot(axis=ax, linestyle="--", color="blue")
````
plot_extrapolate

climada/engine/impact.py
chahank: That is clearer thanks. My question was more to be sure that we do not hard code hidden values for no good reason.

climada/engine/impact.py
chahank: In general, I prefer no hardcoding of values without good reason and rather would work with good defaults.

climada/engine/impact.py
chahank: Yes, basically for the same reasons as mentioned above.

chahank:
Nice! And how would the interpolate look like? It seems to me that for this function interpolation (with extrapolation) would be more informative to smooth out the missing values at high return periods. (Maybe this should also be the default instead of the direct extrapolation).

ValentinGebhart:
> Nice! And how would the interpolate look like? 

Interpolate would look exactly the same but with no values outside the two dashed lines. I think matplotlib would the automatically zoom into the data range, so it would look very much like the original plot() method (ofc the interpolation is a bit different).

> It seems to me that for this function interpolation (with extrapolation) would be more informative to smooth out the missing values at high return periods. (Maybe this should also be the default instead of the direct extrapolation).

What exactly do you mean with "smooth out"? I assume to mark the outside-of-data range, but not sure how you would like to visualize it.

ChrisFairless:
Thanks a lot for this, it's going to be very useful for me!

Something that would be useful in some of the work I'm doing would be an `ImpactFreqCurve.add_points()` method, that returns a new `ImpactFreqCurve` with additional interpolated/extrapolated points using this new functionality.

Also: if that's not within scope of the PR and you'd just like to merge we can leave it for later!

chahank:
Great to hear @ChrisFairless. If you need this, you could add a wrapper method around the added `extroplate` one. Would you like to do this before we merge? Otherwise, the new method gives back the new values which can then be used to create a new object.

ValentinGebhart:
I am hestitating with merging the PR due to two points. What do you think @chahank?

(1) Depending on user input it can happen that the return period curve is interpolated twice, each with different methods. In this code 
````python

fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))

fc.plot_extrapolate(return_periods=np.linspace(10,100,100))
````
The first line creates a frequency curve at the given return periods, interpolated with linear interpolation between (exceedence frequncy, impact) pairs.
The second line interpolates to a second series of return periods (here with higher resolution), by default with linear interpolation between (log(exceedence frequncy), log(impact)) pairs. Apart from the different default settings, I don't think the double interpolation is very elegant. If the first line is instead `fc = imp.calc_freq_curve()`, the above is not an issue.

(2) Is the plot style clear or do we want to show it differently? See two dashed gray lines indicating the range of data.

chahank:
Good points!

> I am hestitating with merging the PR due to two points. What do you think @chahank?
> 
> (1) Depending on user input it can happen that the return period curve is interpolated twice, each with different methods. In this code
> 
> ```python
> fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))
> 
> fc.plot_extrapolate(return_periods=np.linspace(10,100,100))
> ```
> 
> The first line creates a frequency curve at the given return periods, interpolated with linear interpolation between (exceedence frequncy, impact) pairs. The second line interpolates to a second series of return periods (here with higher resolution), by default with linear interpolation between (log(exceedence frequncy), log(impact)) pairs. Apart from the different default settings, I don't think the double interpolation is very elegant. If the first line is instead `fc = imp.calc_freq_curve()`, the above is not an issue.

Should we make it default that `calc_freq_curve` does not take specific interpolations? 

> 
> (2) Is the plot style clear or do we want to show it differently? See two dashed gray lines indicating the range of data.

I think it is a bit unclear what the grey lines mean. Not sure how to make it clearer. Maybe instead make the color of the extrapolated part of the curve dashed? And no vertical dashed lines?

ValentinGebhart:
Perfect, I like both your suggestions. I will implement them. Small question:

A line like
`fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))`
would then result in an error because at this stage we just evaluate at the return periods of the impact object. The user can then inter/extrapolate later. This might break some existing user code or pipelines. Can we just change it or do we need deprecation warning etc?

chahank:
> Perfect, I like both your suggestions. I will implement them. Small question:
> 
> A line like `fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))` would then result in an error because at this stage we just evaluate at the return periods of the impact object. The user can then inter/extrapolate later. This might break some existing user code or pipelines. Can we just change it or do we need deprecation warning etc?

Good point. Then for the moment, let us keep it working, but with a deprecation warning.

ValentinGebhart:
I added a deprecation warning if one creates the ImpactFreqCurve using `Impact.calc_freq_curve()` for specfic return periods, for instance like this:
`fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))`.

Also, I changed the extrapolation plot style, extrapolated values are now shown as a dashed line, see plot above.

@chahank for me this would be ready to merge, let me know if we should go ahead.

chahank:
> I added a deprecation warning if one creates the ImpactFreqCurve using `Impact.calc_freq_curve()` for specfic return periods, for instance like this: `fc = imp.calc_freq_curve(return_per=np.linspace(10, 100, 10))`.
> 
> Also, I changed the extrapolation plot style, extrapolated values are now shown as a dashed line, see plot above.
> 
> @chahank for me this would be ready to merge, let me know if we should go ahead.

Excellent! We now just have one little inconsistency: the extrapolate method returns a numpy array, while the calc_freq_curve returns directly the ImpactFreqCurve. Maybe we can improve the deprecation message to make it clear how to obtain a new ImpactFreqCurve? Or should maybe the extrapolate method be able to return an ImpactFreqCurve too?

ValentinGebhart:
I would say that there are two methods to create an `ImpactFreqCurve`, one is `impact.calc_freq_curve()`, and one is via manually initializing it.
Then, given an `ImpactFreqCurve`, one can inter- and extrapolate to any test return period using `extrapolate()`, returning an array of corresponding impact values.
For me, this behaviour would be okay, as it treats the actual "data-based" frequency curve as the central object, and does not equate it with some extrapolated values.

If you think it would be better to have the extrapolate method take an ImpactFreqCurve and return an ImpactFreqCurve, would could also do that.

chahank:
> I would say that there are two methods to create an `ImpactFreqCurve`, one is `impact.calc_freq_curve()`, and one is via manually initializing it. Then, given an `ImpactFreqCurve`, one can inter- and extrapolate to any test return period using `extrapolate()`, returning an array of corresponding impact values. For me, this behaviour would be okay, as it treats the actual "data-based" frequency curve as the central object, and does not equate it with some extrapolated values.
> 
> If you think it would be better to have the extrapolate method take an ImpactFreqCurve and return an ImpactFreqCurve, would could also do that.

Yes, I am unsure what is best. Maybe we can ask in the developer's call for an opinion?

ValentinGebhart:
After advice from @peanutfun and @emanuel-schmid (thanks!), I changed the implementation to:

- `freq_curve = impact.calc_freq_curve()` returns an `ImpactFreqCurve`. If return periods are given as an argument, a deprecation warning is raised.
- `freq_curve_interp = freq_curve.interpolate(return_periods=np.linspace(10, 300, 100), method="extrapolate")` returns a new `ImpactFreqCurve` object with the inter- and extrapolated impacts at the given return periods. Interpolation is default.
- `freq_curve.plot()` is the original plot function and can be used by the user to combine curves, e.g.
````
freq_curve_extra = freq_curve.interpolate(return_periods=np.linspace(10,300,100), method="extrapolate")
freq_curve_inter = freq_curve.interpolate(return_periods=np.linspace(10,300,100))
ax = freq_curve_inter.plot(color="blue")
freq_curve_extra.plot(axis=ax, linestyle="--", color="blue")
````

@chahank Let me know if this looks good, then we can merge.

chahank (approved):
Excellent! You may merge, thanks for the great contribution.

Note: it seems that we do not show any return period curve in the tutorials (which was a surprise to me, or I missed it). You could add an example, for instance in the Impact tutorial or the 10-minutes climada if you want (or not). After you can simply merge without further review.

ValentinGebhart:
We have a return period curve at the end of the [10min intro](https://climada-python.readthedocs.io/en/stable/user-guide/0_10min_climada.html). That should be sufficient or should we add the extrapolation options?

After finising the GEV/GPD fit options, I would anyways add a (extension of a) tutorial somewhere, so then there would also be something explicit about the extrapolation options.

PR #1256, opened 2026-02-26
Read units information when creating an Exposures from an Excel or data frame

ChrisFairless:
Changes proposed in this PR:

When the user provides creates an Excel file from a dataframe, CSV or Excel file, the value_unit column is ignored. Units currently have to be specified manually or the default "USD" is used.

It looks like early versions of Entity files maybe didn't include a value_unit column, since the entity_template.xlsx file in the system data doesn't have it. However it is now convention to specify the units (see e.g. the San Salvador case study elsewhere in the repo).

This PR
- Updates the Exposures `__init__` method to read the units when they're available in the input dataframe, and deals with conflicts if they are also specified through the `meta` or `value_unit` parameters)
- Updates the entity_template.xlsx file to add a value_unit column. I give it the 'XSD' currency to match the HDF5 Entity file that it's often compared to during tests
- Updates the tests, including
  - new tests for these changes
  - not using the default USD when creating Exposures so they fail if the default unit is used for any reason
  - improving a constructor at one point

When I need a currency in the tests I pick PKR, since this behaviour was forst noted by colleagues in Pakistan – thanks to Dr Salman Saeed for the help.

Note: The units determined from the first row of the data frame. There's no check to make sure they're consistent. This might be worth doing! Probably in the `__init__` method, with something like
```
assert geodataframe.get('value_unit', []).nunique() <= 1, "Exposure input data must all use the same unit"
``` 
worth adding?

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Great fix, thank you for solving this issue! 

Just to be clear (since you tested it): does the reading / writing of the units work correctly for you with the .hdf5 format? 

Checking the uniquenss of the units in the value_unit column in the excell file could be nice, but not key since the value_unit should not be stored in the gdf, but in the metadata.

ChrisFairless:
> Just to be clear (since you tested it): does the reading / writing of the units work correctly for you with the .hdf5 format?

I've only looked at reading here, and that works fine for Excel and HDF5. (The `climada.exposures.test.test_base.test_io_hdf5_pass` test reads the example HDF5 and Excel files and checks that their units match, and an earlier test checks they haven't fallen back to CLIMADA's default values)
 
> Checking the uniquenss of the units in the value_unit column in the excell file could be nice, but not key since the value_unit should not be stored in the gdf, but in the metadata.

Yeah. Currently the Entity file doesn't let you specify metadata so this is the workaround that somehow became convention. When we redesign Excel inputs for v7 there should be an optional 'metadata' sheet with units instead of this `value_units` column.

spjuhel:
FYI the Entity file used in the test is loaded via a constant that is dependent on the config. Hence, local tests fail if you have a data folder different from the one within the repo (as the changes to the xlsx file)

Not a huge problem, but we should probably make sure files loaded for tests are from the internal folder.

PR #1264, opened 2026-03-10
Fixes TypeError when casting arrays to scalar with Numpy>=2.4

### LINKED ISSUE #1263: `numpy>=2.4.0` introduces type errors in the tests.
**Describe the bug**

See [numpy release notes](https://numpy.org/doc/2.4/release/2.4.0-notes.html#raise-typeerror-on-attempt-to-convert-array-with-ndim-0-to-scalar)

**To Reproduce**
1. have numpy>=2.4.0
2. run the tests (e.g. `climada/entity/exposures/test/test_mat.py`)
3. Observe the failure:

```
climada/entity/exposures/test/test_mat.py:37: in test_read_demo_pass
    expo = Exposures.from_mat(ENT_TEST_MAT)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
climada/entity/exposures/base.py:1277: in from_mat
    _read_mat_metadata(exp, data, file_name, var_names)
climada/entity/exposures/base.py:1597: in _read_mat_metadata
    exposures.value_unit = u_hdf5.get_str_from_ref(
climada/util/hdf5_handler.py:110: in get_str_from_ref
    return get_string(file[var])
           ^^^^^^^^^^^^^^^^^^^^^
climada/util/hdf5_handler.py:92: in get_string
    return "".join(chr(int(c)) for c in array)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
climada/util/hdf5_handler.py:92: in <genexpr>
    return "".join(chr(int(c)) for c in array)
                       ^^^^^^
E   TypeError: only 0-dimensional arrays can be converted to Python scalars
```

**Expected behavior**
No failure

**Suggestion**

Numpy 2.4 now requires to specifically access an item to enable conversion to a number. Implement this change.

spjuhel:
Changes proposed in this PR:
Proper item access (or 0-dim array) before casting to float/int.

This PR fixes #1263 and partly #1204

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/tc_tracks_synth.py
+++ b/climada/hazard/tc_tracks_synth.py
@@ -1049,7 +1049,7 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
                 # if there is no further landfall, correct until the end of
                 # the track
                 end_cor = track["time"].size
-            rndn = 0.1 * float(np.abs(np.random.normal(size=1) * 5) + 6)
+            rndn = 0.1 * float(np.abs(np.random.default_rng().normal() * 5) + 6)
             r_diff = (
                 track["central_pressure"][land_sea].values
                 - track["central_pressure"][land_sea - 1].values
--- a/climada/util/hdf5_handler.py
+++ b/climada/util/hdf5_handler.py
@@ -89,7 +89,7 @@ def get_string(array):
     -------
     string
     """
-    return "".join(chr(int(c[0])) for c in array)
+    return "".join(chr(int(c.item())) for c in array)
 
 
 def get_str_from_ref(file_name, var):
--- a/climada/util/hdf5_handler.py
+++ b/climada/util/hdf5_handler.py
@@ -89,7 +89,7 @@ def get_string(array):
     -------
     string
     """
-    return "".join(chr(int(c)) for c in array)
+    return "".join(chr(int(c[0])) for c in array)
 
 
 def get_str_from_ref(file_name, var):
--- a/climada/util/hdf5_handler.py
+++ b/climada/util/hdf5_handler.py
@@ -89,7 +89,7 @@ def get_string(array):
     -------
     string
     """
-    return "".join(chr(int(c.item())) for c in array)
+    return "".join(chr(c.item()) for c in array)
 
 
 def get_str_from_ref(file_name, var):


climada/util/hdf5_handler.py
```diff
@@ -89,7 +89,7 @@ def get_string(array):
     -------
     string
     """
-    return "".join(chr(int(c)) for c in array)
+    return "".join(chr(int(c[0])) for c in array)
```
spjuhel:
```suggestion
    return "".join(chr(int(c.item())) for c in array)
```

spjuhel:
The new error appears to come from an incompatibility between `numpy>=2.4` and `salib<1.5.2` (for the same reason we had an error).

But currently `salib>1.5.2` causes some test to fail (at least on my machine), I assume this is related to #1081 ?
@emanuel-schmid

So we can also pin `numpy<2.4` but this is postponing the problem.

climada/util/hdf5_handler.py
```diff
@@ -89,7 +89,7 @@ def get_string(array):
     -------
     string
     """
-    return "".join(chr(int(c)) for c in array)
+    return "".join(chr(int(c.item())) for c in array)
```
emanuel-schmid:
```suggestion
    return "".join(chr(c.item()) for c in array)
```
my guess ist that the `int` casting was introduced as a workaround of this very problem in the first place.

climada/hazard/tc_tracks_synth.py
```diff
@@ -1049,7 +1049,7 @@ def _apply_decay_coeffs(track, v_rel, p_rel, land_geom, s_rel):
                 # if there is no further landfall, correct until the end of
                 # the track
                 end_cor = track["time"].size
-            rndn = 0.1 * float(np.abs(np.random.normal(size=1) * 5) + 6)
+            rndn = 0.1 * float(np.abs(np.random.default_rng().normal() * 5) + 6)
```
emanuel-schmid, line 1052: er - why?

emanuel-schmid (commented):
🙌 thanks for catching and fixing!

just wonder what is the tc_tracks_synth change for? it doesn't fix any bugs, does it?

btw. : the salib issue has come to a conclusion, which must be implemented yet. no need to pin down numpy

climada/hazard/tc_tracks_synth.py
spjuhel: Two things at once here sorry:
- `np.random.normal(size=1)` generates a numpy array (so it was causing the same bug as the other place)
- `np.random.normal` has been deprecated for >3years now and a random generator should be instantiated instead. And the `.normal()` of a random generator with no arguments does return a number and not a singled-value array.

climada/util/hdf5_handler.py
spjuhel: Possible, I wasn't sure if the value was always an integer, so I kept it the same.

emanuel-schmid:
@spjuhel awesome, thanks!

PR #1269, opened 2026-03-23
Fix TCTracks.from_FAST duplicate loading from year loop

### LINKED ISSUE #1268: TCTracks.from_FAST duplicates tracks by iterating over unrelated year dimension
I found that `TCTracks.from_FAST` multiplies tracks when reading FAST NetCDF files that contain both `n_trk` and `year` dimensions.

Observed behavior:
Input FAST file has n_trk = 23000 and year = 46.
`TCTracks.from_FAST` returns 1058000 tracks.
This equals 23000 × 46, suggesting each track is reconstructed once per year index.

Why:
Loader loops over both `dataset.year` and `dataset.n_trk.`
It then selects `dataset.sel(n_trk=i, year=year),` but key per-track variables `(lon_trks`, `lat_trks`, `vmax_trks`, `tc_years`, `tc_month`, `tc_basins`) are indexed by `n_trk` (and `time`), not by `year`.
So the same track is appended repeatedly for each year entry.

Expected behavior:
Each `n_trk` should produce one track entry.
Returned track count should match `n_trk` total (or sum across files), not `n_trk` × `n_year`.

Impact:
Memory usage and runtime increase massively.

Suggested fix:
Remove the outer loop over `dataset.year`.
Iterate only over `dataset.n_trk` and select `track = dataset.sel(n_trk=i)`.

simonameiler:
Changes proposed in this PR:
This PR fixes duplicate track creation in `TCTracks.from_FAST` for FAST NetCDF files containing both `n_trk` and `year` dimensions.

Root cause:
`from_FAST` iterated over both `dataset.year` and `dataset.n_trk,` then selected `dataset.sel(n_trk=i, year=year)`.
For FAST files where track variables are indexed by `n_trk` and `time,` this repeats each track once per year index.

Changes:
In climada/hazard/tc_tracks.py:
removed outer loop over `dataset.year`
iterate only over `dataset.n_trk`
select tracks with `dataset.sel(n_trk=i)`

Tests:
Added regression test `test_from_FAST_not_multiplied_by_year_dim` in climada/hazard/test/test_tc_tracks.py.
Test builds a tiny temporary FAST-like NetCDF with both `n_trk` and `year` dims and checks that output size equals `n_trk` (not `n_trk * n_year`).
Validation

Ran:
python -m unittest climada.hazard.test.test_tc_tracks.TestIO.test_from_FAST
python -m unittest climada.hazard.test.test_tc_tracks.TestIO.test_from_FAST_not_multiplied_by_year_dim
Both passed locally.

This PR fixes #1268

### PR Author Checklist

### PR Reviewer Checklist


climada/hazard/test/test_tc_tracks.py
```diff
@@ -704,6 +705,70 @@ def test_from_FAST(self):
...
+                        "year",
+                        np.array([1998, 1999, 2000, 2001], dtype=np.int64),
+                    ),
+                    "basin": ("basin", np.array(["NA"], dtype="<U2")),
+                    "month": ("month", np.arange(1, 13, dtype=np.int64)),
+                },
+            )
+
+            path = DATA_DIR.joinpath(tmpdir, "fast_regression.nc")
+            ds.to_netcdf(path)
```
NicolasColombi: Since there is already a test file named `FAST_test_tracks.nc`, with only one year (hence the test did not reveled the bug you are suggesting to fix), it might be a good idea to update such file to contain 2 years, and then test the updated `from_FAST` function on such file with only one test. I do not think this would be an issue size wise, as it is only 62KB at the moment. This would require, either reproducing the small file for two years, or simply fabricating such file by duplicating the existing one, concatenate it, and manually modify the second year number. Lastly, you will need to update the original test to capture this. 

This way I think we can have a single file and single test, reducing the code, since the majority of the code in your test is there to create a temporary file.

NicolasColombi (changes requested):
Hi Simona, thanks for catching this and proposing to fix it! :raised_hands:

As mentioned in the review, my only remark is on the test file, I believe it would be less cumbersome to have a single file and single test.

climada/hazard/test/test_tc_tracks.py
simonameiler: Thanks for the suggestion, this makes a lot of sense.

I’ve updated the existing fixture file (`FAST_test_tracks.nc`) to include two years (year = [2025, 2026]) by duplicating the `seeds_per_month` data along the `year` dimension. The track-related variables (`tc_years`, `lon_trks`, etc.) remain unchanged, so the file structure stays consistent with the original intent.

With this update, I removed the separate regression test and its temporary-file setup. The existing assertion in `test_from_FAST`:

`self.assertEqual(len(tc_track.data), 5)`

now ensures that tracks are not duplicated when a `year` dimension is present. With two years present, the previous buggy implementation would have returned 10 tracks instead of 5.

This keeps the test setup simpler and avoids duplicating logic for temporary file creation.

NicolasColombi (approved):
Thanks for fixing this Simona! :raised_hands:  This is ready to merge from my side.

@chahank do we have your blessing ?

chahank (approved):
Please update the CHANGELOG file : https://github.com/CLIMADA-project/climada_python/blob/main/CHANGELOG.md

chahank:
Thanks for the bugfix!

PR #1273, opened 2026-03-26
Adds `conftest.py` of common fixtures for integration tests

spjuhel:
This is based on data test used for the integration tests of the static and interpolated trajectories.

The objective is to provide:
- "as comprehensible as possible and as minimalistic as possible data (exposure, hazard, impfset)"
- Factories  

Changes proposed in this PR:
- Adds fixtures and factories for minimalistic objects for integration testing
- Adds an quick example of how to use them

We need to decide if this is something we actually want.
Otherwise, I will move the content to the trajectory integration test directly.

### PR Author Checklist

### PR Reviewer Checklist


spjuhel:
Notes from discussion:

- shift the examples to a (possibly existing) page in the documentation

spjuhel:
Note for this PR (or rather for future improvements):

- Have also fixtures with the `DEMO` files
- Plan (a hackathon?) to move tests to use the fixtures

spjuhel:
I've added a documentation page, 
I chose to make a new one because direct integration in the "Writing tests" one did not feel right, but I'm happy to reconsider. Note that there are two mentions/references in the "Writing tests" one!

I tried to find the balance between being explicit and concise (had help from Claude to be honest).

I am wondering about putting actual values in the documentation as this implies that we keep them up to date with the file (although the values should technically not change).

Quick access: https://climada-python--1273.org.readthedocs.build/en/1273/development/Guide_test_fixtures.html

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+# Sanity checks
+for const in [VALUES, CATEGORIES, EXP_LONS, EXP_LATS]:
+    assert len(const) == len(
+        VALUES
+    ), "VALUES, REGIONS, CATEGORIES, EXP_LONS, EXP_LATS should all have the same lengths."
+
+for const in [EVENT_IDS, EVENT_NAMES, DATES, FREQUENCY]:
+    assert len(const) == len(
+        EVENT_IDS
+    ), "EVENT_IDS, EVENT_NAMES, DATES, FREQUENCY should all have the same lengths."
```
peanutfun: If these assertions fail, *importing* this file will result in an error. I think this is dangerous, as the resulting errors are hard to understand an occur at weird points in time. Are those assertions needed at all? Shouldn't the constructed objects like Exposures complain?

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+@pytest.fixture(scope="session")
```
peanutfun: Other scopes than `function` are primarily intended to avoid reconstructing "expensive" objects. The way I see it, this applies to none of the objects here and can have serious side effects, as modifications will persist throughout the test session. I suggest using the default (smallest) scope everywhere, at least in this file.

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+def exposure_values():
```
peanutfun: Do you foresee this being used as a fixture to an actual test? I don't, so I would keep this a "regular" function and only declare fixtures that are intended for use.

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+        * Event 4: quarter max intensity everywhere (impact == 1/4 total value)
+        * Event 5: max intensity everywhere (but zero frequency)
+      With max intensity set at 100
+    - Impact function is the "identity function", x intensity is x% damages
+    - Impact values should be:
+        * AAI = 18 = 1000*1/2*0.006+(1000+2000+3000+4000+5000)*0.25*0.004
+        * RP20 = event1 = 0
+        * RP50 = event2 = 0
+        * RP100 = event3 = 500 = 1000*1/2
+        * RP250 = event4 = 3750 = (1000+2000+3000+4000+5000)*0.25
```
peanutfun, line 43: Fixtures are Python functions, so I suggest adding this information to fixture docstrings and creating an API reference. Individual fixtures can be documented with `autofunction`. See https://docs.pytest.org/en/stable/reference/reference.html#fixtures and https://github.com/pytest-dev/pytest/blob/f93656d611f3c9cd22f346dfeda52cddb778c589/doc/en/reference/reference.rst?plain=1#L325

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+@pytest.fixture(scope="session")
+def categories():
+    return CATEGORIES.copy()
+
+
+@pytest.fixture(scope="session")
+def exposure_geometry():
+    return [Point(lon, lat) for lon, lat in zip(EXP_LONS, EXP_LATS)]
```
peanutfun: https://geopandas.org/en/stable/docs/reference/api/geopandas.points_from_xy.html#geopandas.points_from_xy ?

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+HAZ_LATS = EXP_LATS + HAZ_JITTER
+
+# Hazard events
+EVENT_IDS = np.array([1, 2, 3, 4, 5])
+EVENT_NAMES = ["ev1", "ev2", "ev3", "ev4", "ev5"]
+DATES = np.array([1, 2, 3, 4, 5])
+
+# Frequency are choosen so that they cumulate nicely
+# to correspond to 250, 100, 50, and 20y return periods (for impacts)
+FREQUENCY = np.array([0.03, 0.01, 0.006, 0.004, 0.0])
```
peanutfun, line 92: Do the math for clarity?
```suggestion
FREQUENCY = np.array([1/20, 1/50, 1/100, 1/200, 0.0, dtype="float"])
```

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+# ---------------------------------------------------------------------------
+EXP_DESC = "Test exposure dataset"
+EXPOSURE_REF_YEAR = 2020
+EXPOSURE_VALUE_UNIT = "USD"
+VALUES = np.array([0, 1000, 2000, 3000, 4000, 5000])
+CATEGORIES = np.array([1, 1, 2, 1, 1, 3])
+
+# Exposure coordinates
+EXP_LONS = np.array([4, 4.25, 4.5, 4, 4.25, 4.5])
+EXP_LATS = np.array([45, 45, 45, 45.25, 45.25, 45.25])
```
peanutfun: Hard to read. As values don't matter, choose digits that are easier to distinguish (also against the LON values)

```suggestion
EXP_LATS = np.array([33, 33, 33, 33.25, 33.25, 33.25])
```

climada/test/conftest.py
```diff
@@ -0,0 +1,328 @@
...
+CRS_WGS84 = "EPSG:4326"
+
+# ---------------------------------------------------------------------------
+# Exposure attributes
+# ---------------------------------------------------------------------------
+EXP_DESC = "Test exposure dataset"
+EXPOSURE_REF_YEAR = 2020
+EXPOSURE_VALUE_UNIT = "USD"
+VALUES = np.array([0, 1000, 2000, 3000, 4000, 5000])
+CATEGORIES = np.array([1, 1, 2, 1, 1, 3])
```
peanutfun, line 68: Good first step, but I think we need to parameterize asap: Use multiple CRS, build exposures of size 1 and >1, etc.

peanutfun (changes requested):
This is a good first step towards using common fixtures for all our unit tests! I found a few issues in the code. Other than that, I think we should add an API reference for fixtures to the docs (this might require loading the conftest.py as well. Not sure how to do that atm).

I think most of the functions defined here should *not* be fixtures because they are not intended for use outside this file. The only "real" fixtures I see:
* exposures_factory
* exposures
* hazard_factory
* hazard
* impf_factory
* linear_impact_function
* impfset_factory
* impfset

Comments on the tutorial:
* Add a target at the top of the new tutorial file and use it for linking with a `ref` directive. This will keep links valid if file paths change. See https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#reference-roles
* To make the advantage of a fixture more obvious, add another test to the example that uses `sample_user`
* Should be `attention`: "Never import fixtures directly" https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-attention
   For use of directives in MyST, see https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html#directives-a-block-level-extension-point
* Remove explicit information on fixture contents, like "There are 4 events..." (should be in docstring, as mentioned above) 
* "Tip" should use `tip`, "Note" should use `note` directive https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-tip
* Errors:
   * "Usefu**ll** references"
   * `from climada.entity.exposures` ?
   * `stronger_hazard = ...; # Intensities are halved ...` 😂

climada/test/conftest.py
peanutfun: We should be able to import docstrings from conftest in Sphinx, as the test code is part of the climada package. However, imports may fail (since packages for testing are required).

I asked ChatGPT and it said this:

Screenshot 2026-04-02 at 15 18 56

climada/test/conftest.py
spjuhel: > If these assertions fail, importing this file will result in an error.

That was sort of wanted, in the sense that if the constants you define in the file are inconsistent, then you should have an error as early as possible. 

Though I see your point, I removed it.

climada/test/conftest.py
spjuhel: Done

climada/test/conftest.py
spjuhel: I actually moved these to default arguments of the factory, also because it gives more flexibility to the factory.

climada/test/conftest.py
spjuhel: Done

climada/test/conftest.py
spjuhel: The frequencies are such that their *sorted cumulated sum* is 0, 1/250, 1/100, 1/50, 1/20,
Not their sorted values (inverse order of the array), which are:  0, 1/250, 1/166.666..., 1/100, 1/33.333...

climada/test/conftest.py
spjuhel: Yep, I agree!

spjuhel:
I think I can document fixtures with this: https://sphinx-autofixture.readthedocs.io/en/latest/usage.html#directive-autofixture

Would it be acceptable to push that to a future PR?

climada/test/conftest.py
peanutfun: Follow-up here: https://github.com/CLIMADA-project/climada_python/issues/1278

climada/test/conftest.py
```diff
@@ -0,0 +1,308 @@
...
+
+# ---------------------------------------------------------------------------
+# Impact function
+# ---------------------------------------------------------------------------
+IMPF_ID = 1
+IMPF_NAME = "IMPF_1"
+
+
+@pytest.fixture
+def exposures_factory():
```
peanutfun, line 109: Contrary to the guide, the factories now do **not** support passing arguments!

climada/test/conftest.py
peanutfun: Aahhh I see. Leave as is then!

climada/test/conftest.py
peanutfun: Thanks! At least when I have errors during test discovery with VS Code, they usually are cumbersome to debug 😬

climada/test/conftest.py
peanutfun: Ah sorry, I misunderstood what was happening.

peanutfun:
> I think I can document fixtures with this: https://sphinx-autofixture.readthedocs.io/en/latest/usage.html#directive-autofixture

I don't think that's necessary. It seems to do the same thing as `autofunction`, with the difference that there is a new "role" `fixture` that can be referenced. I don't think we need to add another Sphinx plugin for this functionality.

> Would it be acceptable to push that to a future PR?

Fine given the dependency of #1201.

spjuhel:
Thanks a lot for the swift review @peanutfun, 
Are we good to merge?

peanutfun (approved):

peanutfun:
We are!

PR #1274, opened 2026-04-02
[Option Appraisal Module] [Package 1: Measures] Split 0: Move old measures to _legacy module

spjuhel:
This PRs preps the field for the reworks of the Measure / MeasureSet classes.

Because the design completely changes, tests and code using the legacy code will fail. 

To avoid that while gradually pushing the new code and before the new option appraisal module is fully ready, this PR moves the old code from `measures` to `_legacy_measures` module, and import from there.

### PR Author Checklist

- [x] Read the [Contribution Guide][contrib]
- [x] Correct target branch selected (if unsure, select `develop`)
- [x] Descriptive pull request title added
- [x] Source branch up-to-date with target branch
- [ ] [Documentation](https://climada-python.readthedocs.io/en/latest/guide/Guide_PythonDos-n-Donts.html#2.--Commenting-&-Documenting) updated
- [x] [Tests][testing] updated
- [x] [Tests][testing] passing
- [x] No new [linter issues][linter]
- [x] [Changelog](https://github.com/CLIMADA-project/climada_python/blob/main/CHANGELOG.md) updated

### PR Reviewer Checklist

- [ ] Read the [Contribution Guide][contrib]
- [ ] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/development/Guide_Review.html) passed
- [ ] [Tests][testing] passing
- [ ] No new [linter issues][linter]

[contrib]: https://github.com/CLIMADA-project/climada_python/blob/main/CONTRIBUTING.md
[testing]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html
[linter]: https://climada-python.readthedocs.io/en/latest/development/Guide_continuous_integration_GitHub_actions.html#static-code-analysis

PR #1276, opened 2026-04-07
[Option Appraisal Module] [Package 1: Measures] Split 1: Adds `MeasureConfig` dataclasses for serializable measure configuration.

spjuhel:
Changes proposed in this PR:

This PR introduces a new module defining a family of dataclasses that encode adaptation measure parameters in a serializable, declarative form. This is a preparatory step for the broader `Measure`/`MeasureSet` rework.

The objective is to retain the possibility to define measures from modifiers in addition to the upcoming system that will use python functions directly, and provide a (better) user interface for this.

**no existing code is changed.**

New dataclasses:

- `MeasureConfig` top-level configuration for a single measure, aggregating all modifier configs below.
- `ImpfsetModifierConfig` parameters for modifying impact functions (intensity, MDD, PAA scaling, optional replacement)
- `HazardModifierConfig` parameters for modifying hazard (intensity scaling, frequency cutoff, optional replacement)
- `ExposuresModifierConfig` parameters for modifying exposures (impact function remapping, zeroing regions, optional replacement)
- `CostIncomeConfig` serializable representation of cost/income cash flow parameters

All configs inherit from the abstract `_ModifierConfig` base, which provides shared `to_dict`/`from_dict` and a `__repr__` that highlights non-default fields.

`MeasureConfig` supports full serialization to/from dict, YAML, and legacy Excel rows (from_row), which will serve as the migration path for the existing file-based I/O in `MeasureSet`.

### PR Author Checklist

### PR Reviewer Checklist


climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,318 @@
...
+        ):
+            LOGGER.warning(
+                "Both new impfset object and impfset modifiers are provided, "
+                "modifiers will be applied after changing the impfset."
+            )
+
+
+@dataclass(repr=False)
+class HazardModifierConfig(_ModifierConfig):
+    """Configuration for impact function modifiers."""
```
chahank:
```suggestion
    """Configuration for hazard modifiers."""
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,318 @@
...
+
+    haz_type: str
+    haz_int_mult: Optional[float] = 1.0
+    haz_int_add: Optional[float] = 0.0
+    new_hazard_path: Optional[str] = None
```
chahank, line 286: Wasn't there also a frequency modifier?

chahank:
Excellent idea!

climada/entity/measures/measure_config.py
spjuhel: Either I forgot or didn't see, I can add it :+1:

spjuhel:
@chahank 

Question: Should measures (defined by `MeasureConfig`) be specific to a hazard? (I see reasons for both yes and no)

Currently, they are specific (thus configs require you to provide `haz_type`).

I see how you could want modifications of impact function to be for multiple hazards
But also you should probably not be able to apply a shift in intensity for multiple hazards?

Any opinion on that?

spjuhel:
This PR is officially open for review! Review is advised after #1274 is merged.

Actual changes to be reviewed are in:
- climada/entity/measures/measure_config.py
- climada/entity/measures/test/test_measure_config.py
- doc/user-guide/climada_measure_config.ipynb (and a few other tutos where a target was added at the top)

chahank:
> @chahank
> 
> Question: Should measures (defined by `MeasureConfig`) be specific to a hazard? (I see reasons for both yes and no)
> 
> Currently, they are specific (thus configs require you to provide `haz_type`).
> 
> I see how you could want modifications of impact function to be for multiple hazards But also you should probably not be able to apply a shift in intensity for multiple hazards?
> 
> Any opinion on that?

Good question. I would then to keep it hazard type specific. But maybe best is to ask core users of the model :). 
@ChrisFairless : any opinion?

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+            Fields whose current value equals the dataclass default.
+        """
+
+        non_defaults = {}
+        defaults = {}
+        for defined_field in fields(self):
+            val = getattr(self, defined_field.name)
+            default = defined_field.default
+            if defined_field.default_factory is not field().default_factory:
+                default = defined_field.default_factory()
```
peanutfun: Needs `from dataclasses import MISSING`:
```suggestion
            if default is MISSING:
                default = defined_field.default_factory()
```
See https://docs.python.org/3/library/dataclasses.html#dataclasses.MISSING and https://docs.python.org/3/library/dataclasses.html#dataclasses.Field:
> default, default_factory, init, repr, hash, compare, metadata, and kw_only have the identical meaning and values as they do in the [field()](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) function.

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+class _ModifierConfig(ABC):
+    """
+    Abstract base class for all modifier configuration dataclasses.
+
+    Provides shared serialization, deserialization, and representation
+    logic for all concrete modifier config subclasses. Not intended to
+    be instantiated directly.
+    """
+
+    def _filter_out_default_fields(self):
```
peanutfun:
```suggestion
    def _filter_out_default_fields(self) -> dict[str, Any]:
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+from typing import Dict, Optional, Tuple, Union
+
+import numpy as np
+import pandas as pd
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class _ModifierConfig(ABC):
```
peanutfun: You do not define abstract methods, so you can omit deriving from ABC:

```suggestion
class _ModifierConfig:
```

Also, why private? Just exclude from `all`?

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+            if val != default:
+                non_defaults[defined_field.name] = val
+            else:
+                defaults[defined_field.name] = val
+
+        if "haz_type" in non_defaults:
+            non_defaults.pop("haz_type")
+        return non_defaults, defaults
+
+    def to_dict(self):
```
peanutfun: Judging from its name, I would not expect such a method to omit default values. Even if this is never used, maybe add a parameter?

```suggestion
    def to_dict(self, omit_default : bool = True) -> dict[str, Any]:
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+        -------
+        dict
+            A copy of ``to_filter`` restricted to keys that correspond to declared
+            dataclass fields on this class.
+        """
+
+        filtered = dict(
+            filter(lambda k: k[0] in [f.name for f in fields(cls)], to_filter.items())
+        )
+        return filtered
```
peanutfun: Dict comprehension is easier to read. Also, I am not sure if the list comprehension is evaluated once or each time the lambda runs.
```suggestion
        fields = [f.name for f in fields(cls)]
        return {key: val for key, val in to_filter.items() if key in fields}
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+
+    Warns
+    -----
+    UserWarning
+        If ``new_impfset_path`` is set alongside any non-default modifier
+        values.
+    """
+
+    haz_type: str
+    impf_ids: Optional[Union[int, str, list[Union[int, str]]]] = None
```
peanutfun: I find the operator-style declarations easier to read
```suggestion
    impf_ids: int | str | list[int | str] | None = None
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+    impf_mdd_mult: float = 1.0
+    impf_mdd_add: float = 0.0
+    impf_paa_mult: float = 1.0
+    impf_paa_add: float = 0.0
+    impf_int_mult: float = 1.0
+    impf_int_add: float = 0.0
+    new_impfset_path: Optional[str] = None
```
peanutfun:
```suggestion
    new_impfset_path: str | None = None
```

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+        modifiers are applied after loading the new hazard.
+    impact_rp_cutoff : float, optional
+        Return period (in years) below which hazard events are discarded.
+        If ``None``, no cutoff is applied.
+
+    Warns
+    -----
+    UserWarning
+        If ``new_hazard_path`` is set alongside any non-default modifier
+        values or a non-``None`` ``impact_rp_cutoff``.
```
peanutfun, line 287: Why does this issue a warning at all? Sounds like you are warning about valid and expected behavior. Just needs to be documented clearly.

Also, the warning is issued in more instances than just a non-None impact_rp_cutoff

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+
+    def __post_init__(self):
+        config = self.to_dict()
+        if "new_exposures_path" in config and any(
+            key in config for key in ["reassign_impf_id", "set_to_zero"]
+        ):
+            warnings.warn(
+                "Both new exposures object and exposures modifiers are provided, "
+                "modifiers will be applied after changing the exposures."
+            )
```
peanutfun, line 359: Again, why warn about this?

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+        Recurring cost per period (positive value). Default is ``0.0``.
+    periodic_income : float, optional
+        Recurring income per period. Default is ``0.0``.
+    cost_yearly_growth_rate : float, optional
+        Annual growth rate applied to periodic costs. Default is ``0.0``.
+    income_yearly_growth_rate : float, optional
+        Annual growth rate applied to periodic income. Default is ``0.0``.
+    freq : str, optional
+        Pandas offset alias defining the period length (e.g. ``"Y"`` for
+        yearly, ``"M"`` for monthly). Default is ``"Y"``.
```
peanutfun, line 386: Not sure exactly how these values are used, but `Y` and `M` are deprecated as frequencies. You need to also specify start or end, like `YS` or `ME`, see https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+        The file is structured as ``{"measures": [<this config as dict>]}``,
+        matching the expected format for :meth:`from_yaml`.
+
+        Parameters
+        ----------
+        path : str
+            Destination file path. Will be created or overwritten.
+        """
+
+        import yaml
```
peanutfun: Need to add a dependency, preferably ruamel.yaml. Why no top-level import?

climada/entity/measures/measure_config.py
```diff
@@ -0,0 +1,642 @@
...
+            Path to the YAML file to read.
+
+        Returns
+        -------
+        MeasureConfig
+            The configuration parsed from the first entry in
+            ``measures``.
+        """
+
+        import yaml
```
peanutfun: As above

climada/entity/measures/test/test_measure_config.py
```diff
@@ -0,0 +1,515 @@
...
+
+
+def test_modifier_config_filter_dict_to_fields_filters_extra_keys():
+    d = {"haz_type": "TC", "impf_mdd_mult": 0.5, "not_a_field": 123}
+    filtered = ImpfsetModifierConfig._filter_dict_to_fields(d)
+    assert "not_a_field" not in filtered
+    assert "haz_type" in filtered
+    assert "impf_mdd_mult" in filtered
+    assert filtered["haz_type"] == "TC"
+    assert filtered["impf_mdd_mult"] == 0.5
```
peanutfun: Private methods are private also in the context of tests. Unit tests should not call private methods. A linter will also complain. Private methods must be implicitly tested via calling public methods.

climada/entity/measures/test/test_measure_config.py
```diff
@@ -0,0 +1,515 @@
...
+
+def test_modifier_config_repr_empty_when_all_defaults():
+    config = ImpfsetModifierConfig(haz_type="TC")
+    r = repr(config)
+    assert "Non default fields" not in r
+
+
+# ---------------------------------------------------------------------------
+# ImpfsetModifierConfig
+# ---------------------------------------------------------------------------
```
peanutfun: Create a class to group tests? Would make test names shorter. Pytest creates a sub-suite, grouping the tests accordingly:

```python
class TestImpfsetModifierConfig:

    def test_defaults(self):
        ...
    def test_from_dict_roundtrip(self):
        ...
```

peanutfun (changes requested):
I reviewed mostly on a technical level. Very nice code quality overall, nice class structure! I found a few issues. Sorry for the brevity 😬 ✌️

climada/entity/measures/measure_config.py
spjuhel: Yup, good point

climada/entity/measures/measure_config.py
spjuhel: It was initially a logging warning, and the underlying idea was that you should really be aware of what goes on. In most cases user probably either want to provide a new hazard or modify an existing one, not do both. But it can also happen from e.g., combining measures.
I'm not convinced a warning is the best, but I also think good documentation might be too little?

climada/entity/measures/measure_config.py
spjuhel: I remember this was tricky.
The code uses both offset (the ones you mentionned) and period frequency ([these](https://pandas.pydata.org/docs/user_guide/timeseries.html#period-aliases)), which are different.

Currently, if I set "YS" in a CostIncome object I get the error below.

I did not find a way to make it clean.

I should at least change "offset alias" to "period alias" though! And put a link.

--

```python-traceback
File [~/Repos/climada_python/climada/entity/measures/cost_income.py:401](http://localhost:8888/lab/tree/Projects/MCDM_simona/Repos/climada_python/climada/entity/measures/cost_income.py#line=400), in CostIncome.calc_cash_flows(self, impl_date, start_date, end_date)
    368 """Calculate net cash flows, costs, and incomes over a period.
    369 
    370 Computes cash flow metrics across a specified date range by iterating
   (...)    397         Total incomes for each period.
    398 """
    400 impl_ts = pd.Timestamp(impl_date)
--> 401 periods = pd.period_range(start=start_date, end=end_date, freq=self.freq)
    403 results = [self._calc_at_date(impl_ts, p.start_time) for p in periods]
    404 net, costs, incs = map(np.array, zip(*results))

File [~/miniforge3/envs/climada_env_dev/lib/python3.11/site-packages/pandas/core/indexes/period.py:611](http://localhost:8888/lab/tree/Projects/MCDM_simona/miniforge3/envs/climada_env_dev/lib/python3.11/site-packages/pandas/core/indexes/period.py#line=610), in period_range(start, end, periods, freq, name)
    608 if freq is None and (not isinstance(start, Period) and not isinstance(end, Period)):
    609     freq = "D"
--> 611 data, freq = PeriodArray._generate_range(start, end, periods, freq)
    612 dtype = PeriodDtype(freq)
    613 data = PeriodArray(data, dtype=dtype)

File [~/miniforge3/envs/climada_env_dev/lib/python3.11/site-packages/pandas/core/arrays/period.py:340](http://localhost:8888/lab/tree/Projects/MCDM_simona/miniforge3/envs/climada_env_dev/lib/python3.11/site-packages/pandas/core/arrays/period.py#line=339), in PeriodArray._generate_range(cls, start, end, periods, freq)
    337 periods = dtl.validate_periods(periods)
    339 if freq is not None:
--> 340     freq = Period._maybe_convert_freq(freq)
    342 if start is not None or end is not None:
    343     subarr, freq = _get_ordinal_range(start, end, periods, freq)

File pandas[/_libs/tslibs/period.pyx:1765](http://localhost:8888/_libs/tslibs/period.pyx#line=1764), in pandas._libs.tslibs.period._Period._maybe_convert_freq()

File pandas[/_libs/tslibs/offsets.pyx:4970](http://localhost:8888/_libs/tslibs/offsets.pyx#line=4969), in pandas._libs.tslibs.offsets.to_offset()

ValueError: YS-JAN is not supported as period frequency
```

climada/entity/measures/measure_config.py
spjuhel: Yup, unsure about that, I saw some people deferring these imports to only when they are strictly needed (presumably to make dependencies related to IO optional?)
But yeah probably top-level makes more sense for us.

spjuhel: @peanutfun Why ruamel.yaml over PyYAML? The latter seems more common and with a stronger community/support.

climada/entity/measures/measure_config.py
peanutfun: Maybe just reduce the urgency? Isn't INFO enough?

climada/entity/measures/measure_config.py
peanutfun: Good catch!

climada/entity/measures/measure_config.py
peanutfun: PyYAML once opted for a default unsafe loader and was frowned upon for quite some time. They backpedaled, but I didn't follow-up on the later discussion. Anyway, we have both already in our dependency tree, so you can pick either 😂 
```
$ (climada_env) conda list | grep yaml
pyyaml                              6.0.3            py312h5748b74_0          conda-forge
ruamel.yaml                         0.19.1           pyhcf101f3_0             conda-forge
ruamel.yaml.clib                    0.2.15           py312hb3ab3e3_1          conda-forge
yaml                                0.2.5            h925e9cb_3               conda-forge
```

spjuhel:
@ValentinGebhart This has been reviewed on the technical aspects. Can you give a look at the guide [here](https://climada-python--1276.org.readthedocs.build/en/1276/user-guide/climada_measure_config.html) and tell me if its clear enough?

Note: The "measure tutorial" link at the end does not work yet and is a placeholder.

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c68c6cf1-d0a1-40ae-ae45-838741988ac6",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "CLIMADA uses `Measure` objects to model the effects of adaptation measures. `Measure` objects were formerly defined declaratively (via for instance, a shifting or scaling of the hazard intensity or a change of impact function), and are now defined as python functions to enable more flexibility on the possible changes (see the [tutorial on measure objects](measure-tutorial)'). \n",
+    "\n",
+    "The caveat of defining measure effects as python functions is that it cannot be serialized (written to a file), and also makes reading from a file a challenge.\n",
```
ValentinGebhart, line 22: Is it really impossible to save python functions? That means that measures that do not correspond to MeasureConfigs cannot be saved?

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "\n",
+    "In order to close that gap, the `measure` module now ships `MeasureConfig` objects, which handle the reading, writing and \"declarative\" defining of `Measure` objects.\n",
```
ValentinGebhart: If I understood correctly, this sounds a bit misleading, as if MeasureConfig handle reading and writing of ALL Measure objects, while actually they can only read and write those that correspond to MeasureConfig objects, right?
Maybe rephrase to
"which can be read from and written to files, and which enable a \"declarative\" defining of `Measure` objects."

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "\n",
+    "The definition of measures via `MeasureConfig` is organized into a hierarchy of specialized classes:\n",
+    "\n",
+    "- `MeasureConfig`: The top-level container for a single measure.\n",
+    "- `HazardModifierConfig`: Defines how the hazard is changed (e.g., shifting intensity).\n",
+    "- `ImpfsetModifierConfig`: Adjusts impact functions (e.g., scaling vulnerability curves).\n",
+    "- `ExposuresModifierConfig`: Modifies exposure data (e.g., reassigning IDs or zeroing regions).\n",
+    "- `CostIncomeConfig`: Handles the financial aspects, including initial costs and recurring income.\n",
+    "\n",
+    "Note that everything can be defined and accessed directly from the `MeasureConfig` container, the underlying ones are there to keep things organized.\n",
```
ValentinGebhart: On first reading the meaning was not super clear to me. Maybe rephrase to 
"Note that all effects of a `MeasureConfig` object can be directly defined using the `MeasureConfig` container, the underlying ones are there to keep things organized.

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "| Scope | Parameter | Type | Description |\n",
+    "| :--- | :--- | :--- | :--- |\n",
+    "| **Top-Level** | `name` (required) | `str` | Unique identifier for the measure. |\n",
+    "| | `haz_type` (required) | `str` | The hazard type this measure targets (e.g., \"TC\", \"FL\"). |\n",
+    "| | `implementation_duration` | `str` | Pandas offset alias (e.g., \"2Y\") for implementation time. |\n",
+    "| | `color_rgb` | `tuple` | RGB triple (0-1 range) for plotting and visualization. |\n",
+    "| **Hazard** | `haz_int_mult` | `float` | Multiplier for hazard intensity (default: 1.0). |\n",
+    "| | `haz_int_add` | `float` | Additive offset for hazard intensity (default: 0.0). |\n",
+    "| | `new_hazard_path` | | Path to an HDF5 file to replace the current hazard. |\n",
+    "| | `impact_rp_cutoff` | `float` | Return period (years) threshold; events below this are ignored. |\n",
```
ValentinGebhart:
```suggestion
    "| | `impact_rp_cutoff` | `float` | Return period (years) threshold; events with impacts below this threshold are ignored. |\n",
```

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "# 1. Changing existing Exposures\n",
+    "exp_mod = ExposuresModifierConfig(\n",
+    "    reassign_impf_id={\n",
+    "        \"TC\": {1: 2}\n",
+    "    },  # Reassigns exposures points with impact function id impf_TC == 1 to 2.\n",
+    "    set_to_zero=[\n",
+    "        0,\n",
+    "        25,\n",
+    "        78,\n",
+    "    ],  # Sets the value of exposure points with index 0, 25 and 78 to 0.\n",
```
ValentinGebhart, line 428: This seems in conflict to what you wrote above. Here it sounds that you set to zero exposure points with specific index values, above it sounds that you set to zero exposure points with specific region_ids ("List of Region IDs where exposure value is set to 0."

ValentinGebhart, line 428: This seems in conflict to what you wrote above. Here it sounds that you set to zero exposure points with specific index values, above it sounds that you set to zero exposure points with specific region_ids ("List of Region IDs where exposure value is set to 0."

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "\n",
+    "# This models a measure where costs increase by 2% annually,\n",
+    "# but it generates 100k in yearly income which grows by 3%.\n",
+    "growth_finance = CostIncomeConfig(\n",
+    "    init_cost=500_000.0,\n",
+    "    periodic_cost=20_000.0,\n",
+    "    cost_yearly_growth_rate=0.02,\n",
+    "    periodic_income=100_000.0,\n",
+    "    income_yearly_growth_rate=0.03,\n",
+    "    freq=\"Y\",\n",
```
ValentinGebhart: The freq attribute is not shown in the below print, and it not entirely clear to me what it does. Above, it is written "frequecy of cash flows". That means that if one uses, e.g., freq="2Y", the growth rates etc are still yearly but the the costs and incomes occur only every two years?

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "print(growth_finance)\n",
+    "\n",
+    "\n",
+    "# Custom Cash Flow\n",
+    "# If the investment isn't linear (e.g., a major retrofit in year 5),\n",
+    "# you can define a list of specific events.\n",
+    "custom_schedule = [\n",
+    "    {\"date\": \"2024-01-01\", \"value\": -1000000},  # Initial cost\n",
+    "    {\"date\": \"2029-01-01\", \"value\": -200000},  # Mid-term overhaul\n",
+    "    {\"date\": \"2034-01-01\", \"value\": 500000},  # Terminal value\n",
```
ValentinGebhart: Maybe clarify what negative and positive values are, and above cost were given as positive values. I assume here negative values are one-time costs and negative values are one-time incomes?

ValentinGebhart (approved):
Thanks for this very clear tutorial! I added only a few rephrasings and some comments at points where I thought it was not perfectly clear. However all are only minor things, and I find the explanations very clear about how to define the MeasureConfigs.

doc/user-guide/climada_measure_config.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "\n",
+    "`HazardModifierConfig` allows you to modify the intensity and frequency of the hazard, to apply a cutoff on the return period of impacts, as well as to replace it entirely:\n",
+    "\n",
+    "- The intensity matrix: via `haz_int_mult` to scale it and `haz_int_add` to shift it.\n",
+    "- The frequency array: via `haz_freq_mult` to scale it and `haz_freq_add` to shift it.\n",
+    "- Replacing the hazard: via providing the `new_hazard_path` parameter. It needs to be a valid hazard HDF5 file readable by `Hazard.from_hdf5()`\n",
+    "- Applying a cutoff on frequency based on impacts: via `impact_rp_cutoff` (see the note).\n",
+    "\n",
+    "```{note}\n",
+    "Providing a value for `impact_rp_cutoff` \"removes\" (it sets their intensity to 0.) events from the hazard, for which the exceedance frequency (inverse of return period) of impacts is below the given threshold.\n",
```
ValentinGebhart, line 278: I was actually wondering about this one: AFAIK you want to have each measure consist of three functions that modify each of the risk components. This parameter however looks like it takes into account all hazard, vulnerability and exposure, as it looks at the impacts per event, which seems a bit confusing with the three function structure. My main question is if this cutoff is applied at the start (computing the impact events to be excluded before applying all other stuff), or at the end (after all other effect were taken into account)?

doc/user-guide/climada_measure_config.ipynb
spjuhel: > Is it really impossible to save python functions?

As binary or encoded files, no, not reasonably. 

> That means that measures that do not correspond to MeasureConfigs cannot be saved?

You can still save the `.py` / `.ipynb` files where you define the functions, just as you would save pieces of code to reuse.

doc/user-guide/climada_measure_config.ipynb
spjuhel: Good catch! It is indeed not the region IDs, but the exposures IDs.

doc/user-guide/climada_measure_config.ipynb
spjuhel: Yes, I agree "frequency" is not the best (it is the name used by pandas). This is (or should be) explained in more details in the cost income tutorial.

The "Y" does not appear in the print because it is the default value (I filter them out, to avoid cluttering the output). I see how this can be confusing here, so I've removed the "Y" from the definition. 

I will also add a note on this filtering earlier in the tutorial (I intended to and forgot...)

doc/user-guide/climada_measure_config.ipynb
spjuhel: Done!

doc/user-guide/climada_measure_config.ipynb
spjuhel: Yes this one is tricky.

The actual requirement for the three functions is to return an object of the corresponding component, there's actually no real requirement on the inputs.

In this case the function does return a `Hazard` object, using the other components in its definition.

The created function does this:
```
        Parameters of the returned callable:

        hazard : Hazard
            The hazard object to modify in-place (intensity rows zeroed).
        base_exposures : Exposures
            Exposures used for the reference impact computation.
        base_impfset : ImpactFuncSet
            Impact function set used for the reference impact computation.
        base_hazard : Hazard
            Original hazard used for the reference impact computation.
        exposures_region_id : list of int, optional
            If provided, the impact computation is restricted to exposure
            points whose ``region_id`` is in this list.
```

The "base" triplet is always provided by default to the <>_changes methods of `Measure`. So in that sense the changes are independent of other changes on the hazard.

The actual code implementing this is here (future PR): https://github.com/CLIMADA-project/climada_python/pull/1284/changes#diff-4775895db992dde9e5d10082cd3a65e729ade35b8000fa60cd85048b117faaaaR134

doc/user-guide/climada_measure_config.ipynb
spjuhel: Done

doc/user-guide/climada_measure_config.ipynb
spjuhel: Done

PR #1277, opened 2026-04-07
[Option Appraisal Module] [Package 1: Measures] Split 2: Adds `CostIncome` class for measure financial aspects management

spjuhel:
Changes proposed in this PR:

Introduces `CostIncome` class to represent the financial aspect of adaptation measures (`Measures`) (essentially initial and maintenance costs and possible income) 

Tutorial here: https://climada-python--1277.org.readthedocs.build/en/1277/user-guide/climada_cost_income.html

### PR Author Checklist

### PR Reviewer Checklist


spjuhel:
This PR is officially open for review!

@luseverin would you be so kind to review the tutorial (to spare @chahank who has to review a lot of my code lately)
Feel free to look at the code too, but I think that part needs to be reviewed by a Code Owner in any case.

luseverin:
Hi @spjuhel, yes I'll be happy to review the tutorial :)

doc/user-guide/climada_cost_income.ipynb
luseverin, line 1: Cell 1:
- Do you need to specify the sign convention? I think it might be confusing as the users anyways input the costs with a positive sign.
- Maybe already define the meaning of the values Y, M, Q for the freq parameter here?

Cell 2:
- comment not matching code (20000 vs 50000 and 12000 vs 8000)

Cell 3:
- "Three methods are available **to calculate cash flows**:
- I would maybe repeat the parameters instead of the ... for clarity

Cell 5:
- Consider an option to add a row with totals in the dataframe format

Cell 6:
- ... draws a bar chart **of the cash flows**

Cell 7:
- Is it voluntary that you do not compare the same `init_cost`, `periodic_cost` and `periodic_income` between the with- and without-growth instances? See the comment on cell 2.

Cell 10:
- Not exactly sure of what is meant with "annualised" but it is probably because I am not familiar with the field :) I think I get the idea though

Cell 11:
- Do we care about the days when we consider the monthly case? What happens if we put 2022-01-15 instead of 2022-01-01 for the date? Maybe you could shortly comment on that (if relevant)

Cell 13:
- Maybe add an example of the manual merge of the custom flows?

Cell 15:
- The From YAML part seems to be incomplete here, but you are probably aware of it :)

luseverin (commented):
Hey @spjuhel,

I just reviewed the tutorial, great work! As a new user of the module, with a limited knowledge on the topic I could understand everything, so I think it is a good sign! The module looks very neat and I look forward to experimenting it, congrats! I had only minor comments (see comments on the file). Apologises as I wrote them as a block because I could not find the option to do the line by line (or cell by cell) commenting in the notebook directly.

doc/user-guide/climada_cost_income.ipynb
spjuhel: Hi @luseverin,
Thanks for the review!

Here are some of the changes I made (also based on @chahank remarks):

- Improved the Quickstart to show more of the capabilities right from the start, with minimal amount of scrolling.
- Added meaning of periods aliases as well as url to pandas documentation
- Aligned the values with comment as well as for the different examples
- Improved description of the plotting function
- Improved details on sub-yearly period frequencies
- Improved description of the merging (also now custom cashflows are merged)

I think I adressed all other comments.
For the YAML part, I don't want the notebook to have to write a file, so the code is within the markdown, not as a python cell. I added a bit more to it.

spjuhel:
@ValentinGebhart this one still requires code review. You can ignore the files related to MeasureConfig (they'll be reviewed in the other PR) and focus on `cost_income.py` and its tests.
Of course have a look at the tutorial, but Luca did go through it already.

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+class CostIncome:
+    """
+    Manages costs and incomes related to a measure over time.
+
+    Income are stored a positive numbers and costs as negative
+    ones.
+
+    Attributes
+    ----------
+    mkt_price_year : datetime, default to today's year.
```
ValentinGebhart: I think also str or int or datetime object word, right?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+    periodic_income : float
+        Recurring income per period.
+    cost_yearly_growth_rate : float
+        Yearly growth rate of costs.
+    income_yearly_growth_rate : float
+        Yearly growth rate of income.
+    custom_cash_flows : pd.DataFrame, optional
+        User-defined cash flows indexed by date.
+    freq : str
+        Frequency of the cash flows (e.g., 'Y', '3M', '7D').
```
ValentinGebhart: you could add that all pandas period aliases are allowed?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+
+    """
+
+    def __init__(
+        self,
+        *,
+        mkt_price_year: Optional[int] = None,
```
ValentinGebhart, line 65: see comment above, maybe adapt docstring if only ints work here

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+        cost_yearly_growth_rate: float = 0.0,
+        income_yearly_growth_rate: float = 0.0,
+        custom_cash_flows: Optional[pd.DataFrame] = None,
+        freq: str = "Y",
+    ):
+        """Initialize CostIncome with parameters.
+
+        Parameters
+        ----------
+        mkt_price_year : datetime, default to today's year.
```
ValentinGebhart: see type above

climada/entity/measures/cost_income.py
ValentinGebhart: from code below this should be int

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+        config : CostIncomeConfig
+
+        Returns
+        -------
+        CostIncome
+        """
+
+        df = None
+        if config.custom_cash_flows is not None:
+            df = pd.DataFrame(config.custom_cash_flows)
```
ValentinGebhart, line 205: Just a note for compatibility with config, that you probably already are aware. Here is seems the config.custom_cash_flows must have cost and income columns, but in the config is seems only to have values which can be positive or negative:

This currently breaks (copied definition from other tutorial)
`````python
custom_schedule = [
    {"date": "2024-01-01", "value": -1000000},  # Initial cost
    {"date": "2029-01-01", "value": -200000},  # Mid-term overhaul
    {"date": "2034-01-01", "value": 500000},  # Terminal value
]

custom_finance = CostIncomeConfig(custom_cash_flows=custom_schedule)
ci = CostIncome.from_config(custom_finance)
`````

File: climada/entity/measures/cost_income.py [hunk 6]
```diff
@@ -0,0 +1,568 @@
...
+
+        Returns
+        -------
+        CostIncome
+        """
+
+        import yaml
+
+        with open(path) as f:
+            return cls.from_dict(yaml.safe_load(f)["cost_income"])
```
ValentinGebhart, line 262: why do you require that the yaml have a cost_income section? I guess becasue the yaml might correspond to the whole measure object, including other information as well? Then you could add it somewhere in the docstring, like "Create a `CostIncome` from a yaml file of a MeasureConfig object" or so

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+
+    @classmethod
+    def _freq_to_days(cls, freq: str) -> str:
```
ValentinGebhart: just by trying this method I get
CostIncome._freq_to_days("Y") gives '365d'
CostIncome._freq_to_days("YE") gives '365d'
CostIncome._freq_to_days("YS") gives '366d'

is this intended behaviour?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+
+        The calculation applies compound growth to both costs and incomes based
+        on the number of years elapsed since the market price reference date
+        (`self.mkt_price_year`).
+
+        Parameters
+        ----------
+        impl_date : pd.Timestamp
+            The implementation date that determines which cost/income regime
+            applies. Dates before this use have no cost or income; "at the date" uses
```
ValentinGebhart: typo:
Dates before this use have no cost or income
Dates before this have no cost or income

also: using the custom_cash_flows, couldn't you have a cost before the implementation date? If relevant, you can adapt the docstring here

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+        Computes cash flow metrics across a specified date range by iterating
+        through each period.
+
+        The method creates a period range based on the configured frequency
+        (`self.freq`) and evaluates cash flows at the start time of each period.
+        Results are returned as NumPy arrays for efficient downstream processing.
+
+        Parameters
+        ----------
+        impl_date :
```
ValentinGebhart: is it on purpose that these parameters do not have type suggestions? Because there would be too many that work?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+            * costs : np.ndarray
+                Total costs for each period.
+            * incomes : np.ndarray
+                Total incomes for each period.
+        """
+
+        impl_ts = pd.Timestamp(impl_date)
+        periods = pd.period_range(start=start_date, end=end_date, freq=self.freq)
+
+        results = [self._calc_at_date(impl_ts, p.start_time) for p in periods]
```
ValentinGebhart: the single-date cash flows are here always computed with the start time of the period. Is this consistent with how they are computed? Could it not happen that sometimes a cash flow is left out if it is "in the middle" of a period?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+        end_date :
+            End of the evaluation period.
+        figsize : tuple, optional
+            Figure size as (width, height). Default is (12, 7).
+        title : str, optional
+            Overall figure title. Defaults to 'Cash Flow Analysis'.
+
+        Returns
+        -------
+        plt.Figure
```
ValentinGebhart: the code seems to return tuple of axis (ax_bar, ax_net), which is not the same as plt.Figure?

climada/entity/measures/cost_income.py
ValentinGebhart: It looks to me that this private method is not used anywhere? Will it be used somewhere else?

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def _get_width_days(self) -> float:
+        """Return the number of days in the current frequency."""
+
+        ref = pd.Timestamp("2000-01-01")
+        freq = self._make_offset_compat(self.freq, start=False)
+        offset = pd.tseries.frequencies.to_offset(freq)
+        return float(((ref + offset) - ref).days)
+
+    def _calc_at_date(
```
ValentinGebhart: not clear if this is oblic or private function? looks private here but public in test?

ValentinGebhart (approved):
Thanks @spjuhel for this nice addtition and structure. It think it is well written and seems to compute what it should compute. Also well tested. I added a few comments but I think most of them are minor. Great work!

climada/entity/measures/cost_income.py
spjuhel: Indeed!

climada/entity/measures/cost_income.py
spjuhel: Done!

climada/entity/measures/cost_income.py
spjuhel: Solved!

climada/entity/measures/cost_income.py
spjuhel: Ah good catch! I have to update the config tutorial...

climada/entity/measures/cost_income.py
spjuhel: Indeed, this could be improved to handle different cases, but I think this is low priority.
I've updated the docstring to state that it requires a `MeasureConfig` yaml

climada/entity/measures/cost_income.py
spjuhel: Short answer yes.

Long answer is: years and months are not unambiguous period lengths. So this is a "trick" to transform these periods into a number of days. I spent a lot of time trying to figure out the "best" way to have something seemingly robust and resulted with this.

Yes, it does create some discrepancies. But I think this should be fine in 99.9% of the cases. 

I'm also happy to accept suggestions.

climada/entity/measures/cost_income.py
spjuhel: adapted the docstring

climada/entity/measures/cost_income.py
spjuhel: Nope, that's a miss.

climada/entity/measures/cost_income.py
spjuhel: That's true. I have to think about this.

I think I might restrict the custom cash flows to be on the same periodicity...

climada/entity/measures/cost_income.py
spjuhel: Legacy docstring. I updated it.

climada/entity/measures/cost_income.py
spjuhel: Good question. I have to double check ahah.

climada/entity/measures/cost_income.py
spjuhel: Ah, actually I do this which should cover this case:

```python       
freq = self._make_offset_compat(self.freq)
return df.resample(freq).sum()
```

But indeed I should test this better and maybe document.

climada/entity/measures/cost_income.py
spjuhel: I used it in a former version of the plotting function, but I'm not using it any more. I'll remove it.

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,592 @@
...
+You should have received a copy of the GNU General Public License along
+with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
+
+---
+
+Define the CostIncome class to handle the cash flow of measures.
+"""
+
+from datetime import datetime
+from typing import Any, Optional, Tuple, cast
```
emanuel-schmid, line 23: `typing.Tuple` is out, `tuple` is in

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,592 @@
...
+    freq : str
+        Periodicity (frequency) of the cash flows (e.g., 'Y', '3M', '7D').
+        All pandas aliases are possible. See [pandas documentation](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#period-aliases).
+
+    """
+
+    def __init__(
+        self,
+        *,
+        mkt_price_year: Optional[int] = None,
```
emanuel-schmid, line 65:
```suggestion
        mkt_price_year: int | None = None,
```

suggested by ruff

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,592 @@
...
+        income_yearly_growth_rate : float
+            Yearly growth rate of income.
+        custom_cash_flows : pd.DataFrame, optional
+            User-defined cash flows indexed by date.
+        freq : str
+            Frequency of the cash flows (e.g., 'Y', '3M', '7D').
+        """
+
+        self.freq = freq
+        self.mkt_price_year = datetime(mkt_price_year or datetime.today().year, 1, 1)
```
emanuel-schmid, line 97: (some) linter complain about no tzinfo

climada/entity/measures/cost_income.py
```diff
@@ -0,0 +1,592 @@
...
+        """Combine multiple CostIncomes together.
+
+        Combination sums the costs and incomes from all provided CostIncome
+        objects.
+        """
+
+        first_ci = cost_incomes[0]
+
+        if not all(
+            (
```
emanuel-schmid, line 537: too many paranthesises

emanuel-schmid (changes requested):
just forewarding linter comments 😁 
(what's your linter? mine is ruff)

PR #1279, opened 2026-04-08
[Option Appraisal Module] [Package 1: Measures] Split 3: Adds new base `Measure` class

spjuhel:
Changes proposed in this PR:

This PR brings the basis of the refactored `Measure` class, using closures (functions that return a function) to define the effects of a measure.

Important points:

- Effects are separated by effects on `Exposures`, effects on `Impfset`, and on `Hazard`
- The closures **need to accept keyword arguments** to enable effects that rely on something else than what the effect is applied on (i.e. year, other part of risk, etc...)
- We always pass the Exposures, Impfset and Hazard objects as kwargs to the closures by default
- We pass user-defined function through an `allow_kwargs` to make sure closures always accept kwargs.
- `Measure` instances optionally have a `_config` attribute that contains a `MeasureConfig` if the measure was defined from such an object (and is therefore serializable).

### PR Author Checklist

### PR Reviewer Checklist


doc/user-guide/climada_measures.ipynb
```diff
@@ -0,0 +1,267 @@
...
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "66e0629f-da41-42e9-b3d5-02458c88d693",
+   "metadata": {},
+   "source": [
+    "The measure can then be used to create a new triplet corresponding to the modifications or to compute the impact directly:\n",
+    "\n",
+    "```python\n",
+    "new_exp, new_impf, new_haz = retrofitting.apply(exp, impf, haz, kwargs_hazard={'threshold': 3.5})\n",
+    "impact_with_measure = measure.calc_impact(exp, impf, haz)\n",
```
ValentinGebhart, line 80: This lines is not clear to me. Shouldn't it be
`impact_with_measure = retrofitting.calc_impact(exp, impf, haz)`?

Also, do I need the first line before running the second one? Or are they two separate things you can to with the measure object? It might make sense to clarify with to brief descriptions?

doc/user-guide/climada_measures.ipynb
```diff
@@ -0,0 +1,267 @@
...
+    "\n",
+    "1. **`exposures_changes`**: Modifies the `Exposures` object, the asset location, value, or metadata (e.g., relocating houses from a flood zone).\n",
+    "2. **`impfset_changes`**: Modifies the `ImpactFuncSet` object, the vulnerability of assets (e.g., updating an impact function because houses were elevated).\n",
+    "3. **`hazard_changes`**: Modifies the `Hazard` object, the intensity or frequency of events (e.g., a dam reducing flood depths).\n",
+    "\n",
+    "```{note}\n",
+    "By default, any component not explicitly defined uses the `identity_function`, which simply returns the input object without modification.\n",
+    "```\n",
+    "\n",
+    "The **only** requirement for these python functions is to **take** and **return** an `Exposures` object (or `ImpactFuncSet` or `Hazard`), which enable advanced modeling.\n",
```
ValentinGebhart, line 102: I suggest to make these two sentences a bit clearer. Eg., you do not say what `base_exposures`, `base_impfset`, `base_hazard` are. One suggestion:
"
The **only** requirement for these python functions is to **take** and **return** an `Exposures` object (or `ImpactFuncSet` / `Hazard` objects, respectively),  a structure that enables full modeling flexibility.
Notably, the `apply()` method always passes the entire input triplet of exposure, impact function set, and hazard to every transformation function, so you define measures based on interactions between all three components.
"

doc/user-guide/climada_measures.ipynb
```diff
@@ -0,0 +1,267 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "63b37e27-f4d1-4a0c-8a45-f923aea903e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "\n",
+    "def reduce_exposure_by_depth(exposures, reduction_factor=0.2, threshold=2.0, **kwargs):\n",
```
ValentinGebhart, line 122: by only reading this tutorial and knowing that in the back you also also hand in base_exposures, base_impfset, and base_hazard to all functions, I am just wondering if this is not a bit confusing. E.g., could the exposures input be different to the base_exposures input? What would that mean? (Note that below you actually note this, but it might be nice to give an example for a case where this is necessary/useful). 

Just a naive thought without looking at the code: 
Couldn't each of the three functions take all (exp, impfset, haz) as input, and no hidden base objects? e.g. exposures_changes is always a function that has these three inputs. Anyways if it is usefull to have an expsosure input but also a hidden base_exposure, disregard this suggestion.

doc/user-guide/climada_measures.ipynb
```diff
@@ -0,0 +1,267 @@
...
+    ")\n",
+    "\n",
+    "# When you call apply, the `Hazard` haz is automatically passed to the function\n",
+    "# as 'base_hazard'\n",
+    "new_exp, _, _ = flood_protection.apply(\n",
+    "    exp, impf, haz, \n",
+    "    kwargs_exposures={'reduction_factor': 0.5, 'threshold': 2.0}\n",
+    ")\n",
+    "\n",
+    "# You can also overide this with another \"base\" hazard in the \n",
```
ValentinGebhart, line 170: see comment above: can you give a simple example about when this is necessary or useful?

ValentinGebhart, line 170: see comment above: can you give a simple example about when this is necessary or useful?

doc/user-guide/climada_measures.ipynb
```diff
@@ -0,0 +1,267 @@
...
+    "import numpy as np\n",
+    "\n",
+    "\n",
+    "def reduce_exposure_by_depth(exposures, reduction_factor=0.2, threshold=2.0, **kwargs):\n",
+    "    \"\"\"\n",
+    "    Reduces asset value by a percentage, but only for assets located where\n",
+    "    the peak hazard intensity (e.g., water depth) exceeds the threshold.\n",
+    "    \"\"\"\n",
+    "    # Retrieve the hazard context passed by Measure.apply\n",
+    "    base_hazard = kwargs.get(\"base_hazard\")\n",
```
ValentinGebhart, line 128: as these base_object appear a few times here, it might sense to explain briefly what they represent.

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+    >>> add(1, 2, extra=99)
+    3
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        # Get the names of arguments the original function accepts
+        params = inspect.signature(func).parameters
+
+        # Filter kwargs to only include what the function can handle
```
ValentinGebhart, line 112: reading these comments first I was a bit confused. Could you just write here: 
`# Return function if it already accepts kwargs`

and then below, when you actually filter:
`# Filter kwargs to only include what the function can handle`

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+
+    Finally it holds an `implementation_duration` attribute, in the form of a
+    pandas ``DateOffset``, which is used when the time dimension is considered.
+
+    Notes
+    -----
+
+    The only requirement for each function is to return an object of the same
+    class (e.g. :class:`Hazard` for ``hazard_change``). Functions can accept
+    keyword arguments to enable advanced effect (depending on a year of
```
ValentinGebhart, line 145: could you show this example maybe in the tutorial? The examples there do not yet a time component if I am not mistaken

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+    ----------
+    name : str
+        Name of the measure.
+    exposures_change : ExposuresChange
+        Function to change exposures.
+    impfset_change : ImpfsetChange
+        Function to change impact function set.
+    hazard_change : HazardChange
+        Function to change hazard.
+    sub_measures : list of str, optional
```
ValentinGebhart, line 166: above you write that "A ``Measure`` represents a single adaptation or risk-reduction action". Can you explain the sense in which it can consist of sub_measures and how they combine with the three change functions?

ValentinGebhart: or is this just for bookkeeping but not used operationally?

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+
+    def apply_exposures_changes(
+        self, exposures: Exposures, enforce_copy: bool = True, **kwargs
+    ) -> Exposures:
+        """Apply the changes from the measure to the given :class:`Exposures` object.
+
+        This method applies the `exposures_changes` function of the measure to
+        the provided :class:`Exposures` object. If ``enforce_copy`` is True (default), a
+        deep copy of the exposures is created before modification to ensure
+        immutability of the original object.
```
ValentinGebhart, line 238: Is this not rather "preservation" instead of "immutability"? The original itself could still be mutable?

ValentinGebhart: same thing in the other two docstrings below

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        -------
+        Tuple[Exposures, ImpactFuncSet, Hazard]
+            A tuple containing the transformed entities in the order:
+            (changed_exposures, changed_impfset, changed_hazard). If `enforce_copy`
+            was True, these are new objects; otherwise, they may reference the
+            original inputs or modified versions thereof.
+
+        Notes
+        -----
+        The kwargs merging follows this priority order:
```
ValentinGebhart, line 461: maybe not super clear. In my mind prio 1 is overriding prio 2, but below I think 2. is overriding 1.? maybe rephrase

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        )
+        return changed_exposures, changed_impfset, changed_hazard
+
+    def calc_impact(
+        self, exposures: Exposures, impfset: ImpactFuncSet, hazard: Hazard
+    ) -> Impact:
+        """Compute impact with the active measure.
+
+        This convenience method allow to compute an ``Impact`` object from an
+        (Exposure, ImpactFuncSet, Hazard) triplet by first applying the measure
```
ValentinGebhart, line 501: "after having applied" instead of "by first applying"?

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        if new_haz.centr_exp_col not in new_exp.gdf.columns:
+            LOGGER.warning(
+                "No assigned hazard centroids in exposure object after the "
+                "application of the measure. The centroids will be assigned during impact "
+                "calculation. This is potentiall costly. To silence this warning, make sure "
+                "that centroids are assigned to all exposures."
+            )
+            new_exp.assign_centroids(new_haz)
+        imp = ImpactCalc(new_exp, new_impfs, new_haz).impact(
+            save_mat=False, assign_centroids=False
```
ValentinGebhart, line 533: why do we need save_mat False hard-coded? Could this not be an optional parameter?

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        Function to change exposures.
+    impfset_change : ImpfsetChange
+        Function to change impact function set.
+    hazard_change : HazardChange
+        Function to change hazard.
+    sub_measures : list of str, optional
+        List of measure names that this measure is a combination of.
+    cost_income : climada.entity.measures.cost_income.CostIncome
+        Cost and income object associated with the measure.
+    implementation_duration : pd.DateOffset, optional
```
ValentinGebhart, line 170: as the sub_measures attribute, the implementation_duration attribute is not used anywhere directly? is this here for later additions to the code?

climada/entity/measures/test/test_measure.py
```diff
@@ -0,0 +1,372 @@
...
+            return fn
+
+        m = Measure(
+            "order_test",
+            exposures_changes=track("exposures"),
+            hazard_changes=track("hazard"),
+            impfset_changes=track("impfset"),
+        )
+        m.apply(_make_mock(), _make_mock(), _make_mock(), enforce_copy=False)
+        assert call_order == ["exposures", "hazard", "impfset"]
```
ValentinGebhart, line 372: very cool tests with all the mocking :) 

I'm wondering if it makes sense to eventually also add some tests that actually check numbers, e.g. when you combine two different effects (e.g. using the "base" objects) to see if the effect is what you want? Or should the mock tests above be enough to see that the effects of the change functions (and their interactions?) are the intended ones? Sorry for this vague comment ;)

ValentinGebhart (commented):
This new implementation of the measures looks super cool :) I think it is much cleaner and user-friendly to before! I just commented whenever something was unclear to me. I haven't had time to check the working of the functions with actual data yet , e.g. for some consistency checks.

The only one more conceptual question I have is if you have to work with the (hidden) base_exposure, base_hazard, base_impfunc handling, but maybe I have not understood yet why these can be useful. Happy to discuss!

Note that I first read the tutorial without knowing the code to just check if I understand the usage. Some of the comments there might not be there know after I read the code :)

doc/user-guide/climada_measures.ipynb
spjuhel: Yup, that's just plain wrong! Will rework this.

climada/entity/measures/measure.py
spjuhel: Eh eh, yeah this is some advanced python magic and trickeries.


So, the current structure always passes `base_exposure`, etc., to the `X_changes` functions.
But I don't want to force the users to have to always write their "change functions" such that they accept kwargs, or specifically `base_exposure`, etc.

To clarify, without this wrapper the following function would not be accepted:

```python
def reduce_value(exposures: Exposures, reduction_factor) -> Exposures:
    exposures.gdf["value"] *= reduction_factor
    return exposures
```

Because it would trigger a `TypeError`, when the code supplies `base_exposure`, etc. to it.

One would have to write explicitly:

```python
def reduce_value(exposures: Exposures, reduction_factor, **kwargs) -> Exposures:
    exposures.gdf["vreduction_factoralue"] *= reduction_factor
    return exposures
```

or

```python
def reduce_value(exposures: Exposures, reduction_factor, base_exposures, base_hazard, base_impf) -> Exposures:
    exposures.gdf["value"] *= reduction_factor
    return exposures
```

---

The wrapper just creates a new function from `reduce_value`, say, `wrapped_reduce_value`. This new function accepts any keyword arguments, and discards the one that are not present in the signature of `reduce_value` before calling it.

When a user applies the measure:

```python
measure.apply(exp, impf, haz,  kwargs_exposures={'reduction_factor': 0.95})
```

The code actually will call `wrapped_reduce_value(exp, reduction_factor=0.95, base_exposures, base_hazard, base_impf)` which in turns calls `reduce_value(exp, reduction_factor=0.95)` keeping `reduction_factor`, but discarding  `base_exposures, base_hazard, base_impf` as they are not in its signature.

This way we can always supply `base_exposures, base_hazard, base_impf`, and the users may or may not use them as explicit keyword arguments.

spjuhel:
> This new implementation of the measures looks super cool :) I think it is much cleaner and user-friendly to before! I just commented whenever something was unclear to me. I haven't had time to check the working of the functions with actual data yet , e.g. for some consistency checks.
> 
> The only one more conceptual question I have is if you have to work with the (hidden) base_exposure, base_hazard, base_impfunc handling, but maybe I have not understood yet why these can be useful. Happy to discuss!
> 
> Note that I first read the tutorial without knowing the code to just check if I understand the usage. Some of the comments there might not be there know after I read the code :)

Thanks for these comments.

Answering the conceptual question about the `base_X` arguments. (see also https://github.com/CLIMADA-project/climada_python/pull/1279#discussion_r3387198877):

The idea comes from the fact that I assume there is a whole range of adaptation options effect on `Hazard|Exposures|Impfset` where you have to use some information in your base exposures, hazard or impfset. 

But it is a bit "advanced" to write for instance (ellipsed/simplified):

```python

def my_effect_on_exposures(exposures, req_hazard):
    indices = req_hazard > 10
    exposures.gdf.loc[indices,"value"] *= 0.98
    
measure_a = Measure(..., exposure_change=my_effect_on_exposures)

exp, impf, haz = measure.apply(exp, impf, haz, kwarg_exposures = {req_hazard: haz}) # That part is slightly scary
```

By always providing `base_Hazard|Exposures|Impfset`, you can "simplify":

```python

def my_effect_on_exposures(exposures, base_hazard): # note that the name need to be base_hazard!
    indices = base_hazard > 10
    exposures.gdf.loc[indices,"value"] *= 0.98
    
measure_a = Measure(..., exposure_change=my_effect_on_exposures)

exp, impf, haz = measure.apply(exp, impf, haz) # Scary kwargs are gone!
```

On this small example it may seem not much, but if the three effects all rely on the three components then it is way more relevant.

I agree the current explanations in the docs might be improved ;D

climada/entity/measures/test/test_measure.py
spjuhel: My understanding is that would be integration tests (`climada/test/test_measures.py`).

I'm not sure if they are satisfying yet though.

climada/entity/measures/measure.py
spjuhel: I think I should rephrase, 
"A Measure represents a single adaptation or risk-reduction action" is wrong.
A `Measure` can actually represent multiple actions bundled together.

spjuhel: However, `sub_measures` is indeed rather for bookkeeping :D

climada/entity/measures/measure.py
spjuhel: Yes, I should just remove it for this PR.

climada/entity/measures/measure.py
spjuhel: Probably better indeed!

climada/entity/measures/measure.py
spjuhel:
```suggestion
        (Exposure, ImpactFuncSet, Hazard) triplet after having applied the measure
```

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+
+class Measure:
+    """
+    Contains a measure to be applied to a set of exposures, impact functions,
+    and hazard.
+
+    A ``Measure`` represents a single adaptation or risk-reduction action. It
+    holds three (optional) transformation functions, one each for
+    :class:`Exposures`, :class:`ImpactFuncSet`, and :class:`Hazard`, that are
+    can be applied to a triplet of ``(Exposures, ImpactFuncSet, Hazard)``, to
```
spjuhel, line 131:
```suggestion
    A ``Measure`` represents an adaptation or risk-reduction action
    (or a bundle of actions considered as one). It holds three (optional)
    transformation functions, one each for :class:`Exposures`,
    :class:`ImpactFuncSet`, and :class:`Hazard`, that are
    can be applied to a triplet of ``(Exposures, ImpactFuncSet, Hazard)``, to
```

climada/entity/measures/measure.py
spjuhel:
```suggestion
        preservation of the original object.
```

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+            was True, these are new objects; otherwise, they may reference the
+            original inputs or modified versions thereof.
+
+        Notes
+        -----
+        The kwargs merging follows this priority order:
+
+        1. Default context: The original triplet (exposures, impfset, hazard)
+        2. Entity-specific overrides: Values from ``kwargs_exposures``,
+           ``kwargs_impfset``, or ``kwargs_hazard`` respectively
```
spjuhel, line 465:
```suggestion
         Keyword arguments other than "kwargs_exposures", "kwargs_impfset",
         "kwargs_hazard" are ignored. 
         
         The base exposure, hazard, and impfset provided by default can
         be overriden by explicitely providing a different value within
         ``kwargs_exposures``, ``kwargs_impfset``, or ``kwargs_hazard``
         respectively.
```

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+            LOGGER.warning(
+                "No assigned hazard centroids in exposure object after the "
+                "application of the measure. The centroids will be assigned during impact "
+                "calculation. This is potentiall costly. To silence this warning, make sure "
+                "that centroids are assigned to all exposures."
+            )
+            new_exp.assign_centroids(new_haz)
+        imp = ImpactCalc(new_exp, new_impfs, new_haz).impact(
+            save_mat=False, assign_centroids=False
+        )
```
spjuhel, line 534:
```suggestion
        imp = ImpactCalc(new_exp, new_impfs, new_haz).impact(
            save_mat=kwargs.get("save_mat",False), assign_centroids=kwargs.get("assign_centroids",False)
        )
```

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        changed_hazard = self.apply_hazard_changes(
+            hazard, enforce_copy, **kwargs_hazard
+        )
+        changed_impfset = self.apply_impfset_changes(
+            impfset, enforce_copy, **kwargs_impfset
+        )
+        return changed_exposures, changed_impfset, changed_hazard
+
+    def calc_impact(
+        self, exposures: Exposures, impfset: ImpactFuncSet, hazard: Hazard
```
spjuhel, line 496:
```suggestion
        self, exposures: Exposures, impfset: ImpactFuncSet, hazard: Hazard, **kwargs: dict,
```

climada/entity/measures/measure.py
```diff
@@ -0,0 +1,535 @@
...
+        Parameters
+        ----------
+        exposures : Exposures
+            The initial (no adaptation) exposures
+        impfset : ImpactFuncSet
+            The initial (no adaptation) impact function set
+        hazard : Hazard
+            The initial (no adaptation) hazard set
+
+        Returns
```
spjuhel, line 513:
```suggestion
        hazard : Hazard
            The initial (no adaptation) hazard set
        kwargs: dict
            Keyword arguments passed to the `impact()` method of `ImpactCalc`
            (Currently `save_mat` and `assign_centroids`)

        Returns
```

PR #1281, opened 2026-04-10
Upgrade GitHub Actions and Python versions

peanutfun:
Changes proposed in this PR:
- Python 3.11 to 3.13.

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "climada"
 version = "6.1.1-dev"
 description = "Framework for climate risk assessment and adaptation option appraisal"
 readme = "README.md"
-requires-python = ">=3.10,<3.13"
+requires-python = ">=3.10,<3.14"
```
chahank, line 10:
```
requires-python = ">=3.11,<3.14"
```
Do we want to require above 3.10 since the test fail due to the time needed to build the environment?

peanutfun: I don't have a strong opinion on that 🤷 @spjuhel @emanuel-schmid maybe?

spjuhel: Well technically I think you can install, it just takes forever.

I quickly looked around to see if there's a way to "discourage" a python version without enforcing the requirement, but it does not seem to exist (understandable).

I don't have a strong opinion either, and there are conflicting views on this:
1. Support all python versions that haven't reached End-of-Life is package works with them
2. Only support versions tested against.

I'd slightly favour 2. (because otherwise if we do implement some feature not available in 3.10 at some point, we will not see it fail), also because 3.10 end of life is at the end of 2026, so that's reasonable.

Discuss this at next SCRUM?

peanutfun: Is there a particular reason for the failure? Maybe just a conda hickup that might resolve itself again in a few days?

spjuhel: Edited 12:32:

My guess is that the dependency tree solving for 3.10 takes forever (10-20min) with respect to >3.10 (a few minutes). And we have an auto-cancel at 20 min.

    timeout-minutes: 20

We could increase the duration (but then I would move the test to develop-to-main PRs only, or to manual runs, to avoid the hassle?)

peanutfun:
@emanuel-schmid Why did you merge when tests are still failing?

PR #1290, opened 2026-05-13
Fix #814: raise clear error for empty Hazard in ImpactCalc.impact

### LINKED ISSUE #814: `np.array_split` in `ImpactCalc.impact` fails if Hazard is empty
The error is obscure. There is no helpful error message when computing an impact on an empty Hazard object

ankurmahiwal118028:
Closes #814. Adds an early guard in ImpactCalc.impact that raises a ValueError when the Hazard has no events, replacing the obscure np.array_split crash. Regression test added in test_impact_calc.py. CHANGELOG updated.


spjuhel (approved):
Good! Thanks for tackling this.

climada/engine/test/test_impact_calc.py
```diff
@@ -147,6 +147,22 @@ def test_error_handling_mismatch_haz_type(self):
                 "functions found for hazard type TC in impf_set.",
             )
 
+    def test_error_handling_empty_hazard(self):
+        """An empty Hazard must raise a clear ValueError (see GH #814)."""
+        haz_empty = Hazard("TC")
+        exp = Exposures()
```
peanutfun: Use non-empty exposure
```suggestion
        exp = Exposures(data={"value": [1.0]}, lat=[10.0], lon=[10.0])
```

peanutfun (changes requested):
Thank you for fixing the issue! I would only suggest to use a non-empty Exposure object, as an empty one will inevitably cause the impact calculation to throw an error before the issue with `np.array_split` occurs.

Also, you may want to add your name to the `AUTHORS.md`, at your discretion.

emanuel-schmid:
@ankurmahiwal118028 : the change, @peanutfun suggested makes sense, so I applied it.
I'll merge the pr soon, but if you want to add your name to the AUTHORS.md, you can also do it afterwards.

emanuel-schmid:
Many thanks, @ankurmahiwal118028 !

PR #1297, opened 2026-06-16
Fixes possible incorrect mapping between eai metric values and exposure points in trajectories

spjuhel:
This PR fixes a bug in `calc_eai_gdf` (both for static and interpolated trajectories) where the exposure points coordinates are incorrectly mapped to the corresponding risk value when the exposure `GeoDataFrame` index is discontinuous.

The suggested fix directly uses the index, instead of creating it from a range.

### PR Author Checklist

- [X] No new [linter issues][linter]

### PR Reviewer Checklist

### DIFF
--- a/climada/trajectories/calc_risk_metrics.py
+++ b/climada/trajectories/calc_risk_metrics.py
@@ -255,7 +255,12 @@ class CalcRiskMetricsPoints:
 
         """
 
-        metric_df = pd.DataFrame(self.per_date_eai, index=self._date_idx)
+        metric_df = pd.DataFrame(
+            self.per_date_eai,
+            index=self._date_idx,
+            columns=self.snapshots[0].exposure.gdf.index,
+        )
+
         metric_df = metric_df.reset_index().melt(
             id_vars=DATE_COL_NAME, var_name=COORD_ID_COL_NAME, value_name=RISK_COL_NAME
         )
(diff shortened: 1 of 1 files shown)


peanutfun (commented):
Looks reasonable. But are there no tests for the results?

spjuhel:
There are, just not for the very specific case where the index of the geodataframe of the exposure is not a continuous range of number.

peanutfun:
Is that a hassle to implement? Simply dropping a row from a built exposure object before the test should do it, right?

chahank (approved):
Does this need a test to avoid breaking it in the future?

climada/trajectories/calc_risk_metrics.py
```diff
@@ -255,7 +255,12 @@ def calc_eai_gdf(self) -> pd.DataFrame:
 
         """
 
-        metric_df = pd.DataFrame(self.per_date_eai, index=self._date_idx)
+        metric_df = pd.DataFrame(
+            self.per_date_eai,
+            index=self._date_idx,
+            columns=self.snapshots[0].exposure.gdf.index,
```
emanuel-schmid, line 261: tbh, i don't quite understand the idea behind having both, `self._date_idx` and `self.date_idx`, especially in the combination with the `@data_idx.setter` - nevertheless: why not   `index=self.date_idx?,`

emanuel-schmid: the only relevant difference to the develop branch seems to be the explicit setting of columns. however, there is no hint to the rationale in the PR description.

emanuel-schmid (changes requested):
looks nice - but sorry, I don't get it. 😕 A test would be awesome. But also a more explanatory description could help.

PR #1313, opened 2026-08-25
perf: compute forecast quantiles block-wise to bound memory

### LINKED ISSUE #1203: Column-wise quantiles in forecasts to reduce memory load
In `HazardForecast` and `ImpactForecast`, the entire sparse matrix is densified before computing the quantiles. This will easily crash for larger, sparse datasets. Instead of densifying the entire matrix, we should densify single columns, and compute the respective quantile individually. This should reduce the memory load of `quantile` (but can make it a bit slower).

EnvDroneSense:
Changes proposed in this PR:
- Add `climada.util.forecast.sparse_quantile_axis0`, which takes a quantile along axis 0 of a sparse matrix by densifying a block of columns at a time rather than the whole matrix. The block size is derived from a `max_memory_mb` budget.
- Call it from `HazardForecast._quantile` (intensity and fraction) and `ImpactForecast._quantile` (imp_mat), replacing the `.toarray()` on the full matrix.

This PR fixes #1203

Peak memory no longer scales with the number of centroids. Measured on a full DWD ICON-EU-EPS run, 2560 events x 164,984 cells, float32, with [[icon_full.py](https://github.com/user-attachments/files/31507435/icon_full.py)](https://github.com/user-attachments/files/31507435/icon_full.py) :

```
old    8110.9 ms (best of 3)   peak   3382.3 MB
new    2375.7 ms (best of 3)   peak    280.5 MB
```

The script asserts the two paths return equal arrays on every run.

One constraint worth flagging: implicit zeros have to take part in the quantile. A column of `[0, 0, 0, -1, 9]` has median 0.0, not the 4.0 you get from its two stored values. That is why the helper densifies blocks instead of reducing over `matrix.data`, which would be faster but silently wrong.

### DIFF
--- a/climada/engine/test/test_impact_forecast.py
+++ b/climada/engine/test/test_impact_forecast.py
@@ -19,11 +19,13 @@ with CLIMADA. If not, see <https://www.gnu.org/licenses/>.
 Tests for Impact Forecast.
 """
 
+from unittest.mock import patch
+
 import numpy as np
 import numpy.testing as npt
 import pandas as pd
 import pytest
-from scipy.sparse import csr_matrix
+from scipy.sparse import csc_matrix, csr_matrix
 
 from climada.engine import Impact, ImpactForecast
 
@@ -469,27 +471,15 @@ class TestReduce:
         )
 
 
-def test_quantile_does_not_densify(impact_forecast):
-    """quantile must not densify the whole impact matrix
-
-    The result must still equal the dense reference, so this pins both halves
-    of the fix: same answer, without the full-size temporary.
-    """
+def test_quantile_uses_block_wise_helper(impact_forecast):
+    """quantile routes through the block-wise helper and keeps the same result"""
     expected = np.quantile(impact_forecast.imp_mat.toarray(), 0.5, axis=0)
 
-    full_shape = impact_forecast.imp_mat.shape
-    densified = []
-    original = csr_matrix.toarray
-
-    def spy(self, *args, **kwargs):
-        densified.append(self.shape)
-        return original(self, *args, **kwargs)
-
-    csr_matrix.toarray = spy
-    try:
+    # the helper converts to CSC first, so csc_matrix is what densifies
+    with patch.object(
+        csc_matrix, "toarray", autospec=True, side_effect=csc_matrix.toarray
+    ) as spy:
         reduced = impact_forecast.quantile(0.5)
-    finally:
-        csr_matrix.toarray = original
 
-    assert full_shape not in densified, f"full matrix densified: {densified}"
-    npt.assert_allclose(reduced.imp_mat.toarray().squeeze(), expected)
+    assert spy.call_args_list, "quantile did not go through the block-wise helper"
+    npt.assert_array_equal(reduced.imp_mat.toarray().squeeze(), expected)
--- a/climada/engine/test/test_impact_forecast.py
+++ b/climada/engine/test/test_impact_forecast.py
@@ -467,3 +467,29 @@ class TestReduce:
             imp_fc_reduced.at_event,
             reduction_results_dim[dim][attr]["at_event"],
         )
+
+
+def test_quantile_does_not_densify(impact_forecast):
+    """quantile must not densify the whole impact matrix
+
+    The result must still equal the dense reference, so this pins both halves
+    of the fix: same answer, without the full-size temporary.
+    """
+    expected = np.quantile(impact_forecast.imp_mat.toarray(), 0.5, axis=0)
+
+    full_shape = impact_forecast.imp_mat.shape
+    densified = []
+    original = csr_matrix.toarray
+
+    def spy(self, *args, **kwargs):
+        densified.append(self.shape)
+        return original(self, *args, **kwargs)
+
+    csr_matrix.toarray = spy
+    try:
+        reduced = impact_forecast.quantile(0.5)
+    finally:
+        csr_matrix.toarray = original
+
+    assert full_shape not in densified, f"full matrix densified: {densified}"
+    npt.assert_allclose(reduced.imp_mat.toarray().squeeze(), expected)
--- a/climada/engine/test/test_impact_forecast.py
+++ b/climada/engine/test/test_impact_forecast.py
@@ -25,9 +25,10 @@ import numpy as np
 import numpy.testing as npt
 import pandas as pd
 import pytest
-from scipy.sparse import csc_matrix, csr_matrix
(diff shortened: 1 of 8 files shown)


climada/engine/test/test_impact_forecast.py
```diff
@@ -467,3 +467,29 @@ def test_reduce_dim(
...
+        densified.append(self.shape)
+        return original(self, *args, **kwargs)
+
+    csr_matrix.toarray = spy
+    try:
+        reduced = impact_forecast.quantile(0.5)
+    finally:
+        csr_matrix.toarray = original
+
+    assert full_shape not in densified, f"full matrix densified: {densified}"
```
peanutfun: This is a very convoluted way of mocking the csr_matrix.toarray method and adding `spy` as `side_effect`, see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect. I think it might even be safer (and more intuitive) to use the mock calls and check if it was never called with the entire original matrix as argument.

climada/util/forecast.py
```diff
@@ -200,3 +201,43 @@ def reduce_unique_selection(
...
+        only 7 percent less runtime.
+
+    Returns
+    -------
+    np.ndarray
+        1-D array of length ``matrix.shape[1]``.
+    """
+    n_cols = matrix.shape[1]
+    csc = matrix.tocsc()  # columns are the slow slicing direction for CSR
+    quantiles = np.empty(n_cols, dtype=np.float64)
```
peanutfun: According to the docs (https://numpy.org/doc/stable/reference/generated/numpy.quantile.html), the output of `quantile` is not necessarily float64, so it might be dangerous to set the result type here. How about concatenating the quantiles after the loop?

climada/util/forecast.py
```diff
@@ -200,3 +201,43 @@ def reduce_unique_selection(
         ],
         **concat_kws,
     )
+
+
+def sparse_quantile_axis0(
+    matrix: sparse.spmatrix, q: float, block_size: int = 1024
```
peanutfun: Instead of a block size it might be more convenient to set a maximum memory budget in bytes. The block size can then be derived from the matrix rows, data type and memory budget.

climada/util/test/test_forecast.py
```diff
@@ -93,3 +95,70 @@ def test_idx_lead_time():
...
+    @pytest.mark.parametrize("q", [0.0, 0.25, 0.5, 0.9, 1.0])
+    @pytest.mark.parametrize("block_size", [1, 3, 1000])
+    def test_matches_dense_quantile(self, q, block_size):
+        """Block-wise result must equal the dense reference exactly"""
+        rng = np.random.default_rng(0)
+        dense = rng.random((20, 37))
+        dense[dense < 0.7] = 0.0  # sparse, with implicit zeros dominating
+        mat = csr_matrix(dense)
+        result = sparse_quantile_axis0(mat, q, block_size=block_size)
+        npt.assert_allclose(result, np.quantile(dense, q, axis=0))
```
peanutfun: Are we expecting noise? The results should be equal, not just close, so we should test with `npt.assert_array_equal`. Also applies to all other tests.

peanutfun (changes requested):
@EnvDroneSense Thank you for raising a PR! I am happy about your contribution and your code generally looks good, but your PR description is lengthy, has many unsupported claims, and overall looks AI-generated.

In your "Results" section you claim a speedup and reduction in memory usage with the new code. How did you conduct these measurements? What code did you execute? Later, you claim that setting `overwrite_input=True` is faster. I see that it is reasonable to save memory, but the entire addition of the new function already serves that purpose. Did you measure the additional speedup as well? Also, the entire "Speedup" section never shows speedup but discusses saving memory. Finally, you claim that results are identical, but the unit tests use `assert_allclose` instead of `assert_array_equal`.

> The dim is not None branch of both _quantile methods delegates to reduce_unique_selection and is not changed here. It was not measured, and if it densifies by another route then https://github.com/CLIMADA-project/climada_python/issues/1203 is only partly addressed by this PR. Happy to extend the scope if you would prefer that handled in one go.

This is not necessary. The "dim is not None branch" calls `reduce_unique_selection`, which performs a subselection and then delegates back to the calling attribute, in this case `_quantile`, so there is no extension needed to apply your changes to that particular branch.

I would be happy if you could answer my questions regarding your benchmarks. Please also review my code comments, which I would like to see resolved before approving a merge.

Again, thank you for your contribution! ☺️

EnvDroneSense:
@peanutfun sorry for the messy PR. (I have rewritten it)

On the speedup: part of that was me misinterpreting my own benchmark results, so I have redone them from scratch. This time I measured against real data rather than synthetic [icon_full.py](https://github.com/user-attachments/files/31507435/icon_full.py), and the results look quite good:
```
run 2026-08-27 00:00:00
2560 events x 164,984 cells, density 0.055, dense would be 1,689.4 MB

old    8110.9 ms (best of 3)   peak   3382.3 MB
new    2375.7 ms (best of 3)   peak    280.5 MB
```
On overwrite_input=True: I measured it separately this time. At the default 8 MB budget it makes no difference to peak memory (280.5 MB either way) but is about 2x faster (2078 vs 4985 ms), because without it every block gets copied before sorting. So it is a speed optimization rather than a memory one. [icon_overwrite.py](https://github.com/user-attachments/files/31508867/icon_overwrite.py)
```
2560 events x 164,984 cells, budget 8 MB = 781 columns over 212 blocks, best of 5

overwrite_input=True      2078 ms   peak  280.5 MB
overwrite_input=False     4985 ms   peak  280.5 MB
identical
```
I have also implemented the changes you recommended. One thing I am unsure about: the default for max_memory_mb is currently an arbitrary 8 MB. Is that acceptable, and if not, how would you suggest picking a better one? Should it go into climada.conf instead?

Thank you for your review, I am trying to learn more about code optimization.

peanutfun:
@EnvDroneSense Excellent work! Using your benchmark code, I could replicate these numbers on my computer. The tests also run through on my end. A final request: Please update the test `test_quantile_uses_block_wise_helper` with a mock for `sparse_quantile_axis0`, ensuring that it is called with the correct matrix object, and add a similar test for the `HazardForecast` class. Once this is done, we are ready to merge!

EnvDroneSense:
@peanutfun Great. Both tests mock ``sparse_quantile_axis0`` and assert it receives the matrix object itself; the ``HazardForecast`` one covers both calls.

peanutfun (approved):
Excellent work! Thanks again ☺️

PR #1315, opened 2026-09-01
Only apply the cyclostrophic deprecation shim when the argument is passed

### LINKED ISSUE #1209: 'cyclostrophic' argument
**DeprecationWarning**

The 'cyclostrophic' argument is deprecated and will be removed in a future release. Include it in 'model_kwargs' instead.

**Occurrence**

- climada.hazard.trop_cyclone.trop_cyclone_windfields

(44 times during tests)

dylanpulver:
`compute_angular_windspeeds` declares `cyclostrophic: Optional[bool] = False` but guards the deprecation shim with `if cyclostrophic is not None:`. Since `False is not None`, the shim fires on **every** call, including internal ones that never pass the argument, and then does `model_kwargs["cyclostrophic"] = False` — overwriting the caller's own setting with `False`. That makes `cyclostrophic` unreachable through `model_kwargs`, which is the exact route the deprecation message tells users to take.

Fixes #1209.

**Changes**
- `cyclostrophic` defaults to `None` instead of `False`, so the shim only runs when a caller actually passes it.
- `model_kwargs` is copied before the shim writes to it, so a caller's dict is never mutated.

**Evidence**, on `develop`, ER11 at r = 30 km (r_max = 40 km, v_max = 40 m/s):

| call | v_ang (m/s) |
|---|---|
| `_stat_er_2011(..., cyclostrophic=True)` | 38.400000 |
| `_stat_er_2011(..., cyclostrophic=False)` | 39.670883 |
| `compute_angular_windspeeds(..., model_kwargs={"cyclostrophic": True})` | **39.670883** |

The third row is the bug: the request is silently ignored. The caller's dict also comes back as `{'cyclostrophic': False}`.

**No change to existing behaviour.** Without the injected `cyclostrophic=False` each model falls back to its own signature default, which is what `TropCyclone.from_tracks` already documents. Numbers move only for callers who explicitly asked for `cyclostrophic` and were previously ignored.

**Tests.** Four tests in `TestWindfieldHelpers`, asserting values derived from Emanuel and Rotunno (2011) eq. (36) rather than from the code. The first three took the file 9 passed -> 12 and each fails against the unpatched source. The fourth (the warning *is* raised when the argument is passed) was added in review and is unexecuted here — CLIMADA's geospatial stack will not build in this environment, so CI is the check for it. Inputs and reference values are now defined once and shared with `test_er_2011_pass`; I parsed both revisions to confirm the shared fixture and expected array are identical to the literals that test used before, so it is unchanged by construction. `black` 24.4.2 and `isort`, the pinned versions, report no changes.

Two boxes left unticked deliberately: `test_cross_antimeridian` downloads the coast-distance grid from Zenodo (504 here; CI caches it), and I have not run the full suite for this revision.

Written with AI assistance; the wind speeds are function output and 38.4 m/s is hand-derived from the published equation.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada/hazard/test/test_trop_cyclone_windfields.py
+++ b/climada/hazard/test/test_trop_cyclone_windfields.py
@@ -1,4 +1,5 @@
 import unittest
+import warnings
 
 import numpy as np
 import xarray as xr
@@ -9,6 +10,7 @@ from climada.hazard.trop_cyclone.trop_cyclone_windfields import (
     H_TO_S,
     KM_TO_M,
     MBAR_TO_PA,
+    MODEL_VANG,
     _B_holland_1980,
     _bs_holland_2008,
     _stat_er_2011,
@@ -17,6 +19,7 @@ from climada.hazard.trop_cyclone.trop_cyclone_windfields import (
     _v_max_s_holland_2008,
     _vtrans,
     _x_holland_2010,
+    compute_angular_windspeeds,
     get_close_centroids,
     tctrack_to_si,
 )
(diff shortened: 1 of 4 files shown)


climada/hazard/test/test_trop_cyclone_windfields.py
```diff
@@ -302,6 +305,84 @@ def test_er_2011_pass(self):
...
+                "rad": ("time", KM_TO_M * np.array([75.0, 40.0])),
+                "vmax": ("time", [35.0, 40.0]),
+                "lat": ("time", [20.0, 27.0]),
+                "cp": ("time", [4.98665369e-05, 6.61918149e-05]),
+            }
+        )
+        mask = np.array(
+            [[True, True, True, True], [True, False, True, True]], dtype=bool
+        )
+        return si_track, d_centr, mask
```
peanutfun: This is taken from the `test_er_2011_pass` test. Please share this information between the tests instead of re-defining it.

climada/hazard/test/test_trop_cyclone_windfields.py
```diff
@@ -302,6 +305,84 @@ def test_er_2011_pass(self):
...
+            model_kwargs={"cyclostrophic": True},
+        )
+        np.testing.assert_allclose(
+            v_ang_norm[1], self.ER11_NODE1_CYCLOSTROPHIC, atol=1e-6
+        )
+        # guard against the setting being silently dropped, which yields the
+        # Coriolis-corrected profile instead
+        self.assertFalse(
+            np.allclose(v_ang_norm[1], self.ER11_NODE1_WITH_CORIOLIS, atol=1e-4)
+        )
```
peanutfun: This is trivial. If the data is equal to `ER11_NODE1_CYCLOSTROPHIC`, it cannot be equal to `ER11_NODE1_WITH_CORIOLIS`
```suggestion
```

climada/hazard/test/test_trop_cyclone_windfields.py
```diff
@@ -302,6 +305,84 @@ def test_er_2011_pass(self):
...
+
+    def test_compute_angular_windspeeds_no_spurious_deprecation(self):
+        """No DeprecationWarning unless the deprecated argument is actually passed."""
```
peanutfun, line 336: This only checks if there is no deprecation warning. The check for the warning when passing `cyclostrophic` is missing. (This is what I read the test is intended to do?)

climada/hazard/test/test_trop_cyclone_windfields.py
```diff
@@ -302,6 +305,84 @@ def test_er_2011_pass(self):
...
+        return si_track, d_centr, mask
+
+    # Reference values for the second track node (r_max = 40 km, v_max = 40 m/s,
+    # f = 6.61918149e-05 1/s) from equation (36) of Emanuel and Rotunno 2011,
+    # M = M_max * 2 * (r/r_max)^2 / (1 + (r/r_max)^2) and v = M / r.
+    # Cyclostrophic: M_max = r_max * v_max = 1.6e6 m^2/s, so at r = 30 km,
+    # v = 1.6e6 * 2 * 0.5625 / 1.5625 / 30e3 = 38.4 m/s.
+    ER11_NODE1_CYCLOSTROPHIC = [38.4, 0.0, 3.194888178913738, 10.480349344978167]
+    # Non-cyclostrophic: M_max additionally contains 0.5 * f * r_max^2 = 52953.45 m^2/s.
+    ER11_NODE1_WITH_CORIOLIS = [39.670883, 0.0, 3.300626, 10.827206]
```
peanutfun: Reference values are also already defined in the `test_er_2011_pass`. Please avoid re-definition and use a single data source.

peanutfun (changes requested):
@dylanpulver Thank you for your contribution! I would appreciate if you used your advertised agentic AI systems to shorten your PR description to only include relevant information. It took me some time to figure out what this PR was actually about.

While your changes are obvious and solid, the tests are not. They re-define data that is already present in other tests. Please make sure testing and reference data is only defined once, and then "shared" between tests that are using them. See my comments below.

dylanpulver:
Thanks — all four addressed, and the description is rewritten to lead with what the PR is about rather than burying it.

- **Shared fixture.** The inputs and the reference values now live in one place (`_er_2011_inputs()` and `ER11_EXPECTED`) and `test_er_2011_pass` uses them too, so there is a single data source rather than a copy. `ER11_NODE1_WITH_CORIOLIS` is gone; it was row 1 of that same array.
- **Trivial assertion.** Removed, as suggested.
- **Missing deprecation case.** You read it right, and it was the more useful half. Added `test_compute_angular_windspeeds_cyclostrophic_deprecation`, which asserts the warning *is* raised when `cyclostrophic` is passed directly. The existing test only covered the absence of a spurious warning.
- **Description.** Cut roughly in half, leading with the defect.

One thing I should flag rather than let CI surface it: I could not run the suite for this revision — CLIMADA's geospatial stack does not build in my environment. So instead of asserting the refactor is safe, I parsed both revisions and compared: the shared fixture's literals and `ER11_EXPECTED` are identical to the ones `test_er_2011_pass` had inline (`0` vs `0.0` being the only textual difference), so that test is unchanged by construction rather than by my reading of it. `black` 24.4.2 and `isort` with the profile from `.pre-commit-config.yaml` both report no changes. The new deprecation test is the one thing genuinely unverified locally.

peanutfun (approved):

peanutfun:
@dylanpulver Thank you! I improved the tests to use the unittest fixture scheme and added an explicit test for the ER 2011 model with `cyclostropic=True`. Ready to merge.

PR #1317, opened 2026-09-05
perf: compute local exceedance statistics from the sparse columns

EnvDroneSense:
Changes proposed in this PR:
- Add `climada.util.interpolation.sparse_local_exceedance` and `sparse_local_frequency`, which compute the exceedance curve of every matrix column from that column's nonzero entries, instead of calling `preprocess_and_interpolate_ev` once per column on a densified CSR column.
- Call them from `Impact.local_exceedance_impact`, `Impact.local_return_period`, `Hazard.local_exceedance_intensity` and `Hazard.local_return_period`. `supports_sparse_fast_path` gates the switch; every other configuration keeps the existing code path unchanged.

`getcol` on a CSR matrix scans the whole index array, so calling it once per column made these methods O(n_columns × nnz).

Measured on a real US tropical cyclone dataset (LitPop USA 150 arcsec x historical TC hazard USA, both from the data API; 3,890 events x 643,099 exposure points) with [pr_benchmark_exceedance.py](ATTACH), which times both paths in one process and compares them:

```
                                        old          new
Impact.local_exceedance_impact      62700 ms      869 ms    72x
Impact.local_return_period          63241 ms      882 ms    72x
Hazard.local_exceedance_intensity  109259 ms     1173 ms    93x
Hazard.local_return_period         109263 ms     1210 ms    90x
```

In a full risk assessment on that data the step goes from 83.6% of the run to 6.6%, and the run from 74.7 s to 13.1 s.

The script compares every pair with `np.testing.assert_array_equal`: no differences.

One thing worth flagging: the zeros can be dropped because they cannot influence a retained point. Sorted ascending, the zeros come first, and the cumulative frequency is a reverse cumulative sum, so the entry for the p-th largest value only ever sums the p largest frequencies. `value_threshold=0` then discards exactly the entries the zeros sat in. This is why the helpers read `data[start:end]` rather than densifying; the reasoning depends on the threshold being 0, which is one of the conditions in `supports_sparse_fast_path`.

### PR Author Checklist

### PR Reviewer Checklist


EnvDroneSense:
The benchmarking script used to derive these numbers.
[pr_benchmark_exceedance.py](https://github.com/user-attachments/files/31871297/pr_benchmark_exceedance.py)

Issue #826, opened 2023-12-12
Litpop.from_shape_and_countries does not handle GeoSeries as a shape definition (patch proposed)


IanHopkinson:
I had an issue with `LitPop.from_shape_and_countries` whereby it would not accept `geopandas.geoseries.GeoSeries`

The error message was:

> 2023-12-12 10:32:01,162 - climada.entity.exposures.litpop.gpw_population - WARNING - Reference year: 2018. Using nearest available year for GPW data: 2020
> Traceback (most recent call last):
>   File "C:\Users\ianho\code\2023\hdx-scraper-climada\src\hdx_scraper_climada\create_csv_files.py", line 204, in <module>
>     export_litpop_data_to_csv(country="Haiti")
>   File "C:\Users\ianho\code\2023\hdx-scraper-climada\src\hdx_scraper_climada\create_csv_files.py", line 109, in export_litpop_data_to_csv
>     admin1_litpop = LitPop.from_shape_and_countries(admin1_shape, country_iso3a, res_arcsec=150)
>                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "C:\Users\ianho\code\2023\hdx-scraper-climada\venv\Lib\site-packages\climada\entity\exposures\litpop\litpop.py", line 424, in from_shape_and_countries
>     gdf = gdf.append(exp.gdf.loc[exp.gdf.geometry.within(shp)])
>           ^^^^^^^^^^
>   File "C:\Users\ianho\code\2023\hdx-scraper-climada\venv\Lib\site-packages\pandas\core\generic.py", line 6204, in __getattr__
>     return object.__getattribute__(self, name)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'GeoDataFrame' object has no attribute 'append'. Did you mean: '_append'?

I fixed the issue for me by patching this line:
```gdf = gdf.append(exp.gdf.loc[exp.gdf.geometry.within(shp)])```
to
```gdf = gdf._append(exp.gdf.loc[exp.gdf.geometry.within(shp)])```
changing .append to ._append:

https://github.com/CLIMADA-project/climada_python/blob/e41222210bfb99e92829c0a86f1db2f31355abf5/climada/entity/exposures/litpop/litpop.py#L424

[...]

PR #1320, opened 2026-09-07
fix: replace pandas APIs removed in 2.0 that leave four entry points …

### LINKED ISSUE #1319: pandas 2.0 removals leave four entry points dead
**Describe the bug**
Four code paths call pandas APIs removed in pandas 2.0 (`.append()` & `.iteritems()`), so they raise `AttributeError` on any currently supported pandas. `litpop.py:640` is probably the root cause of #826. None of the four is exercised by the test suite, so CI is green.

| Location | Removed call |
|---|---|
| `climada/entity/exposures/litpop/litpop.py:640` | `gdf.append(...)` |
| `climada/engine/impact_data.py:422` | `hit_countries.append(...)` |
| `climada/engine/calibration_opt.py:406` | `df_result.append(df_out, input)` |
| `climada/engine/calibration_opt.py:94` | `years_in_common.iteritems()` |

**To Reproduce**
1. Confirm the APIs are absent on a supported pandas.
2. Call `LitPop.from_shape_and_countries` with a `GeoSeries` or `list` shape (reaches `litpop.py:640`), or `calib_all` (reaches `calibration_opt.py:406`).

Code example:
```python
import pandas as pd, geopandas as gpd
print(pd.__version__)                                  # 2.2.3
print(hasattr(pd.DataFrame, "append"))                  # False
print(hasattr(gpd.GeoDataFrame, "append"))              # False
print(hasattr(pd.Series, "iteritems"))                  # False

# litpop.py:640, in the GeoSeries/list branch of `shape`:
#     gdf = gdf.append(exp.gdf.loc[exp.gdf.geometry.within(shp)])
# -> AttributeError: 'GeoDataFrame' object has no attribute 'append'
```

**Expected behavior**
These four functions run. `pyproject.toml` pins `pandas` with no upper bound, so any supported version hits this.

**Climada Version:** develop @ dea46fac (`6.1.1-dev`)

 - Operating system and version: Windows 11 Home 10.0.26200
 - Python version: 3.12.14 | packaged by conda-forge | (main, Sep  2 2026, 23:18:20) [MSC v.1944 64 bit (AMD64)]

**Additional context**

The fix differs per site:
- `litpop.py:640` and `calibration_opt.py:406` accumulate DataFrames; collect into a list and `pd.concat` once. This also removes the O(n²) copy the loop currently pays, since `.append` reallocated the accumulated frame on every iteration.
- `impact_data.py:422` appends a dict, not a frame; collect the rows into a list and build one `pd.DataFrame(records, columns=[...])` at the end.
- `calibration_opt.py:94` needs no restructuring: `Series.iteritems()` -> `Series.items()`.

`calibration_opt.py:406` also passes the builtin `input` as the second positional argument (`ignore_index`). Since `bool(input)` is `True`, it happened to mean `ignore_index=True`, so this was a latent bug rather than a visible one.

Two adjacent observations in the `litpop.py` loop, neither part of the fix above:
- `exp.gdf.geometry.within(shp)` scans every exposure point per shape with no spatial index, so the loop is O(n_shapes x n_points). The `Shape` branch at `:626` already uses `geopandas.sjoin` for the same job.
- A point inside two overlapping shapes is matched twice and appears twice in the result, so its value is double-counted in the exposure total. Reproduced on a 5-point toy case with two overlapping squares: 6 rows summing to 180.0, where the unique total is 150.0. `geopandas.sjoin(..., predicate="within", how="inner")` returns the identical row multiset, so switching would not change this. Whether it should be deduplicated is a question for the maintainers, since it would alter exposure totals for overlapping input shapes.

The test-coverage gap is arguably the more significant finding: four user-facing entry points are dead and nothing detected it.

EnvDroneSense:
Changes proposed in this PR:
- Replace the four calls to `DataFrame.append` and `Series.iteritems`, removed in pandas 2.0, that make their enclosing functions raise `AttributeError`: `litpop.py:640`, `impact_data.py:422`, `calibration_opt.py:406` and `calibration_opt.py:94`.
- Add one regression test per site. None of the four lines was previously executed by the test suite.

This PR fixes #1319 and #826

The affected entry points are `LitPop.from_shape_and_countries` (with a `GeoSeries` or `list` shape), `impact_data.hit_country_per_hazard`, `calibration_opt.calib_all`, and `calibration_opt.calib_instance` (with a multi-row `df_out` and `yearly_impact=True`). `pyproject.toml` declares `pandas` with no version constraint, so a fresh install always resolves to an affected version.

`litpop.py:640` is the cause of #826; that issue's traceback names this line, and its reporter proposed the same patch.

The fix differs by site. `litpop.py` and `calibration_opt.py:406` accumulate DataFrames, so they collect into a list and `pd.concat` once, which also removes the quadratic copy the loops paid. `impact_data.py:422` appends a dict rather than a frame, so it collects records and builds one `pd.DataFrame(records, columns=[...])`. `calibration_opt.py:94` is a rename.

Two behaviour details worth review:
- `calib_all` returns the single frame directly when there is exactly one parameter combination. The old code deep-copied the first `df_out` and only applied `ignore_index=True` from the second onward, so a caller-supplied `impact_data_source` index survived; `pd.concat(..., ignore_index=True)` alone would silently replace it.
- In `litpop.py` the type check moved out of the loop, ahead of any selection. Previously `[Polygon, Point]` ran a full point-in-polygon pass for the polygon before raising on the point.

Also `calibration_opt.py:406` passed the builtin `input` as the second positional argument (`ignore_index`). `bool(input)` is `True`, so it happened to mean `ignore_index=True`.

Each test was checked by reverting its own fix alone, with the other three in place:

```
reverted                    failing test                          error
litpop.py                   test_geoseries_shape                  'GeoDataFrame' has no attribute 'append'
impact_data.py              test_hit_country_per_hazard_pass      'DataFrame' has no attribute 'append'
calibration_opt.py:406      test_calib_all_multiple_params        'DataFrame' has no attribute 'append'
calibration_opt.py:94       test_calib_instance_yearly_multirow   'Series' has no attribute 'iteritems'
```

Note there is precedent for this bug class here: `test_emdat_impact_yearlysum_no_futurewarning` was added when the same `.append` problem was fixed in `emdat_impact_yearlysum`. This brings the remaining four sites in line with that.

### PR Author Checklist

### PR Reviewer Checklist


climada/engine/calibration_opt.py
```diff
@@ -393,19 +391,22 @@ def calib_all(
...
-
-    return df_result
+        df_out_list.append(df_out)
+
+    if not df_out_list:
+        return None
+    if len(df_out_list) == 1:
+        # single parameter combination: keep the frame's own index, as before
+        return df_out_list[0]
+    return pd.concat(df_out_list, ignore_index=True)
```
emanuel-schmid, line 409: can we make the changes minimal?
like, just

```python
            df_result = pd.concat([df_result, df_out], ignore_index=True)
```
instead of
```python
            df_result = df_result.append(df_out, input)
```

climada/engine/impact_data.py
```diff
@@ -409,26 +409,20 @@ def hit_country_per_hazard(intensity_path, names_path, reg_id_path, date_path):
...
+                {
+                    # Hit country ISO
+                    "hit_country": u_coord.country_to_iso(hit_ctry, "alpha3"),
+                    "Date_start": date[track],
+                    "ibtracsID": names[track],
+                }
+            )
     # retrun data frame with all hit countries per hazard
-    return hit_countries
+    return pd.DataFrame(records, columns=["hit_country", "Date_start", "ibtracsID"])
```
emanuel-schmid, line 425: Also here, can we keep the changes minimal:
```python
    # create data frame for output
    hit_countries = []
    for track, _ in enumerate(names):
        # Check if track has hit any country else go to the next track
        if len(all_hits[track]) > 0:
            # loop over hit_country
            for hit in range(0, len(all_hits[track])):
                # Hit country ISO
                ctry_iso = u_coord.country_to_iso(all_hits[track][hit], "alpha3")
                # create entry for each country a hazard has hit
                hit_countries.append(
                    {
                        "hit_country": ctry_iso,
                        "Date_start": date[track],
                        "ibtracsID": names[track],
                    }
                )
    # retrun data frame with all hit countries per hazard
    return pd.DataFrame(hit_countries, columns=["hit_country", "Date_start", "ibtracsID"])
```

climada/entity/exposures/litpop/litpop.py
```diff
@@ -632,17 +632,22 @@ def from_shape_and_countries(
...
                         "Not implemented for list or GeoSeries containing "
                         f"objects of type {type(shp)} as `shape`"
                     )
+            gdf_parts = [exp.gdf.loc[exp.gdf.geometry.within(shp)] for shp in shape]
+            gdf = (
+                pd.concat(gdf_parts)
+                if gdf_parts
+                else geopandas.GeoDataFrame(columns=exp.gdf.columns)
+            )
         else:
```
emanuel-schmid, line 651: I think this here bears an additional bug when shapes are overlapping. It should be something like this:

```python
        elif isinstance(shape, (geopandas.GeoSeries, list)):
            idx = np.array([False] * exp.gdf.shape[0], dtype=bool)
            for shp in shape:
                if isinstance(
                    shp, (shapely.geometry.MultiPolygon, shapely.geometry.Polygon)
                ):
                    idx |= exp.gdf.geometry.within(shp)
                else:
                    raise NotImplementedError(
                        "Not implemented for list or GeoSeries containing "
                        f"objects of type {type(shp)} as `shape`"
                    )
            gdf = exp.gdf.loc[idx]
        else:
            raise NotImplementedError()
```

climada/entity/exposures/test/test_litpop.py
```diff
@@ -427,6 +431,37 @@ def test_get_value_unit_pass(self):
         self.assertEqual(lp.get_value_unit("none"), "")
 
 
+class TestFromShapeAndCountries(unittest.TestCase):
```
emanuel-schmid, line 434: 🥇 excellent work.

We should include overlapping shapes though.

climada/engine/test/test_impact_data.py
```diff
@@ -300,9 +304,45 @@ def test_emdat_to_impact_2020format(self):
         self.assertAlmostEqual(1.610687e07, impact_emdat.at_event[4], places=0)
 
 
+class TestHitCountryPerHazard(unittest.TestCase):
```
emanuel-schmid, line 307: perfect

climada/test/test_calibration.py
```diff
@@ -94,7 +95,72 @@ def test_calib_instance(self):
         self.assertTrue(all(df_out_yearly["impact_CLIMADA"].values == [*IYS.values()]))
 
 
+class TestCalibPandas2(unittest.TestCase):
```
emanuel-schmid, line 98: can this be merged with the `TestCalib` class? Seems to have some synergy potential.

climada/test/test_calibration.py
```diff
@@ -94,7 +95,72 @@ def test_calib_instance(self):
...
+    """Cover the two calibration_opt paths that used removed pandas APIs (#1319)."""
```
emanuel-schmid, line 99: or - a plain description of the test per se without historical context?

like "test calib_instance with a non-empty df_out"

climada/test/test_calibration.py
```diff
@@ -94,7 +95,72 @@ def test_calib_instance(self):
...
+        for year in sel_years:
+            self.assertAlmostEqual(
+                df_out.loc[df_out["year"] == year, "impact_CLIMADA"].iloc[0], iys[year]
+            )
+
+    def test_calib_all_multiple_params(self):
+        """Several parameter combinations are concatenated into one frame.
+
+        A single-row `impact_data_source` keeps `calib_instance` on its other
+        branch, so this covers only the accumulation in `calib_all`.
```
emanuel-schmid, line 142: or simply "test_calib_all" as name and description, as the method has not been called yet.

emanuel-schmid (commented):
Thanks for pointing this out, the suggested fixes and writing tests.
I've made some remarks. You can address them as you wish. 

In any case, I'm going to just apply minimal fixes without much ado soon. :shrug: There bugs and ought to be removed.

